PR #97, opened 2020-12-09
Support primitive types

nhachicha:
- [x] Int
- [x] Boolean
- [x] Byte
- [x] Short
- [x] Long
- [x] Float
- [x] Double
- [x] Char
- [x] add integration tests (darwin/jvm)


rorbech (approved):
Looks good. I guess there is an superflous indirection with the `realm_get_value`/`realm_set_value` in my initial implementation that we could optimize away. But might be for another PR that digs a bit more into the options of tweaking paths with other C-API options.

PR #103, opened 2020-12-14
Publish to OJO using maven publish plugin

rorbech:
Using https://www.jfrog.com/confluence/display/JFROG/Gradle+Artifactory+Plugin somehow renders stale files or metadata for `cinterop-...-cinterop-realm_wrapper.klib` and `cinterop-...-sources.jar` files. The files are correct on first release, but on subsequent SNAPSHOT releases the files will not be uploaded or they are uploaded in place of the previous release's files. 

Using plain `maven-plugin` repository fixes the issue.

### DIFF
--- a/buildSrc/src/main/kotlin/Config.kt
+++ b/buildSrc/src/main/kotlin/Config.kt
@@ -16,7 +16,7 @@
  */
 
 object Realm {
-    const val version = "0.0.1-SNAPSHOT"
+    const val version = "0.0.2-SNAPSHOT"
     const val group = "io.realm.kotlin"
     const val projectUrl = "http://realm.io"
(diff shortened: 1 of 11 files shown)


buildSrc/src/main/kotlin/Config.kt
```diff
@@ -16,7 +16,7 @@
  */
 
 object Realm {
-    const val version = "0.0.1-SNAPSHOT"
```
nhachicha, line 19: we can keep `0.0.1-SNAPSHOT`

nhachicha (approved):

cmelchior (approved):
Nice 🚀

PR #107, opened 2020-12-16
Notification support

### LINKED ISSUE #65: Basic notification support
Basic notification support

rorbech:
Initial notification support. For now not trying to finalize API but just trying to hook up public API with underlying C implementation on both Android and iOS.

Current implementation covers: 
```
fun interface Callback {
    fun onChange()
}

interface Cancellable {
    fun cancel()
}

// Realm object notification
class Realm {
    companion object {
        fun <T: RealmModel> observe(obj: T, callback: Callback): Cancellable
    }
}

// ... and as extension method on RealmModels
fun RealmModel.observe(callback: Callback): Cancellable

// Realm result notifications
class RealmResults<T:RealmModel> {
    fun observe(callback: Callback): Cancellable
}
```

Notifications relies on the normal Realm Core schedulers, thus requiring to be registered on a thread with a `Looper` on Android and on the main `RunLoop` on Darwin platforms. 

Registrations are kept alive until explicitly canceled. Failing to cancel the registration will currently leak the callback and associated internals. 

Tests are currently written with coroutines as shared asynchronous abstraction, but are thus limited to be run on the main thread for K/N/

- [x] Object change notifications
- [x] Result change notifications
- [x] Lifecycle management (deregistration, deallocation, etc.)

Postponed:
- [ ] Error handling - `realm_get_async_error` and friends are not available in the C-API yet
- [ ] Finalize API with detailed change set information - Tracked by #115 instead.

Not targeted the below, which should probably just go hand in hand with the specific feature implementation applying the principles from object and result change notifications: 
- List change notifications
- Set change notifications
- Dict change notifications

Closes #65


test/src/androidTest/kotlin/io/realm/util/RunLoopThread.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val thread = HandlerThread("test-thread")
+    private val handler: Handler by lazy { thread.start(); Handler(thread.looper) }
+
+    override val coroutineContext: CoroutineContext by lazy { handler.asCoroutineDispatcher() + exceptionHandler }
+
+    private var error : Throwable? = null
+
+    val exceptionHandler = CoroutineExceptionHandler { _, exception ->
+        error = exception
+        println("CoroutineExceptionHandler got $exception")
```
nhachicha, line 41: should we throw?

examples/kmm-sample/androidApp/src/main/java/io/realm/example/kmmsample/androidApp/MainActivity.kt
```diff
@@ -14,6 +15,10 @@ fun greet(): String {
 }
 
 class MainActivity : AppCompatActivity() {
+
+    private lateinit var countTV: TextView
+    private var registration: Cancellable? = null
```
nhachicha, line 20: why is this nullable? can't we use `var lateinit` instead?

nhachicha (approved):
Nice job handling this, I've left minor comments 👍

packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -538,6 +542,48 @@ actual object RealmInterop {
         throwOnError(realm_wrapper.realm_object_delete(obj.cptr()))
     }
 
+    actual fun realm_object_add_notification_callback(obj: NativePointer, callback: Callback): NativePointer {
+        return CPointerWrapper(realm_wrapper.realm_object_add_notification_callback(
+            obj.cptr(),
+            // Use the callback as user data
+            StableRef.create(callback).asCPointer(),
+            staticCFunction<COpaquePointer?, Unit> { userdata ->
+                val ref: StableRef<Callback> = userdata!!.asStableRef()
```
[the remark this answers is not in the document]
rorbech: At least this is my empirical experience, but I will re-verify with Simon

packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -538,6 +542,48 @@ actual object RealmInterop {
         throwOnError(realm_wrapper.realm_object_delete(obj.cptr()))
     }
 
+    actual fun realm_object_add_notification_callback(obj: NativePointer, callback: Callback): NativePointer {
+        return CPointerWrapper(realm_wrapper.realm_object_add_notification_callback(
+            obj.cptr(),
+            // Use the callback as user data
+            StableRef.create(callback).asCPointer(),
```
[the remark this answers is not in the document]
rorbech: The `StableRef` should increment the reference count and will have to be manually `dispose`d. At least that is my understanding from https://github.com/JetBrains/kotlin-native/blob/master/INTEROP.md#passing-user-data-to-callbacks. 

The `realm_free_userdata_func_t` is only triggered on explicitly unregistration. I have put a note about the leakage if failing to `cancel` on `observe`. We could put some bookkeeping on top if we want to be able to unregister all notifications on close, etc. All internal notification callbacks passed to the platform layer are currently only referenced by the userdata, so there should be only one reference to it and should be released when disposing the stableref (or destructing the Swig director on JVM).

packages/library/src/commonMain/kotlin/io/realm/Callback.kt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm
+
+fun interface Callback {
+    fun onChange()
+    // FIXME API-NOTIFICATION Consider adding an onError(throwable: Throwable), but cannot be done
+    //  with functional interface
```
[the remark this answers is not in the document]
rorbech: Ah, yes. Head is sometimes stuck on Java7 🤦

packages/library/src/commonMain/kotlin/io/realm/Cancellable.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm
+
+interface Cancellable {
```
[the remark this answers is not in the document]
rorbech: We could do that. I would probably wait until we have the other use cases, as it would then be easier to evaluate if we should have a common root interface for all _resource_-like entities.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -51,6 +52,36 @@ class Realm {
...
+         * To receive asynchronous callbacks this must be called:
+         * - Android: on a thread with a looper
+         * - iOS/macOS: on the main thread (as we currently do not support opening Realms with
+         *   different schedulers similarly to
+         *   https://github.com/realm/realm-cocoa/blob/master/Realm/RLMRealm.mm#L424)
+         *
+         * Notes:
+         * - Calls are triggered synchronously on a [beginTransaction] when the version is advanced.
+         * - Ignoring the return value will eliminate the possibility to cancel the registration
+         *   and will leak the [callback] and internals related to the registration.
```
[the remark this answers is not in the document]
rorbech: No. Not unless we put some bookkeeping layer on top of it ourselves. Could consider to suggest it for the C-API, but then it is doable on our side.

test/src/androidTest/kotlin/io/realm/NotificationTests.kt
```diff
@@ -0,0 +1,217 @@
...
+import kotlin.test.assertTrue
+import kotlin.time.ExperimentalTime
+import kotlin.time.seconds
+
+private val INITIAL = "Hello, World!"
+private val FIRST = "FIRST"
+private val SECOND = "SECOND"
+
+@OptIn(ExperimentalTime::class)
+class NotificationTests {
```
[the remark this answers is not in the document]
rorbech: Already made an `RunLoopThread`-abstraction, but we cannot put it in `commonTest` and sym-link as it will then create overlapping symbols. Will add a FIXME with link to the youtrack issue.

test/src/androidTest/kotlin/io/realm/NotificationTests.kt
```diff
@@ -0,0 +1,217 @@
...
+        sample.observe {
+            val stringField = sample.stringField
+            this@run.launch {
+                c.send(stringField)
+            }
+        }
+
+        launch {
+            realm.beginTransaction()
+            assertEquals(INITIAL, c.receive())
```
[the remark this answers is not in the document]
rorbech: Would you like skip validation of the synchronous callback on advancing the versions on `beginTransaction`?

test/src/androidTest/kotlin/io/realm/NotificationTests.kt
```diff
@@ -0,0 +1,217 @@
...
+            sample.stringField = FIRST
+            realm.commitTransaction()
+            assertEquals(FIRST, c.receive())
+            realm.close()
```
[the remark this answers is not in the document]
rorbech: Yeah, they ended up almost being duplicates. Will remove this one. 

`realm_is_closed()` is not exposed through `RealmInterop` yet.

test/src/androidTest/kotlin/io/realm/util/RunLoopThread.kt
rorbech: I guess we should just remove the debug statement. The exception is stored and thrown when signalling the exit condition. This way the error is raised on the test execution thread and displayed correctly in the IDE instead of crashing the test-app.

examples/kmm-sample/androidApp/src/main/java/io/realm/example/kmmsample/androidApp/MainActivity.kt
rorbech: It is cleared when `cancel`-ing the subscription in `onPause`.

nhachicha: No need to preemptively setting it to null, it will be GC'ed once the Activity goes out of scope. Plus the safe call on `onPause` is unnecessary since the lifecycle guarantees to call `onResume` before `onPause` anyway which ensure we set the `registration` var

examples/kmm-sample/shared/src/commonMain/kotlin/io/realm/example/kmmsample/ExpressionRepository.kt
```diff
@@ -42,4 +43,8 @@ class ExpressionRepository {
     fun expressions(): List<Expression> {
         return realm.objects(Expression::class)
     }
+
+    fun listen(block: () -> Unit): Cancellable {
+        return realm.objects(Expression::class).addListener { block() }
```
nhachicha, line 48:
```suggestion
        return realm.objects(Expression::class).observe { block() }
```

test/src/androidTest/kotlin/io/realm/NotificationTests.kt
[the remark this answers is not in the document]
rorbech: I added asserts on the initial values.

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,30 @@ import shared
...
+    var counter: Counter
+    
+    class Counter: ObservableObject {
+        var count: Int
+        init(count: Int) {
+            self.count = count
+        }
+    }
+    
+    init () {
```
jsflax, line 17: extra space after `init`

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,30 @@ import shared
 struct ContentView: View {
     let calculator = Calculator.Companion()
     let greet = Greeting().greeting()
+    @State var token: LibraryCancellable? = nil
+    var counter: Counter
+    
+    class Counter: ObservableObject {
```
jsflax, line 10: As far as I can tell, this class is unneeded and could just be represented as:
`@State var count = 0`

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -35,6 +60,12 @@ struct ContentView: View {
                 Text(sum)
             }
             Text("History count: " + String(count))
+        }.onAppear() {
+            print("onAppear")
```
jsflax, line 64: Up to you if you want to leave print statements in the merged code, but I'd probably take them out

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,30 @@ import shared
...
+            return try calculator.listen {
+                self.counter.count = self.calculator.history().count
+                print("History updated \(self.counter.count)")
+            }
+        } catch {
+            print("Failed to register for notifications: \(error)")
+        }
+    }
     
     @State private var firstNum: String = "0"
```
jsflax, line 32: properties should go at the top of the struct

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,30 @@ import shared
...
+        init(count: Int) {
+            self.count = count
+        }
+    }
+    
+    init () {
+        counter = Counter(count : self.calculator.history().count)
+    }
+    
+    private func listen()-> LibraryCancellable {
```
jsflax, line 21: missing space after close paren

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -4,6 +4,30 @@ import shared
...
+        do {
+            return try calculator.listen {
+                self.counter.count = self.calculator.history().count
+                print("History updated \(self.counter.count)")
+            }
+        } catch {
+            print("Failed to register for notifications: \(error)")
```
jsflax, line 28: This isn't particularly useful to the user– but I'm not sure what alternative you would want in this sample code. Considering this is a precondition of the entire sample working, I would probably just `fatalError` with the appropriate message

examples/kmm-sample/iosApp/iosApp/ContentView.swift
```diff
@@ -35,6 +60,12 @@ struct ContentView: View {
                 Text(sum)
             }
             Text("History count: " + String(count))
```
jsflax, line 62:
```suggestion
            Text("History count: \(count)")
```

PR #110, opened 2021-01-06
Publishable Dokka dokumentation

rorbech:
This adds support for generating documentation for the public consumable API.  

The `dokka` artifact is not immediately consumable from the IDE, but I added the missing source jars, so now it should be available for all targets. This should be sufficient to make the dokka-documentation available in _Quick Documentation_ etc. 

The task of writing meaningful documentation is considered to be part of the individual issues for the major API parts (#77,  #83, etc.)

Includes
- [x] Generate Dokka documentation for `library` and `runtime-api`
- [x] Generate distributable containing the documentation


buildSrc/src/main/kotlin/Config.kt
```diff
@@ -57,6 +57,7 @@ object Versions {
     const val androidxTest = "1.3.0" // https://maven.google.com/web/index.html#androidx.test:rules
     const val autoService = "1.0-rc6"
     const val detektPlugin = "1.14.1"
+    const val dokka = "1.4.20"
```
cmelchior, line 60:
```suggestion
    const val dokka = "1.4.20" // https://github.com/Kotlin/dokka
```

Just adding an easy reference, so it is quick to lookup the source and find out if the version changed

Jenkinsfile
```diff
@@ -228,6 +233,18 @@ def test(task) {
...
 }
 
+def runGenerateDocs() {
+    withEnv(['PATH+USER_BIN=/usr/local/bin']) {
+        node(osx_kotlin) {
+            sh """
+                cd packages
+                ./gradlew --no-daemon :library:dokkaJar --info --stacktrace
+            """
+            archiveArtifacts artifacts: 'packages/library/build/libs/library-*-dokka.jar'
```
cmelchior, line 243: I don't think we normally archive the docs, but I guess it doesn't hurt either 👍

packages/library/build.gradle.kts
```diff
@@ -189,3 +190,28 @@ realmPublish {
     }
     ojo { }
 }
+
+tasks.dokkaHtml.configure {
```
cmelchior, line 194: I don't think this is enough to make the docs part of the maven publications, we had to do extra configs in Realm Java at least, but I can see that is still missing from your todo. You might want to see 

https://github.com/realm/realm-java/blob/master/realm/realm-library/build.gradle#L424 and https://github.com/realm/realm-java/blob/master/realm/realm-library/build.gradle#L252

rorbech: I added the `library-<version>-dokka.jar` to the common publication, but I do not really know if it makes sense as I don't think the IDE can consume it. I also tried with `javadoc` classifier without luck. 

The source jars are automatically published for the various platform variants and I manually added it for the common parts.

At least there is now some browsable option in the IDE and a downloadable bundle until we come up with something better.

Jenkinsfile
rorbech: It was just to have the generated artifact somewhere for inspection. I removed it in favor of adding the artifact to the publication.

nhachicha (approved):

packages/library/io.realm.md
```diff
@@ -0,0 +1,4 @@
+# Package io.realm
+
+Info for package realm
+
```
cmelchior, line 4: Maybe just add a TODO or FIXME, so it is clear this is work in progress?

packages/library/overview.md
```diff
@@ -0,0 +1,9 @@
+# Module Realm Kotlin SDK
+
+Realm Kotlin Multiplatform SDK
+
+## Overview
+
+## Usage
+
```
cmelchior, line 8: Same

cmelchior (approved):

rorbech:
@cmelchior Somehow missed that you put in some final remarks. Will address them in next PR (#107) instead

PR #111, opened 2021-01-06
Adding analytics

nhachicha:


packages/gradle-plugin/src/main/kotlin/io/realm/gradle/RealmAnalytics.kt
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    override fun afterExecute(task: Task, state: TaskState) {
+        if (!state.skipped && task.name.startsWith("compile")) {
+            sendMetricIfNeeded(task.project)
+        }
+    }
+
+    private fun sendMetricIfNeeded(project: Project) {
+        if (!METRIC_PROCESSED) {
```
cmelchior, line 84: The comment for the this variable indicates that we are concerned about race conditions here? In that case, we should probably synchronize access on the METRIC_PROCESSED boolean or some other lock?

packages/gradle-plugin/src/main/kotlin/io/realm/gradle/RealmAnalytics.kt
```diff
@@ -0,0 +1,276 @@
...
+
+    private fun jsonPayload(project: Project): String {
+        val userId: String = ComputerIdentifierGenerator.get()
+        val appId: String = anonymousAppId(project)
+        val osType = System.getProperty("os.name")
+        val osVersion = System.getProperty("os.version")
+
+        return """{
+                   "event": "$EVENT_NAME",
+                   "properties": {
```
cmelchior, line 102: In Java we also attempt to track Android minSdk/targetSdk, would it make sense to attempt to do this here as well?

It would probably also be nice to see if people are consuming sync/non-sync and if it is a KMP/pure android project. But since we are not supporting that just yet, we can probably just postpone it?

packages/gradle-plugin/src/main/kotlin/io/realm/gradle/RealmAnalytics.kt
```diff
@@ -0,0 +1,276 @@
...
+        if (projectName.isEmpty()) {
+            projectName = project.name
+        }
+
+        var packageName = project.group.toString()
+        if (packageName.isEmpty()) {
+            packageName = project.rootProject.group.toString()
+        }
+
+        return hexStringify(sha256Hash("$packageName.$projectName".toByteArray()))
```
cmelchior, line 126: If this works, it seems like a nicer way to track builders 👍 . The transformer approach of hashing model classes felt a bit weird.

packages/gradle-plugin/src/main/kotlin/io/realm/gradle/RealmAnalytics.kt
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    private fun sendAnalytics(json: String) {
+        try {
+            val pool = Executors.newSingleThreadExecutor()
+            try {
+                pool.execute { networkQuery(json) }
+                pool.awaitTermination(CONNECT_TIMEOUT + READ_TIMEOUT, TimeUnit.MILLISECONDS)
```
cmelchior, line 134: This seems dangerous? Wouldn't this block until it completes, which means it would block the Gradle script execution?

cmelchior (commented):
Nice 👍 ... I have 2 concerns though: One about sending metrics twice and one about us accidentally blocking Gradle.

packages/gradle-plugin/src/main/kotlin/io/realm/gradle/RealmAnalytics.kt
nhachicha: we're using the same approach in java https://github.com/realm/realm-java/blob/20f49687f96cfdee025d1ab0d0104ae3dff7cfc1/realm-transformer/src/main/kotlin/io/realm/transformer/RealmTransformer.kt#L171 I can relax the condition to perform an immediate shut

packages/gradle-plugin/src/main/kotlin/io/realm/gradle/RealmAnalytics.kt
nhachicha: Added min/target SDK. We'll add Sync when it's supported

rorbech (approved):
I think this is a reasonable start to ensure that we get some feedback - But, I don't see the need for using a `TaskExecutorAdapter` for the current collected metrics. But maybe just preparation for more detailed stats? 

We could consider installing a hook with `gradle.taskGraph.whenReady` to analyze the task graph once computed and collect relevant information from all the `compile*`/`withType<>` tasks in some scatter-gatter form and then only dispatch the request once all were done/skipped. That would ease collection of details from the individual compilation tasks and also make it more consistent of what was reported and when, compared to just posting on the first executed compile task according to `METRIC_PROCESSED`. 

Since the above would require significant investigations of the task structure I guess we could postpone it for later.

nhachicha:
@rorbech good points. I think I'll revisit this once we have a comprehensive list of things we want to capture (including at the compiler level) the refactor it accordignly. As a first step like you mentioned having an idea of the unique number of builds is a good start.

PR #112, opened 2021-01-07
Update type handling according to new C-API

rorbech:
This updates to the latest C-API with update API (`realm_string_t`->`const char*` and plain typedefs for key types).


packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -45,7 +45,7 @@ actual object RealmInterop {
                 primary_key = clazz.primaryKey
                 num_properties = properties.size.toLong()
                 num_computed_properties = 0
-                key = realm_table_key_t()
+                key = -1 // Use constant directly to avoid JNI round trip to realmc.getRLM_INVALID_CLASS_KEY()
```
nhachicha, line 48: we can also lazy load and cache them

nhachicha (approved):

PR #124, opened 2021-01-19
Fix automatic snapshot publishing

rorbech:


cmelchior (approved):

cmelchior:
Maybe we should also print the branch name in the log? Just to make it easier to see if it breaks again in the future?

Issue #106, opened 2020-12-14
Object import


rorbech:
We need support for setting initial values for properties when objects are created.

rorbech:
Fixed by #143

PR #126, opened 2021-01-22
One-to-one relationship

### LINKED ISSUE #71: Nullable field support

### LINKED ISSUE #120: Add support 1:1 relationships (Links)

rorbech:
This adds support for one-to-one relationships and nullability of fields derived from the nullability of the type. 

Note: Due to lack of support for default values retrieving non-nullable fields before setting it will currently crash. This should be fixed by #106.

Internals
- Generating typed schema in `RealmObjectCompanion`

Closes #71 and #120


test/src/androidTest/kotlin/io/realm/LinkTests.kt
```diff
@@ -0,0 +1,74 @@
...
+    @RealmModule(Parent::class, Child::class)
+    class MySchema
+
+    lateinit var realm: Realm
+
+    @BeforeTest
+    fun setup() {
+        val configuration = RealmConfiguration.Builder(schema = MySchema()).build()
+        realm = Realm.open(configuration)
+        // FIXME Cleaning up realm to overcome lack of support for deleting actual files
```
nhachicha, line 39: Check my GC PR where I introduced the platform-specific routines to create and delete temporary directories for test isolation

test/src/commonMain/kotlin/test/Nullability.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package test
+
+import io.realm.runtimeapi.RealmModel
+import io.realm.runtimeapi.RealmObject
+
+@RealmObject
+class Nullability : RealmModel {
```
nhachicha, line 23: can we extend to other supported types as well? (Long, Byte etc.)

test/src/androidTest/kotlin/io/realm/NullabilityTests.kt
```diff
@@ -0,0 +1,73 @@
...
+        realm.objects(Nullability::class).delete()
+        realm.commitTransaction()
+        assertEquals(0, realm.objects(Nullability::class).size, "Realm is not empty")
+    }
+
+    @Test
+    fun nullability() {
+        realm.beginTransaction()
+        val nullability = realm.create(Nullability::class)
+        assertNull(nullability.stringNullable)
```
nhachicha, line 50: Maybe we should disable this assertion until we implement default values? because it's weird to assert for nullability on `var stringNullable: String? = ""` maybe enable this test if the declared property was `var stringNullable: String? = null`

test/src/androidTest/kotlin/io/realm/NullabilityTests.kt
```diff
@@ -0,0 +1,73 @@
...
+        // What is the default semantics for this? Can we fix it before default values
+        assertNull(nullability.stringNullable)
+
+        nullability.stringNullable = "Realm"
+        assertNotNull(nullability.stringNullable)
+        nullability.stringNullable = null
+        assertNull(nullability.stringNullable)
+
+        // Should we try to verify that compiler will break on this
+        // nullability.stringNonNullable = null
```
nhachicha, line 60: What would be the benefit of this test at this layer (since the user cannot write such code?) we could check at the c-interop level the behaviour of passing null on a non-nullable column

test/src/commonMain/kotlin/test/link/Parent.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package test.link
+
+import io.realm.runtimeapi.RealmModel
+import io.realm.runtimeapi.RealmObject
+
+@RealmObject
+class Parent : RealmModel {
+    var child: Child? = null
```
nhachicha, line 24: What happens if we define a default value `child = Child("foo)` do you think we can prevent such behaviour now or when we implement default values?

nhachicha (approved):
Nice work on refactoring the IR to simplify it 🎉 I've left some minor comments

packages/cinterop/src/androidTest/java/io/realm/interop/CinteropTest.kt
```diff
@@ -130,7 +130,10 @@ class CinteropTest {
         realmc.propertyArrayArray_setitem(props, 1, properties_2)
 
         val realmSchemaNew = realmc.realm_schema_new(classes, 2, props)
-        assertTrue(realmc.realm_schema_validate(realmSchemaNew))
+        // FIXME Fix uint64_t of realm_schema_validate to avoid the need for toBigInteger. Put a
+        //  suggestion for the C-API, but maybe we should just truncate all uint64_t to long in
+        //  realm.i
+        assertTrue(realmc.realm_schema_validate(realmSchemaNew, realm_schema_validation_mode_e.RLM_SCHEMA_VALIDATION_BASIC.toBigInteger()))
```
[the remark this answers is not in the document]
rorbech: BigInteger is just coming from Swig's default typemap for `uint64_t`, and I was just not aware how/if we handled the overflow. From last weeks review meeting discussion I guess we can just go for not handling the overflow as in Java?

packages/cinterop/src/androidTest/java/io/realm/interop/CinteropTest.kt
```diff
@@ -236,4 +239,117 @@ class CinteropTest {
 
         realmc.realm_commit(realm)
     }
+
+    @Test
+    fun parentChildRelationship() {
+        val context = InstrumentationRegistry.getInstrumentation().context
+
+        System.loadLibrary("realmc")
+        println(realmc.realm_get_library_version())
```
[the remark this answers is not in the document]
rorbech: The `CinteropTest` was primarily used for ramping up the interop layer towards the C-API, so not sure they should stay, but I have found it useful to pin point issues in the C-API without any generated schema/accessors, so maybe it could stay in some form. Anyway, I extracted the tests and verified the output instead of printing.

packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt
```diff
@@ -108,19 +111,30 @@ class GenerationExtensionTest {
         val sampleModel = kClazz.newInstance()!!
```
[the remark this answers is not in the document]
rorbech: In 1.4.30? It is not marked deprecated in my environment.

test/src/androidTest/kotlin/io/realm/LinkTests.kt
rorbech: Added a TODO with link to it. Then we can see who gets to merge first.

test/src/commonMain/kotlin/test/Nullability.kt
rorbech: We could. I was a bit in doubt on how to define the axis for these basic test model/cases. The idea here was to define a specific class for the nullability test, to avoid the `AllTypes` as in Java, which is heavily confusing, because it has so many uses. Should I just include all non-containerized types here and verify them all? Or should we try to build the AllTypes++?

test/src/androidTest/kotlin/io/realm/NullabilityTests.kt
rorbech: Yeah. It was to somehow have a test that ensures that we don't mess up the checks for nullability now that I implemented the setters/getters as generic methods. Maybe it is at the wrong place, but it is just so damn easy to write the test when the schema is generated (which requires the compiler plugin). I will see if I can come up with a good pattern.

rorbech: Come to think of it. I guess `RealmObjectHelper` should in fact be in `library`, which would allow us to do the out-commented test, but I didn't want to do that in this PR as it ties a bit into the origanization of `@RealmObject` vs `RealmModel`-interface and similar.

test/src/commonMain/kotlin/test/link/Parent.kt
rorbech: I guess we could just `error` from the compiler plugin!?

test/src/commonMain/kotlin/test/Nullability.kt
nhachicha: Let's figure this out in the next Sprint

test/src/commonMain/kotlin/test/link/Parent.kt
rorbech: Will postpone this till default value implementation in #106

test/src/androidTest/kotlin/io/realm/LinkTests.kt
rorbech: I update all the tests to use your new tmp-dir infrastructure. A small note is that the shared tests cannot use `@ ExperimentalPathApi` as it is only available for JVM. I added the opt-in as a compiler command line arguments instead.

test/src/androidTest/kotlin/io/realm/NullabilityTests.kt
rorbech: I created an issue for reconsidering the module structure in #134. Let's do the clean up there.

test/src/commonMain/kotlin/test/Nullability.kt
rorbech: Yep. I created #133 for coming up with an appropriate pattern for exhaustive test that also breaks when we add new types.

PR #130, opened 2021-02-10
Add support for nullable primitive fields

### LINKED ISSUE #71: Nullable field support

rorbech:
This adds support for nullable (`<type>?`) primitive values. Since this relies heavily on the schema metainformation updated in the implementation of one-to-one relationships this PR currently targets #126.

Note: Due to lack of support for default values retrieving non-nullable fields before setting it will currently crash. This should be fixed by #106.

Closes #71


rorbech:
Merging into #126 as it relies heavily on infrastructure from that PR and contains a lot of clean up that should maybe have gone into #126

PR #131, opened 2021-02-10
Copy to realm by introspection (jvm only)

rorbech:
Experimental copy/copy to realm by introspection. Unfortunately it is JVM only, but maybe worth building our own KMutableProperty1 map in the companion object.


rorbech:
Superseded by #143

PR #132, opened 2021-02-15
Native Resources Cleanup

nhachicha:
- Garbage collector for Kotlin/Native and Kotlin/JVM
- Added `createTempDir` & `deleteTempDir` which are platform-dependent routines to increase test isolation (running each test within its own temp directory)


test/src/androidTest/kotlin/io/realm/InstrumentedTests.kt
```diff
@@ -72,6 +80,8 @@ class InstrumentedTests {
 
     @Test
     fun query() {
+        val realm = openRealmFromTmpDir()
```
rorbech, line 83: Seems like a common thing, so should probably go to the `setup` method.

test/src/androidTest/kotlin/io/realm/InstrumentedTests.kt
```diff
@@ -136,25 +149,156 @@ class InstrumentedTests {
...
+        realm.beginTransaction()
+        for (i in 1..100) {
+            realm.create(Sample::class).apply {
+                stringField = oneMBstring
+            }.also { referenceHolder.add(it) }
+        }
+        realm.commitTransaction()
+
+        mappedMemorySize = numberOfMemoryMappedBytes(command)
+        assertTrue(mappedMemorySize >= 99 * oneMB, "Committing the 100 objects should result in memory mapping of ~ 99 MB. Current amount is ${bytesToHumanReadable(mappedMemorySize)}")
```
rorbech, line 226: Should we have an upper bound, just to ensure that the test does what it is supposed to?

test/src/nativeTest/kotlin/io/realm/InstrumentedTests.kt
```diff
@@ -39,30 +51,30 @@ class InstrumentedTests {
...
+    fun tearDown() {
+        deleteTempDir(tmpDir)
     }
 
     // FIXME API-CLEANUP Do we actually want to expose this. Test should probably just be reeavluated
     //  or moved. Local implementation of pointer wrapper to support test. Using the internal one would
     //  require the native wrapper to be api dependency from cinterop/library. Don't know if the
     //  test is needed at all at this level
-    class CPointerWrapper(val ptr: CPointer<out CPointed>?) : NativePointer
+    class CPointerWrapper(val ptr: CPointer<out CPointed>?, managed: Boolean = true) : NativePointer
```
rorbech, line 70: This seems a bit dangerous, as the implementation now differs significantly. Given the FIXME it should maybe just go or be moved to an internal test in the compiler plugin...maybe something similar is already there.

test/src/nativeTest/kotlin/io/realm/InstrumentedTests.kt
```diff
@@ -80,4 +92,98 @@ class InstrumentedTests {
             assertEquals("Sample", realmModel.`$realm$TableName`)
         }
     }
+
+    // TODO Only run on macOS, filter using https://developer.apple.com/documentation/foundation/nsprocessinfo/3608556-iosapponmac when upgrading to XCode 12
+    @Test
+    fun garbageCollectorShouldFreeNativeResources() {
```
rorbech, line 98: I think these memory related test should go to their own file. There is a lot of infrastructure specific to verifying the memory usage, etc. that pollutes the ordinary test cases.

test/src/androidTest/kotlin/io/realm/InstrumentedTests.kt
```diff
@@ -136,25 +149,156 @@ class InstrumentedTests {
...
-        realmModel.`$realm$IsManaged` = true
-        realmModel.`$realm$ObjectPointer` = LongPointerWrapper(0xCAFEBABE)
-        realmModel.`$realm$Pointer` = LongPointerWrapper(0XCAFED00D)
-        realmModel.`$realm$TableName` = "Sample"
-
-        assertEquals(true, realmModel.`$realm$IsManaged`)
-        assertEquals(0xCAFEBABE, (realmModel.`$realm$ObjectPointer` as LongPointerWrapper).ptr)
-        assertEquals(0XCAFED00D, (realmModel.`$realm$Pointer` as LongPointerWrapper).ptr)
-        assertEquals("Sample", realmModel.`$realm$TableName`)
+    fun garbageCollectorShouldFreeNativeResources() {
```
rorbech, line 153: Should probably go to its own file.

test/src/androidTest/kotlin/io/realm/InstrumentedTests.kt
```diff
@@ -136,25 +149,156 @@ class InstrumentedTests {
...
+            }
+        }.toString()
+        val oneMB = 1048576L
+
+        var mappedMemorySize = numberOfMemoryMappedBytes(command)
+        assertTrue(mappedMemorySize < oneMB, "Opening a Realm should not cost more than 12KB")
+
+        // inserting ~ 100MB of data and keep a strong reference to all allocated objects
+        val referenceHolder = mutableListOf<Sample>()
+        realm!!.beginTransaction()
```
rorbech, line 172: Might be that there are too fundamental differences in the operation on platforms, but it seems like much of the test code is the same on the platforms and only the memory footprint calculations are differs. Could it be commonized into one implementation with some platform specific utility methods?

packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -75,8 +78,16 @@ private fun throwOnError(pointer: CPointer<out CPointed>?): CPointer<out CPointe
 }
 
 // FIXME API-INTERNAL Consider making NativePointer/CPointerWrapper generic to enforce typing
-class CPointerWrapper(ptr: CPointer<out CPointed>?) : NativePointer {
+
+class CPointerWrapper(ptr: CPointer<out CPointed>?, managed: Boolean = true) : NativePointer {
```
rorbech, line 82: I guess we should remember to make the notification tokens unmanaged?

packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/LongPointerWrapper.kt
```diff
@@ -16,13 +16,20 @@
...
 // JVM/Android specific pointer wrapper
 // FIXME Should ideally be moved to jni-swig-stub as we would be able to use Swig to wrap/unwrap
 //  all pointers going in and out of the JNI layer, handling transferring ownership, etc. But,
 //  doing so currently renders Android studio unable to resolve the NativePointer from the
-//  runtiem-api mpp-module, which ruins IDE solving of the while type hierarchy around the
+//  runtime-api mpp-module, which ruins IDE solving of the while type hierarchy around the
 //  pointers, which makes in annoying to work with.
 //  https://issuetracker.google.com/issues/174162078
-class LongPointerWrapper(val ptr: Long) : NativePointer
+class LongPointerWrapper(val ptr: Long, managed: Boolean = true) : NativePointer {
```
rorbech, line 29: Same as for native: I guess we should remember to make the notification tokens unmanaged?

rorbech (commented):
Great. Nothing fundamental, but added some comments.

I removed the `packages/cinterop/c_api_test.realm.management/access_control.control.mx` file in #126, so should go soon.

packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/gc/NativeContext.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.realm.interop.gc
+
+import io.realm.interop.LongPointerWrapper
+import java.lang.ref.ReferenceQueue
+
+object NativeContext {
+    private val referenceQueue: ReferenceQueue<LongPointerWrapper> = ReferenceQueue<LongPointerWrapper>()
+    private val finalizingThread = Thread(FinalizerRunnable(referenceQueue))
```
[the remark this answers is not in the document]
nhachicha: I don't think we have a reliable `shutdown` event we could use to explicitly stop this Thread, since this cleaner is used across Threads, and we don't do the bookkeeping to detect that the Realm is closed all over the process's Threads to reliably close this cleaner.

packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/gc/NativeContext.kt
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.interop.gc
+
+import io.realm.interop.LongPointerWrapper
+import java.lang.ref.ReferenceQueue
+
+object NativeContext {
```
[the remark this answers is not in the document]
nhachicha: This is slightly different from Java where some pointers like `SharedRealm` are not _dtor_ Thread safe. Since we're not using `SharedRealm` & `Table` API in the C-API I don't see why we should reintroduce the dummy context.

rorbech (approved):
Approving after aligning on the comments face to face, but still need to ensure that the notification token's underlying native pointer is not freed from GC. But I guess we would need some other means of cleaning up (maybe on close) if people forget to `cancel` the subscription themselves.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModuleSyntheticMediatorInterfaceGeneration.kt
```diff
@@ -393,9 +392,9 @@ class RealmModuleSyntheticMediatorInterfaceGeneration(private val pluginContext:
                             }
                     }
                 )
-                val mutableIteratorVar = irTemporaryVar(nameHint = "iterator", value = callIterator.expression)
+                val mutableIteratorVar = irTemporary(nameHint = "iterator", value = callIterator.expression)
 
-                val elementVar = irTemporaryVarDeclaration(realmCompanionIrClass.defaultType.makeNullable(), nameHint = "element", isMutable = true)
+                val elementVar = irTemporary(irType = realmCompanionIrClass.defaultType.makeNullable(), nameHint = "element", isMutable = true)
```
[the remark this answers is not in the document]
nhachicha: Probably we'll migrate to `mapOf` during the technical sprint

test/src/androidTest/kotlin/io/realm/NotificationTests.kt
```diff
@@ -56,6 +56,7 @@ class NotificationTests {
         realm.objects(Sample::class).delete()
         realm.commitTransaction()
         assertEquals(0, realm.objects(Sample::class).size, "Realm is not empty")
+        realm.close()
```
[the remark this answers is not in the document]
nhachicha: Like the tests reflects we have two ways of. closing the Realm
1-  Calling `close` will force close the Realm instance in the current Thread even though references are still being used,  using references (RealmObject passed to ViewModel for instance) after closing the Realm will cause a Segmentation fault
2- The Realm pointer is reference-counted which means that when all the RealmObjects goes out of scope and the Realm instance itself goes out of scope it will be closed

test/src/androidTest/kotlin/io/realm/InstrumentedTests.kt
nhachicha: Good refactor next Sprint

test/src/nativeTest/kotlin/io/realm/InstrumentedTests.kt
nhachicha: Let's discuss this in the refactor Sprint

test/src/nativeTest/kotlin/io/realm/InstrumentedTests.kt
nhachicha: Good point, especially for Darwin where some should not be run under iOS and be filtered.

test/src/androidTest/kotlin/io/realm/InstrumentedTests.kt
nhachicha: Didn't want to jump into this until we figure out how to run and filter specific platform test (like for `iOS`)

packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
nhachicha: Good catch 👍

PR #136, opened 2021-02-26
Remove usage of RealmObject to define schema

nhachicha:
Fixing https://github.com/realm/realm-kotlin/issues/123

Realm model classes are defined using the marker interface only 

```Kotlin
class Sample : RealmObject {
     var stringField: String? = "Realm"
}
```


packages/plugin-compiler/src/main/kotlin/io/realm/compiler/Identifiers.kt
```diff
@@ -54,10 +54,10 @@ internal object Names {
 
 internal object FqNames {
     // TODO we can replace with RealmObject::class.java.canonicalName if we make the runtime_api available as a compile time only dependency for the compiler-plugin
-    val REALM_OBJECT_ANNOTATION = FqName("io.realm.runtimeapi.RealmObject")
     val REALM_NATIVE_POINTER = FqName("io.realm.runtimeapi.NativePointer")
     val REALM_MODULE_ANNOTATION = FqName("io.realm.runtimeapi.RealmModule")
-    val REALM_MODEL_INTERFACE = FqName("io.realm.runtimeapi.RealmModelInternal")
+    val REALM_MODEL_INTERNAL_INTERFACE = FqName("io.realm.runtimeapi.RealmModelInternal")
```
rorbech, line 59: Maybe cleaning up the namespace would render us able to reason about what is what and we could eliminate the `_INTERFACE`?

rorbech (commented):
I am not fond of the `RealmModel` name, as model would sound more like the overall model and not a specific class. I get that `RealmObject` could indicate that it somehow relates to a singleton, but IMO I think it is less confusing that `RealmModel`. I guess we could discuss with the team. 

I put some additional notes on #123 a while ago that we had a similar issue with `RealmModule`/`Mediator`. If not targeting that in same go, we should create a new issue for that.

rorbech (approved):

PR #140, opened 2021-03-09
Support Xcode12 and Arm64 ABIs

nhachicha:
- Add support for XCode12
- Building for iOS Arm64 iOS Simulator (x86_64, Arm64) and macOS 
- Fixes https://github.com/realm/realm-kotlin/issues/72 and https://github.com/realm/realm-kotlin/issues/138


examples/kmm-sample/iosApp/iosApp.xcodeproj/project.pbxproj
```diff
@@ -307,7 +307,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cd \"$SRCROOT/..\"\n./gradlew :shared:packForXCode -PXCODE_CONFIGURATION=${CONFIGURATION}\n";
+			shellScript = "cd \"$SRCROOT/..\"\nANDROID_SDK_ROOT=$HOME/Library/Android/sdk ./gradlew :shared:packForXCode -PXCODE_CONFIGURATION=${CONFIGURATION}\n";
```
rorbech, line 310: It worked out locally for me before applying this, so there must be some other mechanism for "injecting" this...but I guess this should ruing anything.

packages/cinterop/build.gradle.kts
```diff
@@ -82,12 +84,26 @@ android {
     }
 }
 
-val nativeLibraryIncludes = mutableListOf(
-    "-include-binary", "$rootDir/../external/core/build-macos_x64/src/realm/object-store/c_api/librealm-ffi-static-dbg.a",
-    "-include-binary", "$rootDir/../external/core/build-macos_x64/src/realm/librealm-dbg.a",
-    "-include-binary", "$rootDir/../external/core/build-macos_x64/src/realm/parser/librealm-parser-dbg.a",
-    "-include-binary", "$rootDir/../external/core/build-macos_x64/src/realm/object-store/librealm-object-store-dbg.a"
+val nativeLibraryIncludesMacosUniversal = listOf(
```
rorbech, line 87: Looks like something that could be defined as a list and mapped onto a specific base platform directory.

packages/cinterop/build.gradle.kts
```diff
@@ -137,7 +154,7 @@ kotlin {
             // ... and def file does not support using environment variables
             // https://github.com/JetBrains/kotlin-native/issues/3631
             // so resolving paths through gradle
-            kotlinOptions.freeCompilerArgs += nativeLibraryIncludes
+            kotlinOptions.freeCompilerArgs += nativeLibraryIncludesMacosUniversal
```
rorbech, line 157: Does these _universal_ builds eliminate the option to use the monorepo CMAKE recipes?

rorbech (approved):
Awesome. Maybe create issues for future optimizing (linking, stripping, etc.) and for consuming monorepo binaries directly if we should consider that.

examples/kmm-sample/iosApp/iosApp.xcodeproj/project.pbxproj
nhachicha: still asking for it locally for me, maybe because I'm using the ` includeBuild("../../packages")` which compiles and build the artefacts from XCode rather than pulling them via maven repo ...

packages/cinterop/build.gradle.kts
nhachicha: I don't think we can use the `listOf().map{}` transformation since the ` "-include-binary"` needs to be emitted as a sperate argument (not concatenated with space `"-include-binary $rootDir/../external/core/$it"`) otherwise it won't work

PR #146, opened 2021-03-16
Fix Android GC crash

### LINKED ISSUE #145: Fix GC crash on Android

nhachicha:
fixes #145

### DIFF
--- a/packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
@@ -244,7 +244,7 @@ actual object RealmInterop {
                 obj.cptr(),
                 object : NotificationCallback {
                     override fun onChange(pointer: Long) {
-                        callback.onChange(LongPointerWrapper(pointer, managed = false))
+                        callback.onChange(LongPointerWrapper(pointer, managed = false)) // FIXME use managed pointer https://github.com/realm/realm-kotlin/issues/147
                     }
                 }
             ),
@@ -258,7 +258,7 @@ actual object RealmInterop {
                 results.cptr(),
                 object : NotificationCallback {
                     override fun onChange(pointer: Long) {
-                        callback.onChange(LongPointerWrapper(pointer, managed = false))
+                        callback.onChange(LongPointerWrapper(pointer, managed = false)) // FIXME use managed pointer https://github.com/realm/realm-kotlin/issues/147
                     }
                 }
             ),
--- a/packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
@@ -240,11 +240,11 @@ actual object RealmInterop {
 
     actual fun realm_object_add_notification_callback(obj: NativePointer, callback: Callback): NativePointer {
         return LongPointerWrapper(
-            realmc.realm_object_add_notification_callbackJNI(
+            realmc.register_object_notification_cb(
                 obj.cptr(),
-                object : io.realm.interop.NotificationCallback() {
+                object : NotificationCallback {
                     override fun onChange(pointer: Long) {
-                        callback.onChange(LongPointerWrapper(pointer))
+                        callback.onChange(LongPointerWrapper(pointer, managed = false))
                     }
                 }
             ),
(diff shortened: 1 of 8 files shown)


packages/jni-swig-stub/realm.i
```diff
@@ -179,63 +210,68 @@ struct realm_size_t {
 // How to
 // - delete all listeners
 %typemap(directorin, descriptor="J") const void* "*(const void **)&$input = $1;"
```
rorbech, line 212: This type map is also not needed anymore if we are not using directors.

packages/jni-swig-stub/realm.i
```diff
@@ -179,63 +210,68 @@ struct realm_size_t {
...
-        realm_object_t* object,
-        NotificationCallback* callback
-        ) {
+realm_notification_token_t*  register_results_notification_cb(realm_results_t* results, jobject callback) {
+        using namespace realm::jni_util;
+    auto jenv = realm::jni_util::get_env();
+    static jclass notification_class = jenv->FindClass("io/realm/interop/NotificationCallback");
+    static jmethodID on_change_method = jenv->GetMethodID(notification_class, "onChange", "(J)V");
+
+        return realm_results_add_notification_callback(
```
rorbech, line 221: Indentation: Should follow the static above

packages/jni-swig-stub/realm.i
```diff
@@ -179,63 +210,68 @@ struct realm_size_t {
...
+                results,
+                // Use the callback as user data
+                static_cast<jobject>(get_env()->NewGlobalRef(callback)),
+        [](void *userdata) {
+            get_env(true)->DeleteGlobalRef(static_cast<jobject>(userdata));
+        },
+        // change callback
+        [](void* userdata, const realm_collection_changes_t* changes) {
+            auto jenv = realm::jni_util::get_env(true);
+            if (jenv->ExceptionCheck()) {
```
rorbech, line 231: What is the intention with this? Should it have been after invoking the callback? Or is the intention to identify issues with resolving the jenv??

packages/jni-swig-stub/realm.i
```diff
@@ -179,63 +210,68 @@ struct realm_size_t {
...
+            static_cast<jobject>(get_env()->NewGlobalRef(callback)),
+    [](void *userdata) {
+        get_env(true)->DeleteGlobalRef(static_cast<jobject>(userdata));
+    },
+    // change callback
+    [](void* userdata, const const realm_object_changes_t* changes) {
+        // FIXME API-NOTIFICATION Consider catching errors and propagate to error callback
+        //  like the C-API error callback below
+        auto jenv = get_env(true);
+        if (jenv->ExceptionCheck()) {
```
rorbech, line 263: Same as above...but here you are also throwing an exception??

packages/jni-swig-stub/realm.i
```diff
@@ -179,63 +210,68 @@ struct realm_size_t {
...
-                // FIXME API-NOTIFICATION Consider catching errors and propagate to error callback
-                //  like the C-API error callback below
-                static_cast<NotificationCallback*>(userdata)->onChange(static_cast<const void*>(changes));
-            },
-            // FIXME API-NOTIFICATION Error callback, C-API realm_get_async_error not available yet
-            [] ( void* userdata, const realm_async_error_t* async_error) { },
-            // FIXME NOTIFICATION C-API currently uses the realm's default scheduler
-            NULL
+            static_cast<jobject>(get_env()->NewGlobalRef(callback)),
+    [](void *userdata) {
```
rorbech, line 255: Indentation should match the line above

packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -240,11 +240,11 @@ actual object RealmInterop {
...
     actual fun realm_object_add_notification_callback(obj: NativePointer, callback: Callback): NativePointer {
         return LongPointerWrapper(
-            realmc.realm_object_add_notification_callbackJNI(
+            realmc.register_object_notification_cb(
                 obj.cptr(),
-                object : io.realm.interop.NotificationCallback() {
+                object : NotificationCallback {
                     override fun onChange(pointer: Long) {
-                        callback.onChange(LongPointerWrapper(pointer))
+                        callback.onChange(LongPointerWrapper(pointer, managed = false))
```
rorbech, line 247: This will leak. Maybe we need to elaborate on why it is throwing, maybe an issue with the C-API, but if we really leave it as manage, I guess we need to release manually after the callback, or alternative find some other convention. I think there is copy/clone that I would suspect just increments the reference count if we need to construct managed pointers higher up the stack.

rorbech (changes requested):
I don't really see why the callback native pointer cannot be managed. Maybe we need to elaborate with the Simon, but as an absolute minimum we should have a FIXME/issue tracking the leakage.

packages/jni-swig-stub/realm.i
nhachicha: In general, we should check for pending exceptions before making another JNI call (in our case invoking the callback)

packages/jni-swig-stub/realm.i
nhachicha: Missed to throw here 👍

rorbech (approved):

PR #148, opened 2021-03-17
Module clean up

### LINKED ISSUE #134: Clean up module structure
We should reevaluate module structure from the current knowledge of constraints and wishes.

- There is probably no need for the `runtime` module as the compiler plugin can resolve symbols from the `library` (jvm) part (though not published due to https://github.com/realm/realm-kotlin/blob/master/packages/library/build.gradle.kts#L58). 
- I also think it would be easier if we keep `cinterop` operation on basis types as `NativePointer`, etc. and push internal classes to `library` to have fewer module dependencies

rorbech:
First stab at cleaning up our module structure from our current insights. 

- Eliminate `runtime-api` module and move content to `cinterop`/`library`
- Compiler plugin now resolves symbols directly from `library`

Closes #134


rorbech:
Aligned with @nhachicha that he will try to eliminate the `@RealmModule` and the need for `Mediator` to be public in another PR.

nhachicha (approved):
LGTM if CI passes 👍

PR #155, opened 2021-04-07
Primary key support

### LINKED ISSUE #122: Add support for Primary keys

rorbech:
This adds primary key support. 

Highlights:
- Primary keys are marked with the `@PrimaryKey` annotation on the field in the model definition
- Objects needs to have primary keys when added to the realm. Either by supplying it when creating it through `create(type, primaryKey)` or by ensuring that unmanaged objects have the right value before importing it by `copyToRealm` or by assigning the object to a field of an already managed objects. 

Modeling:
```
    @PrimaryKey
    var id : String
```

Creation
```
class Realm {
   fun <T : RealmObject> create(type: KClass<T>, primaryKey: Any?): T
}
```

Internals:
```
interface RealmObjectCompanion {
   val `$realm$primaryKey`: KMutableProperty1<*, *>
}
```

TODO: 
- [x] Darwin support 

Closes #122


packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -199,6 +205,14 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
         val fields: MutableMap<String, Pair<String, IrProperty>> =
             SchemaCollector.properties.getOrDefault(irClass, mutableMapOf())
 
+        val primaryKeyFields = fields.filter { it.value.second.backingField!!.hasAnnotation( PRIMARY_KEY_ANNOTATION) }
+
+        val primaryKey = when(primaryKeyFields.size)  {
+            0 -> ""
```
nhachicha, line 211: This should be nullable. similarly `Table#primaryKey` 
```suggestion
            0 -> null
```
```Kotlin
val primaryKey: String? = null,
```

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -199,6 +205,14 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
         val fields: MutableMap<String, Pair<String, IrProperty>> =
             SchemaCollector.properties.getOrDefault(irClass, mutableMapOf())
 
+        val primaryKeyFields = fields.filter { it.value.second.backingField!!.hasAnnotation( PRIMARY_KEY_ANNOTATION) }
+
+        val primaryKey = when(primaryKeyFields.size)  {
```
nhachicha, line 210:
```suggestion
     val primaryKey: String? = when(primaryKeyFields.size) {
```

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -217,12 +231,12 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
                     // Name
                     putValueArgument(arg++, irString(irClass.name.identifier))
                     // Primary key
-                    putValueArgument(arg++, irString(""))
+                    putValueArgument(arg++, irString(primaryKey))
```
nhachicha, line 234:
```suggestion
                    putValueArgument(
                        arg++,
                        if (primaryKey != null) irString(primaryKey) else IrConstImpl.constNull(
                            startOffset,
                            endOffset,
                            pluginContext.irBuiltIns.nothingNType
                        )
                    )
```

test/src/androidTest/kotlin/io/realm/PrimaryKeyTests.kt
```diff
@@ -0,0 +1,110 @@
...
+    @AfterTest
+    fun tearDown() {
+        Utils.deleteTempDir(tmpDir)
+    }
+
+
+    @Test
+    fun string() {
+        realm.beginTransaction()
+        realm.create(PrimaryKeyString::class, primaryKey)
```
nhachicha, line 54: can we assert the value or the DB count to make sure there were no silent errors?

test/src/androidTest/kotlin/io/realm/PrimaryKeyTests.kt
```diff
@@ -0,0 +1,110 @@
...
+        realm.commitTransaction()
+        // Query
+    }
+
+    @Test
+    fun nullPrimaryKey() {
+        realm.beginTransaction()
+        realm.create(PrimaryKeyStringNullable::class, null)
```
nhachicha, line 62: ditto, also make sure using null as a duplicate is correctly detected (i.e throw uniqueness constraints)

test/src/androidTest/kotlin/io/realm/PrimaryKeyTests.kt
```diff
@@ -0,0 +1,110 @@
...
+        realm.commitTransaction()
+    }
+
+    @Test
+    fun missingPrimaryKeyThrows() {
+        realm.beginTransaction()
+        assertFailsWith<RuntimeException> {
+            realm.create(PrimaryKeyString::class)
+        }
```
nhachicha, line 71: cancel or roll back the transaction?

test/src/androidTest/kotlin/io/realm/PrimaryKeyTests.kt
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    fun duplicatePrimaryKeyThrows() {
+        realm.beginTransaction()
+        val create = realm.create(PrimaryKeyString::class, primaryKey)
+        assertFailsWith<RuntimeException> {
+            // C-API semantics is currently to return any existing object if already present
+            val create1 = realm.create(PrimaryKeyString::class, primaryKey)
+        }
+        create.primaryKey = "Y"
+        println(create.primaryKey)
```
nhachicha, line 83: WIP?

nhachicha (commented):
Nice and simple implementation 👍, some minor comments left

packages/library/src/commonMain/kotlin/io/realm/PrimaryKey.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm
+
+@Retention(AnnotationRetention.RUNTIME)
```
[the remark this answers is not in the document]
rorbech: I was hoping that we could maybe use it to derive runtime information about a specific `KProperty` actually was a primary key, but it looks like [annotations](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-annotated-element/annotations.html) are not present at runtime for Kotlin Native

packages/cinterop/src/jvmCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -22,6 +22,14 @@ package io.realm.interop
 private val INVALID_CLASS_KEY: Long by lazy { realmc.getRLM_INVALID_CLASS_KEY() }
 private val INVALID_PROPERTY_KEY: Long by lazy { realmc.getRLM_INVALID_PROPERTY_KEY() }
 
+actual class RealmValue private constructor(val _v: Any?, val _n: realm_value_t?)  {
```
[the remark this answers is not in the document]
rorbech: It would be doable to just expect/actual-ing the internal type, but we would end up needing to expose it as part of the `RealmInterop` interface, thus leaking the internal type through the API. 

Further, I think that it would be easier to handle any platform memory handling if we did the full implementation in the platform specific code. 

I tried to sketch it on native too, but it requires us to handle allocation of not only the actual struct but also the related memory, i.e. strings and binary data. As there is no need for that at the moment since it is a pure input parameter, I just ended up passing it as `Any?`. This way I could do the conversion to `realm_value_t` inside the specific `realm_object_create_with_primary_key` methods by existing mem-scoped methods for conversion.

rorbech:
@nhachicha I address all your comments. Only leftover is finding a way to move `PropertyDesciptor` out of the `library`-module.

packages/library/src/commonMain/kotlin/io/realm/internal/PropertyDescriptor.kt
```diff
@@ -0,0 +1,170 @@
...
+// Core field types with their support level
+@Suppress("LongParameterList")
+enum class CoreFieldType(
+    val type: PropertyType,
+    val nullable: Boolean,
+    val nonNullable: Boolean,
+    val listSupport: Boolean,
+    val primaryKeySupport: Boolean,
+) {
+    INT(PropertyType.RLM_PROPERTY_TYPE_INT, true, true, true, true),
```
nhachicha, line 36: using named argument makes this more readable 👓

packages/library/src/commonMain/kotlin/io/realm/internal/PropertyDescriptor.kt
```diff
@@ -0,0 +1,170 @@
...
+            }
+            if (realmFieldType.nonNullable) {
+                acc.add(ElementType(classifier, false))
+            }
+            acc
+        }
+    )
+}
+
+// Convenience variables holding collections of the various supported
```
nhachicha, line 90:
```suggestion
// Convenience variables holding collections of the various supported types
```

packages/library/src/commonMain/kotlin/io/realm/internal/PropertyDescriptor.kt
```diff
@@ -0,0 +1,170 @@
...
+data class RealmFieldType(
+    val collectionType: CollectionType,
+    val elementType: ElementType
+) {
+    val isPrimaryKeySupported: Boolean =
+        collectionType == CollectionType.RLM_COLLECTION_TYPE_NONE && elementType.realmFieldType.primaryKeySupport
+
+    // Utility method to generate Kotlin code for the specific field
+    fun toKotlinLiteral(): String {
+        val elementType = this.elementType
```
nhachicha, line 111:
```suggestion
```

packages/library/src/commonMain/kotlin/io/realm/internal/PropertyDescriptor.kt
```diff
@@ -0,0 +1,170 @@
...
+}
+
+private fun elementType(type: KType) = when (collectionType(type)) {
+    CollectionType.RLM_COLLECTION_TYPE_NONE ->
+        type
+    CollectionType.RLM_COLLECTION_TYPE_SET,
+    CollectionType.RLM_COLLECTION_TYPE_LIST ->
+        type.arguments[0].type!!
+    CollectionType.RLM_COLLECTION_TYPE_DICTIONARY ->
+        type.arguments[1].type!!
```
nhachicha, line 169: Does this mean we'll be supporting String only as keys?

packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt
```diff
@@ -243,6 +247,85 @@ class GenerationExtensionTest {
...
+                continue
+            }
+
+            val elementType = type.elementType
+            val default = if (!elementType.nullable) defaults[elementType.classifier] ?: error("unmapped default") else null
+
+            val kotlinLiteral = type.toKotlinLiteral()
+            var result = compileFromSource(
+                source = SourceFile.kotlin(
+                    "schema.kt",
```
nhachicha, line 277:
```suggestion
                    "primarykey.kt",
```

test/src/androidTest/kotlin/io/realm/PrimaryKeyTests.kt
```diff
@@ -0,0 +1,255 @@
...
+            // We could expose this through the test model definitions instead if that is better to avoid the internals
+            val realmObjectCompanion = mediator.companionOf(c)
+            realm.copyToRealm(realmObjectCompanion.`$realm$newInstance`() as RealmObject)
+            val type = realmObjectCompanion.`$realm$primaryKey`!!.rType()
+            assertTrue(types.remove(type), type.toString())
+        }
+        assertTrue(types.toTypedArray().isEmpty(), "Untested primary keys: $types")
+    }
+
+    // TODO Test all types that cannot be supported raises compilation error. Probably fits better in
```
nhachicha, line 253: WIP?

nhachicha (approved):

packages/library/src/commonMain/kotlin/io/realm/internal/PropertyDescriptor.kt
```diff
@@ -0,0 +1,177 @@
...
+    val primaryKeySupport: Boolean,
+    val indexSupport: Boolean,
+) {
+    // FIXME Get the support level right
+    INT(PropertyType.RLM_PROPERTY_TYPE_INT, true,  true, true, false, false, false, true, true),
+    BOOL(PropertyType.RLM_PROPERTY_TYPE_BOOL, true, true, true, false, false, false, false, true),
+    STRING(PropertyType.RLM_PROPERTY_TYPE_STRING, true, true, true, false, false, false, true, true),
+    OBJECT(PropertyType.RLM_PROPERTY_TYPE_OBJECT, true, false, true, false, false, false, false, false),
+    FLOAT(PropertyType.RLM_PROPERTY_TYPE_FLOAT, true, true, true, false, false, false, false, false),
+    DOUBLE(PropertyType.RLM_PROPERTY_TYPE_DOUBLE, true, true, true, false, false, false, false, false);
```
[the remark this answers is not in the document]
rorbech: These are the core types. The supported Kotlin classifiers and their mapping to core types are in https://github.com/realm/realm-kotlin/pull/155/files#diff-618c73b5fd17e77a2eb31d96620e6ba9af61ccf7bb63174f2261b2b9b757fa02R83.

packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -257,6 +257,10 @@ actual object RealmInterop {
         checkedBooleanResult(realm_wrapper.realm_commit(realm.cptr()))
     }
 
+    actual fun realm_rollback(realm: NativePointer) {
+        checkedBooleanResult(realm_wrapper.realm_rollback(realm.cptr()))
```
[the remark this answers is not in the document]
rorbech: I already changed it from something like `throwOnError` to this. It is implemented to check the C-API for errors if the value is `false` and if no error is registered just return false. I guess we wouldn't have to propagate the `true` return value when the alternative is that we will throw an exception. Think it will make most sense to adjust as part of #70.

packages/library/src/commonMain/kotlin/io/realm/internal/PropertyDescriptor.kt
rorbech: That is at least the current support level in core and hence also Realm Java. I think it arises from sync where the keys are just the JSON keys. I guess we can just adjust our descriptors when the restriction is lifted.

rorbech:
@nhachicha Finally some green light at the end of the CI tunnel. I have moved the _behavioral_ compiler plugin tests to the `test`'s `jvmTest` and cleaned up the `test` source sets with some slight modifications as a side effect. Don't know if you want to give that a glance? Otherwise I finally will merge it.

Issue #90, opened 2020-11-27
Support default schema creation


nhachicha:
Currently, we have to pass in a class annotated with `RealmModule` which captures some or all the models that we would like to participate in the Realm scheam. This works fines for modules and library but adds an extra step to open a Realm of you want the default behaviour that consist of using all the annotated `RealmObject` classes

cmelchior:
Things we need to tackle:

1) How to locate all classes implementing `RealmObject`? This should also work with incremental compilation steps.
2) How to handle that model classes might be defined in multiple modules. Relying on autogenerated code can be problematic if multiple modules write to the same location.

Some background from Realm Java

- The annotation processor there allows us to find all Realm model classes.
- We write these model classes to a `DefaultMediator` class, which was looked up using reflection at runtime.
- Libraries are required to create a class annotated with `@RealmModule`. This disables the generation of a `DefaultMediator` and also requires that people define a `RealmConfiguration.modules(...)` as a way to set the schema.

This approach means that app projects work out of the box, while library modules require extra work. So far, this seems to have been a sensible tradeoff, but it is unclear if the same would work for Multiplatform where Realm primarily is used within a library module.

nhachicha:
Also bear in mind the following compiler constraints expressed in https://github.com/realm/realm-kotlin/pull/156#issue-610868385

cmelchior:
Maybe some ideas for doing this can be found here:  https://kotlinlang.slack.com/archives/C7L3JB43G/p1634316429144100?thread_ts=1634306190.142200&cid=C7L3JB43G

elv-stav:
Having Realm automatically detect all RealmObjects in a project and use that as a default schema would be great. Any chance this can be re-visited?

PR #156, opened 2021-04-07
Schema definition simplification

nhachicha:
# TL;DR

## Before

```Kotlin 
@RealmModule(A::class, B::class, C::class)
class Entities

val configuration = RealmConfiguration.Builder()
    .schema(Entities())
    .build()
```

## Now 

```Kotlin 
val configuration = RealmConfiguration(schema = listOf(A::class, B::class, C::class))
```

- In this PR we simplified the steps to define a schema, by getting rid of the `RealmModule` as well as the `RealmConfiguration.Builder`.
- `schema` parameter accepts a list of class literal (similar to annotation arguments)

The technique used to perform this is to replace the constructor call in `RealmConfiguration` with an internal constructor which has a map of companion objects necessary to build the `Mediator`

### Default Schema

The ability to automatically have a default schema (similar to Realm-Java) will not be supported here.

Reasoning: 
The current `SchemaCollection` logic is broken at least for the test scenarios since the compiler plugin runs in two different steps: 
the first step is compiling the model classes in the `Task :compileKotlinMacos`, where all the classes are added to the `SchemaCollector`, the second step is compiling the test classes (which includes `RealmConfiguration` construct substitution for instances)  
```
> Task :compileTestKotlinMacos
w: duplicate library name: io.realm.kotlin:cinterop-cinterop-realm_wrapper
i: [Realm Compiler Plugin] 2021-03-27T15:50:30.870Z Collected schema is: {}
```
This is why relying on the current `SchemaCollector` to define the default module is not feasible, since we can not collect reliably all participating classes, even without this multiple steps compilation, incremental compilation will break the `SchemaCollector` logic since the last run will only see the newly added models. I think we should abandon this magic behaviour for now and behave similarly to other libraries like Dagger where you specify the components and modules explicitly using class literals.


cmelchior:
> val configuration = RealmConfiguration(schema = listOf(A::class, B::class, C::class))

If the schema is required, it makes a lot of sense to move to a constructor parameter. I'm not sure we should remove the Builder class though. At least that wasn't the decision we came to when @edualonso investigated the alternatives: https://docs.google.com/document/d/1mM5wZuaXDvHDl-ozh9VGG3VtDbzCJUHOGA1rUD3Z4YQ/edit ... The problem is that discovery + validation gets a lot harder once we start adding support for the more advanced methods currently available in Java.

This definitely seems a step forward, but IMO it still makes the onboarding experience harder than it should be. I.e. the possibility of just doing `val realm = Realm.open()` is extremely compelling for onboarding new users.

Worst case, I think even doing runtime reflection to be able to do this automatically would be preferred to forcing users to manually specify the schema, but I don't have any great ideas on how to achieve this currently.

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,52 +17,54 @@
...
+        RealmInterop.realm_config_set_schema_mode(
+            nativeConfig,
+            SchemaMode.RLM_SCHEMA_MODE_AUTOMATIC
+        )
+        RealmInterop.realm_config_set_schema_version(nativeConfig, version = 0)
+        val schema = RealmInterop.realm_schema_new(mapOfKClassWithCompanion.values.map { it.`$realm$schema`() })
         RealmInterop.realm_config_set_schema(nativeConfig, schema)
-    }
 
-    data class Builder(
```
cmelchior, line 43: See my comment in the PR, but I don't think we should remove the Builder completely

packages/library/src/commonMain/kotlin/io/realm/internal/Mediator.kt
```diff
@@ -16,14 +16,10 @@
...
 import kotlin.reflect.KClass
 
-// FIXME https://github.com/realm/realm-kotlin/issues/90 support default schema creation.
-interface Mediator { // avoid reflection, implemented and defined by compiler plugin for each `@RealmModule`
-    val companionMapping: Map<KClass<*>, RealmObjectCompanion>
-    fun newInstance(clazz: KClass<*>): RealmModelInternal
-    // This is constant so could just be computed directly in IR or at least just be a default method like
-    // fun schema(): List<Table> { return companionMapping.values.map { it.`$realm$schema`() } }
-    fun schema(): List<Table>
+interface Mediator {
```
cmelchior, line 22: Maybe rename this to `SchemaMediator`? Right now the name is very generic and it is hard to guess what the responsibility is for this class.

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,52 +17,54 @@
...
-class RealmConfiguration private constructor(
-    val path: String?, // Full path if we don't want to use the default location
-    val name: String?, // Optional Realm name (default is 'default')
-    val schema: Mediator,
-    val version: Long = 0,
-    val tables: List<Table> = listOf()
+class RealmConfiguration(
+    val path: String? = null, // Full path if we don't want to use the default location
+    val name: String = "default", // Optional Realm name (default is 'default')
+    val schema: List<KClass<out RealmObject>> // classes literal (T::class) //TODO rename ot model?
```
cmelchior, line 30: Wouldn't it be more correct to make this a `Set`?

rorbech (commented):
I agree with Christian that there is no obvious reason for abandoning the builder. We could just do the exact same lowering trick from the builder method, by accepting the schema as
```
fun schema(vararg ... KClass<T: RealmObject>)
```
and build the map as part of instantiating the mediator from this before passing it to the configuration. 

Complete elimination of the need to specify a schema is tracked in #90. I think it would still be possible to achieve something like the pattern for Java, but it requires us to generate files while compiling. We would also then maybe still need the `RealmModule` as an explicit mediator for libraries.

packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt
```diff
@@ -72,6 +83,54 @@ class GenerationExtensionTest {
         }
     }
 
+    @Test
+    fun `RealmConfiguration Schema Argument Lowering`() {
+        val inputs = Files("/schema")
+        val result = compile(inputs)
+        assertEquals(KotlinCompilation.ExitCode.OK, result.exitCode)
```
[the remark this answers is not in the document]
nhachicha: This is just checking the IR output. The real tests are done in `SchemaTests`

test/src/commonTest/kotlin/io/realm/SchemaTests.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    @Test
+    fun usingSingleClassAsPositional() {
+        // Using a single class causes a different input IR to transform (argument not passed as vararg)
+        val conf = RealmConfiguration("name", "path", listOf(Sample::class))
+        assertValidCompanionMap(conf, Sample::class)
+    }
+
+    @Suppress("invisible_member")
+    private fun assertValidCompanionMap(conf: RealmConfiguration, vararg schema: KClass<out RealmObject>) {
```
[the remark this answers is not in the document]
nhachicha: Accessing various elements using reflection in the compiler tests are not optimal, I'm trying to keep compiler test small and limited to IR transformation and push the rest as much towards the integration `test` project

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,52 +17,54 @@
...
-import io.realm.interop.Table
+import kotlin.reflect.KClass
 
-class RealmConfiguration private constructor(
-    val path: String?, // Full path if we don't want to use the default location
-    val name: String?, // Optional Realm name (default is 'default')
-    val schema: Mediator,
-    val version: Long = 0,
-    val tables: List<Table> = listOf()
+class RealmConfiguration(
```
cmelchior, line 27: As discussed. I think it would be a mistake to remove the builder as the builder has better discoverability and autocomplete in the IDE. Required parameters like schema do need to be constructor parameters though and it might make sense to have a shorthand to create common configurations directly, but not sure if having multiple ways to construct the configuration is too confusing?

But IMO having these two options, seems like a good tradeoff to me:

```
// Construct "common" configurations directly
RealmConfiguration(val name = "default.realm",  val schema: Set<KClass<out RealmObject>>)

// Anything else more complicated requires you to go through the builder
RealmConfiguration.Builder(val schema: Set<KClass<out RealmObject>>)
class Builder {
  fun name(...)
  fun directory(...)
  fun otherMethods(...)
}
```

Thoughts?


```

**Comment by nhachicha** [comment]
@rorbech @cmelchior I reintroduced the Builder pattern which accepts a vararg for schema

File: packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt (hunk 4 above)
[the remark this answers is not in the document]
Comment (rorbech) [reply packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt:90]: Just for completion. I see your point about not polluting with weird introspection/reflection and maybe only doing explicit tests here if there it is necessary for explicit testing while development. I sometimes find it difficult to get right without and really tedious to do without immediate debugging support when applying the plugin on `test`.  So I guess it is a matter of balancing it.

File: packages/plugin-compiler/src/test/resources/schema/input/Schema.kt [hunk 7]
```diff
@@ -0,0 +1,30 @@
...
+
+class A : RealmObject
+class C : RealmObject
+class B : RealmObject
+
+val conf1 = RealmConfiguration.Builder()
+    .schema(A::class, B::class, C::class)
+    .path("/some/path").build()
+
+val conf2 = RealmConfiguration(schema = setOf(A::class, C::class))
```
rorbech, line 30:
```suggestion
val conf2 = RealmConfiguration(schema = setOf(A::class, C::class))

```

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
rorbech: I like the concept of having a simple option and a full blown one. I am a bit in favor of the `schema(...)` method instead of required arguments to the constructor, as that is more fluent when there are many classes involved. But I do get the that it is in fact required until we potentially could find a way to gather a default schema behind the scenes, so don't know if it is too loose.

rorbech (approved):
Nice. I didn't get to a conclusion on my experiments about using the direct reference to the companion objects. Also not sure I follow all the IR details of the gathering of classes. But think it should do it for now. Most importantly we eliminated the `RealmModule` 🎉

PR #158, opened 2021-04-12
Move external sources to packages folder

rorbech:
This moves the `external` folder to `packages` as it is really annoying to navigate to the sources outside of the actual project

### DIFF
--- a/packages/cinterop/build.gradle.kts
+++ b/packages/cinterop/build.gradle.kts
@@ -38,7 +38,9 @@ repositories {
 //  Currently just adding the common darwin parts to macos-target.
 val idea = System.getProperty("idea.active") == "true"
 
-val corePath = "../external/core"
+val corePath = "external/core"
+val relativeCorePath = "../$corePath"
+val cinteropCorePath = "$rootDir/$corePath"
 
 android {
     compileSdkVersion(Versions.Android.compileSdkVersion)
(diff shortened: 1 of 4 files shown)


packages/cinterop/build.gradle.kts
```diff
@@ -85,23 +87,23 @@ android {
...
+    "-include-binary", "$rootDir/external/core/capi_macos_universal/src/realm/parser/librealm-parser-dbg.a",
+    "-include-binary", "$rootDir/external/core/capi_macos_universal/src/realm/object-store/librealm-object-store-dbg.a"
 )
 // TODO remove the debug suffix (-dbg) when switching to use release
 val nativeLibraryIncludesIosArm64 = listOf(
-    "-include-binary", "$rootDir/../external/core/build-capi_ios_Arm64/lib/librealm-ffi-static-dbg.a",
-    "-include-binary", "$rootDir/../external/core/build-capi_ios_Arm64/lib/librealm-dbg.a",
-    "-include-binary", "$rootDir/../external/core/build-capi_ios_Arm64/lib/librealm-parser-dbg.a",
-    "-include-binary", "$rootDir/../external/core/build-capi_ios_Arm64/lib/librealm-object-store-dbg.a"
+    "-include-binary", "$rootDir/external/core/build-capi_ios_Arm64/lib/librealm-ffi-static-dbg.a",
```
cmelchior, line 97: Any reason to not use the the `$corePath` variable here? so `$rootDir/$corePath`?

cmelchior (approved):
If CI is happy 🚀

packages/cinterop/build.gradle.kts
rorbech: The cinterop paths has to be absolute so they have to be handled differently. I have included another variable for the cinterop specific prefix.

PR #159, opened 2021-04-13
Upgrade to latest Android Gradle Plugin

rorbech:


cmelchior (approved):

packages/cinterop/build.gradle.kts
```diff
@@ -70,8 +70,7 @@ android {
     }
     defaultConfig {
         ndk {
-            // FIXME MPP-BUILD Extend supported platforms. Currently using local C API build and CMakeLists.txt only targeting x86_64
-            abiFilters("x86_64", "arm64-v8a")
+            abiFilters += setOf("x86_64", "arm64-v8a")
```
nhachicha, line 73: appending to what? we want exactly those ABI to be supported

nhachicha (approved):

packages/cinterop/build.gradle.kts
rorbech: I would assume that it is empty 🤷 . The `abiFilters` is a value, thus not updatable, so we have to append to the existing set. I can `clear` it up front if you are afraid that it holds something already?

nhachicha: Ah ok it should be fine then 👍

PR #162, opened 2021-04-16
Dispatch queue scheduler

rorbech:
Adding a dispatch queue scheduler to allow notifications on background threads (backed by a dispatch queue) on Darwin as well. 

The dispatcher assumes that:
- The queue is only backed by a single thread
- That `set_notify_callback` is only called once and on the same thread as opening the realm and prior to any call to `is_on_thread`

Will try to get guarantees for the above.

NOTE: It is still required to open the and register callbacks on the same thread.


rorbech:
Superseded by #272 as the dispatch queues aren't guaranteed to run on a single thread.

PR #165, opened 2021-04-20
update user guide

nhachicha:
- Update Doc
- Added reified method for `objects` query
- Remove listener for a single object (to avoid similar issue we had in Realm-Java) the single object value should be frozen anyway so it doesn't make sense to register an observer for it
- Now listener callback returns the `RealmResults<T>`


cmelchior (approved):

rorbech (approved):
I see the point about eliminating the `RealmObject.observe` as we eventually will get frozen objects instead. But I guess it is hard to achieve the same without it for now 🤔 

Otherwise only minor comments - mostly setting the context explicitly for being in a KMM project

PR #171, opened 2021-04-21
Logging infrastructure

### LINKED ISSUE #117: Logger infrastructure
We need to have some logging infrastructure. We should have some default implementation that uses Android logger, NSLog, std out for java, etc.

cmelchior:
Closes #117 

This PR addresses a number of issues:

* I added infrastructure to enable logging in our own components with a public API so users can configure it. The API has changed from how it exposed in Realm Java. This was mostly due to threading restrictions in Kotlin Native, which basically makes global singletons with mutable state impossible. The logger is now configured pr. Realm and needs to be accessed through the Realm instance. This will also prepare the API for how Core will expose a unified logging API. 
* I added best effort log implementations for Android, Darwin and JVM. 
* A `LogLevel` enum has been added to the public API.
* A`RealmLogger` interface has been added to the public API.
* In order to support configuring the logs, I had to refactor the `RealmConfiguration` to become more extensible. So constructors have moved around. KDoc has been added to some methods and I added test for the `RealmConfiguration` in general.
* I moved away from `var args` for the schema in `RealmConfiguration.Builder(...)` as it made it impossible to create named arguments in the builder constructor. But with the collection factory methods in Kotlin (`setOf(...)`), this should not be a huge burden on developers.

```
// Usage
class MyLogger(override val tag: String = "REALM") {
   override log(....)
}

val config = RealmConfiguration.Builder()
  .log(level = LogLevel.ERROR, customLoggers = listOf(MyLogger()))
  .build()

// These options are accessible through a new `LogConfiguration` sub class that can be accessed in the `RealmConfiguration
val config = RealmConfiguration(schema = setOf(Person::class))
val otherConfig = RealmConfiguration.Builder(schema = setOf(Person::class)).build()
val yetMoreConfig = RealmConfiguration.Builder().schema(setOf(Person::class)).build()
val yetYetMoreConfig = RealmConfiguration.Builder().schema(Person::class).build() // Only place var args are still used

val logConfig = config.log
logConfig.level
logConfig.loggers
```

Usage from within our components:

```
val realm = Realm.open(config)
realm.log.error("Message")
```

TODO
- [x] It looks like something goes wrong with the compiler plugin when transforming `RealmConfiguration.Builder().build()` need to figure out what.
- [ ] Unclear how to handle logging outside the Realm lifecycle. E.g. when loading native code or before any Realms are created.


cmelchior:
Ready for review. Sorry that this PR exploded a bit in complexity. I should probably have realized this sooner and pulled the locks and RealmConfiguration refactor into its own PR's.

test/src/commonTest/kotlin/io/realm/LogTests.kt
```diff
@@ -0,0 +1,174 @@
...
+        val args: Array<out Any?> = arrayOf("foo")
+        log.error(throwable, message, *args)
+        assertEquals(LogLevel.ERROR, customLogger.logLevel)
+        assertEquals(throwable, customLogger.throwable)
+        assertEquals(message, customLogger.message)
+        assertTrue(customLogger.args.contentEquals(args))
+    }
+
+    @Test
+    fun smallLogEntry() {
```
rorbech, line 64: Is some part of the verification missing, or is it just to ensure that the codepaths works? If so, then maybe note it.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmLog.kt
```diff
@@ -0,0 +1,87 @@
+package io.realm.internal
+
+import io.realm.LogConfiguration
+import io.realm.log.LogLevel
+import io.realm.log.RealmLogger
+
+/**
+ * Logger class used by Realm components. One logger is created for each Realm instance.
+ */
+// FIXME Only public because of testing. Using @Suppress("invisible_members") is not enough.
```
rorbech, line 10: I think it is `invisible_member`. There is also a similar `invisible_reference`, which is probably what you would need here. ... or both

packages/library/src/commonMain/kotlin/io/realm/internal/RealmLog.kt
```diff
@@ -0,0 +1,87 @@
...
+public class RealmLog(val tag: String = "REALM", val configuration: LogConfiguration) {
+
+    @Suppress("JoinDeclarationAndAssignment")
+    private val logLevel: LogLevel
+    private val loggers: MutableList<RealmLogger> = mutableListOf()
+
+    init {
+        logLevel = configuration.level
+        if (!configuration.removeSystemLogger) {
+            loggers.add(TypeFactory.createDefaultSystemLogger(tag))
```
rorbech, line 20: Would it make sense to put the system logger first?

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -22,53 +22,143 @@ import io.realm.internal.RealmObjectCompanion
...
-        internal fun build(companionMap: Map<KClass<*>, RealmObjectCompanion>): RealmConfiguration {
-            return RealmConfiguration(path, name, companionMap)
+        // Called from the compiler plugin
+        internal fun build(companionMap: Map<KClass<out RealmObject>, RealmObjectCompanion>): RealmConfiguration {
+            return RealmConfiguration(
+                companionMap,
+                path,
+                name,
+                schema,
+                LogConfiguration(logLevel, removeSystemLogger, customLoggers)
```
rorbech, line 155: I would probably just have initialized the set of loggers with the system logger dependent on the configuration instead of propagating the knowledge of that concept around.

rorbech (approved):
Nice. Good little _"Good first Issue"_ in Kotlin multiplatform 😅 

Had some comments here and there and I would prefer if @nhachicha took a glance at the updates to the compiler plugin updates around the realm configuration construction. Otherwise, LGTM.

packages/library/build.gradle.kts
```diff
@@ -125,7 +125,7 @@ kotlin {
 
 kotlin {
     sourceSets {
-        create("nativeCommon") {
+        create("darwinCommon") {
```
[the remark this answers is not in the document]
cmelchior: Using that tree makes sense to me, so I assume you want `nativeDarwin` here?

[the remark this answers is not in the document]
rorbech: Yep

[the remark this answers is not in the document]
cmelchior: Turns out this might be a rabbithole, so created https://github.com/realm/realm-kotlin/issues/175 to track this instead.

test/src/commonTest/kotlin/io/realm/LogTests.kt
cmelchior: There is a class comment here https://github.com/realm/realm-kotlin/pull/171/files#diff-8e3f6edc9620b29f9c42eeb34c2bdd2be307144396d4d0e55175ca585c34e9c7R13 

But I can replicate it to all the tests (about 5) if you think that would be better?

test/src/commonTest/kotlin/io/realm/RealmConfigurationTests.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Test
+    fun pathOverrideName() {
+        val realmPath = "<HowToGetPlatformPath>/custom.realm"
+        val realmName = "my.realm"
+
+        val config = RealmConfiguration(realmPath, realmName, setOf(Sample::class))
+        assertEquals(realmPath, config.path)
+        // Correct assert: assertEquals("custom.realm", config.name)
```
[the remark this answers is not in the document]
cmelchior: Yes, it should probably just be fixed, but was't sure if there was hidden complexity involved as involves interacting with platform paths... I'll create and issue and FIXME

cmelchior:
Updated based on PR feedback at the meeting

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,58 +17,154 @@
...
+         * Removes the default system logger from being installed. If no custom loggers have
+         * been configured, not log events will be reported, regardless of the configured
+         * log level.
+         *
+         * @see [RealmConfiguration.Builder.log]
+         */
+        internal fun removeSystemLogger() = apply { this.removeSystemLogger = true }
 
         fun build(): RealmConfiguration {
             @Suppress("SpreadOperator")
```
rorbech, line 145: Not needed anymore, but maybe at https://github.com/realm/realm-kotlin/pull/171/files#diff-2ee4dd35764b2be7198c9da2591563685604ccc9a3580256919303340bdabeb7R118

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,58 +17,154 @@
...
+     * @param name name of the Realm file being created if no [RealmConfiguration.path] is configured. Realm files are
+     *             placed in the default location for the platform. On Android this is in `getFilesDir()`
+     * @param schema set of classes that make up the schema for the Realm. Identified by their class literal `T::class`.
+     */
+    // This constructor is never used at runtime, all calls to it are being rewired by the Realm Compiler Plugin to call
+    // the other secondary instructor with all schema classes mapped to their RealmCompanion.
+    public constructor(path: String? = null, name: String = Realm.DEFAULT_FILE_NAME, schema: Set<KClass<out RealmObject>>) :
+        this(path, name, mapOf())
 
+    internal constructor(path: String? = null, name: String = Realm.DEFAULT_FILE_NAME, schema: Map<KClass<out RealmObject>, RealmObjectCompanion>) :
```
nhachicha, line 92:
```suggestion
    // called by the compiler plugin, with a populated companion map
    internal constructor(path: String? = null, name: String = Realm.DEFAULT_FILE_NAME, schema: Map<KClass<out RealmObject>, RealmObjectCompanion>) :
```

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
```diff
@@ -73,13 +73,17 @@ class RealmSchemaLoweringExtension : IrGenerationExtension {
...
+                        // There exist 3 constructors in RealmConfiguration
+                        // 1) Primary private constructor accepting all builder fields.
+                        // 2) Public secondary constructor accepting a schema as a Set.
+                        // 3) Internal secondary constructor accept the map of <KClass, RealmCompanion>.
+                        // In this step we are finding calls to constructor(2) and transform it into calls
+                        // to constructor(3).
+                        val internalConstructor: IrConstructorSymbol =
                             pluginContext.lookupConstructorInClass(REALM_CONFIGURATION) {
-                                !it.owner.isPrimary
+                                !it.owner.isPrimary && (it.owner.valueParameters[2].type.classFqName?.toString().equals("kotlin.collections.Map"))
```
nhachicha, line 86:
```suggestion
                                !it.owner.isPrimary
                                        && it.valueParameters.size == 3        
                                        && (it.owner.valueParameters[2].type.classFqName == KOTLIN_COLLECTIONS_MAP)
```
Don't forget to add `import io.realm.compiler.FqNames.KOTLIN_COLLECTIONS_MAP` at the begining

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,58 +17,154 @@
...
         }
 
-        // Called from compiler plugin
-        internal fun build(companionMap: Map<KClass<*>, RealmObjectCompanion>): RealmConfiguration {
-            return RealmConfiguration(path, name, companionMap)
+        // Called from the compiler plugin
+        internal fun build(companionMap: Map<KClass<out RealmObject>, RealmObjectCompanion>): RealmConfiguration {
+            val allLoggers = mutableListOf<RealmLogger>()
+            if (!removeSystemLogger) {
+                allLoggers.add(PlatformHelper.createDefaultSystemLogger(Realm.DEFAULT_LOG_TAG))
```
rorbech, line 153: Still not the first ;)

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,58 +17,154 @@
...
+         * log level.
+         *
+         * @see [RealmConfiguration.Builder.log]
+         */
+        internal fun removeSystemLogger() = apply { this.removeSystemLogger = true }
 
         fun build(): RealmConfiguration {
             @Suppress("SpreadOperator")
-            return RealmConfiguration(path, name, setOf(*schema))
+            return RealmConfiguration(path, name, schema)
```
rorbech, line 146: Is this actually replaced by the compiler plugin? Maybe we should define a 
```
@kotlin.internal.InlineOnly
public inline fun REPLACED_BY_IR(s: String): Nothing = throw NotImplementedError(s)
```
to signal this with target and implementation reference
```
REPLACED_BY_IR("Collect companions and invoke build(map) instead [...]")
```

test/src/commonTest/kotlin/io/realm/RealmConfigurationTests.kt
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    fun customLoggers() {
+        val logger = TestLogger()
+        val config: RealmConfiguration = RealmConfiguration.Builder(schema = setOf(Sample::class))
+            .log(customLoggers = listOf(logger))
+            .build()
+        assertEquals(2, config.log.loggers.size)
+        assertEquals(logger, config.log.loggers.last())
+    }
+}
```
rorbech, line 103: I guess you could keep the test of `removeSystemLogger` even though it is internal.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
```diff
@@ -224,6 +144,96 @@ class RealmSchemaLoweringExtension : IrGenerationExtension {
...
 
+/**
+ * Locate the list of schema class literals. This currently support them being defined in-place as var args, or in place
+ * using some of the factory collection methods like setOf and arrayOf, but we don't follow the code if defined further
+ * away than that.
+ *
+ * E.g. all of these are valid
+ * ```
+ * RealmConfiguration(schema = setOf(MyType::class))
+ * RealmConfiguration.Builder(schema = arrayOf(MyType::class)).build()
```
nhachicha, line 155: this is no longer true since you removed the varargs from the `RealmConfiguration` public ctor ...
only `RealmConfiguration(schema = setOf(Sample::class))` mode is supported

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
```diff
@@ -224,6 +144,96 @@ class RealmSchemaLoweringExtension : IrGenerationExtension {
     }
 }
 
+/**
+ * Locate the list of schema class literals. This currently support them being defined in-place as var args, or in place
+ * using some of the factory collection methods like setOf and arrayOf, but we don't follow the code if defined further
```
nhachicha, line 149: arrayOf is not supported

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
```diff
@@ -224,6 +144,96 @@ class RealmSchemaLoweringExtension : IrGenerationExtension {
...
+ * RealmConfiguration(schema = setOf(MyType::class))
+ * RealmConfiguration.Builder(schema = arrayOf(MyType::class)).build()
+ * RealmConfiguration(path, name, MyType::class, MyOtherType::class)
+ * ```
+ * While these are not
+ * ```
+ * val classes = setOf(MyType::class)
+ * RealmConfiguration(schema = classes)
+ *
+ * TODO We should lift this restriction
```
nhachicha, line 163: 🍿

rorbech (approved):
Minor comments, but otherwise LGTM.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
```diff
@@ -246,31 +256,25 @@ private fun populateCompanion(
     pluginContext: IrPluginContext
 ): IrCallImpl {
     if (specifiedModels.isEmpty()) {
-        logError("Schema argument must be a list of class literal (T::class)")
+        logError(
+            "No schema was provided. It must be defined as a set of class literals (MyType::class) either through " +
+                "RealmConfiguration(schema = setOf(...)), RealmConfiguration.Builder(schema = setOf(...).build(), " +
+                "or RealmConfiguration.Builder().schema(...).build()."
```
nhachicha, line 262: varargs is not supported

test/src/commonTest/kotlin/io/realm/LogTests.kt
nhachicha: why not doing something like https://github.com/realm/realm-kotlin/pull/171/files#diff-8e3f6edc9620b29f9c42eeb34c2bdd2be307144396d4d0e55175ca585c34e9c7R33 (i.e using a custom logger that logs to a StringBuffer or a file then make sure the content matches what you've logged?)

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,58 +17,154 @@
...
 import kotlin.reflect.KClass
 
-class RealmConfiguration(
-    val path: String? = null, // Full path if we don't want to use the default location
-    val name: String = "default", // Optional Realm name (default is 'default')
-    schema: Set<KClass<out RealmObject>> // classes literal (T::class)
+/**
+ * Configuration for log events created by a Realm instance.
+ */
+public data class LogConfiguration(
```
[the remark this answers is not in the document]
cmelchior: When we enable Explicit API mode, it is going to be required, so I thought I just wanted to add it sooner rather than later: https://github.com/realm/realm-kotlin/issues/22

But if it confusing I can remove them again?

packages/library/build.gradle.kts
```diff
@@ -95,6 +95,10 @@ android {
             proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
         }
     }
+
+    dependencies {
+        implementation("androidx.startup:startup-runtime:1.0.0")
```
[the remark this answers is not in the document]
cmelchior: Yes, we should 👍

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -17,58 +17,154 @@
...
         fun name(name: String) = apply { this.name = name }
-        fun schema(vararg classes: KClass<out RealmObject>) = apply { this.schema = classes }
+        fun schema(classes: Set<KClass<out RealmObject>>) = apply { this.schema = classes }
+        fun schema(vararg classes: KClass<out RealmObject>) = apply { this.schema = setOf(*classes) }
+
+        /**
+         * Configure how Realm will report log events.
+         *
+         * @param level all events at this level or higher will be reported.
+         * @param customLoggers any custom loggers to send log events to. A default system logger is
```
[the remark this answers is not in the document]
cmelchior: I was thinking about that, but that would mean that each logger must implement a `logLevel >= myLogLevel` check , which seems redundant for 90% of use cases. So I decided to move that under our control to reduce what you had to think about when implementing a `RealmLogger`. But if you think it has value I can move it to be part each logger instead?

packages/library/src/darwinCommon/kotlin/io.realm.internal/NSLogLogger.kt
```diff
@@ -0,0 +1,92 @@
...
+        var formattedMessage = ""
+        val regEx = "%[\\d|.]*[sdf]|[%]".toRegex()
+        val singleFormats = regEx.findAll(message).map {
+            it.groupValues.first()
+        }.asSequence().toList()
+        val newStrings = message.split(regEx)
+        for (i in 0 until args.count()) {
+            val arg = args[i]
+            formattedMessage += when (arg) {
+                is Double -> {
```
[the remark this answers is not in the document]
cmelchior: Yes, it fallsthough to format as Strings

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
cmelchior: Yup, in the public API right now, but this method still supports it if we re-add it.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
cmelchior: I added an override in the builder where it is supported (as the only place).

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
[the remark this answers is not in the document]
nhachicha: Ah forgot about this one, let's keep it

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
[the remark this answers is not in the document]
nhachicha: That is fine 👍

nhachicha (approved):

PR #174, opened 2021-04-26
Introduce MutableRealm

cmelchior:
This PR is the first step towards achieving the frozen architecture described in https://docs.google.com/document/d/1bGfjbKLD6DSBpTiVwyorSBcMqkUQWedAmmS_VAhL8QU/edit

In splits the Realm API into two classes: `Realm` and `MutableRealm` with a shared `BaseRealm` super class. All data is still live, but with this PR the API is now better prepared when we start freezing it.  

- `Realm` now only exposes methods to query data.

- `Realm` exposes two methods `suspend fun write {}` (which will be implemented in a future PR) and `writeBlocking { }` which mirrors the old `executeTransaction()` behavior from Realm Java. I'm not 100% sure if we want `writeBlocking` in the final API, but for now, it works as a good bridge while we move to a more suspendable API. It also means that we are not forced to deal with coroutines everywhere just yet (especially in tests).

- `MutableRealm` works as a wrapper around a write transaction with a lifecycle controlled by us. This also means we can scope all the create/delete methods in the correct context.

- I moved to use `write {}` and `cancelWrite {}` instead of `beginTransaction()` and `cancelTransaction()`. Javascript already does something similar and read/write transactions don't seem like concepts we need to shove into people's faces. 

**Implementation note:**

- Right now we have `BaseRealm -> Realm / MutableRealm` with query methods on `BaseRealm`. I suspect once we add support for DynamicRealms that we will introduce a 2nd layer so it becomes `BaseRealm -> TypedRealm -> Realm` and `BaseRealm -> UntypedRealm -> DynamicRealm`, but there is no reason to introduce this additional layer just yet.

**Review notes:**

- `Realm`, `MutableRealm` and `BaseRealm` are the most relevant classes to review. The rest just changes the test to use the new API.


packages/library/src/commonMain/kotlin/io/realm/RealmModel.kt
```diff
@@ -21,5 +21,5 @@ package io.realm
 //  https://github.com/realm/realm-kotlin/issues/83
 
 fun RealmObject.delete() {
-    Realm.delete(this)
+    MutableRealm.delete(this)
```
rorbech, line 24: Maybe remove this as it is leaking mutability outside the `write/writeBlocking`-scope. If the method was on `RealmObject` itself I guess you could scope by a `MutableRealm`-receiver, so it is only available on `RealmObject` inside the write function.

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * This class represents the writeable state of a Realm file. The only way to modify data in this file is through
+ * instances of this class. These are provided and managed automatically through either [Realm.write] or
+ * [Realm.writeBlocking].
+ */
+class MutableRealm : BaseRealm {
+
+    companion object {
+        // FIXME EVALUATE Should this be on RealmModel instead?
+        fun <T : RealmObject> delete(obj: T) {
```
rorbech, line 33: This is calleable from an arbitrary context. See comment on the `RealmObject.delete` extension method.

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -0,0 +1,81 @@
...
+    internal val log: RealmLog = RealmLog(configuration = configuration.log)
+
+    init {
+        log.info("Realm opened: ${configuration.path}")
+    }
+
+    fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T> {
+        return RealmResults(
+            dbPointer,
+            @Suppress("SpreadOperator") // TODO PERFORMANCE Spread operator triggers detekt
```
rorbech, line 47:
```suggestion
```

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -0,0 +1,81 @@
...
+    // null checks everywhere, when it is rarely needed.
+    private var isClosed: Boolean = false
+    internal val log: RealmLog = RealmLog(configuration = configuration.log)
+
+    init {
+        log.info("Realm opened: ${configuration.path}")
+    }
+
+    fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T> {
+        return RealmResults(
```
rorbech, line 45: Should we also `checkClosed` here?

rorbech (approved):
Nice. 

It is a bit unclean whether the full scheme for realms are laid out (which instances are available when and where). If it is, shouldn't we start enforcing `checkClosed` on (almost) all calls and consider synchronization of the state?

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -48,82 +34,68 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
...
-    fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T> {
-        return RealmResults(
-            dbPointer!!,
-            @Suppress("SpreadOperator") // TODO PERFORMANCE Spread operator triggers detekt
-            { RealmInterop.realm_query_parse(dbPointer!!, clazz.simpleName!!, "TRUEPREDICATE") },
-            clazz,
-            configuration.mediator
-        )
+    @Suppress("TooGenericExceptionCaught")
+    fun <R> writeBlocking(function: MutableRealm.() -> R): R {
```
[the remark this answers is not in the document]
cmelchior: As in `writeBlock` instead? I was going with the pattern set in the Kotlin API like `runBlocking { }` ... IMO that consistency is worth more than saving a few characters especially since the IDE autocomplete it. But I can see it being annoying to read if you do this a lot... I also had the assumption that in real-world apps `realm.write { }` is going to be the predominant use case, so optimizing this name was not worthwhile.  

Thoughts?

packages/library/src/commonMain/kotlin/io/realm/RealmModel.kt
cmelchior: Yes, we should probably just move this to an instance method on MutableRealm 👍

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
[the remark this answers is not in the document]
rorbech: Ahh. Sorry. I was just referring to the argument. `function`->`block`. 

`writeBlocking` is fine as, I agree that, `write` would be the predominant use case.

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -0,0 +1,81 @@
...
+    internal inline fun checkClosed() {
+        if (isClosed) {
+            throw IllegalStateException("Realm has been closed and is no longer accessible: ${configuration.path}")
+        }
+    }
+
+    // Not all sub classes of `BaseRealm` can be closed by users.
+    internal open fun close() {
+        checkClosed()
+        RealmInterop.realm_close(dbPointer)
```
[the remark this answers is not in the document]
cmelchior: Right now when the Realm is life, this should be safe enough, but once we move to a frozen architecture, I suspect it might be required yes. I created https://github.com/realm/realm-kotlin/issues/180 to track it.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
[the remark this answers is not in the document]
cmelchior: Ah okay, I don't have any strong preference there, as long as we use the same word consistently. `block` seems fine to me 👍

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -48,82 +34,68 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
...
+     */
+    suspend fun <R> write(function: MutableRealm.() -> R): R {
+        TODO("Awaiting implementation of the Frozen Architecture")
     }
-    // Convenience inline method for the above to skip KClass argument
-    inline fun <reified T : RealmObject> create(): T { return create(T::class) }
 
     /**
-     * Creates a copy of an object in the Realm.
+     * NOTE Avoid calling this method on the UI thread, instead use [Realm.write].
```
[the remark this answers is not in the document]
cmelchior: Yeah, that is probably a bit premature. I'll remove.

packages/library/src/commonMain/kotlin/io/realm/RealmModel.kt
cmelchior: Also created https://github.com/realm/realm-kotlin/issues/181

cmelchior:
Only thing breaking CI is the sample app...not sure why it breaks as it seems to pull from OJO. Will need to investigate.

PR #182, opened 2021-04-27
Migrate to Maven Central

cmelchior:
Enable releasing snapshots and releases on Maven Central


cmelchior:
Ready for review. The changes in this PR has already been used to publish SNAPSHOT's and 0.1.0 to Maven Central. Not without hickups, but it seems to work.

Unlike Realm Java, we can actually run a fully automated release as all packages are part of the same root project.

The release process is the same as Realm Java, i.e. update changelog, verify version and tag the branch. This is described here: https://github.com/realm/realm-wiki/wiki/Kotlin-Release-Checklist

Note, Dokka docs are not uploaded to the MongoDB website yet. I'm in a dialog with the docs team about where to place them.

build.gradle.kts
```diff
@@ -21,50 +21,161 @@ buildscript {
     }
 }
 
+// Find property in either System environment or Gradle properties.
+// If set in both places, Gradle properties win.
+fun getPropertyValue(propertyName: String, throwIfNotFound: Boolean = false): String {
+    var value: String? = System.getenv(propertyName)
```
rorbech, line 27: Good old 
```
val value: String? = project.findProperty(propertyName) ?: System.getenv(propertyName)
```

build.gradle.kts
```diff
@@ -21,50 +21,161 @@ buildscript {
...
+                "a Gradle property.")
+    }
+    return value ?: ""
+}
+
+// Cache dir for build artifacts that should be stored on S3
+val releaseMetaDataDir = File("${buildDir}/outputs/s3")
+releaseMetaDataDir.mkdirs()
+
+fun readAndCacheVersion(): String {
```
rorbech, line 43: We could also consider to move our own version into top-level `version.txt` and read that from Config.kt. After all we know where to find that and does need the same level of IDE assistance to navigate, etc. as for all the other dependency version, etc. 🤷

rorbech (approved):
A bit hard to follow all the various tiny fractions of this, but I guess it works since there is a release at Maven central. Have you tried to verify that it is consumable? 

Otherwise only minor comments.

CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
+## 0.1.0 (2021-05-07)
+
+This is the first public Alpha release of the Realm Kotlin SDK for Android and Kotlin Multiplatform.
+
+A minimal setup for including this library looks like this:
+
+```
+// Top-level build.gradle file
+buildscript {
```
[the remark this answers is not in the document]
cmelchior: I'm looking into the marker artifact now as part of the Gradle Plugin Portal work. But it pushes the marker artifact to an entirely new location `/realm-kotlin/....` instead of `/io/realm/kotlin/...` so I need to get a hold of Maven Central support first so we can push artifacts to that location.

build.gradle.kts
cmelchior: Yes, I have been thinking the same thing. But right now it is a bit annoying to do as the Config.kt file is built in isolation before anything else, so it is a bit difficult to make it more advanced than just listing constants.

But I do agree it will probably make some common edits easier. I'll create an issue tracking it.

PR #186, opened 2021-05-03
Add VersionId and RealmConfiguration.maxNumberOfActiveVersions

cmelchior:
With a frozen archicture, knowledge about the version is going to be required by our internal code and useful to users for debugging.

So this PR adds explicit support for getting the Core VersionID for all relevant Realm classes: Realm, RealmObject and RealmResults (for now).

It also exposes the `RealmConfiguration.maxNumberOfActiveVersions()` as we know it from Realm Java as well as a `Realm.getActiveVersions()` method. These methods are useful for controlling and debugging how many versions are currently used.

I made a few design choices in this PR, but input on those would be nice

* Added a top-level `VersionId` class instead of `Realm.VersionId` as in my POC. Primary reason is that we will end up with a number of "Realm"-classes like `Realm`, `MutableRealm`, `DynamicRealm` and `MutableDynamicRealm`, so it felt weird to put it in one of those, even if `Realm` is our primary entry point.  Putting it in `BaseRealm` also felt awkward as that is mostly an implementation detail and not really a concept users need to be cognisant about. 

* `VersionId` is the name used in Core, but I have mixed feelings about the name (what does ID mean?). `Version` feels a bit too generic though. Other suggestions could be `DataVersion` (because we also have schema version) or `RealmVersion`, but `realm.realmVersion` is just meh. `TransactionVersion` is probably more accurate, but also very "technical". Input is welcome.

* Right now I'm exposing `Realm.version`, `RealmResults.version` and `RealmObject.version`. The first two feel pretty good, but I have reservations about using such a generic term as `version()` on the RealmObject interface. Mostly because there is a high chance it will conflict with user properties or methods. Right now it is an extension method, so it is safe, but if we move it to a default implementation in the interface, it could spell problems. Alternatives could be to make the method open so people can override it, or choose a different name, but then we should probably make the name consistent.

TODO:

- [x] Add missing tests for getting version on Object/RealmResults
- [x] How to return uint64_t from SWIG
- [x] Add tests for `Realm.getActiveVersions()`.
- [x] Depend on https://github.com/realm/realm-core/pull/4652 being merged.


packages/jni-swig-stub/realm.i
```diff
@@ -153,6 +153,10 @@ struct realm_size_t {
 // bool output parameter
 %apply bool* OUTPUT { bool* out_found };
 
+// uint64_t output parameter
+// FIXME How to convert uint64_t to Long or ULong?
```
cmelchior, line 157: @rorbech I could use some help with figuring this out...the SWIG syntax is a bit magical to me 🙈

rorbech: You can tell Swig to copy a typemap from an existing type to another by ex:
```
%apply int64_t* OUTPUT { uint64_t* out_long };
```
Which will copy the typemap from an "output int64_t pointer" to all `unint64_t*` variables named `out_long`. 

`ULong` is an experimental add-on in Kotlin, so we cannot use that in JNI code.

test/src/commonTest/kotlin/io/realm/VersionIdTests.kt
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+package io.realm
+
+import kotlin.test.Test
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+@ExperimentalUnsignedTypes
```
rorbech, line 23:
```suggestion
```

rorbech (approved):
LGTM. Only a few minor comments.

rorbech (commented):
I guess the `commonTest`s should follow the convention from #187

packages/library/src/commonMain/kotlin/io/realm/RealmModel.kt
```diff
@@ -16,10 +16,25 @@
...
+
+var RealmObject.version: VersionId
+    get() {
+        val internalObject = this as RealmModelInternal
+        internalObject.`$realm$Pointer`?.let {
+            val (version, index) = RealmInterop.realm_get_version_id(it)
+            return VersionId(version, index)
+        } ?: throw IllegalArgumentException("Cannot get version from an unmanaged object.")
+    }
+    private set(_) {
```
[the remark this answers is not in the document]
cmelchior: Apparently `val` requires a backing field (I tried), this was the only way I could make the Kotlin compiler happy... (but I might be missing some certain way of doing it)

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -32,6 +32,16 @@ public abstract class BaseRealm internal constructor(
...
 ) {
 
+    /**
+     * The current data version of this Realm and data fetched from it.
+     */
+    public var version: VersionId = VersionId(0, 0)
+        get() {
+            checkClosed()
+            val (version: Long, index: Long) = RealmInterop.realm_get_version_id(dbPointer)
+            return VersionId(version, index)
```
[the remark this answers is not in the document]
cmelchior: God point 👍

cmelchior:
> I guess the commonTests should follow the convention from #187

Yes, I was just awaiting that PR to be merged before changing it here.

packages/library/src/commonMain/kotlin/io/realm/RealmModel.kt
```diff
@@ -16,10 +16,25 @@
...
+
 // FIXME API Currently just adding these as extension methods as putting them directly into
 //  RealmModel would break compiler plugin. Reiterate along with
 //  https://github.com/realm/realm-kotlin/issues/83
 
 fun RealmObject.delete() {
     MutableRealm.delete(this)
 }
+
+var RealmObject.version: VersionId
```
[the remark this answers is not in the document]
cmelchior: So I tried doing this as an extension variable `var RealmObject.version`. This seems to be a pretty bad situation because if the RealmObject itself has a version field, you have no way of accessing the extension property nor give it another name. The sample problem applies to functions though. 

I would suggest keeping it as is for now, and then create a ticket for going back to look at it. Ideally, we should move all the extension functions into the interface declaration itself, which would solve most of this (I hope).

If we, for whatever reason cannot do interface implementations, we should probably strongly consider renaming it to `realmVersion` or similar. 

What do you think?

cmelchior:
The PR should be ready for final review @rorbech

test/src/androidTest/kotlin/io/realm/shared/MutableRealmTests.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    @AfterTest
+    fun tearDown() {
+        if (!realm.isClosed()) {
+            realm.close()
+        }
+        PlatformUtils.deleteTempDir(tmpDir)
+    }
+
+    @Test
```
rorbech, line 49: For completion we could maybe add a minimal test of actually committing is no errors occurs 🤪 

It is tested elsewhere, but since only a four-liner I guess it could be here. I guess it somehow relates to test organization: If we could organize it in a way so that it is easily identifiable if core concepts like this are in place before digging into why more complex use cases fail.

test/src/androidTest/kotlin/io/realm/shared/RealmObjectTests.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class RealmObjectTests {
+
+    companion object {
+        // Expected version after writing Parent to Realm
+        private val EXPECTED_VERSION = VersionId(3, 2)
```
rorbech, line 35: #magicalnumber. Just out of curiosity: Is there any straight forward way to reason why this is the right number?

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
```diff
@@ -30,12 +31,19 @@ import kotlin.reflect.KClass
...
 //    be raised. Maybe we can get an option to prevalidate queries in the C-API?
 class RealmResults<T : RealmObject> constructor(
+    private val realmConfiguration: RealmConfiguration,
     private val realm: NativePointer,
     private val queryPointer: () -> NativePointer,
     private val clazz: KClass<T>,
     private val mediator: Mediator
 ) : AbstractList<T>(), Queryable<T> {
 
+    public var version: VersionId = VersionId(0, 0)
```
rorbech, line 41: Should we place this in a `Versioned` interface, maybe even along with the db pointer? 🤔 

Maybe questionable when naming of the property might interfere on RealmObject (#219), but we could maybe consider it when we have more insight into the outline of the different realm-pointers/versions flying around.

test/src/androidTest/kotlin/io/realm/shared/RealmTests.kt
```diff
@@ -0,0 +1,120 @@
...
+import kotlin.test.BeforeTest
+import kotlin.test.Ignore
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class RealmTests {
+
+    companion object {
+        // Initial version of any new typed Realm (due to schema being written)
```
rorbech, line 34: Ahh. Here was the description of the magical number.

test/src/androidTest/kotlin/io/realm/shared/RealmTests.kt
```diff
@@ -0,0 +1,120 @@
...
+    fun numberOfActiveVersions() {
+        assertEquals(2, realm.getNumberOfActiveVersions())
+        realm.writeBlocking {
+            assertEquals(2, realm.getNumberOfActiveVersions())
+        }
+        assertEquals(2, realm.getNumberOfActiveVersions())
+    }
+
+    @Test
+    @Ignore // FIXME This fails on MacOS only. Are versions cleaned up more aggressively there?
```
rorbech, line 101: I don't have an immediate explanation. Think it is ok to ignore for now to get on and revisit later.

rorbech (approved):

packages/library/src/commonMain/kotlin/io/realm/RealmModel.kt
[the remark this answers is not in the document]
rorbech: Just to clarify: The comment in the above FIXME about interface methods breaking the compiler plugin is not meaning that we cannot have them as interface methods with default implementation. We would just have to add all the interface methods as fake overrides on all user defined objects. It can probably be done in a generic way, so we don't have to do too many adjustments in the IR generation when adding new methods. But I guess the discussion is postponed to #219

packages/cinterop/src/jvmCommon/CMakeLists.txt
```diff
@@ -1,54 +1,16 @@
...
-#link_directories("${CAPI_BUILD}/src/realm/sync")         # realm-sync
-link_directories("${CAPI_BUILD}/src/realm")              # realm-object-store
-link_directories("${CAPI_BUILD}/src/realm/object-store/c_api") # librealm-ffi-dbg.a
-link_directories("${CAPI_BUILD}/openssl/lib/")           # libcrypto.a libssl.a
+# Build Realm Core
+# Set option flags for Core.
+# See https://github.com/realm/realm-core/blob/master/CMakeLists.txt#L174 for the full list.
+set(REALM_ENABLE_SYNC OFF)
+set(REALM_BUILD_LIB_ONLY ON)
+add_subdirectory("realm-core" EXCLUDE_FROM_ALL)
```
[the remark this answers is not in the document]
cmelchior: `add_subdirectory` are required to be a subdir unless you also specify the binary output dir for the CMake project and since Core has quite a lot of output dirs I wasn't sure it was possible. So this was the quickest way around it.

But we probably have to go back for a 2nd when we want to build on Windows.

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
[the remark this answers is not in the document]
cmelchior: As long as the RealmResults wrap a Live Realm this version can change, but as soon as we switch to frozen results we can turn this into a `val`.

test/src/androidTest/kotlin/io/realm/shared/MutableRealmTests.kt
cmelchior: Yeah, I'll add it :D

test/src/androidTest/kotlin/io/realm/shared/RealmObjectTests.kt
cmelchior: The Realm Version is `(2,1)` when we open a normal Realm (due to us writing the schema), so it becomes version 3 once we write an object. The `index` is some internal accounting used by Core (at least that was the explanation I got).

I have been considering if we should just hide the index since it isn't used when comparing versions anyway, but I wasn't sure if it would be relevant for debugging. 

Do you have any strong opinions either way?

cmelchior: It would mean it couldn't be a data class, but that wouldn't be a big issue

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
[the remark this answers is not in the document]
rorbech: `val` just means that it is _read-only_, not that it has the same value each time you retrieve it...at least according to my understanding.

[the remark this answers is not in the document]
cmelchior: It is the same as `final` in Java which means we couldn't replace the version once it is defined.

cmelchior: Hmm, interesting idea. I think it is a bit premature given that we still don't know exactly how this looks on RealmObject nor how we pass Realms around yet, but I added it to https://github.com/realm/realm-kotlin/issues/219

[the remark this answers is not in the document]
rorbech: Anyway; reading up on https://kotlinlang.org/docs/coding-conventions.html#functions-vs-properties again we should probably consider changing it (back) to a function. Maybe just leaving it to be part of #219.

test/src/androidTest/kotlin/io/realm/shared/RealmObjectTests.kt
cmelchior: Taking to Finn, the `index` is an index into an internal ringbuffer in Core, so it doesn't have any practical use. I'll remove it.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -24,7 +24,9 @@ inline class ColumnKey(val key: Long)
 @Suppress("FunctionNaming", "LongParameterList")
 expect object RealmInterop {
 
+    fun realm_get_version_id(realm: NativePointer): Long
```
rorbech, line 27: I guess we could now go for an inline class. They even seem to be stable in 1.5, but I guess we could just address it when updating. There are also other key types that could be converted to improve type safety.

rorbech (approved):
LGTM. Very nice clean up of the `CMakeList.txt`. Would we also to able to consume prebuilt binaries that way now?

cmelchior:
> Would we also to able to consume prebuilt binaries that way now?

In theory yes, but prebuilt libraries prevent us from using LTO (as far as I am aware), but it would probably be useful for debug builds. I don't know exactly how much work it would be though.

PR #259, opened 2021-05-21
Add Gradle Plugin Portal and Maven Central badges

cmelchior:
Add badges to the README for where to find latest versions. See https://github.com/cmelchior/realm-kotlin/blob/master/README.md for how it will look.


nhachicha (approved):
Can we use a place holder instead of the outdated version in `0.0.1-SNAPSHOT` and get rid of `oss-snapshot-local` repo in the [Setup](https://github.com/cmelchior/realm-kotlin/blob/master/README.md#setup ) section

cmelchior:
> Can we use a place holder instead of the outdated version in 0.0.1-SNAPSHOT and get rid of oss-snapshot-local repo in the Setup section

Will look into updating the changelog in another PR. Not sure if we can inject dynamic text, but it would be ideal.

PR #260, opened 2021-05-25
RealmList - schema creation

edualonso:
- Added tentative API for RealmList.
- Added logic to generate schema for RealmList fields in synthetic properties generation.
- Updated sample classes.


packages/plugin-compiler/src/main/kotlin/io/realm/compiler/IrUtils.kt
```diff
@@ -150,9 +150,42 @@ internal fun IrPluginContext.lookupConstructorInClass(
...
+    RLM_PROPERTY_TYPE_INT,
+    RLM_PROPERTY_TYPE_BOOL,
+    RLM_PROPERTY_TYPE_STRING,
+    RLM_PROPERTY_TYPE_OBJECT,
+    RLM_PROPERTY_TYPE_FLOAT,
+    RLM_PROPERTY_TYPE_DOUBLE
+}
+
+data class SchemaProperty(
+    val type: String,
```
rorbech, line 181: We could consider to just use the declaration type instead of `type` or maybe extract the _element type_ into this `type` to mimic cores way of representing a property by a _type_ and a (potentially NONE) collection type. 

But I guess exposing the public scheme will probably lead to some clean up of this.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -335,23 +356,39 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
...
                                         arg++,
                                         if (type == objectType) {
-                                            irString(backingField.type.classifierOrFail.descriptor.name.identifier)
-                                        } else
+                                            when (collectionTypeSymbol) {
+                                                PROPERTY_COLLECTION_TYPE_NONE ->
+                                                    irString(backingField.type.classifierOrFail.descriptor.name.identifier)
+                                                PROPERTY_COLLECTION_TYPE_LIST ->
+                                                    irString(backingField.symbol.descriptor.type.arguments[0].toString())
+                                                else -> error("Invalid collection type: $collectionTypeSymbol")
```
rorbech, line 387: Same. Maybe improve the message a bit.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -335,23 +356,39 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
                                             symbol = type.symbol
                                         )
                                     )
-                                    // Collection type
+                                    // Collection type: remember to specify it correctly here - the
+                                    // type of the contents itself is specified as "type" above!
+                                    val collectionTypeSymbol = when (value.collectionType) {
+                                        CollectionType.NONE -> PROPERTY_COLLECTION_TYPE_NONE
+                                        CollectionType.LIST -> PROPERTY_COLLECTION_TYPE_LIST
+                                        CollectionType.SET -> error("Sets not ready yet.")
```
rorbech, line 364: Maybe more expressive `Unsupported collection type for field ...` to lead the way for fixing thing when you get this error.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -335,23 +356,39 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
...
                                     putValueArgument(
                                         arg++,
                                         if (type == objectType) {
-                                            irString(backingField.type.classifierOrFail.descriptor.name.identifier)
-                                        } else
+                                            when (collectionTypeSymbol) {
+                                                PROPERTY_COLLECTION_TYPE_NONE ->
+                                                    irString(backingField.type.classifierOrFail.descriptor.name.identifier)
+                                                PROPERTY_COLLECTION_TYPE_LIST ->
+                                                    irString(backingField.symbol.descriptor.type.arguments[0].toString())
```
rorbech, line 386: I am not sure what this covers. Are lists somehow expressed with a _link_?

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -310,4 +404,42 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
+            }
+        }
+
+        // Otherwise just return the generic(s) present in the declaration
+        return declaration.symbol.descriptor.type.arguments.map {
+            it.toString()
+        }
+    }
+
+    private fun getCoreTypeFromKotlinType(type: String): String {
```
rorbech, line 429: I guess we should consider looking up the actual symbols once and map directly to a compile time element instead of passing these strings around and retrieving the enumeration element every time we are generating the property.

packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+package io.realm
+
+import io.realm.interop.NativePointer
+
+/**
+ * TODO
+ */
+class RealmList<E> : MutableList<E>, AbstractMutableCollection<E> {
```
rorbech, line 24: Just for completion: I somehow feel that this is a bit verbose, but of discussion I guess we cannot use delegation from inheritance (as it only works for interfaces and you would eliminate the option of extending `AbstractMutableCollection`) and we want to do the differentiation between managed and unmanaged internally, as having two public types pollutes the public space.

rorbech (approved):
Most notably we should maybe consider only looking up enums in compilation once. I guess we could postpone this to a clean up sweep while exposing the schema publicly, but if doing so please add a comment linking to the issue. 

Otherwise only minor comment and LGTM.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
edualonso: The previous logic would add the link parameter from `backingField.type.classifierOrFail.descriptor.name.identifier` which in the case of a `RealmList<Object>` field, would return `RealmList`. That is wrong, it should be `Object`.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/IrUtils.kt
edualonso: I will add a fixme and point to the schema issue as agreed.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
edualonso: This has changed a bit in the upcoming PRs so I will add a note to address it in one of them.

PR #265, opened 2021-05-25
Add Freeze and Thaw methods to the internal API

cmelchior:
Frozen notifications require that we can freeze and thaw the various Realm types. This PR adds that.

### DIFF
--- a/test/src/macosTest/kotlin/io/realm/InstrumentedTests.kt
+++ b/test/src/macosTest/kotlin/io/realm/InstrumentedTests.kt
@@ -55,12 +55,12 @@ class InstrumentedTests {
             // Accessing getters/setters
             realmModel.`$realm$IsManaged` = true
             realmModel.`$realm$ObjectPointer` = CPointerWrapper(ptr1.ptr)
-            realmModel.`$realm$Owner` = CPointerWrapper(ptr2.ptr)
+            realmModel.`$realm$Pointer` = CPointerWrapper(ptr2.ptr)
             realmModel.`$realm$TableName` = "Sample"
 
             assertEquals(true, realmModel.`$realm$IsManaged`)
             assertEquals(ptr1.rawPtr.toLong(), (realmModel.`$realm$ObjectPointer` as CPointerWrapper).ptr.toLong())
-            assertEquals(ptr2.rawPtr.toLong(), (realmModel.`$realm$Owner` as CPointerWrapper).ptr.toLong())
+            assertEquals(ptr2.rawPtr.toLong(), (realmModel.`$realm$Pointer` as CPointerWrapper).ptr.toLong())
             assertEquals("Sample", realmModel.`$realm$TableName`)
         }
     }


packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectUtil.kt
```diff
@@ -56,15 +59,37 @@ fun RealmObjectInternal.unmanage() {
...
 
-// Create a frozen copy of this object. Expected DB pointer is used by writes where we need to freeze
-// the object before the owner Realm is updated, but the pointer it will be updated with is already know.
-fun <T : RealmObject> RealmObjectInternal.freeze(realm: NativePointer, expectedDbPointer: NativePointer? = null): T {
+/**
+ * Creates a frozen copy of this object.
+ *
+ * @param frozenRealm Pointer to frozen Realm to which the frozen copy should belong.
+ */
+internal fun <T : RealmObject> RealmObjectInternal.freeze(frozenRealm: NativePointer): T {
```
rorbech, line 67: Shouldn't this be `T.freeze(...)`? But I guess you will then have to make an unchecked cast internally. Maybe we should include the type in `RealmObjectInternal<T>`?

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
```diff
@@ -30,29 +30,70 @@ import kotlin.reflect.KClass
...
+        QUERY, // RealmResults created from a query
+        RESULTS // RealmResults wrapping a Realm Core Results.
+    }
+
+    // Create Results from query
+    private constructor(realm: BaseRealm, query: () -> NativePointer, clazz: KClass<T>, schema: Mediator) {
+        this.mode = Mode.QUERY
+        this.owner = realm
+        this.realm = realm.dbPointer
+        this.query = query()
```
rorbech, line 55: If we are always evaluating this we should maybe just supply the native pointer from outside directly. The only need for this is, as far as I remember, to get the `result` pointer so maybe we could skip the `query` and only track the `result`. Might also impact when the majority of the load is performed, but probably have to have some impact from core team to reassess. 

Might be that moving it around and thawing it need the query!? 🤔

cmelchior: I assume the old behavior was to lazy evaluate the query as late as possible, but with this architecture, it probably doesn't make sense and we can just evaluate it directly.

But yes, there is probably room for optimization here.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectUtil.kt
cmelchior: Playing around with the types, we end up being forced to do unchecked casts somewhere since the type system doesn't know that `RealmObjectInternal` is always implemented by a `RealmObject`. So for now it seems simpler to just keep the types as they are?

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
cmelchior: Hmm, thinking about this. We probably need to be a bit careful how this is used. Evaluating the query itself is fairly cheap I believe since it just validates the query itself, but e.g. running a `find_all` is definitely not and is something we would want to move to the background thread if at all possible.

cmelchior: Turns out that our `realm_query_find_all` is lazy. So it doesn't actually run the query until you interact with it. This seems to be the semantics we want, but there isn't a need to store the query pointer itself inside our `RealmResults`...we just need to wrap the underlying Results class which takes care of the rest.

cmelchior:
Ready for review @rorbech @nhachicha

rorbech (approved):

PR #266, opened 2021-05-25
Frozen notifications

cmelchior:
This PR adds the building blocks in place to enable notifications on frozen objects:

- Adding a `RealmId` wrapper around the `Pair<Realm,NativePointer>`. I'll move this to a separate PR and we need to discuss the abstraction
- Added the ability to add changelisteners to frozen realm types by moving them to a notifier thread, thaw them and register the change listener there. This also means that all notifications are triggered from one central Realm.
- Added Flow API's to Realm, RealmResults and RealmObject
- Renamed old observe method to `addChangeListener()`

TODO:
- [ ] Split RealId into its own PR.
- [x] Split helper methods into own pr
- [x] Split frozen methods into own pr
- [ ] Unit tests
- [ ] Cleanup
- [ ] Verify that corner cases around cancelation and closing Realms work in a sensible way.

PR #267, opened 2021-05-26
Add MutableRealm.isInTransaction() and RealmObject.isValid()

cmelchior:
Adding a few helper methods used by the #266. Just splitting it up into multiple PR's.

### DIFF
--- a/examples/kmm-sample/settings.gradle.kts
+++ b/examples/kmm-sample/settings.gradle.kts
@@ -12,6 +12,7 @@ pluginManagement {
         gradlePluginPortal()
         google()
         mavenCentral()
+        mavenLocal()
     }
     resolutionStrategy {
         eachPlugin {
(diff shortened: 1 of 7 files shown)


examples/kmm-sample/settings.gradle.kts
```diff
@@ -12,6 +12,7 @@ pluginManagement {
         gradlePluginPortal()
         google()
         mavenCentral()
+        mavenLocal()
```
cmelchior, line 15: Needed to support building examples locally.

rorbech: If you wan't to build example locally you can just re-insert the 
```
includeBuild("../../packages")
```
In line 3!? Or are there other uses cases for doing local builds?

rorbech (approved):

PR #268, opened 2021-05-27
Introduce TransactionId instead of passing around dbPointer

cmelchior:
Needed for frozen notifications, but we need to discuss if this abstraction is the best way to do this and the naming of it.


cmelchior:
This needs to be totally reworked. Passing around the Realm instance might accidentally freeze it on Kotlin Native.

cmelchior:
Replaced by https://github.com/realm/realm-kotlin/pull/285

PR #269, opened 2021-05-27
RealmList - accessors: getter

edualonso:
Added getter, `size`, `get`, `add` and `clear` methods to list API.


packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -369,6 +374,42 @@ actual object RealmInterop {
...
+                    list.cptr(),
+                    realm_list_size(list).toULong(),
+                    to_realm_value(value).ptr
+                )
+            )
+        }
+    }
+
+    actual fun realm_list_clear(list: NativePointer) {
+        realm_wrapper.realm_list_clear(list.cptr())
```
rorbech, line 410: If the C-API call can fail we should `CheckedBooledResult`!?

packages/jni-swig-stub/realm.i
```diff
@@ -145,10 +146,13 @@ struct realm_size_t {
     size_t value;
 };
 %}
+
```
rorbech, line 149: You can use something like this
```
// size_t output parameter
// The below struct is used to pass size_t output parameters to Java. 
%inline %{
struct realm_size_t {
    size_t value;
};
%}
// The below typemaps are used to convert the realm_size_t into a pointer to the same struct in JNI.
// The type maps are only applied to arguments are named exactly: 'out_count' 
%typemap(jstype) (size_t* out_count) "realm_size_t"                      // Java type
%typemap(javain) (size_t* out_count) "realm_size_t.getCPtr($javainput)"  // Type conversion to JNI interface type
%typemap(jtype) (size_t* out_count) "long"                               // JNI interface type
%typemap(jni) (size_t* out_count) "long"                                 // JNI type
```

packages/jni-swig-stub/realm.i
```diff
@@ -145,10 +146,13 @@ struct realm_size_t {
...
+// TODO - add comment Claus will write
 %typemap(jni) (size_t* out_count) "long"
 %typemap(jtype) (size_t* out_count) "long"
 %typemap(jstype) (size_t* out_count) "realm_size_t"
 %typemap(javain) (size_t* out_count) "realm_size_t.getCPtr($javainput)"
+%apply size_t* out_count { size_t* out_size };
```
rorbech, line 155: Maybe add a comment like 
```
// Apply the above type map for `size_t* out_count` to `size_t* out_size` too
```

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -270,8 +278,8 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
                         //      return backing_field
                         //  }
 
-                        // TODO: use correct getter
-                        modifyAccessor(declaration, getValue, setValue)
+                        // TODO: setter
```
rorbech, line 281: Do we allow updating the list itself?

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -61,6 +64,16 @@ object RealmObjectHelper {
         return null
     }
 
+    fun getList(obj: RealmObjectInternal, col: String): RealmList<Any?> {
```
rorbech, line 67: Maybe add a one line comment on the type issues if you had issues making a generic variant. I know that it is listed in the class comment, but maybe also make it explicit here, so we know where we could improve if we find a way.

rorbech (approved):
Some minor comments/questions. Otherwise LGTM.

packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
edualonso: Yeah, forgot about this one.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
edualonso: We talked about creating a setter that calls `clear` on the list and then `add`s whatever items the source list may have, if that's what you mean? In any case, it isn't really urgent to implement, that's why I wrote the TODO.

PR #270, opened 2021-05-27
RealmList - nullability

edualonso:
Added nullability constraints for RealmLists:
```
// Int represents a Realm primitive type

RealmList<Int>           // allowed
RealmList<Int?>          // allowed
RealmList<RealmObject>   // allowed

RealmList<Int>?          // not allowed
RealmList<RealmObject>?  // not allowed
RealmList<RealmObject?>  // not allowed

```


packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -430,36 +430,53 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
+    private fun getListGenericCoreType(declaration: IrProperty): CoreType {
         // Check first if the generic is a subclass of RealmObject
-        val superTypes = declaration.symbol.descriptor.type.arguments[0].type.constructor.supertypes
-        for (superType: KotlinType in superTypes) {
+        val descriptorType = declaration.symbol.descriptor.type
+        val listGenericType = descriptorType.arguments[0].type
+        for (superType in listGenericType.constructor.supertypes) {
             if (superType.toString() == "RealmObject") {
-                return listOf("RealmObject")
+                // Nullable objects are not supported
```
nhachicha, line 439: Do we have the same constraint on Realm-Java? I know it's a Core limitation but I was under the impression it was still possible ... not sure when this constraint was added ... cc @cmelchior

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -430,36 +430,53 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
+                )
             }
         }
 
-        // Otherwise just return the generic(s) present in the declaration
-        return declaration.symbol.descriptor.type.arguments.map {
-            it.toString()
+        // If not a RealmObject, check whether the list itself is nullable - if so, throw error
+        if (descriptorType.isNullable()) {
+            error("Error in field ${declaration.name} - a RealmList field cannot be marked as nullable.")
```
nhachicha, line 452: I assume we have tests for these use cases?

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -430,36 +430,53 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
+                    "Char" -> PropertyType.RLM_PROPERTY_TYPE_INT
+                    "Short" -> PropertyType.RLM_PROPERTY_TYPE_INT
+                    "Int" -> PropertyType.RLM_PROPERTY_TYPE_INT
+                    "Long" -> PropertyType.RLM_PROPERTY_TYPE_INT
+                    "Boolean" -> PropertyType.RLM_PROPERTY_TYPE_BOOL
+                    "Float" -> PropertyType.RLM_PROPERTY_TYPE_FLOAT
+                    "Double" -> PropertyType.RLM_PROPERTY_TYPE_DOUBLE
+                    "String" -> PropertyType.RLM_PROPERTY_TYPE_STRING
+                    "RealmObject" -> PropertyType.RLM_PROPERTY_TYPE_OBJECT
+                    else -> error("Wrong Kotlin type: '$type'")
```
nhachicha, line 478: more of (_Unsupported Kotlin type_)

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -307,8 +307,14 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
...
                                 val backingField = property.backingField
-                                    ?: error("Property without backing field or type")
-                                val nullable = backingField.type.isNullable()
+                                    ?: error("Property without backing field or type.")
+                                // Nullability applies to the generic type in collections
+                                val nullable = if (value.collectionType == CollectionType.NONE) {
+                                    backingField.type.isNullable()
+                                } else {
+                                    value.genericTypes?.get(0)?.nullable
+                                        ?: error("Missing generic type while processing a collection field.")
```
nhachicha, line 316: These are good candidates for compiler tests (similar with nullability use cases)

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
```diff
@@ -50,17 +50,16 @@ class Sample : RealmObject {
...
-//    var nullableByteListField: RealmList<Byte?> = RealmList()
-//    var nullableCharListField: RealmList<Char?> = RealmList()
-//    var nullableShortListField: RealmList<Short?> = RealmList()
-//    var nullableIntListField: RealmList<Int?> = RealmList()
-//    var nullableLongListField: RealmList<Long?> = RealmList()
-//    var nullableBooleanListField: RealmList<Boolean?> = RealmList()
-//    var nullableFloatListField: RealmList<Float?> = RealmList()
-//    var nullableDoubleListField: RealmList<Double?> = RealmList()
-//    var nullableObjectListField: RealmList<Sample?> = RealmList()
+    // Nullable list types - RealmList<RealmObject?> is not supported
```
nhachicha, line 53: We could have a separate model (generated a-la `unsupported schema argument` test) with this defined to assert the compilation will fail

nhachicha (approved):
Small nitpick around testing 👍

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
cmelchior: Yes, RealmLists that contain RealmObjects are not allowed to contain `null` values. I actually do not remember if RealmLists of embedded objects are allowed.

nhachicha: @finnschiermer is there an issue tracking the above? to keep consistency with other supported types, I think `RealmList` should allow nullable `RealmObject` ?

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
edualonso: I will address this in https://github.com/realm/realm-kotlin/issues/290

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
edualonso: I will address this in https://github.com/realm/realm-kotlin/issues/290

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
edualonso: So far I've only tested it manually, but I'm currently working on how to write compiler tests for RealmLists. The only place where we test compiler stuff now is `GenerationExtensionTest` and the feeling I got is that everything would look all over the place if I put list-based tests in that file too. I created https://github.com/realm/realm-kotlin/issues/290 to address this.

rorbech: There are also compiler tests in `test/jvm/kotlin/io/realm/test/compiler`.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
finnschiermer: There is no such issue, the "anomaly" in how lists handle (or rather does not handle) null references is by design. I have been led to believe over time that some users rely on (or even want) this behavior. I don't. Feel free to open such an issue to start a discussion.

nhachicha: @finnschiermer Java lists (and collection API) supports adding `null` value(s). Since our goal is to be as close as possible to the platform's language we should support this use case. This also needs to be unified across other data structures like Set and Map.
Another argument is persisting an unmanaged RealmList (which contains already nullable elements coming from JSON for instance) into a managed one, since List is ordered the persisted RealmLIst should reflect the original order/elements

edualonso: I've created https://github.com/realm/realm-kotlin/issues/292 to follow up on this.

PR #271, opened 2021-06-07
Add a new sample app

nhachicha:
Demo app using Realm-Kotlin Multiplatform SDK
                    
🎨 UI: using Jetpack compose 
 ---- Shared ---
📡 Network: using Ktor & Kotlinx.serialization
💾 Persistence: using Realm Database

[Screenshots](https://github.com/realm/realm-kotlin/tree/nh/bookshelf_sample/examples/Bookshelf/)


examples/Bookshelf/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Tue Jun 01 17:22:02 BST 2021
+distributionBase=GRADLE_USER_HOME
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
```
rorbech, line 3: Would be nice if we could align the version of gradle throughout all the projects. Every version will trigger its own daemon. But I guess only option is sym-linking the file, so maybe better off to leave them. Just have in mind that we are running many different versions as part of CI build, etc.

examples/Bookshelf/androidApp/src/main/java/io/realm/sample/bookshelf/android/ui/BookshelfViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+class BookshelfViewModel : ViewModel() {
+    private var sdk = BookshelfSDK()
+
+    val savedBooks: StateFlow<List<Book>> = sdk.allBooksAsFlowable()
```
rorbech, line 33: Would it make sense to try and have the state flow in shared code? Or is that pushing to much Android-style to iOS?

examples/Bookshelf/shared/src/commonMain/kotlin/io/realm/sample/bookshelf/BookshelfSDK.kt
```diff
@@ -0,0 +1,61 @@
...
+ */
+package io.realm.sample.bookshelf
+
+import io.realm.Cancellable
+import io.realm.sample.bookshelf.database.RealmDatabase
+import io.realm.sample.bookshelf.model.Book
+import io.realm.sample.bookshelf.network.OpenLibraryApi
+import kotlinx.coroutines.flow.Flow
+
+class BookshelfSDK {
```
rorbech, line 24: I would probably leave the `SDK` out of the name.

examples/Bookshelf/androidApp/src/main/java/io/realm/sample/bookshelf/android/ui/AboutScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+            .padding(8.dp)
+    ) {
+        Column(Modifier.clickable {
+            val browserIntent = Intent(Intent.ACTION_VIEW, Uri.parse("https://www.github.com/realm/realm-kotlin"))
+            context.startActivity(browserIntent)
+        }) {
+            Text(
+                modifier = Modifier.padding(4.dp),
+                text = """
+Demo app using Realm-Kotlin Multiplatform SDK
```
rorbech, line 47: You can leave the indentation to follow the code structure and `trimIndent()` which makes it more readable.

rorbech (commented):
Awesome 🏅

As discussed offline we should probably move this somewhere else. I left some minor comments. 

Maybe @edualonso should also give it a look.

examples/Bookshelf/androidApp/src/main/java/io/realm/sample/bookshelf/android/ui/BookshelfViewModel.kt
edualonso: StateFlow isn't Android-specific and can be used across different platforms. I think it would be a good idea to move to the BookshelfSDK/repository class - which is what John O'Reilly is doing here: https://johnoreilly.dev/posts/state-flow-multiplatform/

examples/Bookshelf/shared/src/commonMain/kotlin/io/realm/sample/bookshelf/BookshelfSDK.kt
edualonso: `BookshelfRepository`?

edualonso:
I only have one comment and I don't think it's a great deal so 👍 from me

edualonso (approved):

examples/Bookshelf/androidApp/build.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    implementation(project(":shared"))
+    implementation("com.google.android.material:material:1.3.0")
+    implementation("androidx.appcompat:appcompat:1.2.0")
+    implementation("androidx.constraintlayout:constraintlayout:2.0.4")
+    implementation("androidx.compose.compiler:compiler:$compose_version")
+    compileOnly("io.realm.kotlin:library:0.1.0")
```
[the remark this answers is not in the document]
nhachicha: If you use the model class in the Android project you will need to resolve `io.realm.RealmObject` otherwise you'll have a compiler error
```
e: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
    class io.realm.sample.bookshelf.model.Book, unresolved supertypes: io.realm.RealmObject
```

examples/Bookshelf/androidApp/src/main/java/io/realm/sample/bookshelf/android/ui/BookshelfViewModel.kt
nhachicha: I think this approach is still WIP since you need to wrap the Flow with Swift Combine as described in https://johnoreilly.dev/posts/kotlinmultiplatform-swift-combine_publisher-flow/ 
I think `ViewModel` still does makes sense in its respective platform, however, we could add another sample exploring this idea. One goal of this project is to be a template/[Kata](https://en.wikipedia.org/wiki/Kata_(programming)) to express such nuanced implementation details

nhachicha:
merged in https://github.com/realm/realm-kotlin-samples

PR #274, opened 2021-06-08
Add timeout to CI jobs

cmelchior:
This adds a 2-hour timeout to CI jobs. The activity flags mean the timeout is counted from the last activity in the logs instead of total duration. This seems to be match closer what we are after and we don't have any tests or otherwise that risk going into a loop and logging things at the same time.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,6 +42,7 @@ pipeline {
     // build.
     options {
         lock resource: 'kotlin_build_lock'
+        timeout(time: 2, activity: true, unit: 'HOURS') 
     }
     environment {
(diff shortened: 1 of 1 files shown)


rorbech (approved):
Two hours seems like very inactive. Any reason why it shouldn't just be a matter of minutes? Or are waiting on other jobs also considered inactive?

Anyway, it is better that it is eventually dropped than just stalling.

cmelchior:
True, I guess I just didn't wanted it to suddenly start timing out in the middle of something, but Gradle is generally pretty verbose, so reducing it to something less is probably fine.

I guess the problem we have on Java where the build might hang in the middle due to the Gradle lock isn't present here as we lock at the start of the pipeline, which hopefully doesn't trigger the timeout counter.

PR #275, opened 2021-06-08
Fix kmm-sample to use version 0.1.0

rorbech:
The sample project cannot even initialize in the IDE when trying to resolve the snapshot version defined in `Config.kt`. Hardcode the sample to use the latest released version until settling on the correct setup.


nhachicha (approved):
can't we define the version once in the root build Gradle of the sample project?

rorbech:
@nhachicha Yes, we can.  I just wanted to bring the sample into an operating state as quickly as possible. 

If we go for a central definition it should also cover all the other version defined in the `Config.kt`, so it started to ripple. Maybe we should have a general discussion about what this sample is for (development vs. fully self contained public sample), etc.., especially when introducing more examples (#271). So, again just a quick fix to get it into an operating state.

PR #282, opened 2021-06-09
RealmList - first steps in testing

edualonso:
First steps towards a type-exhaustive test framework for lists.

The testers are multidimensional:

- API dimension get, clear, removeAt, etc.
- Type safety dimension:
- - Nullable: property used for testing must be nullable and dataset must contain at least a null value
- - Non nullable: property used for testing must not be nullable and dataset must not contains null values
- Mode dimension:
- - Unmanaged: no special requirements attached to this mode
- - Managed: needs a Realm instance in which to persist data

Testers implement the API and mode dimensions. The type safety dimension is handled via constructor parameter.


test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
```diff
@@ -0,0 +1,385 @@
...
+    @Test
+    fun addWithIndex() {
+        for (tester in unmanagedTesters) {
+            tester.addWithIndex()
+        }
+        for (tester in managedTesters) {
+            tester.addWithIndex()
+        }
+    }
+
```
nhachicha, line 80: Is there a possibility to test the order `insertAt` is honored on a managed list?

nhachicha (approved):

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
edualonso: Here I iterate over `typeSafetyManager.getInitialDataSet()`'s elements (which is the original dataset I use to populate the `list`) using indices and check that they are in the rightful position after inserting them.
```
                typeSafetyManager.getInitialDataSet()
                    .forEachIndexed { index, e ->
                        assertEquals(index, list.size)
                        list.add(0, copyToRealmIfNeeded(e))
                        assertEquals(index + 1, list.size)
                    }
```
Is this what you meant?

edualonso: Oh, you're right. The element equality assertion is missing. Sorry, it's coming in the next PR 😅 This one was just to set framework in place.

PR #285, opened 2021-06-09
Wrap native pointer in RealmReference

rorbech:
This is a write up of #268 encapsulating Realm `dbPointer` into a `RealmReference`-object. 

The user facing `Realm` now only holds a reference to the underlying core realm version as an atomic `RealmReference` that is advanced on writes. This allows freezing the realm and passing across threads in Kotlin native too. 

The merge of `cr/background-writes` into #268 was rather complex to untangle, so decides to target `cr/background-writes` instead of updating #268.


packages/library/src/commonMain/kotlin/io/realm/internal/RealmReference.kt
```diff
@@ -0,0 +1,37 @@
...
+ * For both versions, the public Realm reference is allowed to change independently from the underlying
+ * `dbPointer`, so care should be taken when accessing any methods on the public Realm.
+ */
+data class RealmReference(
+    val owner: BaseRealm,
+    val dbPointer: NativePointer
+    // FIXME Should we keep a debug flag to assert that we have the right liveness state
+)
+
+internal interface RealmReferrer {
```
cmelchior, line 27: It looks problematic to mix internal and external methods/properties in this interface. I.e. `realm` is internal, but the rest is not. But it is rather unfortunate that it would have made it quite easy to add default implementations to classes.

Not sure if we can have a public interface extend another internal interface, but I cannot come up with a name that would make sense ... `RealmLifeCycle`?

Also, `RealmReferrer` sounds a bit weird to me, I guess we have a reference to it, but it still feels off somehow. Not sure I have a better name right now though

packages/library/src/commonMain/kotlin/io/realm/internal/RealmReference.kt
```diff
@@ -0,0 +1,37 @@
...
+    val realm: RealmReference
+
+    // TODO Can we add default implementations of this. Remember guidelines for properties vs. functions
+    // fun version()
+    // fun isClosed
+}
+
+interface LiveRealm
```
cmelchior, line 35: Are these used for anything?

test/src/androidTest/kotlin/io/realm/shared/RealmTests.kt
```diff
@@ -52,10 +55,12 @@ class RealmTests {
...
 
     private val configuration: RealmConfiguration by lazy {
-        RealmConfiguration(
-            path = "$tmpDir/default.realm",
-            schema = setOf(Parent::class, Child::class)
-        )
+        RealmConfiguration.Builder()
+            .path( "$tmpDir/default.realm",)
+            .schema( setOf(Parent::class, Child::class))
+            // FIXME Debugging only
```
cmelchior, line 61: Why having a FIXME here?

test/src/androidTest/kotlin/io/realm/shared/RealmTests.kt
```diff
@@ -348,4 +353,23 @@ class RealmTests {
...
+    @Suppress("invisible_member")
+    fun writesOnFrozenRealm() {
+        val dispatcher = newSingleThreadContext("background")
+        runBlocking {
+            realm.write {
+                copyToRealm(Parent())
+            }
+        }
+        runBlocking(dispatcher) {
+            assertTrue { realm.isFrozen }
```
cmelchior, line 367: I didn't thought that `isFrozen` was available on the JVM?

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -123,12 +142,15 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
         realmPointerMutex.withLock {
             log.debug("$version -> $newVersion")
             if (newVersion >= version) {
-                // FIXME Currently we need this to be a live realm to be able to continue doing
-                //  writeBlocking transactions.
-                dbPointer = RealmInterop.realm_thaw(newRealm)
-                // We need to start a read transaction to be able to retrieve version id, etc.
+                // FIXME Currently we need to thaw the realm to be a live realm to be able to
```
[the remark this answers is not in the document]
rorbech: No

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -26,9 +27,27 @@ import kotlinx.coroutines.sync.withLock
...
+     * Realms the `dbPointer` points to a frozen SharedRealm that is guaranteed to remain the same, even if
+     * the public Realm advance to a later version.
+     *
+     * The public Realm instance part of the ID can also mutate, so care must be taken if any methods on that
+     * Realm is used.
+     */
+    val updateableRealm: kotlinx.atomicfu.AtomicRef<RealmReference> = kotlinx.atomicfu.atomic<RealmReference>(
+        RealmReference(this, dbPointer)
+    )
+    override val realm: RealmReference
```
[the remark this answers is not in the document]
rorbech: I have made it a `var`. It could technically be a `val` in `BaseRealm` and then be overwritten as a `var` in `Realm`, but I guess that would still cause confusion if we do generic things on `BaseRealm` over the updates.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmReference.kt
rorbech: I have introduced internal `RealmLifeCycle` interface to hold the common interface of the state of any objects tied to a Realm. The interface is currently only implemented for the internal `RealmReference`, but could also be used to do a common default implementation across that could be used across all objects holding a `RealmReference`. But, if we expose this on any public stuff (which I think would indeed make sense), we would need to break up `Realm`, `BaseRealm`, etc. up into public interfaces and internal implementations (which I think would indeed make sense).  But left for another PR.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmReference.kt
rorbech: No, just WIP. Deleted it.

test/src/androidTest/kotlin/io/realm/shared/RealmTests.kt
rorbech: No, just some quick assert for local native testing. Deleted it again.

test/src/androidTest/kotlin/io/realm/shared/RealmTests.kt
```diff
@@ -54,10 +56,11 @@ class RealmTests {
...
 
     private val configuration: RealmConfiguration by lazy {
-        RealmConfiguration(
-            path = "$tmpDir/default.realm",
-            schema = setOf(Parent::class, Child::class)
-        )
+        RealmConfiguration.Builder()
+            .path("$tmpDir/default.realm",)
+            .schema(setOf(Parent::class, Child::class))
+            .log(level = LogLevel.DEBUG)
```
cmelchior, line 62: Do we use this for anything?

cmelchior (approved):

PR #294, opened 2021-06-17
RealmList - add compiler tests

edualonso:
- Added compiler tests for valid/invalid list types.
- Refactored and added error logging when compilation is unsuccessful due to invalid types or nullability constraints for lists.


nhachicha (approved):
LGTM 👍

PR #300, opened 2021-06-22
Track and close intermediate realm references

rorbech:
This makes the `Realm` frozen and tracks all intermediate frozen realm references and closes them when the realm is closed.

### DIFF
--- a/packages/library/src/commonMain/kotlin/io/realm/Realm.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/Realm.kt
@@ -33,8 +33,7 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
     private val writer: SuspendableWriter = SuspendableWriter(this)
     private val realmPointerMutex = Mutex()
 
-    private var updateableRealm: AtomicRef<RealmReference> =
-        atomic(RealmReference(this, dbPointer))
+    private var updateableRealm: AtomicRef<RealmReference> = atomic(RealmReference(this, dbPointer))
 
     /**
      * The current Realm reference that points to the underlying frozen C++ SharedRealm.
@@ -50,6 +49,8 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
             updateableRealm!!.value = value
         }
 
+    // Set of currently open realms. Storing the native pointer explicitly to enable us to close
+    // the realm when the RealmReference is no longer referenced anymore.
     internal val intermediateReferences: AtomicRef<Set<Pair<NativePointer, WeakReference<RealmReference>>>> =
         atomic(mutableSetOf())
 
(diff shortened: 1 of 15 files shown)


packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -134,9 +140,23 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
     private suspend fun updateRealmPointer(newRealm: NativePointer, newVersion: VersionId) {
         realmPointerMutex.withLock {
             log.debug("Updating Realm version: $version -> $newVersion")
-            if (newVersion >= version) {
-                realmReference = RealmReference(this, newRealm)
+            val newReference = RealmReference(this, newRealm)
+            val intermediateReference = if (newVersion >= version) {
+                val version = Pair(realmReference.dbPointer, WeakReference(realmReference))
+                realmReference = newReference
+                version
```
cmelchior, line 147: You are returning `version` here and a Pair in the else. This seems wrong?

cmelchior: Also the naming in this section is a bit hard to follow, we have`newReference`, `RealmReference`,  `WeakReference`, `intermediateReference` and `intermediateReferences`, at least it makes my head spin a little when trying to reason about it. Not exactly sure I have a better proposal though.

test/build.gradle.kts
```diff
@@ -48,6 +48,7 @@ kotlin {
                 // Our current compiler plugin tests only runs on JVM, so makes sense to keep them
                 // for now, but ideally they should go to the compiler plugin tests.
                 implementation("io.realm.kotlin:cinterop:${Realm.version}")
+                implementation("org.jetbrains.kotlinx:atomicfu:${Versions.atomicfu}")
```
cmelchior, line 51: Does this work? I had problems using atomicfu in the test project

test/src/androidMain/kotlin/io/realm/util/PlatformUtils.kt
```diff
@@ -41,4 +42,36 @@ actual object PlatformUtils {
...
+
+    @Suppress("ExplicitGarbageCollectionCall")
+    actual fun triggerGC() {
+        for (i in 1..2) {
+            allocGarbage(0)
+            SystemClock.sleep(100)
+            System.gc()
+            System.runFinalization()
+        }
+        SystemClock.sleep(5000) // 5 seconds to give the GC some time to process
```
cmelchior, line 54: Is there any, even anecdotal, evidence this does anything useful? Sleeping for 5 seconds is pretty heavy.

cmelchior (changes requested):

test/build.gradle.kts
rorbech: This is required since I am asserting on the internal `realm.intermediateReferences.value`. So not applying it.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
rorbech: The name was wrong, so it was actually returning a reference. I tried to separate the _reference tracking_ into a method to hide building the pairs, weak references, etc. Hopefully it is a bit more clear now.

test/src/androidMain/kotlin/io/realm/util/PlatformUtils.kt
rorbech: I just moved the code from https://github.com/realm/realm-kotlin/pull/300/files#diff-8243d5deaa5b3325464776989c0be50c922a20b96ac486be0e9fdf60f026c011L165 to `PlatformUtils` to enable using `triggerGC` in common tests. Only comment that I accidentally didn't copy is `Empiric approach to trigger GC`. I added it again, but maybe @nhachicha can elaborate a bit?

rorbech:
@cmelchior Ready for another round. The sample app tests are failling as `Notifications are not available on frozen lists since they do not change.`. I guess it doesn't really make sense to try to bring in order before we more of your pieces in play too. So I would consider to just to merge it if it passes review this time.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -131,15 +138,35 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
...
+    // Must only be called with realmPointerMutex locked
+    private fun trackNewAndCloseExpiredReferences(realmReference: RealmReference) {
+        val expiredReferences = intermediateReferences.value.filter { (_, ref) -> ref.get() == null }
+        expiredReferences.map { (pointer, _) ->
+            log.debug("Closing unreferenced version: ${RealmInterop.realm_get_version_id(pointer)}")
+            RealmInterop.realm_close(pointer)
+        }
+        intermediateReferences.value = intermediateReferences.value
+            .plus(Pair(realmReference.dbPointer, WeakReference(realmReference)))
+            .subtract(expiredReferences)
```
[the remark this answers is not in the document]
rorbech: 👍 Tweaked it a bit to only iterate the list once.

We have to build a new set as the atomic value is frozen on assignment, so we cannot update the existing one on Native.

rorbech:
I merged this even though the sample app test fails. The failure is due to listening for notifications on frozen lists that will be targeted in #296. Added a TODO item in #291 on fixing the sample app.

Issue #293, opened 2021-06-16
Enabling shrinker causes a RuntimeException


nhachicha:
Enabling the minifier/shrinker requires providing a custom Proguard configuration, otherwise, the user app will crash at runtime 
```
java.lang.RuntimeException: Unable to start activity ComponentInfo{io.realm.example.kmmsample.androidApp/io.realm.example.kmmsample.androidApp.MainActivity}: java.lang.RuntimeException: Cannot find class: 'c
```

This is needed to address #142 for Android

Issue #142, opened 2021-03-10
Support generating Release Apple Frameworks 


nhachicha:
Currently `capiMacosUniversal` and `capiIosArm64` tasks are using debug Monorepo builds 

maybe we can use 
```Kotlin
val framework = kotlin.targets.getByName<org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget>("ios").binaries.getFramework("CONFIGURATION")
 framework.buildType == NativeBuildType.RELEASE
```
To figure out if XCode is trying to use the build phase script (which runs the Gradle `packForXcode ` task) to build a Release framework. This is generally set via the `CONFIGURATION` flag from XCode

cmelchior:
I ran into this. It means that if you just do 

```
cd examples/kmm-sample
./gradlew assemble
```

It will fail with the 

```
> Task :shared:linkReleaseFrameworkIosArm64
w: duplicate library name: io.realm.kotlin:cinterop-cinterop-realm_wrapper
e: Compilation failed: null
 * Source files:
 * Compiler version info: Konan: 1.4.30 / Kotlin: 1.4.30
 * Output kind: FRAMEWORK
e: java.lang.NullPointerException
```

Using 

```
./gradlew assembleDebug
```

Will work.

PR #308, opened 2021-06-23
Support Release build type for Apple builds

nhachicha:
Also enabling Proguard/shrinker for Android
fixes https://github.com/realm/realm-kotlin/issues/293 and https://github.com/realm/realm-kotlin/issues/142

### DIFF
--- a/packages/cinterop/build.gradle.kts
+++ b/packages/cinterop/build.gradle.kts
@@ -36,11 +36,17 @@ repositories {
 //  somehow derive a `isMainHost` property as proposed in
 //  https://kotlinlang.org/docs/reference/mpp-publish-lib.html
 //  Currently just adding the common darwin parts to macos-target.
-val idea = System.getProperty("idea.active") == "true"
+val idea: Boolean = System.getProperty("idea.active") == "true"
+
+// CONFIGURATION is an env variable set by XCode or could be passed to the gradle task to force a certain build type
+//               * Example: to force build a release
+//               realm-kotlin/packages> CONFIGURATION=Release ./gradlew capiIosArm64
+//               * to force build a debug (default BTW) use
+//               realm-kotlin/packages> CONFIGURATION=Debug ./gradlew capiIosArm64
+val isReleaseBuild: Boolean = (System.getenv("CONFIGURATION") ?: "DEBUG").equals("Release", ignoreCase = true)
 
 val corePath = "external/core"
-val relativeCorePath = "../$corePath"
-val cinteropCorePath = "$rootDir/$corePath"
+val absoluteCorePath = "$rootDir/$corePath"
 
 android {
     compileSdkVersion(Versions.Android.compileSdkVersion)
@@ -60,12 +66,6 @@ android {
                 // "val test by getting" instead
                 // androidTest.java.srcDirs += "src/androidTest/kotlin"
             }
-        }
-    }
-    buildTypes {
-        val release by getting {
-            isMinifyEnabled = false
-            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
         }
     }
     defaultConfig {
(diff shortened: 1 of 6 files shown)


packages/cinterop/build.gradle.kts
```diff
@@ -277,74 +310,102 @@ val capiAndroidRelease by tasks.registering {
...
 }
 
 fun Task.build_C_API_Android(releaseBuild: Boolean = false) {
     val buildType = if (releaseBuild) "Release" else "Debug"
     doLast {
         // x86_64
         exec {
-            workingDir(project.file("$relativeCorePath"))
-            commandLine("tools/cross_compile.sh", "-t", buildType, "-a", "x86_64", "-o", "android", "-f", "-DREALM_ENABLE_SYNC=0 -DREALM_NO_TESTS=ON")
+            workingDir(project.file("$absoluteCorePath"))
```
cmelchior, line 329: Are we even using these tasks for anything? Shouldn't the CMake file handle all of this?

cmelchior (approved):

packages/cinterop/build.gradle.kts
nhachicha: These tasks are using CMake with various toolchains, this also gives us the granularity to control some flags (like `-UseModernBuildSystem=NO` more easily)

packages/cinterop/build.gradle.kts
```diff
@@ -88,24 +88,53 @@ android {
...
 }
 
-val nativeLibraryIncludesMacosUniversal = listOf(
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/object-store/c_api/librealm-ffi-static-dbg.a",
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/librealm-dbg.a",
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/parser/librealm-parser-dbg.a",
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/object-store/librealm-object-store-dbg.a"
+val nativeLibraryIncludesMacosUniversalRelease = listOf(
+    "-include-binary",
+    "$absoluteCorePath/capi_macos_universal/src/realm/object-store/c_api/librealm-ffi-static.a",
```
rorbech, line 93: Really annoying that the directory `capi_macos_universal` doesn't follow the pattern of all the other build directories by prefixing `build_`

packages/cinterop/build.gradle.kts
```diff
@@ -88,24 +88,53 @@ android {
     }
 }
 
-val nativeLibraryIncludesMacosUniversal = listOf(
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/object-store/c_api/librealm-ffi-static-dbg.a",
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/librealm-dbg.a",
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/parser/librealm-parser-dbg.a",
-    "-include-binary", "$cinteropCorePath/capi_macos_universal/src/realm/object-store/librealm-object-store-dbg.a"
+val nativeLibraryIncludesMacosUniversalRelease = listOf(
```
rorbech, line 91: Extremely verbose. I guess there are some veriations, but I guess we could maybe define the libs 
```
val libs = setOf("lib/librealm-ffi-static.a"....)
val libsDbg = setOf("lib/librealm-ffi-static.a"....)
```
And define the compiler arguments by something from these, by some mapping
```
val nativeLibraryIncludesIosArm64Debug = libsDbg.map { ... }
```

packages/cinterop/build.gradle.kts
```diff
@@ -131,7 +160,7 @@ kotlin {
             cinterops.create("realm_wrapper") {
                 defFile = project.file("src/nativeCommon/realm.def")
                 packageName = "realm_wrapper"
-                includeDirs(project.file("$relativeCorePath/src/"))
+                includeDirs("$absoluteCorePath/src/")
```
rorbech, line 163: Have you verified that this still allows compilation from both inside `packages` and `test`. As far as I recall there was some issue with that.

packages/cinterop/build.gradle.kts
rorbech: I don't think these tasks are triggered anymore after building through the monorepo Cmake-infrastructure, so you would need to incorporate stuff into the `CMakeLists.txt`. At least I removed them in the `frozen-architecture`-branch without causing issues.

rorbech (approved):
LGTM, but please verify that we can build/test things both from `packages` and `test` after updating the core paths.

PR #314, opened 2021-06-27
Eliminated notifier, writer and RealmReference to avoid memory leak

rorbech:
`RealmReference` introduced a cycle in the references, which seems to cause GC to fail deallocation the Realm and the RealmReference on native. 

This draft PR is just my experiments of stripping the notifier and writer from `Realm` and then finally eliminating the `realmReference` in favor of a pure `NativePointer` in the `BaseRealm`. Just to show that this is in fact the triggering factor.

### DIFF
--- a/packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
+++ b/packages/cinterop/src/darwinCommon/kotlin/io/realm/interop/RealmInterop.kt
@@ -93,12 +93,18 @@ private fun <T : CPointed> checkedPointerResult(pointer: CPointer<T>?): CPointer
 
 // FIXME API-INTERNAL Consider making NativePointer/CPointerWrapper generic to enforce typing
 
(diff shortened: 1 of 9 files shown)


cmelchior:
Thinking a bit more about this at first it made sense, but we only store `RealmReference` as a WeakReference inside the Realm itself, so the cycle should be breakable 🤔 Something to look further into.

rorbech:
We are only using weak references for the intermediate versions, as the current version must not be released until all objects _and_ the realm are no longer referencing it. 

Seems like the actual issue is that the cyclic reference is stored in a shared atomic reference. From https://blog.jetbrains.com/kotlin/2020/07/kotlin-native-memory-management-roadmap/: 
> Synchronization primitives must internally share a mutable state, which is supported in Kotlin/Native via special atomic references. Yet the existing memory management algorithm does not track cycles through such references.

rorbech:
Closing as outdated as we have now moved on to the new memory model.

PR #316, opened 2021-06-28
Fix README installation instructions

nathan-contino-mongo:
- update "realm-kotlin" plugin name to "io.realm.kotlin"
- remove redundant screenshots
- remove explicit version references, replace with `<VERSION>` placeholder
- point users to config/github releases section for potential version values
- updated all transaction usages to the `write`/`writeBlocking` syntax from the now-removed `beginTransaction`/`commitTransaction` pattern
- removed usages of the `realm.create` method, which has been deprecated
- updated `maven(url = "https://oss.jfrog.org/artifactory/oss-snapshot-local")` dependency references to `mavenCentral()`, since that's a default KMM template setting that most users will understand that they don't need to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ This repository holds the source code for the Kotlin SDK for Realm, which runs o
 
 https://github.com/realm/realm-kotlin-samples
 
-# Quick Startup
+# Quick Start
 
 ## Prerequisite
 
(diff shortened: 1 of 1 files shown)


nhachicha (approved):
Thanks for the update 👏

PR #322, opened 2021-07-01
Fix Gradle Portal login params

cmelchior:
Fix login tokens for uploading to Gradle Plugin Portal during releases.


nhachicha (approved):
🎉

PR #325, opened 2021-07-01
Update README.md

nhachicha:


README.md
```diff
@@ -21,34 +21,13 @@ Start a new [KMM](https://kotlinlang.org/docs/mobile/create-first-app.html) proj
...
-
-allprojects {
-    repositories {
-        mavenCentral()
-    }
-}
-```
-
-- Apply the `io.realm.kotlin` plugin and specify the dependency in the common source set.
+- In the shared module (`shared/build.gradle.kts`), apply the `io.realm.kotlin` plugin and specify the dependency in the common source set.
```
cmelchior, line 24: Not sure this enough? People still need to manually add `gradlePluginPortal()` and `mavenCentral()` to build script repositories and `mavenCentral()` to project repositories?

nhachicha: I tested it with a fresh KMM project. `mavenCentral` and `gradlePluginPortal()` are already included in the root Gradle project.
```Gradle
buildscript {
    repositories {
        gradlePluginPortal()
        google()
        mavenCentral()
    }
    dependencies {
        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.10")
        classpath("com.android.tools.build:gradle:7.0.0-beta04")
    }
}

allprojects {
    repositories {
        google()
        mavenCentral()
    }
}

tasks.register("clean", Delete::class) {
    delete(rootProject.buildDir)
}
```

cmelchior: Okay, just wasn't sure if we should post the entire setup or just assume that people are using KMM project tutorials? But not a big deal.

cmelchior (approved):

README.md
```diff
@@ -21,34 +21,13 @@ Start a new [KMM](https://kotlinlang.org/docs/mobile/create-first-app.html) proj
...
 
 ```Gradle
 plugins {
     kotlin("multiplatform")
     id("com.android.library")
-    id("io.realm.kotlin")
+    id("io.realm.kotlin") version "<VERSION>"
```
nathan-contino-mongo, line 30: Is there a good reason to include the version here as well? In my experience it's just another place to update when you want to increase your version. I'm currently leaving it out in the official docs, but if this is something useful I can always add it.

README.md
```diff
@@ -61,6 +40,13 @@ kotlin {
 }
 ```
 
+- If you use the model classes or query results inside the Android module(`androidApp/build.gradle.kts`) you need to add a compile time dependency as follow:
```
nathan-contino-mongo, line 43:
```suggestion
- If you use the model classes or query results inside the Android module(`androidApp/build.gradle.kts`) you need to add a compile time dependency as follows:
```

README.md
nathan-contino-mongo: I can confirm that `mavenCentral()` and `gradlePluginPortal()` are already in there using the KMM plugin template. However, I wonder if it's worth mentioning them here so that users at least know where the dependencies are coming from if they do choose to create a project without the template, or if the template changes in the future.

README.md
nhachicha: It depends on how you applied the plugin (see https://plugins.gradle.org/plugin/io.realm.kotlin) the legacy way doesn't need to specify the version since it's been provided in the `buildscript` dependency. In this PR we focus on the modern way of applying the plugin

README.md
nhachicha: I think it's fair to assume these are already in since even if the project is created from scratch it needs the `gradlePluginPortal` to apply Android and Multiplatform plugins. We will also point to the detailed doc on the website to cater for different use cases.

Issue #318, opened 2021-06-29
[Dokka] Create package level docs


cmelchior:
We should have package level docs describing the purpose of each package:

![image](https://user-images.githubusercontent.com/406066/123841513-eedb0500-d90f-11eb-9b25-2d60ee84d746.png)

Issue #317, opened 2021-06-29
[Dokka] Revisit first sentance in all classes


cmelchior:
The default Dokka rendering only shows the first sentence of the full Class docs. We should go through all classes and make sure that the first line clearly describes the purpose of the class.

![image](https://user-images.githubusercontent.com/406066/123840963-4dec4a00-d90f-11eb-942d-e06047de16c1.png)

Issue #192, opened 2021-05-05
Throw exception when creating objects with duplicate primary keys


rorbech:
Current C-API is returning any existing object when calling `realm_object_create_with_primary` if an object with the given primary key already exists. This means that the current behavior of our `create`/`copyToRealm` is actually an `copyOrUpdate`. 

This could be fixed by checking for existence of the object prior to creating it and throwing the error ourselves, but requested this to be part of the C-API as it probably also would be useful for other SDKs and error reporting would be more uniform if all constraint violations of the same constraint are raised the same way. 

C-API issue is in https://github.com/realm/realm-core/issues/4595

PR #343, opened 2021-07-12
Dokka cleanups

rorbech:
Cleaned up package info (#318) and first line doc for classes (#317). 

Still lacking overall descriptive content for `io.realm` and especially `Realm`.

Questions
- [x] Package info.md in root of `packages` or inside `src/commonMain/kotlin`? - Ended up putting package specific info inside package directory. 
- [x] Realm constructor or `Realm.open` ... or both? - Fixed `Realm` constructor, so both works for now
- [x] Internal only (at least for now, or maybe required for native?)
  - [x] Queryable? - Kept is but marked likely for change and referenced issue
  - [x] Cancellable? - Internalized until we expose callback based API as I expect the iOS samples to be updated to use flows 
  - [x] Callback<T>? - Internalized until we expose callback based API as I expect the iOS samples to be updated to use flows
- [ ] Should we try do a check for existing primary key objects to throw on duplicate until C-API is throwing #192 - Will try to see if I can make a quick fix in another PR

### DIFF
--- a/packages/library/build.gradle.kts
+++ b/packages/library/build.gradle.kts
@@ -217,7 +217,8 @@ realmPublish {
 }
 
 tasks.dokkaHtml.configure {
-    moduleName.set("realm-kotlin")
+    moduleName.set("Realm Kotlin Multiplatform SDK")
+    moduleVersion.set(Realm.version)
     dokkaSourceSets {
         configureEach {
             moduleVersion.set(Realm.version)
--- a/packages/library/build.gradle.kts
+++ b/packages/library/build.gradle.kts
@@ -217,7 +217,7 @@ realmPublish {
 }
 
 tasks.dokkaHtml.configure {
-    moduleName.set("Realm Kotlin Multiplatform SDK")
+    moduleName.set("realm-kotlin")
     dokkaSourceSets {
         configureEach {
             moduleVersion.set(Realm.version)
@@ -227,9 +227,14 @@ tasks.dokkaHtml.configure {
                 matchingRegex.set("io\\.realm\\.internal\\.*")
                 suppress.set(true)
             }
+            jdkVersion.set(8)
         }
         val commonMain by getting {
-            includes.from("overview.md", "io.realm.md")
+            includes.from(
+                "overview.md",
+                "src/commonMain/kotlin/io/realm/info.md",
+                "src/commonMain/kotlin/io/realm/log/info.md"
+            )
             sourceRoot("../runtime-api/src/commonMain/kotlin")
         }
     }
--- a/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
@@ -69,7 +69,7 @@ public abstract class BaseRealm internal constructor(
     /**
      * Returns the results of querying for all objects of a specific type.
      *
-     * For a [Realm] instance this will reflect the state of the Realm at the invocation time, thus
+     * For a [Realm] instance this reflects the state of the realm at the invocation time, thus
      * the results will not change on updates to the Realm. For a [MutableRealm] the result is live
      * and will in fact reflect updates to the [MutableRealm].
      *
--- a/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
@@ -54,7 +54,7 @@ public abstract class BaseRealm internal constructor(
         }
 
     /**
-     * The current version of the data in this Realm.
+     * The current version of the data in this realm.
      */
     // TODO Could be abstracted into base implementation of RealmLifeCycle!?
     public var version: VersionId = VersionId(0)
--- a/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
@@ -35,7 +35,7 @@ public abstract class BaseRealm internal constructor(
 ) {
 
     companion object {
-        private const val observablesNotSupportMessage = "Observing changes are not supported by this Realm."
+        private const val observablesNotSupportedMessage = "Observing changes are not supported by this Realm."
     }
 
     /**
@@ -54,7 +54,7 @@ public abstract class BaseRealm internal constructor(
         }
 
     /**
-     * The current version of this Realm and data fetched from it.
+     * The current version of the data in this Realm.
      */
     // TODO Could be abstracted into base implementation of RealmLifeCycle!?
     public var version: VersionId = VersionId(0)
@@ -72,6 +72,9 @@ public abstract class BaseRealm internal constructor(
      * For a [Realm] instance this will reflect the state of the Realm at the invocation time, thus
      * the results will not change on updates to the Realm. For a [MutableRealm] the result is live
      * and will in fact reflect updates to the [MutableRealm].
+     *
+     * @param clazz The class of the objects to query for.
+     * @return The result of the query as of the time of invoking this method.
      */
     open fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T> {
         // Use same reference through out all operations to avoid locking
@@ -88,39 +91,44 @@ public abstract class BaseRealm internal constructor(
      * Returns the results of querying for all objects of a specific type.
      *
      * Convenience inline method to catch the reified class type of single argument variant of [objects].
-    */
+     *
+     * @param T The type of the objects to query for.
+     * @return The result of the query. Dependent of the type of the Realm this will either reflect
+     * for [Realm]: the state at invocation or for [MutableRealm] the latest updated state of the
+     * mutable realm.
+     */
     inline fun <reified T : RealmObject> objects(): RealmResults<T> { return objects(T::class) }
 
     internal open fun <T : RealmObject> registerResultsChangeListener(
         results: RealmResults<T>,
         callback: Callback<RealmResults<T>>
     ): Cancellable {
-        throw NotImplementedError(observablesNotSupportMessage)
+        throw NotImplementedError(observablesNotSupportedMessage)
     }
 
     internal open fun <T : RealmObject> registerListChangeListener(
         list: List<T>,
         callback: Callback<List<T>>
     ): Cancellable {
-        throw NotImplementedError(observablesNotSupportMessage)
+        throw NotImplementedError(observablesNotSupportedMessage)
     }
 
     internal open fun <T : RealmObject> registerObjectChangeListener(
         obj: T,
         callback: Callback<T?>
     ): Cancellable {
-        throw NotImplementedError(observablesNotSupportMessage)
+        throw NotImplementedError(observablesNotSupportedMessage)
     }
 
     internal open fun <T : RealmObject> registerResultsObserver(results: RealmResults<T>): Flow<RealmResults<T>> {
-        throw NotImplementedError(observablesNotSupportMessage)
+        throw NotImplementedError(observablesNotSupportedMessage)
     }
     internal open fun <T : RealmObject> registerListObserver(list: List<T>): Flow<List<T>> {
-        throw NotImplementedError(observablesNotSupportMessage)
+        throw NotImplementedError(observablesNotSupportedMessage)
     }
 
     internal open fun <T : RealmObject> registerObjectObserver(obj: T): Flow<T> {
-        throw NotImplementedError(observablesNotSupportMessage)
+        throw NotImplementedError(observablesNotSupportedMessage)
     }
 
     /**
--- a/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
@@ -23,7 +23,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlin.reflect.KClass
 
 /**
- * Base class for all Realm instances.
+ * Base class for all Realm instances ([Realm] and [MutableRealm]).
  */
 @Suppress("UnnecessaryAbstractClass")
 public abstract class BaseRealm internal constructor(
@@ -35,7 +35,7 @@ public abstract class BaseRealm internal constructor(
 ) {
 
     companion object {
-        const val observerablesNotSupportMessage = "Observing changes are not supported by this Realm."
+        private const val observablesNotSupportMessage = "Observing changes are not supported by this Realm."
     }
 
     /**
@@ -54,7 +54,7 @@ public abstract class BaseRealm internal constructor(
         }
 
     /**
-     * The current data version of this Realm and data fetched from it.
+     * The current version of this Realm and data fetched from it.
      */
     // TODO Could be abstracted into base implementation of RealmLifeCycle!?
     public var version: VersionId = VersionId(0)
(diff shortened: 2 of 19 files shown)


packages/library/build.gradle.kts
```diff
@@ -217,7 +217,7 @@ realmPublish {
 }
 
 tasks.dokkaHtml.configure {
-    moduleName.set("Realm Kotlin Multiplatform SDK")
+    moduleName.set("realm-kotlin")
```
cmelchior, line 220: Isn't this used in the title? https://docs.mongodb.com/realm-sdks/kotlin/latest/

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -54,7 +54,7 @@ public abstract class BaseRealm internal constructor(
         }
 
     /**
-     * The current data version of this Realm and data fetched from it.
+     * The current version of this Realm and data fetched from it.
```
cmelchior, line 57: I had `data` in there because we have quite a few version concepts: SDK itself, Schema and our "internal" version which this is. If data is the wrong word, I think we should rephrase this somehow.

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -35,7 +35,7 @@ public abstract class BaseRealm internal constructor(
 ) {
 
     companion object {
-        const val observerablesNotSupportMessage = "Observing changes are not supported by this Realm."
+        private const val observablesNotSupportMessage = "Observing changes are not supported by this Realm."
```
cmelchior, line 38: Nitpick and unrelated, but should probably be `observablesNotSupportedMessage`

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -23,9 +23,17 @@ import kotlinx.coroutines.flow.Flow
...
+ * Represents the writeable state of a Realm file.
+ *
+ * The only way to modify data in a Realm is through instances of this class.
+ * These are provided and managed automatically through either [Realm.write] or
  * [Realm.writeBlocking].
+ *
+ * All objects created and/or obtained from the _mutable realm_ in a write-transaction are bound to
+ * the thread executing the transaction and all operations on the _mutable realm_ or on any of the
+ * objects must be done on the thread executing the transaction. Only exception are objects returned
+ * from [Realm.write] and [Realm.writeBlocking] which will be frozen and kept tied to resulting
```
cmelchior, line 35:
```suggestion
 * from [Realm.write] and [Realm.writeBlocking] which will be frozen and kept tied to the resulting
```

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -134,9 +142,24 @@ class MutableRealm : BaseRealm {
...
     fun <T : RealmObject> copyToRealm(instance: T): T {
         return io.realm.internal.copyToRealm(configuration.mediator, realmReference, instance)
     }
+
+    /**
+     * Returns the results of querying for all objects of a specific type.
+     *
+     * The result is live and are thus also reflecting any update to the MutableRealm.
+     *
+     * The result is only valid on the calling thread.
```
cmelchior, line 157: Missing `@param` and `@return`

packages/library/src/commonMain/kotlin/io/realm/PrimaryKey.kt
```diff
@@ -19,4 +19,13 @@ package io.realm
...
 @Target(AnnotationTarget.FIELD)
 @MustBeDocumented
+/**
+ * Annotation marking a field as a primary key inside Realm.
+ *
+ * Only one field in a RealmObject class can have this annotation, and the field should uniquely
+ * identify the object.
+ *
+ * It is allowed to apply this annotation on the following primitive types: String, Byte, Char,
+ * Short, Int, and Long and are permitted to have null as a primary key value.
```
cmelchior, line 29:
```suggestion
 * Short, Int, and Long, as well as their nullable variants.
```

packages/library/src/commonMain/kotlin/io/realm/Queryable.kt
```diff
@@ -21,6 +21,7 @@ package io.realm
 //    dependent on how the final API is going to look like.
 //  - Query could alternatively be separated into builder to await constructing new results until
 //    actually executing the query
+// FIXME Is this deprecated?
```
cmelchior, line 24: It is part of #206 to revisit this

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -112,9 +119,21 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
      *  this constructor should be the primary way to open Realms (as you only need
      *  to do it once pr. app).
      */
+    // FIXME This is not updated to the frozen architecture. Which way to go? Towards Realm.open or
```
cmelchior, line 122: I think this should be the default way of opening a Realm similar to `Realm.getInstance()`

IMO we should make `Realm.open()` a suspend function and rework it to behave like `Realm.getInstanceAsync()`

A quick fix would be to make `Realm.open()` internal though.

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -33,6 +33,7 @@ import kotlin.reflect.KClass
 /**
  * Configuration for log events created by a Realm instance.
  */
+// FIXME Any reason for this to be public?
```
cmelchior, line 36: It is exposed through `RealmConfiguration.log` so users can inspect the log configurationtion

packages/library/build.gradle.kts
cmelchior: Not 100% sure why this needs to change? But if it does we also need to warn @nathan-contino-mongo as he currently depends on it for some links.

rorbech: Yeah. It was just because the name is also used in the URL and for breadcrumbs so I found it better with a short name. I don't think there is a way to have a different module name and title, so will revert to avoid having to update the docs link.

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
rorbech: I was confused by the `data version` term in the first place. I tried to rephrase, if it doesn't compute in your end, I will just revert.

packages/library/src/commonMain/kotlin/io/realm/Queryable.kt
rorbech: Ok. Updated with a comment that this might change and a reference to the issue.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
rorbech: I added the existing workaround for obtaining the frozen realm from a live one and closing it into the `init`-clause, so `Realm(configuration)` does the right thing. We could basically just eliminate `Realm.open` for now, if we deem the fix ok, but only updated the `RealmTests` to use the constructor directly for now.

packages/library/src/commonMain/kotlin/io/realm/Queryable.kt
```diff
@@ -21,6 +21,12 @@ package io.realm
 //    dependent on how the final API is going to look like.
 //  - Query could alternatively be separated into builder to await constructing new results until
 //    actually executing the query
+//  https://github.com/realm/realm-kotlin/issues/206
+/**
+ * Interface holding common query methods.
+ *
+ * Will most likely change in the future.
```
nhachicha, line 28: I would remove this line. API docs just reflect the current state. Otherwise, this might add confusion for the reader

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -66,7 +66,17 @@ public abstract class BaseRealm internal constructor(
         log.info("Realm opened: ${configuration.path}")
     }
 
-    fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T> {
+    /**
+     * Returns the results of querying for all objects of a specific type.
+     *
+     * For a [Realm] instance this will reflect the state of the Realm at the invocation time, thus
```
nathan-contino-mongo, line 72:
```suggestion
     * For a [Realm] instance this will reflect the state of the realm at the invocation time, thus
```
To avoid confusion between "Realm" (the product name shorthand), "MongoDB Realm" (the product name), "Realm Database" (the client-side database) and "realm" (an instance of a connection to a single file containing Realm Database data), we usually use lowercase "realm" in cases like this. I know it's a little confusing since Kotlin uses uppercase for the first letter of class names, so up to you if you'd like to follow this convention.

nathan-contino-mongo:
```suggestion
     * For a [Realm] instance this reflects the state of the Realm at the invocation time, thus
```
Additionally, active voice can make this easier to read for end users.

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -23,9 +23,17 @@ import kotlinx.coroutines.flow.Flow
 import kotlin.reflect.KClass
 
 /**
- * This class represents the writeable state of a Realm file. The only way to modify data in a Realm is through
- * instances of this class. These are provided and managed automatically through either [Realm.write] or
+ * Represents the writeable state of a Realm file.
+ *
+ * The only way to modify data in a Realm is through instances of this class.
```
nathan-contino-mongo, line 28:
```suggestion
 * To modify data in a [Realm], use instances of this class.
```
Passive -> active rewrite.

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -23,9 +23,17 @@ import kotlinx.coroutines.flow.Flow
...
+ * These are provided and managed automatically through either [Realm.write] or
  * [Realm.writeBlocking].
+ *
+ * All objects created and/or obtained from the _mutable realm_ in a write-transaction are bound to
+ * the thread executing the transaction and all operations on the _mutable realm_ or on any of the
```
nathan-contino-mongo, line 33:
```suggestion
 * the thread executing the transaction. All operations on the _mutable realm_ or on any of the
```
This is a long, complex sentence, so I'd suggest breaking up the two thoughts in it to make it a little more digestible for readers.

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -23,9 +23,17 @@ import kotlinx.coroutines.flow.Flow
...
+ * objects must be done on the thread executing the transaction. Only exception are objects returned
```
nathan-contino-mongo, line 34:
```suggestion
 * objects contained in that realm must execute on the thread executing the transaction. The only exception is objects returned
```

Clarification about objects in a particular realm and subject-verb agreement fix.

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -23,9 +23,17 @@ import kotlinx.coroutines.flow.Flow
...
+ * from [Realm.write] and [Realm.writeBlocking] which will be frozen and kept tied to the resulting
```
nathan-contino-mongo, line 35:
```suggestion
 * from [Realm.write] and [Realm.writeBlocking], which are frozen and remain tied to the resulting
```
Comma makes this a little more readable + small reword for passive voice.

packages/library/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -134,9 +142,27 @@ class MutableRealm : BaseRealm {
...
+    //  currently effectively mimics the copyOrUpdate API from realm-java.
+    //  https://github.com/realm/realm-kotlin/issues/192
     fun <T : RealmObject> copyToRealm(instance: T): T {
         return io.realm.internal.copyToRealm(configuration.mediator, realmReference, instance)
     }
+
+    /**
+     * Returns the results of querying for all objects of a specific type.
+     *
+     * The result is live and are thus also reflecting any update to the MutableRealm.
```
nathan-contino-mongo, line 155:
```suggestion
     * The result is live and thus also reflects any update to the [MutableRealm].
```
Passive -> active voice fix + adding a link.

packages/library/src/commonMain/kotlin/io/realm/PrimaryKey.kt
```diff
@@ -19,4 +19,13 @@ package io.realm
 @Retention(AnnotationRetention.SOURCE)
 @Target(AnnotationTarget.FIELD)
 @MustBeDocumented
+/**
+ * Annotation marking a field as a primary key inside Realm.
+ *
+ * Only one field in a RealmObject class can have this annotation, and the field should uniquely
+ * identify the object.
+ *
+ * It is allowed to apply this annotation on the following primitive types: String, Byte, Char,
```
nathan-contino-mongo, line 28:
```suggestion
 * This annotation applies to the following primitive types: String, Byte, Char,
```
Passive -> active reword.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -34,7 +34,13 @@ import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
+import kotlin.reflect.KClass
 
+/**
+ * A Realm instance is the main entry point for interacting with the persisted Realm.
```
nathan-contino-mongo, line 40:
```suggestion
 * A Realm instance is the main entry point for interacting with a persisted Realm.
```
Definite -> indefinite article fix to match subject.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -107,14 +117,25 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
...
      */
+    // FIXME Figure out how to describe the constructor better
     public constructor(configuration: RealmConfiguration) :
         this(configuration, RealmInterop.realm_open(configuration.nativeConfig))
 
+    /**
+     * Returns the results of querying for all objects of a specific type.
+     *
+     * The result is reflecting the state of the Realm at the invocation time, thus the results
+     * will not change on updates to the Realm and can be accessed from any thread.
```
nathan-contino-mongo, line 130:
```suggestion
     * The result reflects the state of the realm at invocation time, so the results
     * do not change when the realm updates. You can access these results from any thread.
```
Passive -> active voice

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -46,6 +46,16 @@ public data class LogConfiguration(
     public val loggers: List<RealmLogger>
 )
 
+/**
+ * A _Realm Configuration_ defining specific setup and configuration for a Realm instance.
+ *
+ * The RealmConfiguration can, for simple uses cases, be created directly through the constructor,
+ * while more advanced setup requires building the RealmConfiguration through
```
nathan-contino-mongo, line 53:
```suggestion
 * The RealmConfiguration can, for simple uses cases, be created directly through the constructor.
 * More advanced setup requires building the RealmConfiguration through
```
Breaking up a long sentence to make it more readable.

nathan-contino-mongo (approved):
LGTM, just a few rewords to make sentences a little more readable and follow some docs conventions around wording. Thanks for letting me help out!

nhachicha (approved):
LGTM, some functions in `RealmResults` are not documented like `delete()` which will remove all results of a query.

cmelchior (approved):

PR #345, opened 2021-07-13
Update example with frozen arch

nhachicha:
- Using Snapshot release by default (for easy checkout and build)
- Update iOS example with Frozen Architecture 

Note: I started observing the message `Failed to send update to Realm from the Notifier:` displayed in  https://github.com/realm/realm-kotlin/blob/master/packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt#L158 when using the iOS app, this probably needs to be investigated separately


cmelchior (approved):
Nice 👏

edualonso (approved):

PR #347, opened 2021-07-13
Remove Realm.open and MutableRealm.create

rorbech:


README.md
```diff
@@ -72,7 +72,7 @@ val configuration = RealmConfiguration(schema = setOf(Person::class, Dog::class)
 ```
 
 ```Kotlin
-val realm = Realm.open(configuration)
```
nhachicha, line 75: I don't have a strong opinion against it, but one benefit of `Realm.open` is to control the caching strategy (since Realms are frozen we can choose internally to implement a singleton by `RealmConfiguration` for example, doing so via the constructor might feel a bit weird. Also `open` strongly indicates opening a resource that you need to handle and close

cmelchior (approved):

test/src/androidTest/kotlin/io/realm/shared/PrimaryKeyTests.kt
```diff
@@ -133,27 +122,6 @@ class PrimaryKeyTests {
         assertNull(objects[0].primaryKey)
     }
 
-    @Test
-    fun primaryKeyForNonPrimaryKeyObjectThrows() {
```
nhachicha, line 137: Are these tests implemented elsewhere?

nhachicha (commented):
I left a comment regarding `open`, this is a major change, we need to update the Mongo doc and Readme accordingly

README.md
cmelchior: The idea is that we don't really need caching with the frozen architecture...You just create the instance once and use it everywhere.

test/src/androidTest/kotlin/io/realm/shared/PrimaryKeyTests.kt
rorbech: No. There is not longer any API for:
- supplying a primary key when it is not needed, 
- supplying a primary key of the wrong type
- add an primary key object without a primary key

So I just removed the tests.

README.md
rorbech: I don't have any strong opinions either, but I see Nabil's point.

PR #349, opened 2021-07-13
Fix native C-API test for setting empty strings on native

rorbech:


rorbech:
These test are not executed on CI. Reason for doing this was that it wasn't public API and the test mostly acts as C-API alignment examples. But seems like there are now actual schema migration tests, etc. in there. Should we start executing these on CI?

cmelchior:
We probably should. Otherwise the chance of them growing stale is too big.

nhachicha (approved):

PR #350, opened 2021-07-13
Fix realm changed notification on native.

rorbech:
Log message on native introduced in #345 seems to be because the flow cannot deliver events unless `extraBufferCapacity ` is more than 0. Have to investigate why this is the case, but setting to 1 eliminates the error.


cmelchior (approved):
It would be nice to understand the root cause of it, but this change seems safe and we are still logging it if the error pops up. So this seems fine 👍

PR #352, opened 2021-07-14
Add link to doc & remove the production warning

nhachicha:


README.md
```diff
@@ -150,10 +154,6 @@ Next: head to the full KMM [example](./examples/kmm-sample).
 
 NOTE: The SDK doesn't currently support  `x86` - Please use an `x86_64` or `arm64` emulator/device
  
-# Developer Preview
```
rorbech, line 153: Should we have some Alpha-ish statement instead. Just to indicate that APIs can still change, etc.

rorbech (approved):

README.md
nhachicha: I added a section regarding the Alpha status

PR #354, opened 2021-07-14
Throw errors when importing objects with duplicate primary keys

### LINKED ISSUE #192: Throw exception when creating objects with duplicate primary keys

rorbech:
This is adding a manual check for conflicting primary keys before adding new objects. 

Closes #192

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmInterop.kt
@@ -22,8 +22,10 @@ import kotlin.jvm.JvmInline
 // FIXME API-INTERNAL Consider adding marker interfaces NativeRealm, NativeRealmConfig, etc. as type parameter
 //  to NativePointer. NOTE Verify that it is supported for Kotlin Native!
 
+// Wrapper for the C-API realm_class_key_t uniquely identifying the class/table in the schema
 @JvmInline
 value class ClassKey(val key: Long)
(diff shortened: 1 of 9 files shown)


packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -17,11 +17,15 @@
...
 
 import kotlinx.coroutines.CoroutineDispatcher
+import kotlin.jvm.JvmInline
 
 // FIXME API-INTERNAL Consider adding marker interfaces NativeRealm, NativeRealmConfig, etc. as type parameter
 //  to NativePointer. NOTE Verify that it is supported for Kotlin Native!
 
-inline class ColumnKey(val key: Long)
+@JvmInline
+value class ClassKey(val key: Long)
```
cmelchior, line 26: Perhaps add docs for what they represent?

cmelchior (approved):
Good work-around until we get support from the C-API. Only needs an CHANGELOG update as well 👍

nhachicha (approved):
Nice 👍

PR #357, opened 2021-07-15
Fix erroneous realm reference from notifier

rorbech:


packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -147,7 +147,7 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
             // That way, querying the Realm right after the `write` completes will return
             // the written data. Otherwise, we would have to wait for the Notifier thread
             // to detect it and update the user Realm.
-            updateRealmPointer(nativePointer, versionId)
+            updateRealmPointer(RealmReference(this, nativePointer))
```
cmelchior, line 150: So I assume it was this nativePointer that caused it?

cmelchior: But still not 100% sure I see why it is a bug 🤔

rorbech: The issue was that when sending the native pointer to the Realm and creating a `RealmReference` from that, we ended up with multiple references pointing to the same nativePointer. As nobody actually referenced the `RealmReference` created in `updateRealmPointer`, it got cleaned up and closed when moving the Realm current reference to a newer version. 

We should not have multiple RealmReference holding the same pointer. Alternatively we should handle the native pointers to a _realm_ differently from the other native pointers and postpone closing the realm until this is release.

rorbech: Sorry. It was actually not this one causing it. It was the one from notifier from https://github.com/realm/realm-kotlin/pull/357/files#diff-b58a8d37418ba726cc50024a2cecc41960ad1a11bbb0766a53e5f85eab540735R99

cmelchior: That makes sense. Maybe we should add some warnings to the `RealmReferenes` class about that?

cmelchior (approved):

PR #360, opened 2021-07-26
Fix snapshot links

TonnyL:


cmelchior:
Nice, didn't realize you could link to it directly 👏

cmelchior (approved):

PR #365, opened 2021-08-03
Add support for encrypted Realms

clementetb:
With this PR we add support for encrypted Realms. Realms are encrypted with a 64-byte key, it can be passed into the configuration via the configuration builder. The SDK will throw an `IllegalArgumentException` on any wrong-sized key.

Example:

```
val encryptedRealmConfig = RealmConfiguration.Builder()
    .encryptionKey(encryptionKey)
    .build()
    
val encryptedRealm = Realm(encryptedRealmConfig)
```

### DIFF
--- a/packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
@@ -125,7 +125,7 @@ public class RealmConfiguration private constructor(
     /**
      * 64 byte key used to encrypt and decrypt the Realm file.
      */
-    public val encryptionKey: ByteArray?
+    public val encryptionKey get() = RealmInterop.realm_config_get_encryption_key(nativeConfig)
 
     // Internal properties used by other Realm components, but does not make sense for the end user to know about
     internal var mapOfKClassWithCompanion: Map<KClass<out RealmObject>, RealmObjectCompanion>
@@ -149,7 +149,6 @@ public class RealmConfiguration private constructor(
         this.writeDispatcher = writeDispatcher
         this.schemaVersion = schemaVersion
         this.deleteRealmIfMigrationNeeded = deleteRealmIfMigrationNeeded
-        this.encryptionKey = encryptionKey?.copyOf()
 
         RealmInterop.realm_config_set_path(nativeConfig, this.path)
 
@@ -166,6 +165,10 @@ public class RealmConfiguration private constructor(
 
         RealmInterop.realm_config_set_schema(nativeConfig, nativeSchema)
         RealmInterop.realm_config_set_max_number_of_active_versions(nativeConfig, maxNumberOfActiveVersions)
+
+        encryptionKey?.let {
+            RealmInterop.realm_config_set_encryption_key(nativeConfig, it)
+        }
 
         mediator = object : Mediator {
             override fun createInstanceOf(clazz: KClass<*>): RealmObjectInternal = (
--- a/packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
@@ -68,6 +68,7 @@ public class RealmConfiguration private constructor(
     writeDispatcher: CoroutineDispatcher,
     schemaVersion: Long,
     deleteRealmIfMigrationNeeded: Boolean,
+    encryptionKey: ByteArray?,
 ) {
     // Public properties making up the RealmConfiguration
     // TODO Add more elaborate KDoc for all of these
@@ -121,6 +122,11 @@ public class RealmConfiguration private constructor(
      */
     public val deleteRealmIfMigrationNeeded: Boolean
 
+    /**
+     * 64 byte key used to encrypt and decrypt the Realm file.
+     */
+    public val encryptionKey: ByteArray?
+
     // Internal properties used by other Realm components, but does not make sense for the end user to know about
     internal var mapOfKClassWithCompanion: Map<KClass<out RealmObject>, RealmObjectCompanion>
     internal var mediator: Mediator
@@ -143,6 +149,7 @@ public class RealmConfiguration private constructor(
         this.writeDispatcher = writeDispatcher
         this.schemaVersion = schemaVersion
         this.deleteRealmIfMigrationNeeded = deleteRealmIfMigrationNeeded
+        this.encryptionKey = encryptionKey?.copyOf()
 
         RealmInterop.realm_config_set_path(nativeConfig, this.path)
 
@@ -207,7 +214,8 @@ public class RealmConfiguration private constructor(
         singleThreadDispatcher(name),
         singleThreadDispatcher(name),
         0,
-        false
+        false,
+        null,
     )
 
     /**
@@ -228,6 +236,7 @@ public class RealmConfiguration private constructor(
         private var writeDispatcher: CoroutineDispatcher? = null
         private var deleteRealmIfMigrationNeeded: Boolean = false
         private var schemaVersion: Long = 0
+        private var encryptionKey: ByteArray? = null
 
         /**
          * Sets the absolute path of the realm file.
@@ -340,6 +349,12 @@ public class RealmConfiguration private constructor(
             apply { this.schemaVersion = validateSchemaVersion(schemaVersion) }
 
         /**
+         * Sets the 64 byte key used to encrypt and decrypt the Realm file.
+         */
+        fun encryptionKey(encryptionKey: ByteArray) =
+            apply { this.encryptionKey = validateEncryptionKey(encryptionKey) }
+
+        /**
          * TODO Evaluate if this should be part of the public API. For now keep it internal.
          *
          * Removes the default system logger from being installed. If no custom loggers have
@@ -376,7 +391,8 @@ public class RealmConfiguration private constructor(
                 notificationDispatcher ?: singleThreadDispatcher(name),
                 writeDispatcher ?: singleThreadDispatcher(name),
                 schemaVersion,
-                deleteRealmIfMigrationNeeded
+                deleteRealmIfMigrationNeeded,
+                encryptionKey
             )
         }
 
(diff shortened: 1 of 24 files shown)


packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -121,6 +122,11 @@ public class RealmConfiguration private constructor(
      */
     public val deleteRealmIfMigrationNeeded: Boolean
 
+    /**
+     * 64 byte key used to encrypt and decrypt the Realm file.
+     */
+    public val encryptionKey get() = RealmInterop.realm_config_get_encryption_key(nativeConfig)
```
cmelchior, line 128: We should probably make the return type explicit here for better discoverability.

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -339,6 +351,12 @@ public class RealmConfiguration private constructor(
         fun schemaVersion(schemaVersion: Long) =
             apply { this.schemaVersion = validateSchemaVersion(schemaVersion) }
 
+        /**
+         * Sets the 64 byte key used to encrypt and decrypt the Realm file.
```
cmelchior, line 355: We should probably expand this description a little bit with advice on how to do this best, see e.g https://github.com/realm/realm-core/issues/4683.

Maybe something like:
```
Sets a 64 byte key that will be used to encrypt and decrypt the Realm file. 

It is important that this key is created and stored securely. See [this link](https://docs.mongodb.com/realm/sdk/android/advanced-guides/encryption/) for suggestions on how to do that.
```

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -386,5 +405,12 @@ public class RealmConfiguration private constructor(
             }
             return schemaVersion
         }
+
+        private fun validateEncryptionKey(encryptionKey: ByteArray): ByteArray {
+            if (encryptionKey.size < Realm.ENCRYPTION_KEY_LENGTH) {
```
cmelchior, line 410: We should check for equality, not just lesser than?

test/src/androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.realm.RealmConfiguration
+import io.realm.util.PlatformUtils
+import test.Sample
+import kotlin.random.Random
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertFails
+
+class EncryptionTests {
```
cmelchior, line 29: Maybe put some doc explaining what this class exactly tests compared to RealmConfigurationTests?

test/src/androidTest/kotlin/io/realm/shared/RealmConfigurationTests.kt
```diff
@@ -241,4 +244,33 @@ class RealmConfigurationTests {
...
+            .encryptionKey(encryptionKey)
+            .build()
+
+        // Validate that the key stored in core is the same that the one we provided
+        assertContentEquals(encryptionKey, config.encryptionKey)
+    }
+
+    @Test
+    fun wrongEncryptionKeyThrowsIllegalArgumentException() {
+        val encryptionKey = Random.nextBytes(8)
```
cmelchior, line 268: Should also test for longer keys

cmelchior (approved):
Nice 👏 I only had a few minor comments

test/src/androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1,82 @@
...
+
+        // Initializes an encrypted Realm
+        Realm(encryptedConf).close()
+
+        // Should be possible to reopen an encrypted Realm
+        Realm(encryptedConf).close()
+    }
+
+    @Test
+    fun openEncryptedRealmWithWrongKey() {
```
rorbech, line 61: Do we also need the opposite test? Opening an unencrypted realm with a key.

test/src/macosTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1 @@
+../../../../../androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```
rorbech, line 1: I thought the whole `shared` directory was symlinked. Would that be possible?

rorbech (approved):
Same. Only minor comments

test/src/macosTest/kotlin/io/realm/shared/EncryptionTests.kt
clementetb: Yes it is, I am going to address it on another PR

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -121,6 +122,11 @@ public class RealmConfiguration private constructor(
      */
     public val deleteRealmIfMigrationNeeded: Boolean
 
+    /**
+     * 64 byte key used to encrypt and decrypt the Realm file.
```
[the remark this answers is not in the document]
cmelchior: This doesn't seem to have been addressed?

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -339,6 +351,17 @@ public class RealmConfiguration private constructor(
         fun schemaVersion(schemaVersion: Long) =
             apply { this.schemaVersion = validateSchemaVersion(schemaVersion) }
 
+        /**
+         * Sets the 64 byte key used to encrypt and decrypt the Realm file. Not setting an encryption key would
```
cmelchior, line 355:
```suggestion
         * Sets the 64 byte key used to encrypt and decrypt the Realm file. If no key is provided the Realm file 
```

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -339,6 +351,17 @@ public class RealmConfiguration private constructor(
...
+         * default to an unencrypted Realm.
```
cmelchior, line 356:
```suggestion
         * will be unencrypted.
```

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -339,6 +351,17 @@ public class RealmConfiguration private constructor(
...
+         *
+         * It is important that this key is created and stored securely. See [this link](https://docs.mongodb.com/realm/sdk/android/advanced-guides/encryption/) for suggestions on how to do that.
+         *
+         * @param encryptionKey 64-byte key encryption key.
```
cmelchior, line 360:
```suggestion
         * @param encryptionKey 64-byte encryption key.
```

packages/library/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -386,5 +410,12 @@ public class RealmConfiguration private constructor(
             }
             return schemaVersion
         }
+
+        private fun validateEncryptionKey(encryptionKey: ByteArray): ByteArray {
+            if (encryptionKey.size != Realm.ENCRYPTION_KEY_LENGTH) {
+                throw IllegalArgumentException("The provided key must be ${Realm.ENCRYPTION_KEY_LENGTH} bytes.")
```
cmelchior, line 416:
```suggestion
                throw IllegalArgumentException("The provided key must be ${Realm.ENCRYPTION_KEY_LENGTH} bytes. The provided key was ${encryptionKey.size} bytes.")
```

We should always try to give as much information as possible in these crash messages. It becomes much easier to debug.

test/src/androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1,121 @@
...
+                .path("$tmpDir/default.realm")
+                .encryptionKey(randomKey)
+                .build()
+
+            Realm(conf)
+        }
+    }
+
+    @Test
+    fun openUnencryptedRealmWithWrongKey() {
```
cmelchior, line 100:
```suggestion
    fun openUnencryptedRealmWithEncryptionKey() {
```

test/src/androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1,121 @@
...
+            .Builder(
+                path = "$tmpDir/default.realm",
+                schema = setOf(Sample::class)
+            )
+            .build()
+        Realm(unencryptedConf).close()
+
+        // Assert fails opening with encryption key
+        val randomKey = Random.nextBytes(64)
+        assertFails {
```
cmelchior, line 112: We should probably use `assertFailsWith(XXX::class)` here to provide more context about exactly how it fails.

test/src/androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1,121 @@
...
+            val conf = RealmConfiguration.Builder(schema = setOf(Sample::class))
+                .path("$tmpDir/default.realm")
+                .build()
+            Realm(conf)
+        }
+
+        // Assert fails with wrong encryption key
+        val randomKey = Random.nextBytes(64)
+        assertFails {
```
cmelchior, line 89: We should probably use assertFailsWith(XXX::class) here to provide more context about exactly how it fails.

test/src/androidTest/kotlin/io/realm/shared/EncryptionTests.kt
```diff
@@ -0,0 +1,121 @@
...
+            .Builder(
+                path = "$tmpDir/default.realm",
+                schema = setOf(Sample::class)
+            )
+            .encryptionKey(actualKey)
+            .build()
+        Realm(encryptedConf).close()
+
+        // Assert fails with no encryption key
+        assertFails {
```
cmelchior, line 80: We should probably use assertFailsWith(XXX::class) here to provide more context about exactly how it fails.

cmelchior (approved):
A few minor comments around the tests and some docs, otherwise 🚀

PR #367, opened 2021-08-04
Update to Java 11

### LINKED ISSUE #366: Upgrade to Java 11 / AGP 7 / Gradle 7
With Artic Fox, Java 11 is now the default java version. 

Since there is quite a few breaking changes between Java 8 and Java 11, it seems easier to just require Java 11 for launch. If there is sufficient pressure we can consider adding support for previous versions.

This includes:
- Upgrade to use Java 11 to build on CI (and fix any issues)
- Upgrade to use Android Gradle Plugin 7.0.0
- Upgrade to use Gradle 7.1.1.

The corresponding Java issue is here: https://github.com/realm/realm-java/pull/7535

cmelchior:
Closes #366 

Update Gradle to 7.1.1.
Update Android Gradle Plugin to 4.1.0. Upgrading past this is not possible due to various bugs (reasons added in Config.kt).
Update CI to use Java 11

**TODO**
- [x] Update README to clarify requirement about Java 11
- [x] Update CHANGELOG
- [x] Verify that building with Java 8 still works.


Jenkinsfile
```diff
@@ -76,6 +79,15 @@ pipeline {
                 runBuild()
             }
         }
+        stage('Build Android on Java 8') {
```
rorbech, line 82: Would it make sense to move it near the `Tests Android Sample App` stage. Seems more like a safe guard than something we should verify early. At least it fails more often on the `Static Analysis` task on my PRs(🤦), so I'd prefer to postpone this test for later.

packages/cinterop/build.gradle.kts
```diff
@@ -65,40 +65,42 @@ val corePath = "external/core"
...
         ndk {
             abiFilters += setOf("x86_64", "arm64-v8a")
         }
+
         // Out externalNativeBuild (outside defaultConfig) does not seem to have correct type for setting cmake arguments
         externalNativeBuild {
             cmake {
                 arguments("-DANDROID_STL=c++_shared")
+                targets.add("realmc")
+                println("targets: $targets")
```
rorbech, line 95: Is this needed?

rorbech (approved):
LGTM. Only minor comments.

Jenkinsfile
cmelchior: Yes, that makes sense. I'll move it to the end.

PR #369, opened 2021-08-05
Linked shared test folder instead of files

clementetb:
Symlink the shared platform test folder instead of individual tests.


cmelchior (approved):
Nice 🥇

rorbech (approved):

PR #370, opened 2021-08-09
Clean up multiplatform configuration

### LINKED ISSUE #358: Clean up KMM configuration
We should clean up the KMP module configuration, leaving the old pattern of keeping separate platform configurations closures. 

There are also still issues with unresolved symbols for some platform specific parts in the cinterop/library modules.

rorbech:
The multiplatform source sets of the SDK should be organized into the following hierarchy:
```
- common
  - jvm
    - jvmMain
    - androidMain
  - native
    - darwin
      - ios
        - iosX64Main
        - iosArm64Main
      - macosX64
```

We dont need hierarchical setup per se as our API is available to all platforms, but would be nice to allow writing common code for JVM/Android and all Darwin platforms. The following issues currently prevents this: 
- Common symbols not recognized by IDE in consuming Android source sets with composite project setup - https://youtrack.jetbrains.com/issue/KT-48153
- Cannot share implementation across JVM and Android targets - https://youtrack.jetbrains.com/issue/KT-42466
  According to https://kotlinlang.org/docs/mpp-share-on-platforms.html#configure-the-hierarchical-structure-manually
  "Kotlin doesn’t currently support sharing a source set for these combinations:
     Several JVM targets
     JVM + Android targets
     Several JS targets"
- Commonizer cannot be applied at Darwin level - https://youtrack.jetbrains.com/issue/KT-40975 (targeted in 1.5.30-M1, but not verified)
  According to https://github.com/JetBrains/kotlin/blob/1.4.0/native/commonizer/README.md 
  "It works only for a native source set that is shared among platform-specific source sets and these source sets themselves. It doesn’t work for native source sets shared at higher levels of the source set hierarchy. For example, if you have nativeDarwinMain that includes iosMain with iosArm64Main and iosX64Main, and watchosDeviceMain with watchosArm64Main and watchosArm32Main, the KLIB Commonizer will work for separately iosMain and watchosDeviceMain and won’t work for nativeDarwinMain."

Currently used workarounds:
- Since we don't have platform specific APIs we don't per se need the hierarchical setup
- When platform share code without any differentiation, just add folders directly to all source set that needs it instead of defining intermediate source set
- Live with the rest. Platform symbols in `darwinCommon` are not resolved

TODO:
- [x] Rename file structure according to https://kotlinlang.org/docs/mpp-share-on-platforms.html#share-code-in-libraries. Should we follow standard layout with `<target>Main/Test` (and how should common `jvm` be differentiated from target `jvm`) or use `coroutine` convention of having `src` and `test` inside target directories instead (https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-core)
  - Renamed according to the above shown hierarchy. 
- [x] Document outcome into README
- [x] Clean up `PlatformHelper.kt` - Some methods are scope to the `PlatformHelper` while others live in the global scope. We cannot differentiate implementation of `PlatformHelper` across multiple source sets, so either we should have all methods in the global scope or we would need to duplicate code into the platform specific source sets!? (Duplicating the code would also fix the unresolved commonized symbols 😅 ) 
   - Eliminated `PlatformHelper` object but moved all platform differentiated implementations into `platform` sub packages. 

Closes #358


cmelchior (approved):
Great improvement 🥇 🚀 :shipit:

README.md
```diff
@@ -276,6 +276,28 @@ We use the offical [style guide](https://kotlinlang.org/docs/reference/coding-co
...
+          |   ├── iosArm64Main
+          |   └── iosX64Main
+          └── macosX64Main
+```
+
+All source sets ending with `Main` is platform specific source sets, while the others are intermediate source sets shared between multiple targets. Only exception is `commonMain` which is kept to follow the Kotlin MPP gradle convention.
+
+It is currently not possible to enable hierarchical setup due to various issues rendering the IDE unable to resolve common symbols, so for now we are just adding shared source sets to the individual platform specific targets they belong to. (Issues to track: https://youtrack.jetbrains.com/issue/KT-48153, https://youtrack.jetbrains.com/issue/KT-42466, https://youtrack.jetbrains.com/issue/KT-40975, see description of https://github.com/realm/realm-kotlin/pull/370 for details).
+
+All platform differentitated implementations is kept in `platform`-packages with their current package hierarchi, to make it easier to keep track of the level of platform differentiation.
```
clementetb, line 300:
```suggestion
All platform differentiated implementations are kept in `platform`-packages with their current package hierarchy, to make it easier to keep track of the level of platform differentiation.
```

clementetb (approved):
🧙🏻‍♂️ Good work!

rorbech:
As highlighted in the description there are still open issues, so will settle with the current clean up. It doesn't automatically resolve all symbols as some has to be built on all platforms and the composite build needs the updated intermediate metadata descriptors. Current trick to resolve the cinterop/library symbols all over is to build all SDK library artifacts 
(`./gradlew assemble` in `packages`) and trigger at least a single test in the root project (`test`). 

Will have to revisit when there are updates to any of the issues listed here.

PR #371, opened 2021-08-09
RealmList - notifications as flows

edualonso:
Added support for list notifications as flows using `list.observe()`.
Internal support for freezing/thawing lists.


packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -138,6 +139,27 @@ internal class SuspendableNotifier(private val owner: Realm, private val dispatc
         }
     }
 
+    /**
+     * Listen to changes to a RealmList through a [Flow]. If the list is deleted, `null` is emitted and the flow will complete.
```
cmelchior, line 143: This comment doesn't match the method signature as it is non-nullable. Not 100% sure which of them needs to change.

packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
```diff
@@ -92,34 +103,48 @@ class RealmList<E> private constructor(
...
+    fun thaw(liveRealm: RealmReference): RealmList<E>
+}
+
 /**
  * Represents an unmanaged [RealmList] backed by a [MutableList] via class delegation.
  */
 private class UnmanagedListDelegate<E>(
     list: MutableList<E> = mutableListOf()
-) : MutableList<E> by list
+) : MutableList<E> by list, RealmListApi<E> {
```
rorbech, line 134: Couldn't we guard about reaching this situation further out the call stack if we just checked if the list was managed or unmanaged when observing? I guess the whole observability only needs to be implemented for managed lists. 

Maybe an `Observable` interface with `observe(): Flow<T>` that we could reuse on all observables (result, object, lists) to generically dispatch to the notifier.

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
cmelchior: If I remember correctly, we changed it to the Flow just being closed instead of emitting `null`

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
```diff
@@ -225,6 +230,20 @@ class RealmListTests {
         managedTesters[0].setFailsIfClosed(getCloseableRealm())
     }
 
+    @Test
```
cmelchior, line 233: Missing test for what happens if the object owning the list is deleted.

packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
rorbech: Maybe we should find a way to ensure by type that it was only the managed list that was passed around in the internal implementation.

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
```diff
@@ -806,6 +827,91 @@ internal abstract class ManagedListTester<T>(
...
+
+            runBlocking {
+                val channel = Channel<RealmList<*>>(capacity = 1)
+                val observer = async {
+                    val container = realm.objects<RealmListContainer>()
+                        .first()
+
+                    typeSafetyManager.getList(container)
+                        .observe()
+                        .onEach { flowList ->
```
cmelchior, line 847: Why sending this as a side effect instead of inside the `collect` method. IMO that would be more readable:

```
typeSafetyManager.getList(container)
                        .observe()
                        .collect { flowList ->
                             channel.send(flowList)
                         }
```

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
```diff
@@ -806,6 +827,91 @@ internal abstract class ManagedListTester<T>(
...
+
+            runBlocking {
+                val channel = Channel<RealmList<*>>(capacity = 2)
+                val observer = async {
+                    val container = realm.objects<RealmListContainer>()
+                        .first()
+
+                    typeSafetyManager.getList(container)
+                        .observe()
+                        .onEach { flowList ->
```
cmelchior, line 881: Same

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
```diff
@@ -806,6 +827,91 @@ internal abstract class ManagedListTester<T>(
...
+                            println("--- ONEACH SIZE: ${flowList.size}")
```
cmelchior, line 882: Debug log that can be removed?

cmelchior (changes requested):
Nice 💯 ... A changelog entry is missing and it seems we are not testing the case where the object owning the list is missing. Otherwise only a few minor comments.

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -34,19 +35,26 @@ import kotlinx.coroutines.withContext
...
     companion object {
         val NO_OP_NOTIFICATION_TOKEN = object : Cancellable {
-            override fun cancel() { /* Do Nothing */ }
+            override fun cancel() { /* Do Nothing */
+            }
         }
     }
 
     // FIXME Work-around for the global Realm changed listener not working.
     // FIXME Adding extra buffer capacity as we are otherwise never able to emit anything. Needs to
```
rorbech, line 51: I guess we can remove the fixme to investigate. Reason for the buffer capacity is in the docs: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt#L78

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -138,6 +139,27 @@ internal class SuspendableNotifier(private val owner: Realm, private val dispatc
         }
     }
 
+    /**
```
rorbech, line 142: This looks extremely like the other implementations. Can't we somehow generic-alize into `<T: Observable> changed(t: T): Flow<T> { ... }` and put the differentiation into the `Observables` (object, results, lists).

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -245,6 +270,42 @@ internal class SuspendableNotifier(private val owner: Realm, private val dispatc
         return NotificationToken(callback, token)
     }
 
+    /**
+     * Listen to changes to a RealmList through a change listener. The callback will happen
+     * on the configured [SuspendableNotifier.dispatcher] thread.
+     *
+     * FIXME Callers of this method must make sure it is called on the correct [SuspendableNotifier.dispatcher].
+     */
+    internal fun <T> registerListChangedListener(
```
rorbech, line 279: Again exactly the same code. I don't see anything specific for lists. Can't we generic-alize it.

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
rorbech, line 882: Debug?

test/src/androidTest/kotlin/io/realm/shared/RealmListTests.kt
```diff
@@ -806,6 +827,91 @@ internal abstract class ManagedListTester<T>(
...
+
+                            channel.send(flowList)
+
+                            // Update list if the flow list is empty
+                            if (flowList.isEmpty()) {
```
rorbech, line 887: Why is this block in here. Couldn't you just do the updates after you have asserted on the empty list below. The idea with the channel-pattern is that you can write actions and assertions sequentially.

rorbech (commented):
Cool. 

Highlevel comments:
- Missing changelog entry ;)
- It would be more for an interface for the observable stuff (`Observer` (object, list, results) and `OberservableRealm` (Realm) and then try to generalize the implementation. We have a huge replica of code. 
- Also somehow separate the managed list in the type system, so we didn't have to guard for the unmanaged in the implementation besides the entry points for specific behavior (`observe`) 

Don't know if it all makes sense. Maybe consider and ping me for a chat.

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
edualonso: Makes sense. It looks like it was scaffolding copy/paste from the method above for listening to Realm objects 🙊

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -120,14 +120,17 @@ public abstract class BaseRealm internal constructor(
         throw NotImplementedError(observablesNotSupportedMessage)
     }
 
-    internal open fun <T : RealmObject> registerResultsObserver(results: RealmResults<T>): Flow<RealmResults<T>> {
+    internal open fun <T : RealmObject> registerResultsObserver(
```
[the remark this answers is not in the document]
cmelchior: Having different error messages is pretty important IMO.... especially for the case of MutableRealm as we don't allow change listeners inside write transactions. I think that needs to be spelled out explicitly in the error, but maybe there is a better way to support this 🤔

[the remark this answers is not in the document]
rorbech: Maybe if we generalized the methods with `<T: Observable> registerObserver(t: T): Flow<T>` it wouldn't feel as bloated.

packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
edualonso: I recall a conversation where it sounded to me like we had mixed feelings about the use of interfaces like `RealmObservable`, `RealmFreezable`, etc., that's why I didn't do it. But I 100% agree with the approach.

cmelchior: Maybe we should merge this PR as is, and then discuss this abstraction while we are looking at the other interfaces?

rorbech: Yep. Think it would be easier to play around if we rearrange the interfaces in one PR. We just have to remember all of the _point of interests_.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,208 @@ class RealmListNotificationsTests : NotificationTests {
...
-    @Ignore
     override fun initialElement() {
-        TODO("Waiting for RealmList support")
+        val dataSet = OBJECT_VALUES
+
+        realm.writeBlocking {
+            val managedContainer = copyToRealm(RealmListContainer())
+            val managedDataSet = dataSet.map { copyToRealm(it) }
+            managedContainer.objectListField
+                .addAll(managedDataSet)
```
rorbech, line 57: Couldn't you just add the unmanaged set?

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,208 @@ class RealmListNotificationsTests : NotificationTests {
...
+            val observer = async {
+                val container = realm.objects<RealmListContainer>()
+                    .first()
+
+                container.objectListField
+                    .observe()
+                    .collect { flowList ->
+                        channel.send(flowList)
+
+                        realm.writeBlocking {
```
rorbech, line 103: Why not do the write imperatively in between the asserts instead of here?

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,208 @@ class RealmListNotificationsTests : NotificationTests {
...
+
+            // Assert container got populated correctly
+            val emittedList = channel.receive()
+            assertNotNull(emittedList)
+            assertEquals(OBJECT_VALUES.size, emittedList.size)
+
+            realm.write {
+                delete(findLatest(managedContainer)!!)
+            }
+
```
rorbech, line 221: What are we asserting here? Can't recall whether we aligned on just closing it (assert something about no events) or actually completing it (assert that it is completed with onCompletion).

rorbech (commented):
Some minor questions on the tests.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
edualonso: That's right, I'm still getting used to using channels as blocking mechanisms 🐒

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
edualonso: We can't add unmanaged objects to a managed list as per the restriction we have in `RealmInterop.to_realm_value`.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
edualonso: My assumption is that given the owner object is deleted the list will not be valid anymore and therefore shouldn't be emitted at all - since we emit frozen lists, freezing an invalid list would throw, so there is no assertion on anything. I guess that makes sense...?

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
rorbech: But shouldn't we just import them like we do when assigning fields in https://github.com/realm/realm-kotlin/blob/master/packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt#L107

packages/library/src/commonMain/kotlin/io/realm/internal/RealmUtils.kt
```diff
@@ -105,6 +108,31 @@ fun <T : RealmObject> create(
     }
 }
 
+fun <T> copyToRealmIfNeeded(
```
rorbech, line 111: Shouldn't this just not just be collapsed with `copyToRealm`? I think the validity check could might as well be in `copyToRealm` and `copyToRealm` already returns the same element if it is already managed. So looks like you should just also check if it actually is an `RealmObjectInternal` and otherwise return the element. Then we would have multiple methods with almost same functionality.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
rorbech: But isn't it a bit like https://github.com/realm/realm-kotlin/blob/master/test/src/androidTest/kotlin/io/realm/shared/notifications/RealmObjectNotificationsTests.kt#L128, where deletion is signalled by _completion_? @cmelchior, would it make sense to complete the flow for list on deletion of an object too?

rorbech (commented):
I still can't settle on whether deletion of the object should somehow terminate the flow, but otherwise only a single comment.

rorbech:
Also missing a changelog entry on the ability to import objects by adding them to a managed list.

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -176,73 +177,88 @@ internal class SuspendableNotifier(private val owner: Realm, private val dispatc
...
+        list: RealmList<T>,
+        callback: Callback<RealmList<T>>
+    ): Cancellable {
+        val liveList = list.thaw(realm.realmReference)
+        return registerChangedListener(
+            liveComponentPointer = liveList.listPtr,
+            notifyComponentUpdate = { frozenRealm ->
+                // Skip notifications when list is not valid
+                if (!liveList.isValid()) {
+                    return@registerChangedListener
```
cmelchior, line 249: The behavior here is different than for Objects. We should make them consistent. I.e. also return `null` here. That will probably result in the flow completing further up, but we will also have a normal callback based API on top of this, where we probably want to expose the `null` directly.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
     @Test
-    @Ignore
     override fun initialElement() {
-        TODO("Waiting for RealmList support")
+        val dataSet = OBJECT_VALUES
+
+        realm.writeBlocking {
+            val managedContainer = copyToRealm(RealmListContainer())
+            managedContainer.objectListField
+                .addAll(dataSet)
```
cmelchior, line 56: Nitpick: Why not on the above line as above?

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
+        }
+
+        runBlocking {
+            val channel = Channel<RealmList<*>>(capacity = 1)
+            val observer = async {
+                val container = realm.objects<RealmListContainer>()
+                    .first()
```
cmelchior, line 63: Nitpick: Same?

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
+
+        realm.writeBlocking {
+            // Just create an empty container with empty lists
+            copyToRealm(RealmListContainer())
+        }
+
+        runBlocking {
+            val channel = Channel<RealmList<*>>(capacity = 1)
+            val observer = async {
+                val container = realm.objects<RealmListContainer>()
```
cmelchior, line 94: It seems safer to use the result from the `writeBlocking` block?

```
val container = realm.writeBlocking {
            // Just create an empty container with empty lists
            copyToRealm(RealmListContainer())
        }

        runBlocking {
            val channel = Channel<RealmList<*>>(capacity = 1)
            val observer = async {
                container.objectListField.
            }
```

Here and elsewhere

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
+            }
+
+            // Assertion after empty list is emitted
+            val firstEmittedList = channel.receive()
+            assertNotNull(firstEmittedList)
+            assertEquals(0, firstEmittedList.size)
+
+            // Trigger update
+            realm.writeBlocking {
+                val queriedContainer = objects<RealmListContainer>()
```
cmelchior, line 111: You can use `findLatest(container)` if using the result from writeBlocking

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
+            }
+
+            // Ignore first emission with empty lists
+            channel1.receive()
+            channel2.receive()
+
+            // Trigger an update
+            realm.write {
+                val objects = objects<RealmListContainer>()
+                val queriedContainer = objects
```
cmelchior, line 160: You can use `findLatest` instead here.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
+            }
+            assertEquals(OBJECT_VALUES.size, channel1.receive().size)
+            assertEquals(OBJECT_VALUES.size, channel2.receive().size)
+
+            // Cancel observer 1
+            observer1.cancel()
+
+            // Trigger another update
+            realm.write {
+                val queriedContainer = objects<RealmListContainer>()
```
cmelchior, line 173: Same

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
```diff
@@ -33,38 +47,207 @@ class RealmListNotificationsTests : NotificationTests {
...
+                        objectListField.addAll(values.map { copyToRealm(it) })
+                    }.let { container ->
+                        copyToRealm(container)
+                    }
+            }
+            val observer = async {
+                managedContainer.objectListField
+                    .observe()
+                    .collect { flowList ->
+                        channel.send(flowList)
```
cmelchior, line 208: This doesn't seem to check any behavior for when the List is deleted?

cmelchior (changes requested):
Nice refactors. I have a few nitpicks around code style. I assume that ktlint doesn't enforce anything, but we probably shouldn't change unrelated style things as part of another PR.

But more importantly, it seems that the case of the object holding the RealmList isn't really tested and thus the behaviour is undefined. From the implementation, it looks like nothing happens. For Objects we complete the flow when that happens.

test/src/androidTest/kotlin/io/realm/shared/notifications/RealmListNotificationsTests.kt
edualonso: I personally find it easier to have an overview of the entire call when things are in separate lines, my eye-brain interaction is much smoother 🤷 I think this may look/feel weird in case of accessing properties instead of method calls though.

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -34,19 +35,25 @@ import kotlinx.coroutines.withContext
...
+internal class SuspendableNotifier(
+    private val owner: Realm,
+    private val dispatcher: CoroutineDispatcher
+) {
 
     companion object {
         val NO_OP_NOTIFICATION_TOKEN = object : Cancellable {
-            override fun cancel() { /* Do Nothing */ }
+            override fun cancel() { /* Do Nothing */
+            }
```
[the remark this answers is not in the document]
edualonso: Cmd + option + L sometimes does weird things 🤷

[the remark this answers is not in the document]
cmelchior: I would be fine with us just using whatever Kotlin should be the default style 👍 So feel free to ignore. Would be nice if we could autoformat as part of a commit hook though. But obviously not part of this PR.

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -176,73 +181,88 @@ internal class SuspendableNotifier(private val owner: Realm, private val dispatc
...
-                    }
+    internal fun <T> registerListChangedListener(
+        list: RealmList<T>,
+        callback: Callback<RealmList<T>?>
+    ): Cancellable {
+        val liveList = list.thaw(realm.realmReference)
+        return registerChangedListener(
+            liveComponentPointer = liveList.listPtr,
+            notifyComponentUpdate = { frozenRealm ->
+                // Close channel when list is not valid
```
cmelchior, line 251: This part of the code doesn't really know about Flows, so this comment is a bit misleading here.

cmelchior (approved):
Nice 🥇 

Ktlint seems to have a few complaints that need to be addressed, but those seem like minor things.

PR #372, opened 2021-08-09
Move non-cinterop fields from RealmObjectInterop to RealmObjectInternal

rorbech:


edualonso (changes requested):
LGTM - but `GenerationExtensionTest` is failing when comparing the IR output

cmelchior (approved):
When CI is happy 👍

PR #376, opened 2021-08-17
Initial split of public classes into interface and implementation

rorbech:


edualonso (changes requested):
Looks amazing 🎸 

I haven't gone crazy in-depth on the whole thing since it mostly is about moving logic from A to B. I've written minor comments with regards to documentation (there are lots of overridden methods documented both in their respective interfaces and in their `-Impl` classes, I guess side effect from massive copy/paste-ing) and a couple of Kotlin-related syntax issues.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -125,9 +63,7 @@ class Realm private constructor(configuration: RealmConfiguration, dbPointer: Na
      * @param clazz The class of the objects to query for.
      * @return The result of the query as of the time of invoking this method.
      */
-    override fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T> {
-        return super.objects(clazz)
-    }
+    override fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T>
```
cmelchior, line 66: Is there a reason we override the interface method here? Is it just for discoverability when browsing the source code or does it provide some other benefit? 

(I'm not against doing it just for discoverability, but then we should be consistent and it for all methods in BaseRealm)

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
```diff
@@ -1,91 +1,19 @@
...
-        internal fun <T : RealmObject> fromQuery(realm: RealmReference, query: NativePointer, clazz: KClass<T>, schema: Mediator): RealmResults<T> {
-            // realm_query_find_all doesn't fully evaluate until you interact with it.
-            return RealmResults(realm, RealmInterop.realm_query_find_all(query), clazz, schema)
-        }
-
-        internal fun <T : RealmObject> fromResults(realm: RealmReference, results: NativePointer, clazz: KClass<T>, schema: Mediator): RealmResults<T> {
-            return RealmResults(realm, results, clazz, schema)
-        }
-    }
+interface RealmResults<T : RealmObject> : List<T>, Queryable<T> {
```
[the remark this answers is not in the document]
cmelchior: I'm not sure that is correct. It seems that Queryable is the common wording? https://www.google.com/search?q=queriable+vs+queryable&oq=quer&aqs=chrome.0.69i59l2j69i57j0i512l3j69i61j69i60.1080j0j7&sourceid=chrome&ie=UTF-8

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
```diff
@@ -103,26 +31,7 @@ class RealmResults<T : RealmObject> : AbstractList<T>, Queryable<T> {
...
-     *
-     * Observe changes to a Realm result.
-     *
-     * Follows the pattern of [Realm.addChangeListener]
-     */
-    internal fun addChangeListener(callback: Callback<RealmResults<T>>): Cancellable {
-        realm.checkClosed()
-        return realm.owner.registerResultsChangeListener(this, callback)
-    }
+    override fun query(query: String, vararg args: Any): RealmResults<T>
```
cmelchior, line 34:
```suggestion
    override fun query(query: String, vararg args: Any?): RealmResults<T>
```

I think?

packages/library/src/commonMain/kotlin/io/realm/internal/BaseRealmImpl.kt
```diff
@@ -0,0 +1,130 @@
...
+import io.realm.RealmObject
+import io.realm.RealmResults
+import io.realm.VersionId
+import io.realm.interop.NativePointer
+import io.realm.interop.RealmInterop
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+@Suppress("UnnecessaryAbstractClass")
+public abstract class BaseRealmImpl internal constructor(
```
cmelchior, line 31: Any reason this is public?

packages/library/src/commonMain/kotlin/io/realm/internal/RealmConfigurationImpl.kt
```diff
@@ -0,0 +1,196 @@
...
+    internal val nativeConfig: NativePointer = RealmInterop.realm_config_new()
+
+    init {
+        this.path = if (path == null || path.isEmpty()) {
+            val directory = PlatformHelper.appFilesDirectory()
+            // FIXME Proper platform agnostic file separator: File.separator is not available for Kotlin/Native
+            //  https://github.com/realm/realm-kotlin/issues/75
+            "$directory/$name"
+        } else path
+        this.name = name // FIXME Should read name from end of path
```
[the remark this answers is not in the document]
cmelchior: Yes, but I put all of the logic here as it otherwise gets spread out across too many places. Having all the logic in one place makes it easier to reason about. At least that was the reason for doing it this way.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmResultsImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+import io.realm.interop.RealmInterop
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+// FIXME API-QUERY Final query design is tracked in https://github.com/realm/realm-kotlin/issues/84
+//  - Lazy API makes it harded to debug
+//  - Postponing execution to actually accessing the elements also prevents query parser errors to
+//    be raised. Maybe we can get an option to prevalidate queries in the C-API?
+
+class RealmResultsImpl<T : RealmObject> : AbstractList<T>, RealmResults<T> {
```
cmelchior, line 35:
```suggestion
internal class RealmResultsImpl<T : RealmObject> : AbstractList<T>, RealmResults<T> {
```

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -34,7 +31,7 @@ import kotlinx.coroutines.withContext
  * inside writes. Users can therefor not register change listeners on live objects, but it is assumed that other
  * layers check that invariant before methods on this class are called.
  */
-internal class SuspendableNotifier(private val owner: Realm, private val dispatcher: CoroutineDispatcher) {
+internal class SuspendableNotifier(private val owner: RealmImpl, private val dispatcher: CoroutineDispatcher) {
```
cmelchior, line 34: Couldn't we just keep the Realm interface here?

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -49,9 +46,9 @@ internal class SuspendableNotifier(private val owner: Realm, private val dispatc
     private val _realmChanged = MutableSharedFlow<RealmReference>(onBufferOverflow = BufferOverflow.SUSPEND, extraBufferCapacity = 1)
 
     // Must only be accessed from the dispatchers thread
-    private val realm: BaseRealm by lazy {
+    private val realm: BaseRealmImpl by lazy {
```
cmelchior, line 49: Same. Is there a reason we are restricting these to the implementation class? or just a side-effect of refactoring in Android Studio?

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
```diff
@@ -153,15 +153,15 @@ class RealmSchemaLoweringExtension : IrGenerationExtension {
  *
  * E.g. all of these are valid
  * ```
- * RealmConfiguration(schema = setOf(MyType::class))
```
cmelchior, line 156: It feels a bit of a shame to completely removing this. Having to always go through the Builder is quite a lot of overhead.

In Java we have a factory constructor like `SyncConfiguration.defaultConfig(....)` 

Doing something similar here would seem nice?

cmelchior (commented):
Really nice... I have left a few comments here and there. Most of it is minor.

I would probably suggest we do something about RealmConfiguration. Always going through the Builder looks really annoying.

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -1,70 +1,35 @@
...
     // TODO Could be abstracted into base implementation of RealmLifeCycle!?
-    public var version: VersionId = VersionId(0)
-        get() { return realmReference.version() }
-
-    internal val log: RealmLog = RealmLog(configuration = configuration.log)
-
-    init {
-        log.info("Realm opened: ${configuration.path}")
-    }
+    var version: VersionId
```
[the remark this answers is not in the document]
rorbech: Yep. It's in #383

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -76,88 +41,25 @@ public abstract class BaseRealm internal constructor(
...
-        throw NotImplementedError(observablesNotSupportedMessage)
-    }
-    internal open fun <T : RealmObject> registerListObserver(list: List<T>): Flow<List<T>> {
-        throw NotImplementedError(observablesNotSupportedMessage)
-    }
-
-    internal open fun <T : RealmObject> registerObjectObserver(obj: T): Flow<T> {
-        throw NotImplementedError(observablesNotSupportedMessage)
-    }
+    open fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T>
```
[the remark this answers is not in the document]
rorbech: It is overridden in `Realm` and `MutableRealm` to be able to differentiate the docs as the result is frozen for `Realm` but live for `MutableRealm`.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
rorbech: Just to be able to differentiate the docs.

packages/library/src/commonMain/kotlin/io/realm/internal/BaseRealmImpl.kt
rorbech: This PR is targeting an integration branch, but cannot track the other tasks in a PR until the branch differs from master. 

I started to internalize stuff as part of #383. Rest will be handled as part of #56 and #22.  I have added it in the TODOs of this PR and will merge the spill overs into the TODOs of the integration PR.

packages/library/src/commonMain/kotlin/io/realm/Realm.kt
cmelchior: Ok 👍

packages/library/src/commonMain/kotlin/io/realm/internal/RealmResultsImpl.kt
rorbech: Interlization is done in some of the other PRs

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
rorbech: Yeah. Probably nice to keep something like that.

packages/library/src/commonMain/kotlin/io/realm/RealmResults.kt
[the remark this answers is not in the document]
edualonso: I see. It looks odd to me though 🙈

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
rorbech: The internal classes accesses internal stuff. In this case the `owner.configuration.nativeConfig`. This is only exposed because the `RealmImpl` overrides `configuration` offering access to the configuration as a `RealmConfigurationImpl`. 

Alternatively, we need to cast all the public types when passing them to internal classes, but found it easiest to just use the Impl-types as soon as we have them. But technically the internal API should also be separated into an interface and implementation, as we could end up in troubles separating API from implementation at a later stage. 

@edualonso @cmelchior Any better patterns for this?

cmelchior: Hmm, yeah, I can see it being tricky. It is probably fine to leave this as it is? Then we can iterate on it in a later stage?

packages/library/src/commonMain/kotlin/io/realm/BaseRealm.kt
```diff
@@ -76,88 +41,25 @@ public abstract class BaseRealm internal constructor(
...
+}
 
-    // Not all sub classes of `BaseRealm` can be closed by users.
-    internal open fun close() {
-        val reference = realmReference
-        reference.checkClosed()
-        RealmInterop.realm_close(reference.dbPointer)
-        log.info("Realm closed: ${configuration.path}")
-    }
+inline fun <reified T : RealmObject> BaseRealm.objects(): RealmResults<T> {
```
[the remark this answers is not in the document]
rorbech: Added `TypedRealm`.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt
rorbech: Added `RealmConfiguration.defaultConfig(...)`. Implementation just forwards any arguments to the builder constructor `RealmConfiguration.Builder(...)` and returns the built configuration.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -107,9 +107,11 @@ expect object RealmInterop {
     fun <T> realm_list_set(list: NativePointer, index: Long, value: T): T
     fun realm_list_clear(list: NativePointer)
     fun realm_list_erase(list: NativePointer, index: Long)
+    fun realm_list_freeze(liveList: NativePointer, frozenRealm: NativePointer): NativePointer
+    fun realm_list_thaw(frozenList: NativePointer, liveRealm: NativePointer): NativePointer
```
edualonso, line 111: Just noticed these are already in master but not in the detination branch. Have you left it outdated intentionally?

edualonso (commented):
Just one more small comment about the diff with the destination branch.

PR #378, opened 2021-08-17
Sync - Added script and configuration for local sync server

edualonso:


cmelchior (approved):

dependencies.list
```diff
@@ -0,0 +1,3 @@
+# Version of MongoDB Realm used by integration tests
+# See https://github.com/realm/ci/packages/147854 for available versions
+MONGODB_REALM_SERVER=2021-04-22
```
clementetb, line 3: Shall we define the `MONGODB_REALM_SERVER` version? We might be able to use the one defined in Core `dependecies.list`.

clementetb (approved):

dependencies.list
edualonso: I guess that could work, but ATM that file points to `2021-03-24` which is even older than ours.

edualonso: On top of that, we should probably have a look at updating the server at some point but the last few versions don't even boot when running our script 😑

Issue #382, opened 2021-08-17
copyToRealm returns object with default values


shalva97:
okay so this is really weird behavior or im too tired, but anyway I have a test and its failing so here it is

```
import androidx.test.ext.junit.runners.AndroidJUnit4
import io.realm.Realm
import io.realm.RealmConfiguration
import io.realm.RealmObject
import io.realm.delete
import org.junit.Test
import org.junit.runner.RunWith

@RunWith(AndroidJUnit4::class)
class RepoTest {

    class Task : RealmObject {
        var name: String = "new task"
        var status: String = "Open"

        fun toAnotherTask(): AnotherTask {
            return AnotherTask(name, status)
        }
    }

    data class AnotherTask(
        val name: String,
        var status: String
    )

    class TaskRepository {

        private val config = RealmConfiguration(schema = setOf(Task::class))
        private val realm = Realm(config)

        init {
            realm.writeBlocking {
                objects(Task::class).delete()
            }
        }

        fun getAll(): List<Task> {
            return realm.objects(Task::class)
        }

        fun addTask(task: Task): AnotherTask {
            return realm.writeBlocking {
                copyToRealm(task)
                    .toAnotherTask()
            }
        }

        fun deleteTask(task: Task) {
            realm.writeBlocking {
                findLatest(task)?.delete() // this works
            }
        }
    }

    private val myRepo = TaskRepository()


    @Test
    fun testRepo() {
        val myTask = Task().apply {
            name = "old task"
        }

        val mySavedTask = myRepo.addTask(myTask)

        assert(mySavedTask.name == "old task")

    }

}
```

I have changed `name` to "old task" for the myTask variable, so I expect mySavedTask's name property to be also "old task".
addTask directly maps returned object to AnotherTask, I dont get it why but at the moment it gets mapped `Task` object has default values. 

Why would it still have default values?

rorbech:
Looks like a duplicate of #375, where other methods somehow uses the wrong getter for the field. Will have to investigate.

shalva97:
found the workaround, mapping the objects with extension functions work correctly

```
    fun Task.toAnotherTask(): AnotherTask {
        return AnotherTask(name, status)
    }
    
```

rorbech:
Yes, anything that uses accessors from outside the scope of the class seems to work. It seems to be a JVM IR backend error in the compiler, as it isn't failing on macos.

rorbech:
Closing as duplicate of #375

Issue #375, opened 2021-08-16
Custom methods in model classes use initializer values


promanowicz:
```
class SomeObj : RealmObject(){
   ...
   var someVar : Int=100
   
   fun customPrinter():String{
      return "myInt: " + someVar
   }
}

```
use of customPrinter() will print 100 altough some different value has been set to the field
eg:
realm.objects(SomeObj::class).findFirst()?.apply{
   customPrinter() + " " + someVar 
}
will print "myInt: 100 9" if actual value of someVar is 9

build android
realm version "0.4.1"
kotlin + plugin "1.5.21"
java version "11.0.12"

rorbech:
Interesting 🤔. I will have to inspect the IR to understand what is going on here. Maybe there is some optimization in effect or some other reason why our getter is not taken into account.

rorbech:
It was in fact a missing property on the accessors that triggered optimization on the JVM IR backend.

PR #384, opened 2021-08-18
Reproducing bypass of accessors for internal access of fields

rorbech:
Just trying to reproduce and identifying the root cause of #375 and #382.


rorbech:
Filed https://youtrack.jetbrains.com/issue/KTIJ-19438.

rorbech:
Test and fix was migrated to #386

PR #386, opened 2021-08-19
Setting accessor origin to prevent optimizations

### LINKED ISSUE #375: Custom methods in model classes use initializer values

rorbech:
Closes #375


cmelchior (approved):
Nice :shipit:

PR #388, opened 2021-08-23
Design exception types/hierarchy

clementetb:
This PR introduces Realm exception handling. Our concept is that the SDK should treat Realm-core as a separate library with a defined boundary. We achieve this by defining an intermediate representation of core errors to be consumed by the SDK, instead of mapping directly Core errors to end-user exceptions.

The common C-interop package maps raw Core errors types into Kotlin throwables and each c-interop specialization (swig, native) map the exceptions to their internal representation of the core error types. Those exceptions are exposed to the SDK that would decide if it has to [wrap the exception in some more user-friendly one](https://github.com/realm/realm-kotlin/blob/9ccfd0092c3a1e92ae8fb92b3dd11d84b7d6a0dc/packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt#L110) or to [perform some specific logic](https://github.com/realm/realm-kotlin/blob/9ccfd0092c3a1e92ae8fb92b3dd11d84b7d6a0dc/packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectUtil.kt#L109).

Here is an example of how a Core error maps into a end-user exception:

```
SDK                    -> IllegalArgumentException(cause: RealmCoreDuplicatePrimaryKeyException)
C-interop (Common)     -> RealmCoreDuplicatePrimaryKeyException      
C-interop (Darwin)     -> ErrorType.RLM_ERR_DUPLICATE_PRIMARY_KEY_VALUE
Core (Darwin wrapper)  -> realm_wrapper.RLM_ERR_DUPLICATE_PRIMARY_KEY_VALUE
```

This design gives flexibility and enough granularity to the SDK on how to react to Core exceptions. For example, the same Core exception can be rethrown to the user as an `IllegalStateException` on regular Realms, or as an `IllegalArgumentException` on Dynamic Realms. 

**Status**
- [x] C-Interop to expose core exceptions.
- [x] Support multi-platform.
- [x] Refactor `RuntimeExceptions` to actual exception types. #399


packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmEnums.kt
```diff
@@ -65,3 +103,92 @@ expect enum class SchemaValidationMode {
...
+    RLM_ERR_CALLBACK;
+}
+
+/**
+ * Maps core error types to their Kotlin exception representations.
+ */
+@Suppress("ComplexMethod")
+fun errorTypeToThrowable(errorType: ErrorType, message: String?): Throwable {
+    return when (errorType) {
+        ErrorType.RLM_ERR_NONE -> throw IllegalStateException("Cannot map error type $errorType.")
```
cmelchior, line 154: I wonder if this is correct or if `RLM_ERR_NONE` is some roundabout way for core to signal success? 🤔

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmEnums.kt
```diff
@@ -65,3 +103,92 @@ expect enum class SchemaValidationMode {
...
+    RLM_ERR_INVALID_QUERY,
+    RLM_ERR_CALLBACK;
+}
+
+/**
+ * Maps core error types to their Kotlin exception representations.
+ */
+@Suppress("ComplexMethod")
+fun errorTypeToThrowable(errorType: ErrorType, message: String?): Throwable {
+    return when (errorType) {
```
cmelchior, line 153: Still trying to think if doing this mapping is a good idea or if it just introduces an extra layer for no particular reason 🤔 .

I can see specific exceptions being easier to catch in the library layer though, and any unhandled case will just bubble up. While if we just have e.g `RealmCoreException(ErrorType)` then all call sites will need to do a switch for the type they want to handle.

So I'm probably leaning towards this being fine, but would like others opinion on this as well.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmEnums.kt
```diff
@@ -65,3 +103,92 @@ expect enum class SchemaValidationMode {
...
+        ErrorType.RLM_ERR_COLUMN_ALREADY_EXISTS -> RealmCoreColumnAlreadyExistsException(message)
+        ErrorType.RLM_ERR_KEY_ALREADY_USED -> RealmCoreKeyAlreadyUsedException(message)
+        ErrorType.RLM_ERR_SERIALIZATION_ERROR -> RealmCoreSerializationErrorException(message)
+        ErrorType.RLM_ERR_INVALID_PATH_ERROR -> RealmCoreInvalidPathErrorException(message)
+        ErrorType.RLM_ERR_DUPLICATE_PRIMARY_KEY_VALUE -> RealmCoreDuplicatePrimaryKeyValueException(message)
+        ErrorType.RLM_ERR_INDEX_OUT_OF_BOUNDS -> RealmCoreIndexOutOfBoundsException(message)
+        ErrorType.RLM_ERR_INVALID_QUERY_STRING -> RealmCoreInvalidQueryStringException(message)
+        ErrorType.RLM_ERR_INVALID_QUERY -> RealmCoreInvalidQueryException(message)
+        ErrorType.RLM_ERR_CALLBACK -> RealmCoreCallbackException(message)
+        else -> throw IllegalStateException("Error type $errorType not implemented.")
```
cmelchior, line 192: Maybe also append the message here, so the real error isn't completely lost.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/errors/RealmCoreErrors.kt
```diff
@@ -0,0 +1,133 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.interop.errors
+
+/**
+ * Generic representation of a Realm-Core exception.
+ */
+open class RealmCoreException(message: String?) : Exception(message)
```
cmelchior, line 22: Here everything is turned into Exceptions. In Realm Java we signal some things as Errors. I don't have strong preferences, either way, and maybe throwing Errors is a bit too strong. Not sure if you thought about this?

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -99,7 +100,18 @@ object RealmObjectHelper {
...
-        RealmInterop.realm_set_value(o, key, value, false)
+        try {
+            RealmInterop.realm_set_value(o, key, value, false)
+        }
+        // The catch block should catch specific Core exceptions and rethrow them as Kotlin exceptions.
+        // Core exceptions meaning might differ depending on the context, by rethrowing we can add some context related
+        // info that might help users to understand the exception.
+        catch (exception: RealmCoreNotInATransactionException) {
+            throw IllegalStateException(
+                "Cannot set `${obj.`$realm$TableName`}.$col` to `$value`: changing Realm data can only be done from inside a transaction.",
```
cmelchior, line 111:
```suggestion
                "Cannot set `${obj.`$realm$TableName`}.$col` to `$value`: changing Realm data can only be done on a live object from inside a write transaction. Frozen objects can be turned into live using the 'MutableRealm.findLatest(obj)' API.",
```

I think this is going to be a very common failure when first starting using the SDK, so we need to be very descriptive in the error IMO.

cmelchior (commented):
I think the general principle in this PR is great. I have slight concerns about us ending up with a ton of internal exception classes, but maybe it is the best compromise? 🤔 

Otherwise, I just have a few minor comments.

rorbech:
> I think the general principle in this PR is great. I have slight concerns about us ending up with a ton of internal exception classes, but maybe it is the best compromise? 🤔

Yeah, to me it also looks like the wrong way to go with all these specific core exceptions. If it is a matter of being able to selectively only catch some core exceptions and rethrow based on specific core exception type I would imagine that we could do a utility method that is rethrowing based on specific error codes

```
mapError(
  RLM_ERR_INVALID_ARGUMENT to { it -> IllegalArgumentException("....", it) },
  RLM_ERR_LOGIC to { it -> IllegalStateException("....", it) },
)
```
Or something like that. Wouldn't that enable us to achieve the same, but without having to maintain an unrealistic amount of exceptions.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmEnums.kt
clementetb: Bad copy-paste, it is a roundabout.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/errors/RealmCoreErrors.kt
edualonso, line 22: Could this not be done using sealed classes instead?

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmEnums.kt
clementetb: I mapped this error as an `IllegalStateException` because we should never be throwing a `RLM_ERROR_NONE`, it would mean that it has failed with no error.

edualonso (commented):
I'm more inclined towards what Claus said, all those specific core exceptions don't really need to emerge all the way up. Maybe we could group certain exceptions coming from core using codes similarly to how HTTP clients handle status codes so that handling them is grouped by exception type, i.e. client errors (400–499), server errors (500–599), etc.?

rorbech (changes requested):
Nice. There are critical loss of information by turning an error into a null pointer exception if we should ever forget to map a value. I think that needs to be fixed. 

Otherwise also some optimizations and hints on how to share common code with Swig.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -99,7 +100,18 @@ object RealmObjectHelper {
...
         //  instead of generating a typed path for each type.
-        RealmInterop.realm_set_value(o, key, value, false)
+        try {
+            RealmInterop.realm_set_value(o, key, value, false)
+        }
+        // The catch block should catch specific Core exceptions and rethrow them as Kotlin exceptions.
+        // Core exceptions meaning might differ depending on the context, by rethrowing we can add some context related
+        // info that might help users to understand the exception.
+        catch (exception: RealmCoreException) {
+            throw IllegalStateException(
```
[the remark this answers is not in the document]
clementetb: Yes, it should be like that.

packages/cinterop/src/darwin/kotlin/io/realm/interop/RealmEnums.kt
```diff
@@ -74,3 +76,55 @@ actual enum class SchemaValidationMode(override val nativeValue: UInt) : NativeE
     RLM_SCHEMA_VALIDATION_SYNC(realm_wrapper.RLM_SCHEMA_VALIDATION_SYNC),
     RLM_SCHEMA_VALIDATION_REJECT_EMBEDDED_ORPHANS(realm_wrapper.RLM_SCHEMA_VALIDATION_REJECT_EMBEDDED_ORPHANS),
 }
+
+actual enum class ErrorType(override val nativeValue: UInt) : NativeEnumerated {
+    RLM_ERR_NONE(realm_wrapper.RLM_ERR_NONE),
+    RLM_ERR_UNKNOWN(realm_wrapper.RLM_ERR_UNKNOWN),
```
cmelchior, line 82: Isn't this enum defined in Core? Is there a difference between Core UNKNOWN and Kotlin UNKNOWN?

I.e do we need to be able to tell the difference in an exception message?

packages/jni-swig-stub/realm.i
```diff
@@ -96,32 +96,41 @@ return $jnicall;
 // we have a distinction (type map, etc.) in the C API that we can use for targeting the type map.
 bool realm_object_is_valid(const realm_object_t*);
 
+%{
+void throw_as_java_exception(JNIEnv *jenv) {
+    realm_error_t error;
+    if (realm_get_last_error(&error)) {
+        std::string message(error.message);
+        realm_clear_last_error();
```
cmelchior, line 104: Maybe a question for the C API folks, but is there a chance of a race condition using this kind of API?, i.e. shouldn't `realm_get_last_error` also clear it?

packages/cinterop/src/commonMain/kotlin/io/realm/interop/RealmEnums.kt
```diff
@@ -65,3 +67,44 @@ expect enum class SchemaValidationMode {
     RLM_SCHEMA_VALIDATION_SYNC,
     RLM_SCHEMA_VALIDATION_REJECT_EMBEDDED_ORPHANS,
 }
+
+expect enum class ErrorType {
```
cmelchior, line 71: This needs to be added to the Proguard config as we lookup this class from JNI

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
cmelchior, line 110: This still seems to throw IllegalState unconditionally? Shouldn't it test for `RLM_ERR_NOT_IN_A_TRANSACTION` or is this done in another PR?

cmelchior (approved):
Nice 🎉 ... only minor comments.

packages/jni-swig-stub/realm.i
rorbech: It's thread local, so should be ok: https://github.com/realm/realm-core/blob/master/src/realm.h#L364

packages/jni-swig-stub/realm.i
```diff
@@ -96,32 +96,41 @@ return $jnicall;
...
-        if (realm_get_last_error(&error)) {
-            std::string message("[" + std::to_string(error.error) + "]: " + error.message);
-            realm_clear_last_error();
-            // TODO API-SCHEMA Cache class lookup
-            // FIXME Extract all error information and throw exceptions based on type
-            //  https://github.com/realm/realm-kotlin/issues/70
-            jclass clazz = (jenv)->FindClass("java/lang/RuntimeException");
-            (jenv)->ThrowNew(clazz, message.c_str());
-        }
+        throw_as_java_exception(jenv);
```
rorbech, line 126: 🎉

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectUtil.kt
```diff
@@ -101,14 +102,10 @@ internal fun <T : RealmObject> RealmObjectInternal.thaw(liveRealm: BaseRealm): T
                 thawedObject
             )
         }
-    } catch (e: Exception) {
+    } catch (exception: RealmCoreUnknownException) {
```
rorbech, line 105: Is this right? I would expect error code 2 to be mapped to RealmCoreOtherException 🤔

packages/cinterop/src/commonMain/kotlin/io/realm/interop/Errors.kt
```diff
@@ -0,0 +1,136 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.interop
+
+/**
+ * Generic representation of a Realm-Core exception.
+ */
+sealed class RealmCoreException(message: String?) : Exception(message)
```
rorbech, line 22: Don't know if it is strictly needed, but I would be somehow more comfortable if we actually included the error id of the core error somewhere in the `RealmCoreException`. But if the enum mappings are actually validated in the test, then it is maybe ok to leave out. Don't know if other share the same gut feeling though.

rorbech (approved):
Nice. Some minor comments, and CI should of course be green.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/Errors.kt
cmelchior: I was thinking the same thing...but as long as we have automation in place for making sure that we don't skip mapping some codes, then it is probably safe, and the enums provide more immediate context.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectUtil.kt
clementetb: It is addressed in #399

packages/cinterop/src/commonMain/kotlin/io/realm/interop/Errors.kt
clementetb: We should be safe as we have exhaustive mapping tests. I am going to check if we can append the error code into the message.

clementetb: I have reverted to the previous messaging: [$errorcode]: $message.

PR #393, opened 2021-08-24
Sync - project setup, sync module and sync test module

edualonso:
- Added `library-sync` module under `packages` and configured it to depend on `library-base` (former `library`).
- The `test` module now hosts a `base` module which contains all previously existing tests that target local db use cases and a `sync` module for sync-related tests. For now `sync` only contains a sample test.

TODO:
- [x] CI: run sync tests
- [ ] Dokka: publish documentation only for sync https://github.com/realm/realm-kotlin/issues/394
- [ ] Release: analyse release process to include changes for sync https://github.com/realm/realm-kotlin/issues/421


packages/settings.gradle.kts
```diff
@@ -19,6 +19,7 @@ include("gradle-plugin")
 include("plugin-compiler")
 include("plugin-compiler-shaded")
 include("library")
+include("sync")
```
cmelchior, line 22: Maybe `sync-library` and `base-library`? Just to make it clear what they actually are?

packages/sync/build.gradle.kts
```diff
@@ -0,0 +1,303 @@
...
+            }
+            jdkVersion.set(8)
+        }
+        val commonMain by getting {
+            sourceRoot("../runtime-api/src/commonMain/kotlin")
+        }
+    }
+}
+
+tasks.register("uploadDokka") {
```
cmelchior, line 248: I suspect we want to disable this in the base-library and only upload these from here. But maybe that is going to be done in another PR. If so, we should probably create an issue so we remember it.

packages/sync/proguard-rules-consumer-common.pro
```diff
@@ -0,0 +1,20 @@
...
+-keep class * implements io.realm.RealmObject { *; }
+#-keep class **.$* implements io.realm.RealmObject { *; }
+
+# Preserve all native method names and the names of their classes.
+-keepclasseswithmembernames,includedescriptorclasses class * {
+    native <methods>;
+}
+
+# Notification callback
+-keep class io.realm.interop.NotificationCallback {
```
cmelchior, line 12: Do we need these here in the sync variant? I assumed they would be inherited?

packages/sync/src/androidMain/kotlin/io/realm/sync/transport/NetworkTransport.kt
```diff
@@ -0,0 +1,2 @@
+package io.realm.sync.transport
```
cmelchior, line 1: In Java, this package is called `io.realm.mongodb` with a `io.realm.mongodb.sync` subpackage. IMO we should do the same here.

test/settings.gradle.kts
```diff
@@ -9,24 +9,26 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied.
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
 
 includeBuild("../packages")
 
+include("local")
```
cmelchior, line 19: What was your thinking here? To have a `local` and `sync` package with completely separate tests? Or let a `sync` module depend on the `local` module?

In Java we run the local tests for both sync and local builds as they are using two different `.so` files. It does feel a bit extreme, but I suspect we need to do something similar here to be sure we actually test what we ship?

packages/sync/proguard-rules-consumer-common.pro
edualonso: True, forgot about proguard cleanup.

packages/sync/build.gradle.kts
edualonso: Sure, just created https://github.com/realm/realm-kotlin/issues/394

packages/sync/build.gradle.kts
```diff
@@ -0,0 +1,303 @@
+import io.realm.getPropertyValue
```
rorbech, line 1: You should redo this from the kmm-configuration cleaned-up one from `library`.

packages/sync/build.gradle.kts
```diff
@@ -0,0 +1,303 @@
...
+//        exec {
+//            commandLine("xcrun", "simctl", "spawn", device, binary.absolutePath)
+//        }
+//    }
+// }
+
+realmPublish {
+    pom {
+        name = "Sync"
+        description = "Sync code for Realm Kotlin. This artifact is not " +
```
rorbech, line 222: I guess this is not true at the moment. We have a `TODO AUTO-SETUP` for these things that needs to be aligned when we inject the dependencies.

packages/sync/gradlew.bat
```diff
@@ -0,0 +1,100 @@
+@rem
```
rorbech, line 1: This file is needed?

packages/sync/gradle.properties
```diff
@@ -0,0 +1,20 @@
+#
```
rorbech, line 1: I think we should just have these settings in the `packages` specific file?

packages/sync/build.gradle.kts
```diff
@@ -0,0 +1,303 @@
...
+            "supposed to be consumed directly, but through " +
+            "'io.realm.kotlin:gradle-plugin:${Realm.version}' instead."
+    }
+}
+
+tasks.dokkaHtml.configure {
```
rorbech, line 228: I guess there is more to Dokka-generation as we would need one artifact containing both `library-base` and `library-sync`. Don't really know how to do this when it is in different modules. So we should probably have a complete separate issue to do the dokka-generation.

packages/settings.gradle.kts
rorbech: Maybe `library-base` and `library-sync` just to keep the source trees as neighbours in the IDE?

packages/sync/gradlew.bat
rorbech, line 1: We don't need this!?

packages/sync/gradle.properties
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+kotlin.code.style=official
```
rorbech, line 18: Couldn't we just have these settings in the `packages`-level properties file?

packages/sync/src/androidMain/AndroidManifest.xml
```diff
@@ -0,0 +1,33 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    package="io.realm.sync">
+
+    <application>
+        <provider
```
rorbech, line 23: We don't need to put a startup provider here unless we really need on. Manifest-merging will include the one from `library-base`.

packages/sync/src/jvmMain/kotlin/io/realm/sync/transport/NetworkTransport.kt
```diff
@@ -0,0 +1,2 @@
+package io.realm.sync.transport
```
rorbech, line 1: Can't we have a common one for Android and JVM in `src/jvm/kotlin` instead? See the updated kmm-setup for `library` in master. 

Maybe there are some `ktor` dependencies that aren't resolvable if we do it that way, but then it might work by just adding `src/jvm/kotlin` to both `androidMain` and `jvmMain` source sets.

test/build.gradle.kts
```diff
@@ -9,185 +9,27 @@
...
-import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithSimulatorTests
-
 plugins {
-    id("org.jetbrains.kotlin.multiplatform")
-    id("com.android.library")
-    id("io.realm.kotlin")
+    id("com.android.library") apply false
     id("realm-lint")
+    `java-gradle-plugin`
+    id("realm-publisher")
```
rorbech, line 21: Why do we need the publisher in test?

test/build.gradle.kts
```diff
@@ -9,185 +9,27 @@
...
-    }
-}
 
-// Needs running emulator
-tasks.named("iosTest") {
-    val device: String = project.findProperty("iosDevice")?.toString() ?: "iPhone 11 Pro Max"
-    dependsOn(kotlin.targets.getByName<KotlinNativeTargetWithSimulatorTests>("ios").binaries.getTest("DEBUG").linkTaskName)
-    group = JavaBasePlugin.VERIFICATION_GROUP
-    description = "Runs tests for target 'ios' on an iOS simulator"
+    version = Realm.version
```
rorbech, line 29: Why do we need these?

test/local/gradle.properties
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+org.gradle.jvmargs=-Xms512m -Xmx2048m
```
rorbech, line 18: Put these in `test/gradle.properties` instead of having to replicate them.

test/local/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
rorbech, line 1: Not needed.

test/settings.gradle.kts
rorbech: I advices that we should have a separate module for `base` and `sync`. Didn't know that we actually did `base` test on `sync` too. If we initially ship the same `.so`-file it isn't blocking, but it would be a good idea to evaluate if we need to rethink the split.

test/settings.gradle.kts
```diff
@@ -9,24 +9,26 @@
...
+
 pluginManagement {
+    plugins {
+    }
```
rorbech, line 23:
```suggestion
```

rorbech (changes requested):
Added some comments here and there. Most importantly you should use the cleanup multiplatform setup from master as a base for this.

packages/settings.gradle.kts
cmelchior: Yeah, that would be better 👍

cmelchior: Yeah, that would be better 👍

packages/sync/src/jvmMain/kotlin/io/realm/sync/transport/NetworkTransport.kt
edualonso: I haven't added anything related to Ktor in this PR yet, it will come in the next one so I'll have a look at that then.

test/build.gradle.kts
edualonso: Wrong copy-paste.

test/build.gradle.kts
edualonso: Wrong copy-paste.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/NetworkTransport.kt
```diff
@@ -0,0 +1,3 @@
+package io.realm.mongodb.sync
```
rorbech, line 1: Internal? Or is this actually needed by the user for setup?

test/local/build.gradle.kts
```diff
@@ -0,0 +1,193 @@
...
+    google()
+    jcenter()
+    mavenCentral()
+    mavenLocal()
+}
+
+// Common Kotlin configuration
+kotlin {
+    sourceSets {
+        getByName("commonMain") {
```
rorbech, line 37: Hmm. Apparently I missed cleaning up the configuration for the tests. We should probably just leave that for another PR as it will be easier to track the renaming if we not rework the configuration completely.

rorbech (commented):
Looks better, but did you forget to add `test/sync`?

Also seems weird with `test/local` when the library is called `library-base`. So maybe `test/base` and `test/sync`?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/NetworkTransport.kt
edualonso: I believe it should be internal, yes.

edualonso:
`test/sync` will be added in the next PR.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/NetworkTransport.kt
edualonso: Now that I think of it, I don't know if it can be internal since it will be the C-API that will be calling the `sendRequest` function. I guess we will have to try whether it can be done 🤷

edualonso:
I agree with your module naming comment but implementing that change in this PR would pretty much flood my upcoming PR with conflicts so I will do it in the next one.

test/sync/src/commonMain/kotlin/io/realm/mongodb/sync/SyncDispatcher.kt
```diff
@@ -0,0 +1,5 @@
+package io.realm.mongodb.sync
+
+import kotlinx.coroutines.CoroutineDispatcher
+
+expect fun getDispatcher(): CoroutineDispatcher
```
rorbech, line 5: You could just use `io.realm.internal.platform.singleThreadDispatcher` from the library?

rorbech (commented):
Overall LGTM - Some comments:
- Why do we need platform differentiated dispatcher specifically for sync?
- Have you verified that things are still in place for `base`? 
  - All tests are actually run
  - Dokka for `base` is still generated/published correctly

If things are still in place for `base` I think we can just track the remainders in new issues.

test/sync/src/commonMain/kotlin/io/realm/mongodb/sync/SyncDispatcher.kt
edualonso: Yes, that's what I'm doing in the next PR.

edualonso:
There will be some cleanup related to the comments in the PR we will put up next.
Tests for base are run, just checked the build log for build no. 14 and everything looks normal.
Dokka seems to be working and generates the artifact correctly, but the docs format is not correct. @rorbech says it is likely this issue has been introduced after the build script cleanup task.
Apart from that nothing that I'm aware of.

rorbech (approved):

rorbech:
> Dokka seems to be working and generates the artifact correctly, but the docs format is not correct. @rorbech says it is likely this issue has been introduced after the build script cleanup task.

Created #438 to fix this separately as it doesn't seem related to this PR.

PR #396, opened 2021-08-30
Interface based API for lists

rorbech:
This moves implementation details of `RealmList` into `internal` and changes internal implementation to use `ManagedRealmList` to avoid having to handle unmanaged lists in the Realm-faced implementation.


packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
```diff
@@ -16,232 +16,31 @@
...
-            }
-        }
-    }
+interface RealmList<E> : MutableList<E> {
+    fun observe(): Flow<RealmList<E>>
 }
 
 /**
- * Delegate interface for Realm-specific methods.
+ * Instantiates a [RealmList] in **unmanaged** mode.
```
cmelchior, line 43:
```suggestion
 * Instantiates an **unmanaged** [RealmList].
```

Talking about a "mode" make it seems like you can easily switch it IMO.

packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
```diff
@@ -16,232 +16,31 @@
...
-        return RealmList(liveList, metadata.copy(realm = liveRealm))
-    }
-
-    private fun rangeCheckForAdd(index: Int) {
-        if (index < 0 || index > size) {
-            throw IndexOutOfBoundsException("Index: '$index', Size: '$size'")
-        }
-    }
-}
+fun <T> realmListOf(vararg elements: T): RealmList<T> =
```
cmelchior, line 45: Should we also have an override that takes other collections?

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -240,19 +240,30 @@ internal class SuspendableNotifier(
...
+            liveComponentPointer = liveListPointer,
             notifyComponentUpdate = { frozenRealm ->
-                if (!liveList.isValid()) {
+                if (!liveListDelegate.isValid()) {
                     callback.onChange(null)
                 } else {
                     // Notifications need to be delivered with the version they where created on, otherwise
                     // the fine-grained notification data might be out of sync.
-                    val frozenList = liveList.freeze(frozenRealm)
+                    val frozenListPointer = RealmInterop.realm_list_freeze(
```
cmelchior, line 262: Shouldn't this logic be inside the `liveListDelegate.freeze` method?

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -377,17 +377,17 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
-                                // This is due to Kotlin Native being unable to retrieve the related
-                                // supertypes in runtime
-                                if (collectionType != CollectionType.NONE) {
-                                    val supertypes = (type as IrSimpleType).classifier.descriptor
-                                        .typeConstructor.supertypes
-                                    if (inheritsFromRealmObject(supertypes)) {
-                                        putValueArgument(2, irBoolean(true))
-                                    }
-                                }
+//                            // Check if the collection type is a subclass of RealmObject
```
cmelchior, line 380: Why are this commented out?

packages/library/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
rorbech: Yep. Just migrated a previous sample implementation, so didn't gave it too much thought.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
rorbech: According to @edualonso (who did the initial prototype of this PR) we don't longer need this parameter as the `RealmObject` case is now handled as a fallback if the argument is not a primitive type.

packages/library/src/commonMain/kotlin/io/realm/RealmList.kt
rorbech: I added a `Iterable<T>.toRealmList()` extension method following the conventions from Kotlin.

edualonso (approved):
🚀

PR #437, opened 2021-09-03
Sync Public API definition

cmelchior:
This PR is just a draft and I'm not sure if we should work towards merging it or we should just copy things from here a we get to the individual API's.

The intent of this PR is to get an overview of the public API surface of the Sync API and discover any rough edges. I attempted to move our public Sync classes and convert them to interfaces / data classes where it made sense. I left out the MongoClient API's, but they should be easily added later.

So far, I have discovered the following issues we need to take a closer look at:

- [ ] I reworked all Async methods to just be `suspend` functions, drastically simplifying the API surface, but I fear that we might be hiding the AppExceptions that can result from it. We can choose to do something similar to the Kotlin `Result` class as a nice wrapper, but on the other hand, the exception is also rather hidden in Java. Suggested code:

```
val app = App.create("appId")
scope.launch {
  val credentials = Credentials.emailPassword("user", "password")

 // Explicit error with Result wrapper
  val result: Result<User> = app.login(credentials)
  result.
    .onSuccess { user: User -> showSuccess(user) }
    .onError( error: AppException -> showError(error }

 // Throw runtime AppException. Happy path requires less code,  but people might forget to do error handling?
 try {
  val user: User = app.login(credentials)
  showSuccess(user)
 } catch(ex: AppException) {
  showError(error)
 } 
}
```
See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/ and https://github.com/Kotlin/KEEP/blob/master/proposals/stdlib/result.md#error-handling-alternative

- [ ] If `App` is an interface we cannot do `App("appId")` like in Java. `App.create("appId")` and `App.create(AppConfiguration.Builder("appId").build())` doesn't seem that bad, but we should probably strive to use a similar naming pattern across our API. Configurations seem to gravitate towards `defaultConfig` which doesn't work for `App` and collections use `appOf("appId")` which might also work? 🤔 

- [ ] How many things should be `val`/`vars`, e.g. in the `User` class?

- [ ] Dokka. For `val`s, I mostly just took the `@returns` description, but that means that most sentences start with `The ...`. Should this be cleaned up?

- [ ] When do we turn classes into data classes vs. keeping them as interfaces? e.g. `ApiKey`.

- [ ] How to represent callbacks? As SAM interfaces? Typealiases? Inline? Generally, I have seen issues with Kotlin/Java interop depending on what you do. E.g. named SAM interfaces don't seem to autocomplete very nicely in Kotlin if they have more than one return value. E.g AuthenticationListener, ProgressListener.

Other TODO's

- [ ] Rework Javadoc to Dokka in order to fix links and other mistakes.
- [ ] Importing the bson dependency doesn't seem to work
- [ ] Generally, this PR is  very much a rough draft and doesn't compile.


cmelchior:
@edualonso @clementetb Probably relevant for your work

PR #439, opened 2021-09-03
Added support for @indexed properties

finnschiermer:
Perhaps it should be "@index" instead? I'm fine with changing it.

### DIFF
--- a/buildSrc/src/main/kotlin/Config.kt
+++ b/buildSrc/src/main/kotlin/Config.kt
@@ -61,7 +61,7 @@ object Versions {
         // Cannot upgrade past 4.1.0 due to https://issuetracker.google.com/issues/187134648
         // Fix has not been released in lastest 7.1.0-alpha06.
         const val buildTools = "4.1.0" //
-        const val ndkVersion = "22.1.7171670"
+        const val ndkVersion = "21.1.6352462"// "22.1.7171670"
     }
     const val androidxStartup = "1.1.0-beta01" // https://maven.google.com/web/index.html?q=startup#androidx.startup:startup-runtime
     const val androidxJunit = "1.1.3-beta02" // https://maven.google.com/web/index.html#androidx.test.ext:junit
--- a/packages/library/src/commonMain/kotlin/io/realm/annotations/Indexed.kt
+++ b/packages/library/src/commonMain/kotlin/io/realm/annotations/Indexed.kt
@@ -14,17 +14,17 @@
  * limitations under the License.
  */
 
-package io.realm.internal
+package io.realm.annotations
 
-import io.realm.interop.Table
-import kotlin.reflect.KMutableProperty1
-
-// TODO MEDIATOR/API-INTERNAL Consider adding type parameter for the class
-@Suppress("VariableNaming")
-interface RealmObjectCompanion {
-    val `$realm$fields`: List<KMutableProperty1<*, *>>?
-    val `$realm$indexedProps`: List<KMutableProperty1<*, *>>?
-    val `$realm$primaryKey`: KMutableProperty1<*, *>?
-    fun `$realm$schema`(): Table
-    fun `$realm$newInstance`(): Any
-}
+@Retention(AnnotationRetention.SOURCE)
+@Target(AnnotationTarget.FIELD)
+@MustBeDocumented
+/**
+ * Annotation marking a field as indexed.
+ *
+ * Multiple fields in a RealmObject class can have this annotation.
+ *
+ * This annotation applies to the following primitive types: String, Byte, Char,
+ * Short, Int and Long, as well as their nullable variants.
+ */
+annotation class Indexed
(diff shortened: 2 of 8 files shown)


buildSrc/src/main/kotlin/Config.kt
```diff
@@ -61,7 +61,7 @@ object Versions {
         // Cannot upgrade past 4.1.0 due to https://issuetracker.google.com/issues/187134648
         // Fix has not been released in lastest 7.1.0-alpha06.
         const val buildTools = "4.1.0" //
-        const val ndkVersion = "22.1.7171670"
+        const val ndkVersion = "21.1.6352462"// "22.1.7171670"
```
cmelchior, line 64: Looks like a b
```suggestion
        const val ndkVersion = "22.1.7171670"
```

packages/library/src/commonMain/kotlin/io/realm/annotations/Indexed.kt
```diff
@@ -0,0 +1,30 @@
...
+
+package io.realm.annotations
+
+@Retention(AnnotationRetention.SOURCE)
+@Target(AnnotationTarget.FIELD)
+@MustBeDocumented
+/**
+ * Annotation marking a field as indexed.
+ *
+ * Multiple fields in a RealmObject class can have this annotation.
```
cmelchior, line 25: Maybe add a comment about it being added automatically on primary keys?

cmelchior: And maybe combine it with the one below? As they basically say the same thing?

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectCompanion.kt
```diff
@@ -23,6 +23,7 @@ import kotlin.reflect.KMutableProperty1
 @Suppress("VariableNaming")
 interface RealmObjectCompanion {
     val `$realm$fields`: List<KMutableProperty1<*, *>>?
+    val `$realm$indexedProps`: List<KMutableProperty1<*, *>>?
```
cmelchior, line 26: We should probably use the same naming for properties? The above is called `$realm$fields` so this should probably be `$realm$indexedFields`.

Or we rename `$realm$fields` to `$realm$properties`...either seems fine to me.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/Identifiers.kt
```diff
@@ -85,6 +85,7 @@ internal object FqNames {
     val COLLECTION_TYPE = FqName("io.realm.interop.CollectionType")
     val PROPERTY_FLAG = FqName("io.realm.interop.PropertyFlag")
     val PRIMARY_KEY_ANNOTATION = FqName("io.realm.annotations.PrimaryKey")
+    val INDEXED_ANNOTATION = FqName( "io.realm.annotations.Indexed")
```
cmelchior, line 88:
```suggestion
    val INDEXED_ANNOTATION = FqName( "io.realm.annotations.Index")
```

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -346,6 +353,14 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
                                         "Primary key ${property.name} is of type ${backingField.type.classifierOrFail.owner.symbol.descriptor.name} but must be of type ${validPrimaryKeyTypes.map { it.owner.symbol.descriptor.name }}",
                                     )
                                 }
+                                val indexableTypes = with(pluginContext.irBuiltIns) {
+                                    setOf(byteType, charType, shortType, intType, longType, stringType).map { it.classifierOrFail }
+                                }
+                                if (isIndexed && backingField.type.classifierOrFail !in indexableTypes) {
+                                    logError(
+                                        "Indexed key ${property.name} is of type ${backingField.type.classifierOrFail.owner.symbol.descriptor.name} but must be of type ${validPrimaryKeyTypes.map { it.owner.symbol.descriptor.name }}",
```
cmelchior, line 361: The error message map the list of `validPrimaryKeyTypes`, we should probably have a `validIndexKeyTypes` instead

test/src/jvmTest/kotlin/io/realm/test/compiler/IndexedTests.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.realm.util.CompilerTest.compileFromSource
+import io.realm.util.TypeDescriptor.allFieldTypes
+import org.junit.Test
+import kotlin.reflect.KClassifier
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class IndexedTests {
+    @Test
+    fun `indexed supportness`() {
```
cmelchior, line 31: Shouldn't we also have a test for the opposite? i.e links and link-lists do not support it.

Also, we should define the behavior for fields with `@PrimaryKey`... I would be tempted to say we just ignore it there, since it doesn't really do anything.

But just to make sure we verify the behaviour if (when) it changes again ;)

cmelchior (changes requested):
Nice 💯 

I have a few minor comments about naming and testing.

packages/library/src/commonMain/kotlin/io/realm/internal/RealmObjectCompanion.kt
finnschiermer: turns out it is even needed, so will remove.

cmelchior (approved):
Looks good to me 🚀 . Only missing thing is a CHANGELOG.md entry

rorbech (approved):

PR #440, opened 2021-09-03
Update to latest C-API for thawing objects and lists

rorbech:
Update to latests _Sync C-API_ from `yg/sync-c-api`. 

C-API updates in core: 
- Updated C-API for resolving objects, results, list from https://github.com/realm/realm-core/pull/4658
- _de-factory GenericNetworkTransport::NetworkFactory_ https://github.com/realm/realm-core/commit/ec0de20c065ff66f2ba35d62120c926d0d85dcfb

PR #445, opened 2021-09-08
Commonize notification implementation

rorbech:
This implements notifications by an `Observable` interface with type specific implementations hidden into the various implementations of `Observable` (`RealmObjectInternal`, `RealmResult` and `ManagedRealmList`).


packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectInternal.kt
```diff
@@ -41,6 +48,89 @@ internal interface RealmObjectInternal : RealmObject, RealmStateHolder, io.realm
...
+        @Suppress("UNCHECKED_CAST")
+        val type: KClass<*> = this::class
+        val mediator = `$realm$Mediator`!!
+        val managedModel = mediator.createInstanceOf(type)
+        val dbPointer = liveRealm.dbPointer
+        @Suppress("TooGenericExceptionCaught")
+        try {
+            val realmObjectThaw: NativePointer? =
+                RealmInterop.realm_object_thaw(`$realm$ObjectPointer`!!, dbPointer)
+            val realmObjectThaw1: NativePointer = realmObjectThaw!!
```
cmelchior, line 78: Why not just `!!` in `realmObjectThaw`?

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectInternal.kt
```diff
@@ -41,6 +48,89 @@ internal interface RealmObjectInternal : RealmObject, RealmStateHolder, io.realm
...
+        channel: SendChannel<RealmObjectInternal>
+    ): ChannelResult<Unit>? {
+        @Suppress("TooGenericExceptionCaught")
+        val f: RealmObjectInternal? = try {
+            this.freeze(frozenRealm)
+        } catch (e: RuntimeException) {
+            // FIXME C-API is currently throwing an error if the object has been deleted, so currently just
+            //  catching that and returning null. Only treat unknown null pointers as non-existing objects
+            //  to avoid handling unintended situations here.
+            if (e.message?.startsWith("[2]: null") ?: false) {
```
cmelchior, line 117: This is going to break with the new error handling. Is there some way to make a connection so we remember updating it here?

cmelchior (approved):
Nice 👍

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectInternal.kt
nhachicha: I guess according to https://github.com/realm/realm-kotlin/pull/440 the `thaw` will no longer throw but rather return a `null` for invalid/deleted object

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -92,20 +90,40 @@ internal class SuspendableNotifier(
 //        }
     }
 
-    /**
-     * Listen to changes to a RealmResults.
-     */
-    internal fun <T : RealmObject> resultsChanged(results: RealmResultsImpl<T>): Flow<RealmResultsImpl<T>> {
+    internal fun <T> registerObserver(observable: Observable<T>): Flow<T> {
+        println("registerObserver")
```
nhachicha, line 94:
```suggestion
```

nhachicha (approved):
👏

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectInternal.kt
rorbech: Looks like I just introduced an intermediate variable for some debugging/experiments - Removed it again. #440 will update API to match that the underlying C-API will return null on delete objects.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectInternal.kt
rorbech: Cleaning up the actual code for the new error handling is already tackled in #399, but it will also change in #440 when the C-API is actually returning `null`, so I think it will resolve itself when merging that ... 🙈

PR #446, opened 2021-09-09
Move internal and test implementations to separate packages

### LINKED ISSUE #56: Internalize non-public API
Move non-public API out of public accessible namespaces. 

Don't know if this is possible with our current module approach, but maybe a good time to re-evaluate that then. 

An option could be to just more everything to internal packages, which is at least hinting that it shouldn't be used, but could maybe be backed by _opt in requirements_ https://kotlinlang.org/docs/reference/opt-in-requirements.html

rorbech:
This moves
- non-public API to `io.realm.internal.` no matter module
- all test related classes to `io.realm.test....`

Intellij allows auto completion for `internal` packages for `project` dependencies but not for _Maven dependencies_, so this still renders us able to easily access the internals when developing, while hiding internal packages for users (even if we need to expose `cinterop` as an `api` dependency). 

As Kotlin doesn't utilize package protected methods, tests are not bound to be in the same package as the class under test, so should be safe to do. We could maybe keep all the actual tests in `io.realm` (without `test`) just to make it really easy to identify the tests of a class, but all non-API symbols related to test should go into `io.realm.test` to make it obvious what is API and what is not. 

Questions for reviewers?
- Should we keep the actual tests back to `io.realm` again.
- Should we keep all _data model_-related classes in a separate package `io.realm.test.data`? 

Closes #56


rorbech:
One benefit of sharing package name with the actual implementation is that we won't have to add imports for all the classes in the same package.

cmelchior (approved):
Looks great to me 🎉 

> Should we keep the actual tests back to io.realm again.

IMO having a clean separation between tests and non-tests is more beneficial. Imports are hidden in the IDE anyway.

> Should we keep all data model-related classes in a separate package io.realm.test.data

In Java we uses a `io.realm.entities` package. This is super helpful there as we have 80+ model classes for tests. Hopefully that will be lower with Kotlin, but IMO it still makes it cleaner and easer to reason about the classes if they are in their own package. I don't have a strong opinions on wether that package is called `data` or `entities`.

nhachicha (approved):

PR #447, opened 2021-09-10
Sync App API

rorbech:
This adds:
- Network transport implementation for JVM and KN including:
  - Reorginazation of JNI utility source code to keep it out of `realm.i`
  - `CinteropCallback` for error propagation from C-API
    ```
    interface CinteropCallback {
        fun onSuccess(pointer: NativePointer)
        fun onError(throwable: Throwable)
    }
    ```
- Minimal `App`, `AppConfig`, `User` and `Credentials` to do an `app.login(credentials)`  

There are still some rough corners, but as `library-sync` is not released we can postpone them for later. 
Most notable discussion point is:
- User facing API for errors
  `suspend fun App.login(credentials: Credentials): Result<User>`


packages/cinterop/src/commonMain/kotlin/io/realm/interop/NetworkTransport.kt
```diff
@@ -0,0 +1,43 @@
...
+        const val POST = "post"
+        const val PATCH = "patch"
+        const val PUT = "put"
+        const val DELETE = "delete"
+    }
+
+    val authorizationHeaderName: String?
+    val customHeaders: Map<String, String>
+
+    fun sendRequest(
```
cmelchior, line 16: This API is not enough to support refreshing the access token. See https://github.com/realm/realm-java/pull/7552 for details on how this was solved in Java. We should do something similar here.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/NetworkTransport.kt
```diff
@@ -0,0 +1,43 @@
...
+    val customResponseCode: Int,
+    val headers: Map<String, String>,
+    val body: String
+) {
+    // Returns the HTTP headers in a JNI friendly way where it is being serialized to a
+    // String array consisting of pairs of { key , value } pairs.
+    fun getJNIFriendlyHeaders(): Array<String?> {
+        val jniHeaders = arrayOfNulls<String>(headers.size * 2)
+        var i = 0
+        for ((key, value) in headers) {
```
nhachicha, line 36: This could be done using 
```Kotlin
return headers.flatMap { listOf(it.key) + it.value }

File: packages/cinterop/src/jvm/kotlin/io/realm/interop/RealmInterop.kt [hunk 3]
```diff
@@ -431,6 +430,54 @@ actual object RealmInterop {
...
+        networkTransportFactory: () -> Any,
+        baseUrl: String?
+    ): NativePointer {
+        val config = realmc.new_app_config(appId, networkTransportFactory)
+
+        baseUrl?.let { realmc.realm_app_config_set_base_url(config, it) }
+
+        // TODO parameterize
+        realmc.realm_app_config_set_platform(config, "kotlin")
+        realmc.realm_app_config_set_platform_version(config, "PLATFORM_VERSION")
```
cmelchior, line 463: Add TODO for support in AppConfiguration. There should be issues for it.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -0,0 +1,333 @@
+/*
```
cmelchior, line 1: Like discussed, should we move these files outside the `jni-swig-stub` folder or find a pattern so we don't accidentally delete files here?

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    return realm_results_add_notification_callback(
+            results,
+            // Use the callback as user data
+            static_cast<jobject>(get_env()->NewGlobalRef(callback)),
+            [](void *userdata) {
+                get_env(true)->DeleteGlobalRef(static_cast<jobject>(userdata));
+            },
+            // change callback
+            [](void *userdata, const realm_collection_changes_t *changes) {
```
cmelchior, line 36: This callback is shared between `list` and `results`, would be nice if it was only defined once. But can just add a FIXME for now

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -0,0 +1,333 @@
...
+void register_login_cb(realm_app_t* app, realm_app_credentials_t* credentials, jobject callback) {
+    auto jenv = get_env();
+    static jclass notification_class = jenv->FindClass("io/realm/interop/OperationCallback");
+    static jmethodID on_success_method = jenv->GetMethodID(notification_class, "onSuccess", "(Lio/realm/interop/NativePointer;)V");
+    static jmethodID on_error_method = jenv->GetMethodID(notification_class, "onError", "(Ljava/lang/Throwable;)V");
+
+    realm_app_log_in_with_credentials(
+            app,
+            credentials,
+            [](void* userdata, realm_user_t* user, realm_error_t* error) {
```
cmelchior, line 142: These callbacks are shared between a lot of API calls. In java we have helper methods for creating these:

https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/java_network_transport.hpp#L95

https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/java_network_transport.hpp#L119

Callsite, then becomes much less code:

https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsApp.cpp#L192

We should do something similar, but just adding FIXME for now is fine.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+static void network_request_lambda_function(void *userdata, // Network transport
+                                            const realm_http_request_t request, // Request
+                                            void *completion_data,// Call backs
+                                            realm_http_completion_func_t completion_callback) {
+    auto jenv = get_env(true);
+
+    // Initialize pointer to JVM class and methods
+    jobject network_transport = static_cast<jobject>(userdata);
+    static jclass network_transport_class = jenv->FindClass("io/realm/interop/NetworkTransport");
```
cmelchior, line 197: We need to make sure that this class isn't obfuscated

packages/library-sync/src/androidMain/AndroidManifest.xml
```diff
@@ -15,4 +15,9 @@
   ~ limitations under the License.
   -->
 
-<manifest package="io.realm.mongodb.sync" />
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="io.realm.mongodb.sync">
```
cmelchior, line 19: Maybe just `io.realm.mongodb` as this contains more than just sync.

We should probably figure out if `library-sync` is the best name for this module/package since it contains a lot more than just sync, but neither `library-app` nor `library-mongodb` seems that much better.

Something to think about before we announce a public release.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
```diff
@@ -0,0 +1,54 @@
...
+         * TODO
+         */
+        fun create(
+            configuration: AppConfiguration,
+        ): App = AppImpl(configuration as AppConfigurationImpl)
+
+        /**
+         * TODO
+         */
+        fun create(
```
cmelchior, line 43: This looks like an internal factory. Move to internal or extension method just for tests?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
```diff
@@ -0,0 +1,54 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+
+/**
+ * TODO
+ */
+interface App {
+
+    val configuration: AppConfiguration
+
+    suspend fun login(credentials: Credentials): Result<User>
```
cmelchior, line 30: Missing Dokka, or as a minimum a TODO about it.

cmelchior, line 30: We need to discuss the `Results` API used here, but no need to block the PR for it, but we should add a FIXME about it.

Personally, I think we might need to duplicate the stdlib Result class to something like `AppResult`, just so we can control the error case better. But it is annoying that it ends up being a copy just for the sake of specializing the Error case.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/Credentials.kt
```diff
@@ -0,0 +1,48 @@
...
+
+package io.realm.mongodb
+
+import io.realm.interop.NativePointer
+import io.realm.interop.RealmInterop
+
+/**
+ * TODO
+ */
+sealed class Credentials(val id: String) {
```
cmelchior, line 25: This looks like it need a bigger refactor. I tried to migrate the Java API's directly here: https://github.com/realm/realm-kotlin/pull/437/files#diff-1052d1ac32f6e72c1fccfefcf6262a978f35df4e663d72b7a16da9d7708355bdR54

But I can see some advantage to using sealed classes.

But we should add a TODO about revisiting the API here.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/User.kt
```diff
@@ -14,15 +14,11 @@
...
+import io.realm.interop.NativePointer
 
-internal class SampleTest {
-
-    @Test
-    fun sample() {
-        assertEquals(4, 2 + 2)
-    }
+interface User {
+    val nativePointer: NativePointer
```
cmelchior, line 22: This should be private, so should probably be moved to UserImpl

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+                val requestBuilderBlock: HttpRequestBuilder.() -> Unit = {
+                    headers {
+                        // 1. First of all add all custom headers
+                        customHeaders.forEach {
+                            append(it.key, it.value)
+                        }
+
+                        // 2. Then replace default authorization header with custom one if present
+                        headers
+//                            .filter {
```
cmelchior, line 61: Remove?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    override fun sendRequest(
+        method: String,
+        url: String,
+        headers: Map<String, String>,
+        body: String,
+        usesRefreshToken: Boolean
+    ): Response {
+        try {
+            return runBlocking(dispatcher) {
```
cmelchior, line 51: As mentioned. This kind of usage here will end up blocking the UI thread when refreshing the access token. See the Java PR mentioned elsewhere on suggestions on how to refactor this.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+                        }
+
+                        // 2. Then replace default authorization header with custom one if present
+                        headers
+//                            .filter {
+//                                // Ignore Content-Type header as it's handled by the engine
+//                                it.key != HttpHeaders.ContentType
+//                            }
+                            .toMutableMap().also { receivedHeaders ->
+                                val authorizationHeaderValue =
```
cmelchior, line 66: Maybe refactor into its own method. Something like `replaceAuthorizationHeaderIfNeeded` or similar. At least I had a bit of hard time following the logic here.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+                                    headers[DEFAULT_AUTHORIZATION_HEADER_NAME]
```
cmelchior, line 67: This constant should be in `AppConfiguration` as it is user facing.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+                        "get" -> client.get<HttpResponse>(url, requestBuilderBlock)
+                        else -> throw IllegalArgumentException("Wrong request method: '$method'")
+                    }.let {
+                        processHttpResponse(it)
+                    }
+                } catch (e: ServerResponseException) {
+                    // 500s are thrown as ServerResponseException
+                    processHttpResponse(e.response)
+                } catch (e: CancellationException) {
+                    println("EXCEPTION 1: $e")
```
cmelchior, line 108: Leftovers?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+                    println("MESSAGE 1: ${e.message}")
+                    Response(-1, -1, mapOf(), "")
```
cmelchior, line 110: In Java we use `0` instead of `-1` to indicate not in use, we should probably do the same here.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+            "get" -> this.method = HttpMethod.Get
+        }
+    }
+
+    private fun getClient(): HttpClient {
+        // Need to freeze value as it is used inside the client's init lambda block, which also
+        // freezes captured objects too, see:
+        // https://youtrack.jetbrains.com/issue/KTOR-1223#focus=Comments-27-4618681.0-0
+        val frozenTimeout = timeoutMs.freeze()
+        return HttpClient(CIO) {
```
cmelchior, line 152: Where is CIO defined?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+    companion object {
+
+        const val DEFAULT_AUTHORIZATION_HEADER_NAME = "Authorization"
+
+        // Custom error codes. These must not match any HTTP response error codes
+        const val ERROR_IO = 1000
+        const val ERROR_INTERRUPTED = 1001
+        const val ERROR_UNKNOWN = 1002
+
+        private fun httpResponse(
```
cmelchior, line 193: Maybe `createHttpResponse` to make it more readable?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -0,0 +1,65 @@
...
+
+const val TEST_APP_1 = "testapp1"       // Id for the default test app
+const val BASE_URL = "http://127.0.0.1:9090"
+
+class AppTests {
+
+    @Test
+    fun login() {
+        // Send request directly to the local server to get the actual app ID
+        val applicationId = KtorNetworkTransport(
```
cmelchior, line 37: In Java all of this setup is hidden behind a `TestApp` class. We should probably copy that and associated `close` method to here. But it is fine to just add TODO about it:

https://github.com/realm/realm-java/blob/master/realm/realm-library/src/syncTestUtils/kotlin/io/realm/TestApp.kt#L34

https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/mongodb/AppExt.kt#L29

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -0,0 +1,65 @@
...
+            "",
+            true
+        ).let { response ->
+            when (response.httpResponseCode) {
+                200 -> response.body
+                else -> throw IllegalStateException(response.toString())
+            }
+        }
+
+        println("App config")
```
cmelchior, line 53: Leftovers?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -0,0 +1,65 @@
...
+        val configuration: AppConfiguration = appConfigurationOf(applicationId, BASE_URL, singleThreadDispatcher("asdf"))
+        println("App create")
+        val app = App.create(configuration)
+
+        println("Login")
+        runBlocking {
+            app.login(EmailPassword("asdf@asdf.com", "asdfasdf"))
+                .onSuccess { it: User -> }
```
cmelchior, line 61: This doesn't actually test anything?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/internal/KtorNetworkTransportTest.kt
```diff
@@ -0,0 +1,137 @@
...
+            timeoutMs = 5000,
+            dispatcher = singleThreadDispatcher("ktor-test")
+        )
+    }
+
+    @Test
+    fun requestSuccessful() {
+        val url = "$BASE_URL/okhttp?success=true"
+
+        // FIXME test fails on native due to method ordering, e.g.:
```
cmelchior, line 52: I believe this was fixed and was a problem with Ktor setup?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/internal/KtorNetworkTransportTest.kt
```diff
@@ -0,0 +1,137 @@
...
+            val body = if (method == HTTPMethod.GET) "" else "{ \"body\" : \"some content\" }"
+
+            val response = transport.sendRequest(
+                method.nativeKey,
+                url,
+                mapOf(),
+                body,
+                true
+            )
+            println("------------> METHOD: $method")
```
cmelchior, line 65: Leftovers?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/internal/KtorNetworkTransportTest.kt
```diff
@@ -0,0 +1,137 @@
...
+                true
+            )
+            assertEquals(200, response.httpResponseCode)
+            assertEquals(0, response.customResponseCode)
+            assertEquals("${method.name}-success", response.body)
+        }
+    }
+
+//    @Test
+//    fun requestInterrupted() {
```
cmelchior, line 115: We don't need to test Interrupt pr. se, but in Java this test is here to verify that exceptions from the network layer correctly propagate back to the user.

We should have a similar test for Kotlin. It can probably be done by mocking the correct places. But it is fine to leave a TODO for now.

cmelchior (changes requested):
Awesome start 🎉 

I don't think there is anything strictly blocking, but I left a lot of comments about things that needs to fixed/cleaned up. But for now, just adding TODO/FIXME's for them is probably fine.

packages/library-sync/build.gradle.kts
```diff
@@ -62,6 +62,21 @@ kotlin {
...
+
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.coroutines}") {
+                    version { strictly(Versions.coroutines) }
+                }
+                implementation("org.jetbrains.kotlinx:atomicfu:${Versions.atomicfu}")
+                implementation("org.jetbrains.kotlinx:kotlinx-serialization-core:${Versions.serialization}")
+                implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:${Versions.serialization}")
+
+                implementation("io.ktor:ktor-client-core:${Versions.ktor}")
+                implementation("io.ktor:ktor-client-cio:${Versions.ktor}")
```
nhachicha, line 74: We are already specifying engines per target below, why pulling CIO in the common?

packages/library-sync/build.gradle.kts
```diff
@@ -62,6 +62,21 @@ kotlin {
...
+                implementation("io.ktor:ktor-client-serialization:${Versions.ktor}")
+                implementation("io.ktor:ktor-client-logging:${Versions.ktor}")
+
+                implementation("ch.qos.logback:logback-classic:${Versions.logback}")
```
nhachicha, line 78: WHy? we already have a multiplatform logger?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppConfigurationImpl.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+
+internal class AppConfigurationImpl(
+    override val appId: String,
+    override val baseUrl: String = DEFAULT_BASE_URL,
+    override val networkTransportDispatcher: CoroutineDispatcher
+) : AppConfiguration {
+
+    private val networkTransport: io.realm.interop.NetworkTransport = KtorNetworkTransport(
+        timeoutMs = 5000,
```
nhachicha, line 32: `timeoutMs` could be stored in `Constants`? if there's one for the Sync module ?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+import io.ktor.http.*
+import io.realm.interop.NetworkTransport
+import io.realm.interop.Response
+import io.realm.internal.platform.freeze
+import io.realm.internal.platform.runBlocking
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineDispatcher
+
+class KtorNetworkTransport(
+    override val authorizationHeaderName: String = "Authorization",
```
nhachicha, line 35: Could be stored in sync `Constants` or use the one defined below `DEFAULT_AUTHORIZATION_HEADER_NAME`

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
nhachicha, line 108: Use our internal logger instead

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+                    println("MESSAGE 1: ${e.message}")
+                    Response(-1, -1, mapOf(), "")
+                } catch (e: Exception) {
+                    println("EXCEPTION 2: $e")
+                    println("MESSAGE 2: ${e.message}")
+                    Response(-2, -2, mapOf(), "")
+                }
+            }
+        } catch (e: Exception) {
+            println("EXCEPTION 3: $e")
```
nhachicha, line 118: I suppose this is waiting for exception handling to be added?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
nhachicha, line 152: the choice of the engine should be done via (expect/actual) in a non-platform specific way.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -0,0 +1,199 @@
...
+            followRedirects = true
+
+            // TODO connectionPool?
+        }
+    }
+
+    private fun parseHeaders(headers: Headers): Map<String, String> {
+        val parsedHeaders: MutableMap<String, String> = mutableMapOf()
+        for (key in headers.names()) {
+            parsedHeaders[key] = requireNotNull(headers[key]) { "Headeer '$key' cannot be null" }
```
nhachicha, line 179:
```suggestion
            parsedHeaders[key] = requireNotNull(headers[key]) { "Header '$key' cannot be null" }
```

nhachicha (commented):
Nice progress  👏

packages/cinterop/src/jvm/CMakeLists.txt
```diff
@@ -11,6 +11,22 @@ set(REALM_ENABLE_SYNC ON)
 set(REALM_BUILD_LIB_ONLY ON)
 add_subdirectory("realm-core" EXCLUDE_FROM_ALL)
 
+# -DREALM_ANDROID -DREALM_HAVE_CONFIG -DPIC -fdata-sections -pthread -frtti -fvisibility=hidden -fsigned-char -fno-stack-protector -std=c++17
```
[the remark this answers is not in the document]
rorbech: @clementetb Do we need keep these for anything?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: Apparently in `package io.ktor.client.engine.ci` 😅

rorbech: The _Coroutine based I/O_ implementation is available for KMP so I don't see a reason for doing our own platform agnostic wrapper around a client.

packages/cinterop/src/jvm/jni/utils.h
```diff
@@ -0,0 +1,118 @@
...
+    operator realm::StringData() const
+    {
+        // To solve the link issue by directly using Table::max_string_size
+        static constexpr size_t max_string_size = realm::Table::max_string_size;
+
+        if (m_is_null) {
+            return realm::StringData();
+        }
+        else if (m_size > max_string_size) {
+            // TODO: Throw an actual exception
```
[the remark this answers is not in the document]
rorbech: Apparently the utility C++ code for throwing exception hasn't been migrated yet. Will have to investigate what can be done.

packages/cinterop/src/jvm/jni/utils.h
```diff
@@ -0,0 +1,118 @@
...
+            throw 20;
+//            THROW_JAVA_EXCEPTION(
```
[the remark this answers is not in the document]
rorbech: Apparently the utility C++ code for throwing exception hasn't been migrated yet. Will have to investigate what can be done.

packages/cinterop/src/jvm/CMakeLists.txt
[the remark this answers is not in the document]
clementetb: They were ported from realm-java, kept them as a reference, they can be gone.

packages/cinterop/src/jvm/jni/utils.h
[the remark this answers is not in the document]
clementetb: Commented it out to fix it later, but did not add a Fixme comment

packages/cinterop/src/jvm/CMakeLists.txt
```diff
@@ -11,6 +11,22 @@ set(REALM_ENABLE_SYNC ON)
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${REALM_COMMON_CXX_FLAGS} -std=c++17")
```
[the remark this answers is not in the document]
clementetb: This value comes from the Realm-java project, had to add it to get the imported code working.

I am not sure if it is enough as the includes contain object-store code, will try and see.

packages/jni-swig-stub/build.gradle.kts
```diff
@@ -77,7 +77,10 @@ tasks.create("cleanJvmWrapper") {
                 exclude("LongPointerWrapper.java") // not generated
             }
         )
-        delete("$projectDir/src/main/jni/")
+        delete(fileTree("$projectDir/src/main/jni/").matching {
+            include("*")
+            exclude("realm_api_helpers.*") // not generated
```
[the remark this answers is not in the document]
clementetb: The other files are not affected because they sit in another module (cinterop).

packages/cinterop/src/jvm/CMakeLists.txt
```diff
@@ -11,6 +11,22 @@ set(REALM_ENABLE_SYNC ON)
...
+        "jni/realm_api_helpers.cpp"
+        "jni/utils.cpp"
+        )
+
 # Create shared FFI library that is consumed by the C-Interop layer.
-add_library(realmc SHARED ${SWIG_JNI}/realmc.cpp)
-target_link_libraries(realmc log android RealmFFI)
+add_library(realmc SHARED
+        ${SWIG_JNI}/realmc.cpp
+        ${SWIG_JNI}/realm_api_helpers.cpp
```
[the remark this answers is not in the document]
clementetb: jni_SRC files contain the cinterop helper files (environment, string accessor), that file is for swig.

I'm going to create a new blob with the swig files for clarity.

packages/cinterop/src/darwin/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -883,6 +891,140 @@ actual object RealmInterop {
...
+            },
+            StableRef.create(callback).asCPointer(),
+            staticCFunction { userdata ->
+                userdata?.asStableRef<Callback>()?.dispose()
+                    ?: error("Login callback data should never be null")
+            }
+        )
+    }
+
+    private val newRequestLambda = staticCFunction<COpaquePointer?,
```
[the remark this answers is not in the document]
rorbech: Yeah. Pointers passed in and out of the C-API are nullable as there is no way to mark them non null yet (https://youtrack.jetbrains.com/issue/KT-48083)

packages/cinterop/src/commonMain/kotlin/io/realm/interop/Callback.kt
```diff
@@ -19,3 +19,9 @@ package io.realm.interop
 interface Callback {
     fun onChange(change: NativePointer)
 }
+
+interface OperationCallback {
```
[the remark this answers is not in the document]
rorbech: Chose `CinteropCallback` as it is a generic interface that either returns a `NativePointer` or a `Throwable`

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
rorbech: Added a fix referencing this. Probably fits will into implementing _set_ and _dict_ change listeners ;)

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
rorbech: Adding a FIXME referencing the Java implementation. Makes most sense to do when introducing the next call that follows this pattern.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
rorbech: Removed it. Not that big a deal to re-introduce it when it is actually needed.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
rorbech: There is no Dokka-documentation in any of the classes yet, so makes most sense to address at the same time. Added a TODO.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/Credentials.kt
rorbech: Added a TODO referencing this comment

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/internal/KtorNetworkTransportTest.kt
rorbech: Added an empty `@Ignore`d `errorPropagation` tests with a FIXME and comments about intention.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
rorbech: Added a FIXME to revisit this with link to this comment.

packages/cinterop/src/commonMain/kotlin/io/realm/interop/NetworkTransport.kt
rorbech: Created https://github.com/realm/realm-kotlin/issues/450 to track this

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: Created https://github.com/realm/realm-kotlin/issues/450 to track this.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
rorbech: Would probably suit well into https://github.com/realm/realm-kotlin/issues/251. Added a TODO referencing this comment.

packages/cinterop/src/jvm/CMakeLists.txt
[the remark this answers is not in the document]
clementetb: We need c++17 to build it.

packages/cinterop/src/jvm/jni/utils.h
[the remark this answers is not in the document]
clementetb: We have addressed this by catching the runtime exception and transform it into a custom error code. The issue is that custom error codes are not propagated via the app callbacks, see #451.

packages/cinterop/src/jvm/jni/utils.h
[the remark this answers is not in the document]
clementetb: Fixed throwing a runtime exception and catch as explained @ https://github.com/realm/realm-kotlin/pull/447#discussion_r708553781

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: I just simplified the implementation so that you should now be able to reason about it

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: Translated the various exceptions into error response codes according to Java.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: Removed the actual client, as ktor will resolve it automatically from the available clients at runtime. Added a TODO about fixating it, so we can control it even though people provide additional dependencies.

packages/library-sync/build.gradle.kts
rorbech: The `realm-client-cio` seems to work for macos even though not listed in https://ktor.io/docs/http-client-engines.html#desktop, but from your advice I removed the common client to follow the official documentation.

packages/library-sync/build.gradle.kts
rorbech: Replaced http-logger with our own custom one

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppConfigurationImpl.kt
rorbech: Added a TODO around being able to provide this as a `Duration` through the `AppConfigruation` and define the default there.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: Removed the explicit client as ktor automatically resolves it from the runtime available onse. We might however want to fixate this, so people don't change it under the hood. 

Weird part is that `CIO` is not listed in https://ktor.io/docs/http-client-engines.html#desktop, but it is the only one working for `macos` 🤷. Added a TODO for fixating the client here and an explanatory TODO around using `ktor-client-cio` for `macos` in `build.gradle.kts`

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/NetworkTransport.kt
```diff
@@ -0,0 +1,43 @@
...
+        const val POST = "post"
+        const val PATCH = "patch"
+        const val PUT = "put"
+        const val DELETE = "delete"
+    }
+
+    val authorizationHeaderName: String?
+    val customHeaders: Map<String, String>
+
+    fun sendRequest(
```
cmelchior, line 16:
```suggestion
     // FIXME https://github.com/realm/realm-kotlin/issues/450
    fun sendRequest(
```

packages/cinterop/src/darwin/kotlin/io/realm/interop/RealmInterop.kt
```diff
@@ -883,6 +892,138 @@ actual object RealmInterop {
...
+        CValue<realm_http_request_t>,
+        COpaquePointer?,
+        realm_http_completion_func_t?,
+        Unit>
+    { userdata, request, completionData, callback ->
+        safeUserData<NetworkTransport>(userdata).let { networkTransport ->
+            request.useContents {
+                val headerMap = mutableMapOf<String, String>()
+                for (i in 0 until num_headers.toInt()) {
```
cmelchior, line 931: Where is `num_headers` defined? 🤔

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -0,0 +1,62 @@
...
+                200 -> response.body
+                else -> throw IllegalStateException(response.toString())
+            }
+        }
+
+        val configuration: AppConfiguration = appConfigurationOf(applicationId, BASE_URL, singleThreadDispatcher("asdf"))
+        val app = App.create(configuration)
+
+        runBlocking {
+            app.login(EmailPassword("asdf@asdf.com", "asdfasdf")).getOrThrow()
```
cmelchior, line 59: I think this will just throw right now until we add support for the `EmailPasswordAuth` provider. So maybe wrap in try/catch and add TODO?

cmelchior (approved):
Looks really nice 🚀 ... There is a lot of TODOs/FIXME's but nothing that should prevent us from merging this IMO.

The only concern I have is that I don't think the `login` test actually works as this PR doesn't have support for registering users. But If I remember correctly you tested the implementation with an existing user on the test server, correct?

In that case, just modifying the test to capture the failure seems good enough for now, and then we can iterate in another PR.

In that case, it is probably fine to

packages/cinterop/src/commonMain/kotlin/io/realm/interop/NetworkTransport.kt
rorbech: Less lines of code, bu it feels quite inefficient (though I don't know if it is optimized in some way under the hood), so will just keep the existing code.

packages/cinterop/src/darwin/kotlin/io/realm/interop/RealmInterop.kt
rorbech: The `CValue<T>.useContents(block: T.() -> R): R` promotes the `request` as the receiver, so all referencess like `num_headers`, `headers`, `method`, etc. are from the `request`. I have tried to change some of the `apply`s to `let` and name the target explicitly, but it is not possible with `useContents`. Only option would be to prepend `this`, but I didn't find it more explanatory.

cmelchior: Ah okay... that's fine. One of the problems with too much Kotlin magic in Github reviews I guess :D

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/internal/KtorNetworkTransportTest.kt
rorbech: Works with the `ktor-client-cio` for `macos`. Don't really know why, as the client isn't listed for native desktop.

rorbech:
Merged `master` and fixed linting, added issue references to TODOs, etc. So _only_ lacking the actual server tests running on CI.

nhachicha (approved):
Let's sync 🚰 🚀

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
rorbech: We don't spin up the server as part of CI builds yet, so I `@Ignore`d all the sync tests for now with comments about how to manually spin the server up.

rorbech:
I have ignored the sync tests as the CI build doesn't spin up the server image. We will target automation of tests in a separate PR as it not only involves spinning up the server, but also requires ability to do admin logins and register users programatically as in https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/admin/ServerAdmin.kt and https://github.com/realm/realm-java/blob/master/realm/realm-library/src/syncTestUtils/kotlin/io/realm/mongodb/SyncTestUtils.kt#L57.

PR #452, opened 2021-09-16
Add support for running Sync Integration Tests

cmelchior:
This PR starts the Docker container on Jenkins that contains the MongoDB Test Server and run Sync integration tests on it.

### TODO
- [x] See how much logic can be moved to shell scripts so it can be shared with local testing -> CONCLUSION: Doesn't seem immediate obvious how, so punting to another issue.
- [x] Figure out if it would be easy to split Sync and Base tests into two different test runs -> CONCLUSION: Looks like we need to create two `test-xxx` projects. That is too much work for this PR, so punting to another PR.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -334,7 +334,7 @@ def testWithServer(dir, task) {
 
     try {
         testAndCollect(dir, task)
-    } catch (err) {
+    } finally {
         // We assume that creating these containers and the docker network can be considered an atomic operation.
         if (mongoDbRealmContainer != null && mongoDbRealmCommandServerContainer != null) {
             archiveServerLogs(mongoDbRealmContainer.id, mongoDbRealmCommandServerContainer.id)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -81,11 +81,11 @@ pipeline {
                 runScm()
             }
         }
-//         stage('Build') {
-//             steps {
-//                 runBuild()
-//             }
-//         }
+        stage('Build') {
+            steps {
+                runBuild()
+            }
+        }
 //         stage('Static Analysis') {
 //             when { expression { runTests } }
 //             steps {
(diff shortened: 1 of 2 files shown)


cmelchior:
As part of this PR I also setup a new CI builder (atlanta_host14). This PR has run successfully against both atlanta_host13 and atlanta_host14

Jenkinsfile
```diff
@@ -91,18 +97,22 @@ pipeline {
...
         }
-        stage('Tests Android') {
+        stage('Tests Android - Unit Tests') {
             when { expression { runTests } }
             steps {
                 testAndCollect("packages", "connectedAndroidTest")
-                testAndCollect("test",     "connectedAndroidTest")
+            }
+        }
+        stage('Integration Tests') {
```
rorbech, line 112: This grouping goes against what I think would be the better option for parallelising platform specific parts, but I guess I see the motivation for running this together now.

Jenkinsfile
```diff
@@ -294,6 +306,49 @@ def runCompilerPluginTest() {
...
+    mongoDbRealmCommandServerContainer = commandServerEnv.run("--rm -i -t -d --network container:${mongoDbRealmContainer.id} -v$tempDir:/apps")
+    sh "timeout 60 sh -c \"while [[ ! -f $tempDir/testapp1/app_id || ! -f $tempDir/testapp2/app_id ]]; do echo 'Waiting for server to start'; sleep 1; done\""
+
+    try {
+        tasks.each { task ->
+            testAndCollect(dir, task)
+        }
+    } finally {
+        // We assume that creating these containers and the docker network can be considered an atomic operation.
+        if (mongoDbRealmContainer != null && mongoDbRealmCommandServerContainer != null) {
```
rorbech, line 341: Shouldn't this be `||` and shouldn't the `image.run(...)` part above also be in the try-finally block?

rorbech (approved):
Nice 🎉

Do we have a list of things to do while provisioning a new host?

Jenkinsfile
cmelchior: Yes, this needs to be refactored once we start to parallelize builds across multiple machines, but for a single machine this setup makes the JenkinsFile less complicated.

Jenkinsfile
cmelchior: I'm using `&&` as it otherwise becomes very complicated to reason about storing server logs and shutting down things. It is true that this logic breaks down if CI creates the first container but fails to create the 2nd. However, I never noticed this happening in practice, so it didn't seem worthwhile to complicate this code for that.

cmelchior:
Unfortunately, there is no checklist for creating a new CI host. In Realm Java this is done by using a Docker image so we can run on any Linux host.

When we move to Github Actions, the machine configuration is determined in their configuration files.

PR #453, opened 2021-09-16
Sync test infrastructure

rorbech:


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -45,6 +45,7 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlin.collections.set
 
+// TODO Consider adding a logger
```
cmelchior, line 48: We definitely need logging, but there are issues with the setup right now as it is tied to Realm lifecycles in order to be able to control the debug level.

Maybe we need to add a System logger that users have very little control over, but that is for another PR no matter what.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -16,49 +16,38 @@
...
+    @BeforeTest
+    fun setup() {
+        app = TestApp()
+        admin = ServerAdmin(app)
+        admin.createUser("asdf@asdf.com", "asdfasdf")
+    }
 
+    @AfterTest
+    fun teaddown() {
+        admin.deleteAllUsers()
```
cmelchior, line 46: In Java, this is part of an App.close() extension method.  I suspect we need to add this eventually here as well. We might as well just do it in this PR as it should be a trivial amount of code.

cmelchior (approved):
Nice 👍

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
rorbech: I guess we should just add an option to put in a logger in the `AppConfiguration` and that should be propagated to as the default for the `SyncConfiguration`? But I guess it fits best in on of the AppConfiguration PRs

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
rorbech: I added an `asTestApp` extension method and bundled all the `AdminApp` implementations up on `TestApp`, so it becomes 
https://github.com/realm/realm-kotlin/pull/453/files#diff-7789c96c2ad7327e4d0f561d78b70e1cee7f84b15af77df4bf1852da42d686f0R35

to avoid polluting the actual `App` api with extension methods for test.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -16,49 +16,35 @@
...
-        val configuration: AppConfiguration = appConfigurationOf(applicationId, BASE_URL, singleThreadDispatcher("asdf"))
-        val app = App.create(configuration)
+    @BeforeTest
+    fun setup() {
+        app = TestApp()
+        app.asTestApp.createUser("asdf@asdf.com", "asdfasdf")
+    }
 
+    @AfterTest
+    fun teaddown() {
```
cmelchior, line 42:
```suggestion
    fun tearDown() {
```

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -0,0 +1,59 @@
...
+        private fun getAppId(appName: String): String {
+            val client =  defaultClient("test-app-initializer")
+            return runBlocking {
+                client.get("http://127.0.0.1:8888/$appName")
+            }
+        }
+    }
+}
+
+val App.asTestApp : TestApp
```
cmelchior, line 58: This looks slightly odd, but I also hated putting the extension methods directly on App. At least this makes it clear when you are touching APIs not public which is nice.

cmelchior (approved):

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.realm.internal.platform.singleThreadDispatcher
+import io.realm.mongodb.App
+import io.realm.mongodb.appConfigurationOf
+import io.realm.test.mongodb.util.AdminApi
+import io.realm.test.mongodb.util.AdminApiImpl
+import io.realm.test.mongodb.util.defaultClient
+import kotlinx.coroutines.CoroutineDispatcher
+
+/**
+ * This class wraps various methods making it easier to create an App that can be used
```
clementetb, line 32: Maybe we can be more explicit in the comment:
```suggestion
 * This class merges the classes `App` and `AdminApi` making it easier to create an App that can be used
```

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/util/AdminApi.kt
```diff
@@ -0,0 +1,189 @@
...
+import kotlinx.serialization.InternalSerializationApi
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.serializer
+
+private const val baseUrl = "http://127.0.0.1:9090/api/admin/v3.0"
```
clementetb, line 44: Couldn't we have these centralized in some place?

clementetb:
Great!! 🚀

test/sync/build.gradle.kts
```diff
@@ -22,6 +22,7 @@ plugins {
     id("com.android.library")
     id("io.realm.kotlin")
     id("realm-lint")
+    kotlin("plugin.serialization") version "1.5.21"
```
cmelchior, line 25: Would it be possible to move this to our Config file in buildSrc?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -16,49 +16,36 @@
...
 import kotlin.test.Test
 
-const val TEST_APP_1 = "testapp1" // Id for the default test app
-const val BASE_URL = "http://127.0.0.1:9090"
-
 // Cannot run on CI yet, as it requires sync server to be started with
-// tools/sync_test_server/start_server.sh and manual creation of a user "asdf@asdf.com"/"asdfasdf"
-// through the web ui
-@Ignore
+// tools/sync_test_server/start_server.sh
```
cmelchior, line 29: This no longer seems accurate?

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.realm.internal.platform.runBlocking
+import io.realm.internal.platform.singleThreadDispatcher
+import io.realm.mongodb.App
+import io.realm.mongodb.appConfigurationOf
+import io.realm.test.mongodb.util.AdminApi
+import io.realm.test.mongodb.util.AdminApiImpl
+import io.realm.test.mongodb.util.defaultClient
+import kotlinx.coroutines.CoroutineDispatcher
+
+const val BASE_CMD_SRV_URL = "http://127.0.0.1:8888"
```
cmelchior, line 31: I don't think that letters are so expensive that we cannot call this `BASE_COMMAND_SERVER_URL`?  ;)

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/util/AdminApi.kt
```diff
@@ -0,0 +1,191 @@
...
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.serializer
+
+private const val ADMIN_PATH = "/api/admin/v3.0"
+
+/**
+ * Wrapper around MongoDB Realm Server Admin functions needed for tests.
+ */
+interface AdminApi {
```
cmelchior, line 49: Since this is just test code. Is there a good reason to keep the same interface/impl split? I don't have a strong opinion, just wondering

cmelchior (approved):

test/sync/build.gradle.kts
rorbech: Assuming you just referred to the version number!? Yeah, seems like we can now actually put `Versions` references here 🎉 ... must have been fixed in Gradle.

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/util/AdminApi.kt
rorbech: It was just from the attempt of putting all the App related test methods on `TestApp` and provide easy access to the REST API while also separating the REST specific implementation into `AdminApi`. This way `TestApp` can implement the `AdminApi by AdminApiImp(...)` while the implementation is separate and we don't need to always define both a _test app_ and a _server admin_ instance (the latter is always needed to delete users on close, etc). Again I am up for counter proposals, but was just trying to address the `realm-java` annoyances of mixing test specific APIs on `App` and also always having to define an `admin` instance to access the REST api

PR #456, opened 2021-09-18
Enable incremental builds on CI

### LINKED ISSUE #454: [CI] Incremental Builds on CI
Fallout from #452 

We should consider if running incremental builds on CI would be acceptable? It would boost subsequent builds on PR branches by quite a bit (at least 10+ minutes from my experiments). But there is a chance that incremental builds occasional break more than clean builds.

No matter what, we should probably only do clean builds om master/release branches.

It is controlled by these two Jenkins workflow options for the `runScm` step.
```
[$class: 'WipeWorkspace'],
[$class: 'CleanCheckout'],
```

cmelchior:
Closes #454 

This PR does two things:

- Enable CCache support
- Don't wipe the workspace between PR builds .These options clear all files not part of the repo, including those in `.gitignore`. In our case, this specifically means any `.gradle` or `.cxx` caches already built.

By keeping those around, we can significantly increase build times for PR's, but it relies on the build script being able to handle incremental builds. Some changes needed for this were added in #452 (zip file handling), so this PR will fail until that PR is merged.

With multiple nodes also present, I had to refactor how we do locking. Turns out that locking on a node is rather difficult, but fortunately doesn't seem needed either. But I did had to refactor the pipeline to make sure that all steps run on the same node.

### Metrics

With this change:

If `packages` are not modified and we hit the Gradle cache the `build` step goes from 15-17 minutes to 1-2 minutes. 

On a clean build where we hit the CCache, a full build goes from ~30 minutes to ~20 minutes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -164,7 +164,6 @@ def runScm() {
             [$class: 'CleanCheckout'],
         ]       
     }
-    echo repoExtensions
     checkout([
             $class           : 'GitSCM',
             branches         : scm.branches,
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,10 +58,10 @@ pipeline {
         }
      }
     // The Gradle cache is re-used between stages, in order to avoid builds interleave,
-    // and potentially corrupt each others cache, we grab a global lock for the entire
-    // build.
+    // and potentially corrupt each others cache, we grab a node lock for the entire
+    // build. 
     options {
-        lock resource: 'kotlin_build_lock'
+        lock resource: "${env.NODE_NAME}-kotlin_build_lock"
         timeout(time: 15, activity: true, unit: 'MINUTES')
     }
     environment {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,12 +51,7 @@ node_label = 'osx_kotlin'
 workspacePath = "/Users/realm/workspace-realm-kotlin/${currentBranch}"
 
 pipeline {
-    agent {
-        node {
-            label node_label
-            customWorkspace workspacePath
-        }
-     }
+     agent none
      options {
         // In Realm Java, we had to lock the entire build as sharing the global Gradle
         // cache was causing issues. We never discovered the root cause, but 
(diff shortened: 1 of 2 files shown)


nhachicha:
I run twice the build https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/27/pipeline/ and https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/26/pipeline/ (on top of the previously successful https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/25/pipeline/) but both builds took ~30 minutes?

cmelchior:
Not 100% sure what happened there, but the replay configuration was setup with:

```
isReleaseBranch = true // releaseBranches.contains(currentBranch)
```

Which disables the incremental builds and clear the workspace before building. It was something I played around with during testing, but it shouldn't be in the actual PR.

nhachicha:
I think I'm still experiencing the same issue described previously. 

- Entire build [#36](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/37/pipeline) on `atlanta_14` 
- Triggered another  [#37](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/37/pipeline) and it's doing a complete build even though it's running on the same host `atlanta_14`

rorbech:
> think I'm still experiencing the same issue described previously.
> 
> * Entire build [#36](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/37/pipeline) on `atlanta_14`
> * Triggered another  [#37](https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-456/37/pipeline) and it's doing a complete build even though it's running on the same host `atlanta_14`

But you are also still replaying a build with custom modification to do a full build. Go see the replay diff in Jenkins.

nhachicha:
@cmelchior can you push a new commit to make sure I'm not replaying the last commit that prevents incremental builds?

cmelchior:
@nhachicha Done

nhachicha (approved):
Nicely done! CI 👑

rorbech (approved):

PR #457, opened 2021-09-20
Anonymous login support

rorbech:
Adds support for 
```
app.login(Credentials.anynomous())
```

Also, internalizes implementation details of `Credentials`


packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -135,27 +135,30 @@ register_object_notification_cb(realm_object_t *object, jobject callback) {
...
-
     realm_app_log_in_with_credentials(
             app,
             credentials,
             // FIXME Refactor into generic handling of network requests, like
             //  https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsApp.cpp#L192
             [](void *userdata, realm_user_t *user, realm_error_t *error) {
+                // TODO Investigate why method look ups in callbacks seems to need to be non-static
+                //  to avoid throwing
+                //  "JNI DETECTED ERROR IN APPLICATION: use of deleted local reference 0x55"
```
cmelchior, line 146: This seems like it would hit us in a lot of places? Maybe create an issue, so we can reference that and also have more visibility around it.

packages/library-base/src/commonMain/kotlin/io/realm/internal/util/Validation.kt
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal.util
+
+object Validation {
```
cmelchior, line 19: Maybe add a class doc explaining what this class is for

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/Provider.kt
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.mongodb.internal
+
+/**
+ * This enum contains the list of identity providers supported by MongoDB Realm.
+ * All of these except [.EMAIL_PASSWORD] must be enabled manually on MongoDB Realm to
```
cmelchior, line 21: Previously you also had to enable this one. It wasn't on as a default. Has this changed?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/CredentialImpl.kt
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.mongodb.internal
+
+import io.realm.internal.interop.NativePointer
+import io.realm.internal.interop.RealmInterop
+import io.realm.mongodb.Credentials
+
+internal open class CredentialImpl(
```
cmelchior, line 23: We should probably provide a nice `toString()` method for debugging. In Java there is a `asJson()` method. I think that was mostly added because it was hard to keep it package protected, but it also serves as a nice debugging facility. 

With Kotlin we have `internal`, but debugging still seems a valid use case?

rorbech: In Java the `toJson` was dispatching to the C++ implementation (`credentials->serialize_as_json()`). Do we need something similar to have a standardized output. If so, then we need to expose the method in the C-API.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
rorbech: Isolated the issue to be that `FindClass` returns a local reference while `GetMethodID` returns a stable ID that can be statically referenced. We should make a global cache of these as global refs. Created #460 to track this.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/Provider.kt
rorbech: I removed any mentioning of defaults but just highlighted that the providers need to be enabled in the app with a reference to docs on that.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/CredentialImpl.kt
rorbech: Don't know if the java-identifiers for these have anything to do with the json-representation, but the representation towards the C-API has changed to use an enumeration, so don't know what to pretty-print until we have a `asJson` in the C-API. I will skip this for now, but left at TODO.

cmelchior: Looking into the implementation. Prettyprinting the info does require us going to C++, so probably not worth it for now. Keeping this comment for now seems fine 👍

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/AuthProvider.kt
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal.interop.sync
+
+@Suppress("filename")
```
cmelchior, line 19: What is this for?

cmelchior (approved):
Yay, more TODO's 😆 .... I only had a few minor comments, otherwise this looks like a good start.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/AuthProvider.kt
rorbech: I tried to follow the convention from the non-sync enums and define them in one file, but linting didn't allow naming the file differently than the class, when there was only one top level definition, so I tried to suppress that. It didn't work our so I just renamed the file ... and forgot to remove the annotation. Will do.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AuthenticationProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+    ;
+
+    internal val id: io.realm.internal.interop.sync.AuthProvider = id
+
+    internal companion object {
+        /**
+         * Create the identity provider from the ID string returned by MongoDB Realm.
+         *
+         * @param id the string identifier for the provider
+         * @return the enum representing the provider or [.UNKNOWN] if no matching provider
```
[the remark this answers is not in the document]
rorbech: Actually this is just internal implementation and the mapping is now done from the C-API defined enums, so we should never end up with unknown/unable to map. Moved the companion method to internal `Impl` class to make it clear that it is not part of the public API.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
```diff
@@ -40,11 +40,12 @@ internal class AppImpl(
         )
 
     override suspend fun login(credentials: Credentials): Result<User> {
+        val credentialsInternal: CredentialImpl = io.realm.internal.util.Validation.checkType(credentials, "credentials")
         return RealmInterop.runCatching {
             suspendCoroutine { continuation ->
                 realm_app_log_in_with_credentials(
                     nativePointer,
-                    credentials.nativePointer,
+                    (credentials as CredentialImpl).nativePointer,
```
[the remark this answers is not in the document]
rorbech: I see your point, but lets discuss. I guess it would be nice to clarify the classes that would make sense to be able to mock and then be very explicit about that the rest of the interfaces are not ment for user implementation/extension. 

Maybe we can put together a generic error message signaling this in `Validation`.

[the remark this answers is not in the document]
cmelchior: Discussed offline. 

This is fine for now. There should be virtually zero reasons for people to pass in their own custom implementations, and it creates a nice separation in our codebase, where it is very easy to see when the public API is affected.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/CredentialImpl.kt
```diff
@@ -0,0 +1,53 @@
...
+    authenticationProvider: AuthenticationProvider,
+    internal val nativePointer: NativePointer
+) : Credentials {
+
+    override val authenticationProvider: AuthenticationProvider = authenticationProvider
+        get() {
+            return if (field == AuthenticationProvider.fromId(
```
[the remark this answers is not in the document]
rorbech: I just copied it from Java. I removed it and initialize the property from the value returned from the C-API. It is verified for each of the credential types in the test, so should always be sound.

cmelchior (approved):

PR #458, opened 2021-09-22
Use proper java enums for C-API enums on JVM

rorbech:


rorbech:
Closing as outdated. This just showed that it was actually doable to have swig and cinterop to exposed Kotlin enums, but since we didn't pick it up I will just close it.

PR #465, opened 2021-09-24
Revert using Result as transport for resulting data on sync calls

clementetb:
Error handling in sync calls would follow the standard error catching. This PR reverts from using `Result` to throwing exceptions.

Users can at any moment wrap the calls with `runCatching` to convert the data/exceptions in a `Result` instance.

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
@@ -20,6 +20,7 @@ import io.realm.internal.interop.CinteropCallback
 import io.realm.internal.interop.NativePointer
 import io.realm.internal.interop.RealmInterop
 import io.realm.internal.platform.appFilesDirectory
+import io.realm.internal.util.Validation
 import io.realm.mongodb.App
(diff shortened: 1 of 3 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
```diff
@@ -39,24 +39,23 @@ internal class AppImpl(
...
-                realm_app_log_in_with_credentials(
-                    nativePointer,
-                    (credentials as CredentialImpl).nativePointer,
-                    object : CinteropCallback {
-                        override fun onSuccess(pointer: NativePointer) {
-                            continuation.resume(UserImpl(pointer))
-                        }
+    override suspend fun login(credentials: Credentials): User {
+        val credentialsInternal: CredentialImpl =
+            io.realm.internal.util.Validation.checkType(credentials, "credentials")
```
cmelchior, line 44: Nitpick, but add this to imports so we can write ` val credentialsInternal: CredentialImpl = Validation.checkType(credentials, "credentials")` in one line

cmelchior (approved):
Nice 🚀

PR #469, opened 2021-09-26
Asynchronous http response to c-api

### LINKED ISSUE #450: Thread handling for background/async Sync jobs
We need to ensure that callbacks like refreshing of auth token is not blocking the users thread like: https://github.com/realm/realm-kotlin/pull/447#discussion_r707163678

rorbech:
Make http responses to C-API asynchronous. 

Closes #450


edualonso (approved):
Looks good to me. Only one comment related to the use of `Result`.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
```diff
@@ -43,21 +42,27 @@ internal class AppImpl(
...
+            object : CinteropCallback {
+                override fun onSuccess(pointer: NativePointer) {
+                    result.trySend(Result.success(UserImpl(pointer, this@AppImpl)))
+                }
 
-                    override fun onError(throwable: Throwable) {
-                        continuation.resumeWithException(throwable)
-                    }
+                override fun onError(throwable: Throwable) {
+                    result.trySend(Result.failure(throwable))
```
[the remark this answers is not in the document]
rorbech: This is a callback, so `onError` is triggered on the C-API/network dispatcher thread. Posting it to the channel will bring it into the context of the caller.

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -932,10 +954,11 @@ actual object RealmInterop {
                 for (i in 0 until num_headers.toInt()) {
                     headers?.get(i)?.let { header ->
                         headerMap[header.name!!.toKString()] = header.value!!.toKString()
-                    } ?: error("Header at index $i within range ${num_headers.toInt()} should not be null")
+                    }
+                        ?: error("Header at index $i within range ${num_headers.toInt()} should not be null")
```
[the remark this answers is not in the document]
rorbech: It is the IDE's suggestion 🤷

I usually try to avoid pushing reformatting of the whole file, but this was from an old _WIP_ commit, so did not go back and clean it up. I have given up on reasoning on this as I don't get the conventions. If we don't want to follow these, we should maybe reconfigure it so that we got assistance on fulfilling the rules.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
```diff
@@ -43,21 +42,27 @@ internal class AppImpl(
...
+        // ... and cannot be resumed on another thread (we probably also want to guarantee that we
+        // are resuming on the same dispatcher), so run our own implementation using a channel
+        val credentials =
+            Validation.checkType<CredentialImpl>(credentials, "credentials").nativePointer
+        val result = Channel<Result<User>>(1)
+        // Should we do this on the users context or an internal dispatcher
+        RealmInterop.realm_app_log_in_with_credentials(
+            nativePointer,
+            credentials,
+            object : CinteropCallback {
```
cmelchior, line 55: We are going to replicate this pattern all over, with only difference between the mapping of onSuccess, would be nice if we could somehow extract this into a helper class or method

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
```diff
@@ -43,21 +42,27 @@ internal class AppImpl(
...
-                        continuation.resumeWithException(throwable)
-                    }
+                override fun onError(throwable: Throwable) {
+                    result.trySend(Result.failure(throwable))
                 }
-            )
-        }
+            }.freeze()
+        )
+        return result.receive().getOrThrow()
```
cmelchior, line 65: Shouldn't we also close the channel?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/internal/KtorNetworkTransportTest.kt
```diff
@@ -47,36 +50,49 @@ internal class KtorNetworkTransportTest {
...
                 mapOf(),
-                body
+                body,
+                object : ResponseCallback {
+                    override fun response(response: Response) {
+                        result.trySend(response)
+                    }
+                }
             )
+            val response = result.receive()
```
cmelchior, line 70: Also close channel

cmelchior (commented):
Nice 👍 . It is slightly annoying that we duplicate the logic for creating the response, but not sure there is a good way around it.

Also had a few minor suggestions for refactors.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppImpl.kt
rorbech: I have tried to address this with helper method that builds the callback from a channel. Combined with `Channel<T>.autoClose` it becomes reasonable condense: https://github.com/realm/realm-kotlin/pull/469/files#diff-8f52344e1cfad3228c6bb6989f5d253af6127732d6bf1ed8dcc90103e11f6038R52

We could try to combine the channel pattern with the callback to make it even shorter, but I could only come up with 
```
suspend fun <T> appCall(success: (NativePointer) -> T, appCall: (CinteropCallback) -> Unit): T {
    return Channel<Result<User>>(1).autoClose { channel ->
        appCall(channelResultCallback(channel, success))
        channel.receive().getOrThrow()
    }
}
```
It requires multiple lambda arguments which in my opinion does not make it more readable or easier to write
```
appCall(
    { pointer -> UserImpl(pointer, this@AppImpl) },
    { callback -> 
        RealmInterop.realm_app_log_in_with_credentials(
            nativePointer,
            credentials,
            callback
    )}
)
```
I would say `autoClose` and `channelResultCallback` does a decent job until we have more use cases.

packages/library-base/src/commonMain/kotlin/io/realm/internal/util/Channel.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal.util
+
+import kotlinx.coroutines.channels.Channel
+
+// Conventional try-with-resource wrapper for channels
+public inline fun <T : Channel<*>, R> T.autoClose(block: (channel: T) -> R): R {
```
cmelchior, line 22: Nitpick: Would `use` be better and fit how try-with-resources is normally called?

cmelchior (approved):

packages/library-base/src/commonMain/kotlin/io/realm/internal/util/Channel.kt
rorbech: I had some issues with autocompletion trying to import `kotlin.io.use` instead and was annoyed by the overlap, but might be that it was because I accidentially trigger that import on from experiments on `File` os something like that. I have renamed it to `use`, then we can see if it causes issues.

PR #475, opened 2021-10-01
Sync - config builder

edualonso:
Added `SyncConfiguration` and builder.
Abstracted `RealmConfiguration` so that "sync configs" are also "local configs".
Added `PartitionValue` container to mimic BsonValue until we have a viable multiplatform library for that.
Added `InternalRealmConfiguration` to avoid exposing certain properties to the public API.
A number of classes (e.g. `Mediator`) are not internal anymore due to `library-base` and `library-sync` being two different modules.

TODO:
- [ ] add jvm test support

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncConfiguration.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncConfiguration.kt
@@ -16,19 +16,16 @@
 
 package io.realm.mongodb
 
-import io.realm.LogConfiguration
 import io.realm.Realm
 import io.realm.RealmConfiguration
 import io.realm.RealmObject
 import io.realm.internal.REPLACED_BY_IR
 import io.realm.internal.RealmConfigurationImpl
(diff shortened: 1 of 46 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncConfiguration.kt
```diff
@@ -0,0 +1,130 @@
...
+        internal fun build(
+            companionMap: Map<KClass<out RealmObject>, RealmObjectCompanion>
+        ): SyncConfiguration {
+            val allLoggers = mutableListOf<RealmLogger>()
+            if (!removeSystemLogger) {
+                allLoggers.add(createDefaultSystemLogger(Realm.DEFAULT_LOG_TAG))
+            }
+            allLoggers.addAll(userLoggers)
+            @Suppress("invisible_member")
+            val localConfiguration = RealmConfiguration.Builder(path, name, schema)
```
rorbech, line 121: This isn't taking all the `SharedBuilder` properties into account!

PR #476, opened 2021-10-01
Use correct ktor client on macos. 

cmelchior:
As part of enabling the Sync demo, I discovered several problems. This PR fixes one of them: namely that using `cio` as network client doesn't work inside macOS apps. It did work inside our macOS tests for some reason, but when installing a shared module in a real macOS app, it resulted in `Duplicate symbols for architecture x86_64 under Xcode` during build time.

Moving to Curl however also has issues as there seems to be some problem with cleanup or a bug in Ktor. I have not yet found the root cause, but creating a new HttpClient for every request fixes the issues (but it adds significant overhead).

The latest Ktor is 1.6.4, but that doesn't work right now due to https://youtrack.jetbrains.com/issue/KTOR-3216, so I just bumped to latest working version.

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
@@ -55,7 +55,8 @@ class KtorNetworkTransport(
     private val dispatcher: CoroutineDispatcher,
 ) : NetworkTransport {
 
-    private val client: HttpClient = getClient()
+    // FIXME Figure out how to reuse the HttpClient across all network requests.
+    // private val client: HttpClient = getClient()
 
     @Suppress("ComplexMethod", "TooGenericExceptionCaught")
     override fun sendRequest(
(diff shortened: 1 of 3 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/KtorNetworkTransport.kt
```diff
@@ -66,6 +67,14 @@ class KtorNetworkTransport(
         usesRefreshToken: Boolean
     ): Response {
         try {
+            // FIXME When using a shared HttpClient we are seeing sporadic
```
rorbech, line 70: Is this a placeholder for investigations as part of this PR? If not, then I think that it at least should have its own issue to track it. 

Maybe also only add the work around for `macos` if that is the only place we are seeing it.

cmelchior: Yes, I was thinking about making some sort of platform-specific strategy here as well. Will do that. I'll also create an issue for investigating the root cause.

packages/library-sync/src/iosMain/kotlin/io/realm/mongodb/internal/HttpClientCache.kt
```diff
@@ -0,0 +1,14 @@
+package io.realm.mongodb.internal
+
+import io.ktor.client.HttpClient
+
+/**
+ * Cache HttpClient on iOS.
+ * https://github.com/realm/realm-kotlin/issues/480 only seem to be a problem on macOS.
+ */
+actual class HttpClientCache actual constructor(timeoutMs: Long) {
+    private val client = createClient(timeoutMs)
```
nhachicha, line 10: isn't iOS also experiencing the same behaviour as in macOS that needs to create a new client each time as a workaround?

nhachicha (approved):

PR #478, opened 2021-10-01
Add easy way to create App without needing to create a Builder.

cmelchior:
Little helper function that avoids the annoying overhead of creating a Builder. We have a similar method in Realm Java.

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
@@ -38,7 +38,7 @@ interface App {
     companion object {
         /**
          * Create an [App] with default settings.
-         * @return
+         * @param appId The MongoDB Realm App ID.
          */
         fun create(appId: String): App {
             Validation.checkEmpty(appId, "appId")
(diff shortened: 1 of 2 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
```diff
@@ -36,10 +37,21 @@ interface App {
...
+        }
+
+        /**
+         * Create an [App] according to the given [AppConfiguration].
+         *
+         * @param configuration The configuration to use for this [App] instance.
+         * @see AppConfiguration.Builder
+         */
+        fun create(configuration: AppConfiguration): App =
+            AppImpl(configuration as AppConfigurationImpl)
```
rorbech, line 55: We don't have test for this. I wouldn't be against leaving it like that as I wouldn't expect that people throw their own `AppConfiguration`-implementation. But maybe do `AppImpl(Validation.checkType<AppConfigurationImpl>(configuration, "configuration))` so we get a better error message.

rorbech (approved):
There is a minor on the argument check, but otherwise LGTM.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
cmelchior: For now, I would probably prefer not to add code like this. Mostly because we would need to add it in a lot of places to be consistent.

Maybe wait to see if people accidentally start passing their own implementations around before doing it?

nhachicha (approved):

PR #484, opened 2021-10-05
Allow returning unmanaged realm object from write

rorbech:


cmelchior (approved):
Nice 👍 

We also need a CHANGELOG entry.

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
```diff
@@ -63,6 +64,11 @@ class MutableRealmTests {
         assertEquals("N.N.", parents[0].name)
     }
 
+    @Test
+    fun writeReturningUnmanaged() {
+        assertTrue(realm.writeBlocking { Parent() } is Parent)
```
nhachicha, line 69: what was the returning type before this change?

rorbech: I didn't change the return type. Just checking that the object also reached the user.

The problem was that we were trying to freeze the resulting object even if it wasn't managed, so just a test that we don't throw pointer exceptions inside freeze.

nhachicha (approved):

PR #517, opened 2021-10-25
App logout

### LINKED ISSUE #245: Add User.logout()

rorbech:
This adds support for logging out users. 

Updated to latest core commit as per 4 November 2021.

TODO:
- [x] Darwin support
- [x] Extra methods to extend test coverage
  - [x] `User.equals`
  - [x] `AppConfiguration.equals`
  - [x] `User.state`

Closes #245

### DIFF
--- a/test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
+++ b/test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
@@ -31,7 +31,6 @@ import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
-import kotlin.test.assertNotEquals
 import kotlin.test.assertNotNull
 import kotlin.test.assertNull
 import kotlin.test.assertTrue


packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -915,10 +919,42 @@ actual object RealmInterop {
...
+        realm_wrapper.realm_app_log_out(
+            app.cptr(),
+            user.cptr(),
+            staticCFunction { userdata, error: CPointer<realm_app_error_t>? ->
+                val userDataCallback = safeUserData<AppCallback<Unit>>(userdata)
+                if (error == null) {
+                    userDataCallback.onSuccess(Unit)
+                } else {
+                    val message = with(error.pointed) {
+                        "${message?.toKString()} [error_category=${error_category.value}, error_code=$error_code, link_to_server_logs=$link_to_server_logs]"
```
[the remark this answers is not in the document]
clementetb: The whole content of the `staticCFunction` could be refactored into another helper function that manages the callbacks, something like:

```
    private fun <R> handleAppCallback(userdata: COpaquePointer?, error: CPointer<realm_app_error_t>?, getValue: ()->R) {
        val userDataCallback = safeUserData<AppCallback<R>>(userdata)
        if (error == null) {
            userDataCallback.onSuccess(getValue())
        } else {
            val message = with(error.pointed) {
                "${message?.toKString()} [error_category=${error_category.value}, error_code=$error_code, link_to_server_logs=$link_to_server_logs]"
            }
            userDataCallback.onError(AppException(message))
        }
    }
```

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/UserTests.kt
```diff
@@ -0,0 +1,510 @@
...
+        assertFalse(user.isLoggedIn())
+    }
+
+    @Test
+    fun equals() = runBlocking {
+        // val user: User = app.registerUserAndLogin(TestHelper.getRandomEmail(), "123456")
+        val email = TestHelper.randomEmail()
+        val password = "123456"
+        app.asTestApp.createUser(email, password)
+        val user = app.login(Credentials.emailPassword(email, password))
```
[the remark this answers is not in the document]
edualonso: I'd rather not change it in this PR since it will be refactored in the `EmailPasswordAuth` PR (https://github.com/realm/realm-kotlin/pull/530)

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/User.kt
```diff
@@ -24,4 +26,77 @@ package io.realm.mongodb
...
+
+    /**
+     * A user's potential states.
+     */
+    enum class State {
+        LOGGED_OUT,
+        LOGGED_IN,
+        REMOVED;
+
+        companion object {
```
rorbech, line 88: We should at least make this internal, but maybe even better just to define it in `UserImpl`?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/UserImpl.kt
```diff
@@ -16,10 +16,60 @@
...
+
+    // TODO Can maybe fail, but we could also cache the return value?
+    override val identity: String
+        get() = RealmInterop.realm_user_get_identity(nativePointer)
+
+    override val loggedIn: Boolean
+        get() = RealmInterop.realm_user_is_logged_in(nativePointer)
+
+    override suspend fun logOut() {
+        return suspendCoroutine { continuation ->
```
rorbech, line 43: I have reworked the pattern for C-API async remote calls in #469 to ensure that requests are performed on a background thread. We should adopt it for all these remote calls.

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -37,32 +37,62 @@ const val TEST_APP_1 = "testapp1" // Id for the default test app
...
+    ) : this(
+        App.create(
+            builder(testAppConfigurationBuilder(appId, logLevel))
+                .dispatcher(dispatcher)
+                .build()
+        )
+    )
 
     fun close() {
+        // This is needed to "properly reset" all sessions across tests since deleting users
```
rorbech, line 74: Is this also needed if we wipe the app cache data #540? If so, where did the discussion on getting `reset` exposed in the C-API?

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmLog.kt
```diff
@@ -75,4 +75,22 @@ class RealmLog(val tag: String = "REALM", val configuration: LogConfiguration) {
             }
         }
     }
+
+    override fun equals(other: Any?): Boolean {
```
rorbech, line 79: Where are `RealmLog`s compared? If essential, then we might need tests for these

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppConfigurationImpl.kt
```diff
@@ -69,4 +69,24 @@ internal class AppConfigurationImpl(
         platformVersion = OS_VERSION,
         sdkVersion = io.realm.internal.SDK_VERSION
     ).freeze()
+
+    override fun equals(other: Any?): Boolean {
```
rorbech, line 73: I guess we would need tests for these too. Maybe there are some from realm-java that we can grab.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/UserTests.kt
```diff
@@ -0,0 +1,449 @@
...
+
+        currentUser.logOut()
+
+        assertNotEquals(User.State.LOGGED_OUT, currentUser.state)
+        assertNotEquals(User.State.LOGGED_OUT, anonUser.state)
+        assertNull(app.currentUser)
+    }
+
+//    @Test
+//    fun repeatedLogInAndOut() = runBlocking {
```
rorbech, line 136: I guess we have all APIs to include this

rorbech (commented):
Nice. Added some comments throughout the PR. Nothing fundamental but in my opinion still something to address.

I am also not sure on how we should treat the missing API when copying the docs from realm-java. Seems like a waste if we need to iterate it every time just to get it 100% correct for each intermediate release 🤔 

Also, are the core reference correct. I wouldn't have expected it to change, but maybe just bumping to master?

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
edualonso: The conversation is here: https://github.com/realm/realm-core/pull/5014
In short, Jonathan doesn't seem too keen about exposing it, nor is Yavor. However, Nikola seemed pretty certain exposing `reste_for_testing` is needed in case we are to make Realm metadata-related assertions in case we change app configurations across test runs. The status now is waiting for Jonathan to comment since he was the most adamant one against the proposed solution.

In any case, I discussed it with Christian and he suggested we could remove the files manually after each test, which is also nice especially after running macos tests. That's what #540 does.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/UserTests.kt
edualonso: Not really, we are missing `User.profile` but I think it's still valid to enable it using the email and password used to create the user.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/AppConfigurationImpl.kt
edualonso: There aren't any equality tests. I've written a couple of basic ones instead.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmLog.kt
edualonso: Done

edualonso:
> Also, are the core reference correct. I wouldn't have expected it to change, but maybe just bumping to master?

While investigating the `invalid_session` issue I updated to the latest core master as per 4 November to see if this was caused by other bugs but it wasn't. I decided to bump it regardless and updated the changelog since there was not that many changes.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/UserImpl.kt
```diff
@@ -17,9 +17,68 @@
...
+    override suspend fun logOut() {
+        Channel<Result<Unit>>(1).use { channel ->
+            RealmInterop.realm_app_log_out(
+                app.nativePointer,
+                nativePointer,
+                channelResultCallback<Unit, Unit>(channel) {
+                    // No-op
+                }.freeze()
+            )
+        }
```
rorbech, line 54: I guess we still need some suspending call until the underlying operation is done, so that we don't return before the operation has actually been performed.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/UserTests.kt
rorbech: Sorry, missed that. I guess whatever makes most sense for you then.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/UserTests.kt
```diff
@@ -0,0 +1,510 @@
...
+
+    @Test
+    fun equals() = runBlocking {
+        // val user: User = app.registerUserAndLogin(TestHelper.getRandomEmail(), "123456")
+        val email = TestHelper.randomEmail()
+        val password = "123456"
+        app.asTestApp.createUser(email, password)
+        val user = app.login(Credentials.emailPassword(email, password))
+        assertEquals(user, user)
+        assertFalse(user == app)
```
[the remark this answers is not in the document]
edualonso: I removed it since it didn't make sense to me either.

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -915,10 +919,42 @@ actual object RealmInterop {
...
+                    }
+                    userDataCallback.onError(AppException(message))
```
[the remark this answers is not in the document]
rorbech: This was added to `cinterop` in the initial implementation, which was technically sufficient before we had docs as `cinterop` is an `api`-dependency. We don't generate dokka for this, so it will not be part of the documentation, and agree that we should get to a point where there is no public API in `cinterop`. 

Putting `AppException` in `library-sync` will require us to have a separate cinterop-exception, that is caught and rethrown in `library-sync`. This specifically ties into the issue of proper abstraction for all our error codes of #522, so an initial layering cinterop-exception -> AppException is probably just one step towards making an appropriate mapping in `library-sync`. 

I would argue that this rework should go into another PR as it also relates to `login`.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
```diff
@@ -54,6 +54,18 @@ interface App {
 
     val configuration: AppConfiguration
 
+    /**
+     * Returns the current user that is logged in and still valid.
+     *
+     * A user is invalidated when he/she logs out or the user's refresh token expires or is revoked.
+     *
+     * If two or more users are logged in, it is the last valid user that is returned by this method.
```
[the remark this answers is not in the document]
edualonso: I've written the following test and seems to confirm what the docs say:
```
        val user1 = createUserAndLogin(email1, password1)
        assertEquals(app.currentUser, user1)
        val user2 = createUserAndLogin(email2, password2)
        assertNotEquals(app.currentUser, user1)
        assertEquals(app.currentUser, user2)
        user2.logOut()
        assertNotEquals(app.currentUser, user2)
        assertEquals(app.currentUser, user1)
```

cmelchior (approved):

PR #530, opened 2021-10-29
Sync - email-password auth wrapper

edualonso:
Added `EmailPasswordAuth` wrapper to `App` class. It allows interacting with the `EMAIL_PASSWORD` provider for operations such as registering users, resetting password, etc. This first implementation only supports `registerUser(email, password)`.

### TODO
- [x] Refactor interop callbacks to use logout PR approach (see https://github.com/realm/realm-kotlin/pull/517/files#diff-2d4545b3313381a3e914ead2739f8816101e15b2e56da8678c624d15249e9c7bR26 and https://github.com/realm/realm-kotlin/pull/517/files#diff-b191a2b8280990335be6159a3281434b292df6d13e86ce0e05ff0c55085e1161R46)


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/EmailPasswordAuth.kt
```diff
@@ -0,0 +1,46 @@
...
+import io.realm.internal.interop.CinteropVoidCallback
+import io.realm.internal.interop.NativePointer
+import io.realm.internal.interop.RealmInterop
+import io.realm.internal.util.Validation
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+/**
+ * Class encapsulating functionality provided when [User]s are logged in through the
```
rorbech, line 12:
```suggestion
 * Class encapsulating functionality for managing [User]s through the
```

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -228,6 +228,7 @@ void register_login_cb(realm_app_t *app, realm_app_credentials_t *credentials, j
             credentials,
             // FIXME Refactor into generic handling of network requests, like
             //  https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsApp.cpp#L192
+            //  see 'app_register_email_password'
```
rorbech, line 231: This is generalized in https://github.com/realm/realm-kotlin/pull/517/files#diff-b191a2b8280990335be6159a3281434b292df6d13e86ce0e05ff0c55085e1161R46

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -509,3 +510,54 @@ void sync_set_error_handler(realm_sync_config_t* sync_config, jobject error_hand
                                             get_env(true)->DeleteGlobalRef(static_cast<jobject>(userdata));
                                         });
 }
+
+void app_register_email_password(realm_app_t* app,
```
rorbech, line 514: We should merge this one after #517 to piggyback on the refactor of app callbacks: https://github.com/realm/realm-kotlin/pull/517/files#diff-b191a2b8280990335be6159a3281434b292df6d13e86ce0e05ff0c55085e1161R46

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -1101,6 +1101,34 @@ actual object RealmInterop {
...
+        password: String,
+        callback: CinteropVoidCallback
+    ) {
+        memScoped {
+            realm_wrapper.realm_app_email_password_provider_client_register_email(
+                app.cptr(),
+                email,
+                password.toRString(this),
+                staticCFunction { userData, error ->
+                    val userDataCallback = safeUserData<CinteropVoidCallback>(userData)
```
rorbech, line 1116: Even though the code is more compact than for JVM we should probably abstract this _callback handling_ into a method too.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -0,0 +1,552 @@
...
+    @AfterTest
+    fun tearDown() {
+        if (this::app.isInitialized) {
+            app.asTestApp.close()
+        }
+    }
+
+    @Test
+    fun registerUser() {
+        runBlocking {
```
rorbech, line 34: You could use expression body for this

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -0,0 +1,552 @@
...
+    @Test
+    fun registerUser_invalidServerArgsThrows() {
+        runBlocking {
+            val email = "invalid-email"
+            val password = "1234"
+
+            // TODO do exhaustive exception assertion once we have all AppException fields in place
+            assertFailsWith<AppException> {
+                app.emailPasswordAuth.registerUser(email, password)
+                fail("Should never reach this.")
```
rorbech, line 53: This is already asserted by the `assertFailsWith`

rorbech (commented):
Looks good, but should probably be merged after #517 as the app callback conventions have been addressed there.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -0,0 +1,552 @@
...
+    fun tearDown() {
+        if (this::app.isInitialized) {
+            app.asTestApp.close()
+        }
+    }
+
+    @Test
+    fun registerUser() {
+        runBlocking {
+            val email = TestHelper.randomEmail()
```
clementetb, line 35: We could statically import the randomEmail function.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -0,0 +1,552 @@
...
+            val credentials = Credentials.emailPassword(email, password)
+            val user = app.login(credentials)
+            assertNotNull(user)
+        }
+    }
+
+    @Test
+    fun registerUser_invalidServerArgsThrows() {
+        runBlocking {
+            val email = "invalid-email"
```
clementetb, line 47: Maybe we should test independently invalid username and invalid password?

clementetb (approved):
LGTM, some minor changes 🚀

clementetb:
Just noticed that we have some email auth tests in AppTests ex: `AppTests.loginEmailPassword`, that should be removed as part of this PR.

rorbech (commented):
Nice. Still needs to update to the async pattern from #469 when brought up to master and also needs check for C-API errors in the Darwin c-interop.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
edualonso: The reason why I didn't convert it to an expression body is because our custom `runBlocking` function returns a `T` produced inside the lambda block and if you use it in a void function (which is the case in these tests) turns out you need to manually add `Unit` as the last expression inside the lambda, which is kind of annoying. That's why I left it as is. I suspect I forgot to write this explanation 🙈

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -1079,6 +1079,26 @@ actual object RealmInterop {
...
     }
 
+    actual fun realm_app_email_password_provider_client_register_email(
+        app: NativePointer,
+        email: String,
+        password: String,
+        callback: AppCallback<Unit>
+    ) {
+        memScoped {
+            realm_wrapper.realm_app_email_password_provider_client_register_email(
```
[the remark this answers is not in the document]
edualonso: This function returns `realm_app_credentials_t` so what you mention doesn't apply here.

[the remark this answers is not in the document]
rorbech: In my checkout it returns bool. Also seems to do that in core master https://github.com/realm/realm-core/blob/master/src/realm.h#L2276 🤔 

If the C-API method returns `realm_app_credentials_t` shouldn't we then return this?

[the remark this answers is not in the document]
edualonso: Weird, I must have looked at `realm_app_credentials_new_email_password` instead 🐒

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -85,6 +87,15 @@ class TestApp(
         deleteAllUsers()
     }
 
+    suspend fun createUserAndLogIn(email: String, password: String): User =
```
rorbech, line 90: I think it would be nice if we could hold all methods using our actual API on the `App` and leave the `TestApp` for stuff that is only related to test. At least that was the motivation of making the explicit `app.asTestApp()` entry point, as I found it confusing in `realm-java` that we had a lot of extension methods on `App` with test infrastructure code. 

All these methods are practically just using our actual API, so could just be extension methods on `App`.

rorbech (approved):
LGTM. Only some minor comments on keeping a clear separation of API/test method namespace.

PR #531, opened 2021-11-01
Update Android Build dependencies

cmelchior:
This PR updates a lot of dependencies used for building to Android:

- Gradle 7.2
- Android Gradle Plugin 7.1.0-beta03 (This can now build Realm Core from source)
- Updating AGP, also means that we no longer have restrictions on CMake. Latest 3.21.4 can be used. Updated description in README about this.
- Updated to latest NDK release (which is also an LTS release): 23.1.7779620
- Updated to Android targets 31 and compatible build tools 31.0.0

I do not remember exactly why we had this:

```
    resolutionStrategy {
        eachPlugin {
            if (requested.id.namespace == "com.android") {
                useModule("com.android.tools.build:gradle:4.1.0")
            }
        }
    }
```

But it no longer seems to be needed.


rorbech (approved):
🚀  Awesome ... when CI is green ;)

> I do not remember exactly why we had this:
> 
> ```
>     resolutionStrategy {
>         eachPlugin {
>             if (requested.id.namespace == "com.android") {
>                 useModule("com.android.tools.build:gradle:4.1.0")
>             }
>         }
>     }
> ```
> 
> But it no longer seems to be needed.

This was part of the template for multi platform projects when we booted the project. Probably some compatibility constraint that has been fixed somewhere in the KMP gradle plugin.

cmelchior:
Converting back to Draft. We need to rethink how we verify Java 8 support. Most likely by creating local m2-folder that e.g. the kmm-sample and the test project can use instead of using `includeBuild`

cmelchior:
With the upgrade to use AGP 7.*, it also brings a requirement for Java 11. So we cannot use the same versions for testing backwards compatibility.

I'm going to clone the `kmm-sample` into a `min-kmm-sample` which will be configured with our minimum supported versions of Gradle and AGP (this is also unknown right). This also ensure that we actually catch it if our minimum supported compatibility changes.

cmelchior:
This is now ready for review. Sorry for the size. I added a new `min-android-sample` project that is setup with what seems our currently minimum supported setup. This project only compiles if packages have been deployed to a local maven repo. 

It isn't possible to use `buildSrc` in this project as the buildSrc now requires Gradle 7 to compile.

examples/kmm-sample/build.gradle.kts
```diff
@@ -1,5 +1,8 @@
 buildscript {
     repositories {
+        if (Realm.ciBuild) {
```
rorbech, line 3: Can remember if we are allowed to have logic before the `buildScript` section, but maybe define one variable for the configuration and initialize it with `Realm.ciBuild`. 
```
val addBuildFolderRepository = Realm.ciBuild
```

test/base/build.gradle.kts
```diff
@@ -20,16 +20,10 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithSimulatorTes
...
     id("realm-lint")
 }
 
-repositories {
-    google()
-    jcenter()
-    mavenCentral()
-    mavenLocal()
-}
+apply(plugin = "io.realm.kotlin")
```
rorbech, line 26: Can't you apply it in the `plugins` section even for local repositories when you have the plugin on the classpath with `classpath("io.realm.kotlin:gradle-plugin:${Realm.version}")` (as configured in `test/build.gradle.kts`?

test/build.gradle.kts
```diff
@@ -20,9 +20,26 @@ plugins {
...
+
 allprojects {
     repositories {
+        if (Realm.ciBuild) {
+            maven("file://${rootProject.rootDir.absolutePath}/../packages/build/m2-buildrepo")
+        }
+        google()
         jcenter()
+        mavenCentral()
+        mavenLocal()
```
rorbech, line 42: Not needed? If added for convenience I guess it would be better with a flag like `useBuildFolderRepository` in the top of the file, so it is easy _flip_-able.

test/sync/build.gradle.kts
```diff
@@ -20,17 +20,11 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithSimulatorTes
...
     kotlin("plugin.serialization") version Versions.kotlin
 }
 
-repositories {
-    google()
-    jcenter()
-    mavenCentral()
-    mavenLocal()
-}
+apply(plugin = "io.realm.kotlin")
```
rorbech, line 27: Same. Can't we apply in `plugins` when it is already on the class path?

rorbech (approved):
LGTM. Minor comments here and then, but nothing blocking.

Jenkinsfile
```diff
@@ -399,6 +396,13 @@ def runCompilerPluginTest() {
             cd packages
             ./gradlew --no-daemon :plugin-compiler:test --info --stacktrace
         """
+        // See https://stackoverflow.com/a/51206394/1389357
+        script {
+            def testResults = findFiles(glob: "packages/plugin-compiler/build/**/TEST-*.xml")
```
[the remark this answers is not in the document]
cmelchior: Yeah, I wasn't exactly sure where this logic belonged. It seemed to be a Jenkins issue, which was why I fixed it in the JenkinsFile. I created https://github.com/realm/realm-kotlin/issues/546 to track it.

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -52,26 +53,21 @@ object Realm {
...
     const val androidxStartup = "1.1.0-beta01" // https://maven.google.com/web/index.html?q=startup#androidx.startup:startup-runtime
     const val androidxJunit = "1.1.3-beta02" // https://maven.google.com/web/index.html#androidx.test.ext:junit
     const val androidxTest = "1.4.0-beta02" // https://maven.google.com/web/index.html#androidx.test:rules
     // Must be built with same (major.minor!?) kotlin version as 'kotlin' variable below, to be binary compatible with kotlin
     const val atomicfu = "0.16.3" // https://github.com/Kotlin/kotlinx.atomicfu
     const val autoService = "1.0" // https://mvnrepository.com/artifact/com.google.auto.service/auto-service
-    // Cannot upgrade past this due to https://stackoverflow.com/questions/67120904/cannot-properly-link-c-project-with-gradle-exception-during-working-with-exte
-    // Upgrading might be possible if we can bump Android Gradle Plugin, but that is blocked on another bug.
-    const val cmake = "3.18.1" // Core requires minimum 3.15, but 3.18.1 is available through the Android SDK
+    // Not currently used, so mostly here for documentation. Core requires minimum 3.15, but 3.18.1 is available through the Android SDK.
```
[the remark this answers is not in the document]
cmelchior: I did consider that, but that forces a strict requirement on a specific version. Not sure we want to do that? Currently what determines the CMake version is the minimum_version setup in the CMake file itself, and I don't think there is a way to hook the value here up to that.

I created https://github.com/realm/realm-kotlin/issues/547 to track it.

CHANGELOG.md
```diff
@@ -7,16 +7,25 @@
...
 
 ### Compatibility
 * This release is compatible with:
-  * Kotlin 1.5.31
-  * Coroutines 1.5.2-native-mt
-  * AtomicFu 0.16.3
+  * Kotlin 1.5.31.
+  * Coroutines 1.5.2-native-mt.
+  * AtomicFu 0.16.3.
+* Minimum Java version: 11.
```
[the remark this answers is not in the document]
cmelchior: Good point. This line is misleading. I think I added it because AGP 7 requires it, but you should still be able to use Java 8. I'll remove it.

examples/kmm-sample/androidApp/src/main/AndroidManifest.xml
```diff
@@ -22,10 +22,9 @@
             android:allowBackup="false"
             android:supportsRtl="true"
             android:theme="@style/AppTheme">
-        <activity android:name="io.realm.example.kmmsample.androidApp.MainActivity">
+        <activity android:name="io.realm.example.kmmsample.androidApp.MainActivity" android:exported="true">
```
nhachicha, line 25: We don't have any broadcast receiver, why add `exported=true`?

buildSrc/src/main/kotlin/io/realm/RealmPublishPlugin.kt
```diff
@@ -82,6 +94,7 @@ class RealmPublishPlugin : Plugin<Project> {
             configureRootProject(project)
         } else {
             configureSubProject(project, signBuild)
+            configureCustomRepository(project)
```
nhachicha, line 97: Can we guard this by the CI environment since the repo is already consumed when building from CI

nhachicha (approved):
👍

examples/kmm-sample/androidApp/src/main/AndroidManifest.xml
cmelchior: It is now required by Android 12 for all activities started from the launcher.

examples/kmm-sample/build.gradle.kts
cmelchior: it isn't possible to have a global variable, but I could configure the extra properties instead.

test/base/build.gradle.kts
cmelchior: Yes, I remember running into issues in the plugin section and this fixed it, but I just tried moving it back and that also worked. So probably some other caching issue was at play.

buildSrc/src/main/kotlin/io/realm/RealmPublishPlugin.kt
cmelchior: All this does is setting up new publication targets, so it doesn't really impact either CI or local development unless you call them (which right now only CI does). But I think we should keep it for both cases, just in case of debugging.

PR #533, opened 2021-11-01
Separating generated swig code from manual code

rorbech:
This just puts all swig-generated content into `jni-swig-stub/build/generated/sources/{jni,java}` and adds appropriate incremental build dependencies to input files.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -284,7 +284,7 @@ def genAndStashSwigJNI() {
         ../gradlew assemble
         """
     }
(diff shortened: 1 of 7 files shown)


cmelchior (commented):
Nice 💯 , it does seem like this break building the JVM libraries though?

rorbech:
> Nice 💯 , it does seem like this break building the JVM libraries though?

Of course. I only fixed one out of four cmakefiles 🤦‍♂️🤣🤪

clementetb (approved):
Great refactor, makes totally sense to move it out.

nhachicha (approved):
🥇

PR #535, opened 2021-11-04
Move Index to public source folder

rorbech:


cmelchior (commented):
Ups 👍 ... Should also be noted in the changelog

cmelchior (approved):

rorbech:
> Ups 👍 ... Should also be noted in the changelog

The actual package is correct. It was just in the wrong source folder, so no changelog notice is needed.

PR #540, opened 2021-11-05
Cleanup of files related to metadata and mongodb apps after every test

edualonso:
Added cleanup of metadata and mongodb app-related files after testing. This is done using Okio multiplatform.

Additionally, the `realm_clear_cached_apps` function was added as we need to clear the app cache before deletion.

### DIFF
--- a/test/sync/build.gradle.kts
+++ b/test/sync/build.gradle.kts
@@ -54,6 +54,8 @@ kotlin {
                 implementation("io.ktor:ktor-client-core:${Versions.ktor}")
                 implementation("io.ktor:ktor-client-serialization:${Versions.ktor}")
                 implementation("io.ktor:ktor-client-logging:${Versions.ktor}")
+
+                implementation("com.squareup.okio:okio:${Versions.okio}")
             }
         }
 
--- a/test/sync/build.gradle.kts
+++ b/test/sync/build.gradle.kts
@@ -54,7 +54,6 @@ kotlin {
                 implementation("io.ktor:ktor-client-core:${Versions.ktor}")
                 implementation("io.ktor:ktor-client-serialization:${Versions.ktor}")
                 implementation("io.ktor:ktor-client-logging:${Versions.ktor}")
-
                 implementation("com.squareup.okio:okio:${Versions.okio}")
             }
         }
--- a/test/sync/build.gradle.kts
+++ b/test/sync/build.gradle.kts
@@ -153,6 +153,25 @@ kotlin {
 }
 
 kotlin {
+    jvm()
+    sourceSets {
+        getByName("jvmMain") {
+            dependencies {
+                implementation("org.jetbrains.kotlin:kotlin-compiler-embeddable:${Versions.kotlin}")
+                implementation("io.realm.kotlin:plugin-compiler:${Realm.version}")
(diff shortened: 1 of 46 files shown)


test/sync/build.gradle.kts
```diff
@@ -54,6 +54,8 @@ kotlin {
                 implementation("io.ktor:ktor-client-core:${Versions.ktor}")
                 implementation("io.ktor:ktor-client-serialization:${Versions.ktor}")
                 implementation("io.ktor:ktor-client-logging:${Versions.ktor}")
+
```
cmelchior, line 57:
```suggestion
```

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -37,7 +38,7 @@ class AppTests {
 
     @BeforeTest
     fun setup() {
-        app = TestApp()
+        app = TestApp(fileSystem = FileSystem.SYSTEM)
```
cmelchior, line 41: It seems needlessly complicated to have to inject this everywhere? Why not just set the default value in the constructor?

cmelchior (approved):

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
edualonso: Not quite. `FileSystem.SYSTEM` is platform-specific so it cannot be referenced from `commonMain` directly and `TestApp` is placed there. I will write an `expect` variable and use it in the constructor and implement it for each platform instead.

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -41,20 +45,29 @@ const val TEST_APP_1 = "testapp1" // Id for the default test app
...
     AdminApi by (runBlocking(dispatcher) { AdminApiImpl(TEST_SERVER_BASE_URL, appId, debug, dispatcher) }) {
 
     fun close() {
         deleteAllUsers()
+
+        // Make sure to clear cached apps before deleting files
+        RealmInterop.realm_clear_cached_apps()
+
+        // Delete metadata Realm files
+        fileSystem.deleteRecursively((appFilesDirectory() + "/mongodb-realm").toPath())
```
rorbech, line 70: This looks a bit coincidental. We don't specify `appFilesDirectory` as part of the AppConfiguration. Are we sure that it always works on all platforms?

edualonso: I added jvmTest (sync didn't have it yet 🙊) and it does delete the files 👍

rorbech: Ok. I guess manually validating it on `iOS` too would be ok for now. 

It would be nice to understand how the this worked, but I guess either this data is only created along with the realm or there must be something corresponding to our `appFilesDirectory` somewhere in core ... maybe just `./` somehow ending up the same place 🤔

PR #542, opened 2021-11-08
Consolidating CMakes build files 

nhachicha:
Keeping one CMake file to build Android and JVM targets (Win/Linux/Darwin)


packages/cinterop/src/jvmMain/CMakeLists.txt
```diff
@@ -4,18 +4,40 @@ find_package(JNI)
...
+    set(CMAKE_CXX_STANDARD 17)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(REALM_INCLUDE_DIRS "${CMAKE_BINARY_DIR}/src" "${JAVA_INCLUDE_PATH}" "${JAVA_INCLUDE_PATH}/win32" "${CINTEROP_JNI}" "${SWIG_JNI_GENERATED}" "${SWIG_JNI_HELPERS}")
 
+elseif (CMAKE_SYSTEM_NAME MATCHES "^Android")
+    MESSAGE("Building JNI for Android")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
+    set(CAPI_BUILD "${CMAKE_SOURCE_DIR}/../../../external/core}/build-android-${ANDROID_ABI}-${CMAKE_BUILD_TYPE}")
+    set(REALM_INCLUDE_DIRS ${CAPI_BUILD}/src ${CINTEROP_JNI} ${SWIG_JNI_GENERATED} ${SWIG_JNI_HELPERS})
+    set(REALM_TARGET_LINK_LIBS log android RealmFFIStatic Realm::ObjectStore)
```
cmelchior, line 20: It feels a bit akward that we also define `RealmFFIStatic Realm::ObjectStore`.... would it instead make sense to  rename `REALM_TARGET_LINK_LIBS` to `SHARED_REALM_TARGET_LINK_LIBS` (or something) and then set `REALM_TARGET_LINK_LIBS` for each target?

packages/cinterop/src/jvmMain/CMakeLists.txt
```diff
@@ -4,18 +4,40 @@ find_package(JNI)
 set(SWIG_JNI_GENERATED "${CMAKE_SOURCE_DIR}/../../../jni-swig-stub/build/generated/sources/jni")
 set(SWIG_JNI_HELPERS "${CMAKE_SOURCE_DIR}/../../../jni-swig-stub/src/main/jni")
 set(CINTEROP_JNI "${CMAKE_SOURCE_DIR}/../jvm/jni")
+set(REALM_TARGET_LINK_LIBS RealmFFIStatic Realm::ObjectStore)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
+if (CMAKE_SYSTEM_NAME MATCHES "^Windows")
+    MESSAGE("Building JNI for Windows")
+    set(CMAKE_CXX_STANDARD 17)
```
cmelchior, line 11: Is this the only way to set `-std=c++17"`on Windows? It would have been nice if we could have move `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")` outside all the targets

cmelchior (approved):
Nice 👏 ... I only had a few minor suggestions for refactorings.

packages/cinterop/src/jvmMain/CMakeLists.txt
```diff
@@ -4,18 +4,40 @@ find_package(JNI)
...
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(REALM_INCLUDE_DIRS "${CMAKE_BINARY_DIR}/src" "${JAVA_INCLUDE_PATH}" "${JAVA_INCLUDE_PATH}/win32" "${CINTEROP_JNI}" "${SWIG_JNI_GENERATED}" "${SWIG_JNI_HELPERS}")
```
rorbech, line 13: Wouldn't it be possible to extract the common part of these and only append the platform specifics inside the configuration. This still smells like forgetting to add includes for a single platform.

rorbech (approved):
LGTM. Only some minor comments. 

Also, I would place the common `CmakeLists.txt` in `src/jvm` as this is the common JVM sources (JVM+Android) whereas `src/jvmMain` is for the non-Android JVM targets.

edualonso (approved):
👍

packages/cinterop/src/jvmMain/CMakeLists.txt
nhachicha: Splitting this variable will force us to set the target for each platform, whereas now only Android is the exception (the other 3 inherit the default linking behaviour)

PR #545, opened 2021-11-09
Fixing iOS tests

nhachicha:
- Fixed iOS Test, previously Kotlin/Native linker was linking against a macos klib instead of iOS causing 

```
file or directory '/Users/nabil/Dev/realm/KMP/realm-kotlin/test/base/src/iosMain/kotlin', not found
Run in-process tool "konanc"
```

Before:
```
Task :base:compileTestKotlinIos FAILED
file or directory '/Users/nabil/Dev/realm/KMP/realm-kotlin/test/base/src/iosTest/kotlin', not found
Caching disabled for task ':base:compileTestKotlinIos' because:
 Build cache is disabled
Task ':base:compileTestKotlinIos' is not up-to-date because:
 No history is available.
file or directory '/Users/nabil/Dev/realm/KMP/realm-kotlin/test/base/src/iosTest/kotlin', not found
Run in-process tool "konanc"
Entry point method = org.jetbrains.kotlin.cli.utilities.MainKt.daemonMain
Arguments = [
       -g
       -ea
       -target
       ios_x64
       -p
       library
       -o
       /Users/nabil/Dev/realm/KMP/realm-kotlin/test/base/build/classes/kotlin/ios/test/klib/base_test.klib
       -l
       /Users/nabil/Dev/realm/KMP/realm-kotlin/test/base/build/classes/kotlin/ios/main/klib/base.klib
       -l
       /Users/nabil/Dev/realm/KMP/realm-kotlin/test/base/build/classes/kotlin/macos/main/klib/base.klib 💣💣
       -l

```

- Fixed `mkdtemp` POSIX call that was causing an issue on iOS since the absolute path returned did contain invalid char
```
PATH = �ar/folders/2p/3zz6mbpn33xbl69w3jrr547c0000gp/T/com.apple.CoreSimulator.SimDevice.F265A95F-F1FA-4856-AC7D-EBEA9BCD3F27.Standalone.3B9D4847-0304-4CD2-A348-C00A6B9B12CA/realm_test_.8Utd0n
Invalid connection: com.apple.coresymbolicationd
```

- Enhance debugging

Core exception are more verbose to help investigate 

Before: we get only 
```
kotlin.IllegalArgumentException: Could not open Realm with the given configuration
```

Now: we have type of exception and the original configuration dumped 
```
kotlin.IllegalArgumentException: RealmCoreException [RLM_ERR_OTHER_EXCEPTION]: Directory at path '/Users/nabil/Library/Developer/CoreSimulator/Devices/F265A95F-F1FA-4856-AC7D-EBEA9BCD3F27/data/Documents/default.realm.lock' does not exist. Could not open Realm with the given configuration: path=/Users/nabil/Library/Developer/CoreSimulator/Devices/F265A95F-F1FA-4856-AC7D-EBEA9BCD3F27/data/Documents/default.realm name=default.realm maxNumberOfActiveVersions=9223372036854775807 schemaVersion=0 deleteRealmIfMigrationNeeded=false schema=[class io.realm.entities.Sample]
```


packages/library-base/src/commonMain/kotlin/io/realm/internal/InternalRealmConfiguration.kt
```diff
@@ -33,4 +33,13 @@ interface InternalRealmConfiguration : RealmConfiguration {
     val nativeConfig: NativePointer
     val notificationDispatcher: CoroutineDispatcher
     val writeDispatcher: CoroutineDispatcher
+
+    fun debug(): String {
```
edualonso, line 37:
```suggestion
    override fun toString(): String {
```

edualonso (approved):
Good to go whenever CI is happy

packages/library-base/src/commonMain/kotlin/io/realm/internal/InternalRealmConfiguration.kt
nhachicha: This is not possible on an interface. "An interface may not implement a method of 'Any'"

test/base/src/commonTest/kotlin/io/realm/test/NotificationTests.kt
```diff
@@ -39,23 +39,27 @@ interface NotificationTests {
     fun closingRealmDoesNotCancelFlows()
 
     @Test
-    fun addChangeListener() {
+    @Ignore
```
rorbech, line 42: Doesn't ignoring the test goes completely against the concept of having the test in the interface. I think the issue is that we didn't get to the end designing this. There should probably never be an empty default implementation here and if none of the implementation classes actually implements it then it should probably just not be defined. At least not without an issue reference stating what is missing. 

There are also other issues, that implementations mess up the JVM test, so maybe we should just leave it how you have fixed it and then do a separate issue to revisit the pattern.

test/sync/src/androidTest/kotlin/io/realm/test/shared/SyncedRealmTests.kt
```diff
@@ -93,6 +94,7 @@ class SyncedRealmTests {
     }
 
     @Test
+    @Ignore // FIXME Until https://github.com/realm/realm-kotlin/issues/555 is fixed
```
rorbech, line 97: This seem quite essential for our sync verification test. If we ignore this then we have no sync test. Is it only an issue with iOS, since it hasn't been an issue before? If so, then we should maybe move the test out of `shared` and include it on the platform where it works, or can we just somehow ignore in on iOS

rorbech (commented):
I would be a bit hesitant to ignore our only sync test on all platforms. Is #555 a general problem or only related to iOS test?

test/sync/src/androidTest/kotlin/io/realm/test/shared/SyncedRealmTests.kt
nhachicha: As described in the [issue](https://github.com/realm/realm-kotlin/issues/555) it's not related to iOS only (experienced on Android & macOS) so fixing the issue is high priority IMO

rorbech: Yes, then it sounds like we should jump on it. 

I don't think I have seen it in any of my PR builds yet. Have we any idea about when it started or has it always been there?

nhachicha: I re-enabled the ios test

test/base/src/commonTest/kotlin/io/realm/test/NotificationTests.kt
nhachicha: test re-enabled

PR #553, opened 2021-11-16
Data migration support

### LINKED ISSUE #87: Design Migration API
Realm Java right now exposes a manual migration API, while Cocoa/JS/.NET as an almost automatic migration.

Back when these API's where developed we discussed the pros and cons for both approaches. After 3-4 years with both of these API's it is probably safe to conclude that neither negative consequences for these API's have materialized in any meaningful other.

So for the sake of moving fast it is probably easier to expose automatic migration, since that would require less API's from us.

**TODO/Questions**

* Figure out exactly how migrations should work in Realm Kotlin. What interfaces/API's are needed to support an automatic upgrade process
* If we later decide to re-add support for manual migrations. How can this be done? Ideally it can be done without a breaking change.

rorbech:
This add support for callbacks when schema needs to be migrated. I relies on automatic schema migration but does a callback when the schema is changed to allow the user to perform any migration of data. 

Probably requires:
- [ ] Dynamic (string based) API to allow to objects from the old instance. 

Closes #87


packages/library-base/src/commonMain/kotlin/io/realm/schema/RealmMigration.kt
```diff
@@ -0,0 +1,78 @@
...
+//    /**
+//     * This method will be called if a migration is needed. The entire method is wrapped in a
+//     * write transaction so it is possible to create, update or delete any existing objects
+//     * without wrapping it in your own transaction.
+//     *
+//     * @param realm the Realm schema on which to perform the migration.
+//     * @param oldVersion the schema version of the Realm at the start of the migration.
+//     * @param newVersion the schema version of the Realm after executing the migration.
+//     */
+    fun migrate(old: BaseRealm?, new: MutableRealm?)// , oldVersion: Long, newVersion: Long)
```
cmelchior, line 77: I'm not sure this is the exact API we want. At least looking at the Cocoa API with a (Migration, NewRealm) block that seems to be better suited for this use case. Or we need to find a good way to surface the functionality required. These are the features available in Cocoa: https://docs.mongodb.com/realm-sdks/objc/latest/Classes/RLMMigration.html

rorbech:
Superseded by #674

PR #557, opened 2021-11-18
RealmQuery POC

cmelchior:
POC of a possible Query API based on our discussions. This PR is just intended to facilitate discussion and should not be merged.

This API makes it possible to run both object and scalar queries in blocking and non-blocking variants. 

The main idea is to have 3 classes: `RealmQuery`, `RealmElementQuery` and `RealmScalarQuery` which then offer different "execute" methods.

Some design considerations:

1) What should we call the `query/filter` method? We should strive to make it the same across all classes. Other SDK's use `filter`. Which is also offered by the Kotlin stdlib. Overriding it offers easier discoverability. Downside is that people might use the wrong method by accident.

2) Should we expose `min/max/cont/average` directly on RealmResults. They are slower than queries, but faster than the similar Kotlin stdlib methods?

3) Should we support helper methods like `findFirst()` or let people use `findAll().firstOrNull`?

4) I renamed `observe` to `asFlow`. This was a discussion I had with @edualonso a few weeks ago. `observe` feels like `collect` and `asFlow` are used by other frameworks.

5) The biggest tradeoff is that we cannot prevent people from calling `asFlow()` inside a write transaction. Runtime exceptions are needed here. This seems acceptable.

6) This API did try to take into consideration that we also need to support queries on Lists of primitives, but without a specific `PrimitiveRealmList` it is always going to be a bit ugly, e.g. people need to put a placeholder in where properties are required.

### DIFF
--- a/packages/library-base/build.gradle.kts
+++ b/packages/library-base/build.gradle.kts
@@ -56,6 +56,7 @@ kotlin {
     sourceSets {
         commonMain {
             dependencies {
+                implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.3.1")
                 implementation(kotlin("stdlib-common"))
                 implementation(kotlin("reflect"))
                 // If runtimeapi is merged with cinterop then we will be exposing both to the users
--- a/packages/library-base/src/commonMain/kotlin/io/realm/prototype/RealmQuery.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/prototype/RealmQuery.kt
@@ -1,0 +1,248 @@
+package io.realm.prototype
+
+import io.realm.MutableRealm
+import io.realm.RealmObject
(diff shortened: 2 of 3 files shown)


cmelchior:
Summary of review discussion:

- We are pursuing this direction
- Remove aggregate methods on Collections. Nudge people towards a query instead. We can re-evaluate based on feedback.
- Add a `first` (name TBD) builder and a `RealmSingleQuery` to bridge between Collection and Object fine-grained notifications
- Keep using Strings as property input. Property name support can be added through extension methods and a Realm IntelliJ plugin might be a better way to solve this. Keep it minimal for now.
- Use `query` to start queries. Matches what happens and prevents people from accidentally using the wrong "filter" method.

edualonso:
No longer relevant - see https://github.com/realm/realm-kotlin/pull/603

Issue #544, opened 2021-11-09
Multiplatform build fails when not using Realm in all source sets


FERNman:
When using the Realm Gradle plugin (0.7.0) in a multiplatform project, the build fails for source sets not using the Realm library (in our case, the JVM) with the following error message:

```
java.lang.IllegalStateException: Cannot find io.realm.RealmConfiguration.Builder on platform JVM (JVM_1_8).
```

This was not the case with older versions of the plugin (we upgraded from 0.4.1). As far as I can tell, it was introduced by #395.
The issue comes from the schema lowering extension that tries to find [the realm configuration builder](https://github.com/realm/realm-kotlin/blob/2b07af06553b49509c1d26c620f3fd0c009b6547/packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt#L74) (and throws if it doesn't exist) before the builder was actually used for the first time.

rorbech:
Thanks for the report. We tried to cache class lookups early to avoid duplicate look ups, but didn't think of the use case that the plugin is running on source sets that doesn't depend on our library.

nhachicha:
@FERNman This has been fixed, you can try the SNAPSHOT [release](https://github.com/realm/realm-kotlin#using-snapshots), or wait for the upcoming `0.8.0` release.

FERNman:
@nhachicha Thanks for the quick fix! 🙂

PR #593, opened 2021-11-24
Disable symbol  lookup for source set not using Realm

nhachicha:
Fixes https://github.com/realm/realm-kotlin/issues/544


cmelchior (approved):
Missing a changelog entry. Otherwise looks fine to me 🚀

rorbech (changes requested):
I foresee that we end up reintroducing this issue again at some point if we don't address it generically. Maybe we can just add a test case that safeguards this, but wouldn't it somehow be possible to avoid triggering any internals of the sub compiler plugin at all. It seems a bit of waste that we actually have the registrar and register extensions, etc. if there is no need. I have absolutely no idea on how/if this is possible, but would be the preferred solution ... and as said alternatively a test to ensure that we don't introduce it again.

nhachicha:
@rorbech I updated the minimal Android project to include a JVM target not using the Realm dependencies as a way to catch similar issues.

rorbech (approved):

Issue #118, opened 2021-01-16
Design ChangeListener and Finegrained Notifications API


cmelchior:
First draft of this API was added here #107, but this doesn't add support for also providing detailed changes.

There are two orthogonal discussions:

1) How to register and deregister any changelistener?

```
// Option A: Return "token" like 
// Same pattern as RxJava
// Less API methods on top-level classes
class Realm {
  fun addChangelistener(listener: RealmChangeListener): Cancellable
  fun cancelAllChangeListeners() // Or maybe introduce something like CompositeCancellable like RxJava
}

val token = realm.addChangelistener { } 
token.cancel(); // onDestroy
realm.addChangeListener { }
realm.cancelAllChangeListeners()

// Option B: Java like API
// 1 more method in the API
// Must define listener in a class variable to be able to cancel listener
// Mirrors current Java API
class Realm {
  fun addChangelistener(listener: RealmChangeListener)
  fun removeChangelistener(listener: RealmChangeListener)
  fun removeAllChangelisteners()
}

// Option C: Something else?
// ...
```

2) How should fine-grained notification listeners look like?

```
class RealmResults {

    // Option A: Two arguments returned
    // Depends on autocomplete to work "out-of-the-box"
    // Top-level access to the collection
    fun addChangeListener1(listener: (listener: (collection: RealmResults<E>, change: OrderedRealmCollection<E>) -> Unit) { TODO() }

    // Option B: One argument returned
    // `it` is available 
    // collection access is one layer away.
    // Flows only allow one argument. So the API for change listeners and Flows would be consistent
    fun addChangeListener1(listener: (OrderedCollectionChange<E, RealmResults<E>>) -> Unit) { TODO() }
}

```

cmelchior:
![image](https://user-images.githubusercontent.com/406066/104852531-d3e57d00-58fb-11eb-9d51-5d765d088559.png)

It seems that using a named interface causes problems with how easy it is to autocomplete in Android Studio for Kotlin code at least. There you manually have to add the implementation.

Using an anonymous lambda seems to offer the easiest way to autocomplete for Kotlin, i.e.

```
    // Autocomplete support for this isn't great. Even though it is marked as a functional interface, 
    // autocomplete in Android Studio doesn't work.
    // Java support is superior though as the IDE can autocomplete the implementation
    fun interface Listener<E: BaseRealmModel<E>, T: RealmResults<E>> {
[...]

PR #594, opened 2021-11-24
Query API - draft implementation

edualonso:
This tentative API and implementation builds upon https://github.com/realm/realm-kotlin/pull/557 with concrete implementations that respect the baseline proposed there.

In short, the API looks like this:
```
// Blocking - ordinary query
val results: RealmResults<Sample> = realm.query(Sample::class)
    .find()

// Blocking - scalar query
val count: Long? = realm.query(Sample::class)
    .count()
    .find()

// Asynchronous - ordinary query
realm.query(Sample::class)
    .asFlow()
    .collect { results: RealmResults<Sample> -> /* use your results */ }

// Asynchronous - scalar query
realm.query(Sample::class)
    .count()
    .asFlow()
    .collect { count: Long? -> /* use your scalar value */ }
```

### API usage
The way to obtain `RealmResults` is through a `RealmQuery`. A query object can be obtained from a Realm instance.

Through `RealmQuery` we can obtain `RealmResults` via its `find` and `asFlow` functions.

### Query hierarchy
`RealmQuery` offers to create subqueries via the `query` function. Likewise, it gives the possibility of adding query descriptors via `sort`, `limit`, `distinct`, and provides aggregation operators on the query itself via `count`, `sum`, `min`, `max`, and `average`.

Aggregation operators represent a different type of query: `RealmScalarQuery`. This type of query returns scalar values not wrapped inside a `RealmResults` instance, via the `find` and `asFlow` functions.

It is important to observe the following information about scalar queries:
- `count` **always returns `Long`**.
- `average` **returns either a `Double`, a generic numeric type matching the specified property, or `null` in case there are no rows for the given property** - more on this and potential precision loss below.
- `min` and `max` **always return a numeric type matching the specified property** or `null` in case there are no rows for the given property.
- `sum` **always return a numeric type matching the specified property** or `0` in case there are no rows for the given property.

~~-### Average~~-
~~-When using the `Double` variant the result will respect the precision provided by Core. If the generic type variant is used there may be precision loss should the result be a fraction, e.g. for `Int` values `1` and `2` the result will be the value returned by `coreAverageValue.roundToInt()`. Other numerical types will behave similarly.~~- Average will not be available in the first iteration until we have aligned both query and results API in core - see https://github.com/realm/realm-core/issues/5077

### TBD:
- [x] ~~lazy evaluation of queries, sub-queries and results.~~ lazy evaluation of results. Queries must be parsed immediately.
- [x] query composition ~~- `query.query()` requires having a raw query builder similar to the one we have for Java or alternatively get Core to add this for us: https://github.com/realm/realm-core/issues/5067.~~ Pending a fix for this [bug](https://github.com/realm/realm-core/issues/5082) in `realm_query_append_query`.
- [ ] single queries for `first`.
- [ ] fine-grained notifications via `asFlow` - blocked by https://github.com/realm/realm-kotlin/issues/118.
- [ ] remove old objects/query API.
- [ ] documentation.
- [x] removed `average` aggregation functions - add again after https://github.com/realm/realm-core/issues/5077 is resolved.
- [ ] subqueries on `RealmResults`.


packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -67,6 +67,15 @@ interface MutableRealm : TypedRealm {
      */
     override fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T>
 
+    /**
+     * TODO : query
+     */
+    override fun <T : RealmObject> query(
+        clazz: KClass<T>,
+        query: String,
```
cmelchior, line 75: I think this needs a default value like `TRUEPREDICATE` otherwise you cannot run queries like `mutableRealm.query(Sample::class).find()`

packages/library-base/src/commonMain/kotlin/io/realm/Queryable.kt
```diff
@@ -25,6 +25,6 @@ package io.realm
 /**
  * Interface holding common query methods.
  */
-interface Queryable<T : RealmObject> {
+interface Queryable<T> {
     fun query(query: String = "TRUEPREDICATE", vararg args: Any?): RealmResults<T>
```
cmelchior, line 29: Not sure it makes sense that this have a default value. It would only allow patterns like `realm.query(Sample::class, "id == 42").query().find` which doesn't seem useful?

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -69,6 +70,15 @@ interface Realm : TypedRealm {
      */
     override fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T>
 
+    /**
+     * TODO : query
+     */
+    override fun <T : RealmObject> query(
+        clazz: KClass<T>,
+        query: String,
```
cmelchior, line 78: Same as MutableRealm. Probably need a default value.

packages/library-base/src/commonMain/kotlin/io/realm/RealmQuery.kt
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * TODO : query
+ */
+interface RealmQuery<E> : RealmElementQuery<E> {
+
+    fun query(filter: String, vararg arguments: Any?): RealmQuery<E>
+    fun sort(property: String, sortOrder: Sort = Sort.ASCENDING): RealmQuery<E>
+    fun sort(vararg propertyAndSortOrder: Pair<String, Sort>): RealmQuery<E>
+    fun distinct(property: String): RealmQuery<E>
```
cmelchior, line 66: If we have SORT and DISTINCT at this level, we should probably also have `limit`? 🤔 

So even though these can be supplied in the query string, it might make sense that we primarily treat `query` as predicates and leave handling the result set (sort, distinct, limit) to typed functions?

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
```diff
@@ -0,0 +1,168 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+/**
+ * TODO : query
+ */
+// FIXME API-QUERY Final query design is tracked in https://github.com/realm/realm-kotlin/issues/84
+//  - Lazy API makes it harded to debug
+//  - Postponing execution to actually accessing the elements also prevents query parser errors to
+//    be raised. Maybe we can get an option to prevalidate queries in the C-API?
```
cmelchior, line 36: I'm pretty sure that Core can validate queries while we build them up using the query parser. That is what we do in Java.

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
```diff
@@ -0,0 +1,168 @@
...
+
+    override val size: Int
+        get() = RealmInterop.realm_results_count(nativePointer).toInt()
+
+    @Suppress("SpreadOperator")
+    override fun query(query: String, vararg args: Any?): BaseResults<E> {
+        try {
+            return instantiateResults(
+                realm,
+                RealmInterop.realm_query_find_all(
```
cmelchior, line 66: I suspect we don't want to call find_all here, but just do query_parse.

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
```diff
@@ -0,0 +1,168 @@
...
+        nativePointer: NativePointer,
+        clazz: KClass<E>,
+        mediator: Mediator
+    ): BaseResults<E> = ElementResults(realm, nativePointer, clazz, mediator)
+}
+
+/**
+ * TODO : query
+ */
+internal class ScalarResults<E : Any> constructor(
```
cmelchior, line 153: This looks a bit odd? Why do we need a ScalarResults.  It either a Query object or a simple value?

packages/library-base/src/commonMain/kotlin/io/realm/internal/Observable.kt
```diff
@@ -33,11 +33,33 @@ import kotlinx.coroutines.flow.Flow
...
+ * TODO : query
+ */
+interface Freezable<T> {
+    fun freeze(frozenRealm: RealmReference): Observable<T>?
+}
+
+/**
+ * TODO : query
+ */
+interface Thawable<T> {
```
cmelchior, line 56: Do we really want two interfaces for this? Doesn't freeze and thaw always come in pairs?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,291 @@
...
+            // Queries inside a write transaction are live and can be reused within the closure
+            val query = query(Sample::class)
+
+            assertEquals(0, query.find().size)
+            copyToRealm(Sample().apply { stringField = stringValue })
+            assertEquals(1, query.find().size)
+        }
+
+        // No filter
+        realm.query(Sample::class)
```
cmelchior, line 75: Any reason to assert the same as the first one in this test?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,291 @@
...
+            .find()
+            .let { results -> assertEquals(1, results.size) }
+
+        // Filter by string
+        realm.query(Sample::class, "stringField = $0", stringValue)
+            .find()
+            .let { results -> assertEquals(1, results.size) }
+
+        // Filter by invalid string
```
cmelchior, line 84: Maybe name to "filter by string that doesn't match"...the other sounds like something that should throw

cmelchior: Also missing tests for wrong syntax, e.g. missing fieldName and wrong type for argument

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
edualonso: An overriding function is not allowed to specify default values for its parameters. The parent function is specifying `TRUEPREDICATE` though so that should be enough, I think.

packages/library-base/src/commonMain/kotlin/io/realm/Queryable.kt
edualonso: Remember this is the "old" query API and it's only implemented by `RealmResults` - notice it returns `RealmResults` too. The new API has a default `TRUEPREDICATE` parameter as mentioned above.

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
edualonso: See comment above on `MutableRealm`.

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
edualonso: This is the old "objects/query" API and should be removed.

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
edualonso: I think we do. Before introducing this API we only supported `RealmResults<RealmObject>` and its `get` function would assume by default the value returned by `realm_results_get` was a `Link`. Now it can be either a `Link` or a primitive/scalar due to the addition of flows for aggregation queries. The only way we could make it with only one `RealmResults` implementation is if `RealmInterop.realm_results_get` were able to instantiate a `Link` from the interop and preserve the generic type on the library side, which unfortunate cannot be done due to the way things are structured across modules :/

packages/library-base/src/commonMain/kotlin/io/realm/internal/Observable.kt
edualonso: No, they don't come in pairs.

This is related to how we subscribe to notifications. We can only register for notifications using a results object. This means that when we call `query.asFlow()` the notifier will "thaw" the query, which makes no sense in and of itself, so instead, we create a live results object from the query itself which is then used to register for notifications. This means that a query needs only to be `Thawable` and not `Freezable`, since the output of `query.thaw()` is a `RealmResults` object and that object will be the one that will be "frozen", "thawed" and whatnot.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: It isn't quite the same assert. Here we should receive a result, thus the `assertEquals(1, results.size)` whereas in the first one we have no results since it's before the write transaction that populates the Realm.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmQueryImpl.kt
```diff
@@ -0,0 +1,255 @@
...
+        CountQuery(realmReference, queryPointer, mediator)
+
+    override fun thaw(liveRealm: RealmReference): BaseResults<E> {
+        val liveResults = RealmInterop.realm_results_resolve_in(resultsPointer, liveRealm.dbPointer)
+        return ElementResults(liveRealm, liveResults, clazz, mediator)
+    }
+
+    override fun asFlow(): Flow<RealmResults<E>> = realmReference.owner
+        .registerObserver(this)
+        .onStart { realmReference.checkClosed() }
```
cmelchior, line 169: Is this needed?

cmelchior (commented):
Just went over this quick, so I'm pretty sure I missed something... But I would suggest even splitting this PR up if possible. There are so many things going on that it is hard to keep track of it all.

Perhaps split it into:
- Public API classes + Very simple Element queries (query + find)
- Scalar queries (min, max, count)
- Scalar queries (average)
- Flowable variants

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -733,7 +789,87 @@ actual object RealmInterop {
...
+                    property,
+                    average.ptr,
+                    found
+                )
+            )
+            return from_realm_value(average)
+        }
+    }
+
+    actual fun <T> realm_results_average_with_found(
```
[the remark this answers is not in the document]
edualonso: Sure. I was doing some testing with both at some point and was planning on removing the other one.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmQueryImpl.kt
```diff
@@ -0,0 +1,255 @@
...
+    private val mediator: Mediator,
+    private val descriptors: List<QueryDescriptor> = listOf(),
+    private val subQuery: NativePointer? = null,
+    private val filter: String,
+    private vararg val args: Any?
+) : RealmQuery<E>, Thawable<BaseResults<E>> {
+
+    private val queryPointer: NativePointer = when {
+        subQuery != null -> subQuery
+        else -> parseQuery()
```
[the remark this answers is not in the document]
edualonso: It is being cached, this isn't a property with explicit getter. It is setting the value upon instantiation just once.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmQueryImpl.kt
edualonso: I found the other flows do this check. I think this implementation might not respect the original intent since the check will be called as the first thing after calling `collect`, whereas in the other places where we do it is called right when calling `asFlow` and not as the first thing upon collection.

edualonso:
No longer relevant - see https://github.com/realm/realm-kotlin/pull/603

PR #595, opened 2021-11-24
Add support for RealmInstant 

cmelchior:
I was trying to gauge the complexity of the new data types and ended up adding one in the process. Turns out it wasn't hard.

It does raise a lot of architectural questions though:
* Creating and readingan internal timestamp currently requires 3 JNI calls + 1 to release. This seems very excessive and can hopefully be optimized (in a future PR).
* Adding a more "advanced" type that doesn't exists in `CInterop` revealed a lot of architectural flaws that is also relevant for adding support for Type Converters. Basically all places we move known API types to CInterop, we need to convert them to CInterop types because CInterop doesn't have access to the public API's. There are various solutions to this we can pursue .  
* Lots of duplicated code in RealmObjectHelper now...not sure if we should refactor it?

Also a few design ones:
*  Should the name be `RealmInstant`, `RealmDate` or something else entirely? I chose `RealmInstant` as it seems Kotlinx are using that as well even though it is very tied to the JVM ecosystem.
* Core and Java/Kotlin Instant represent time before epoch in different ways. What should we do here? Completely hide that Core does it differently or expose the difference in the public API?  

**TODO**
* [x] Dokka
* [x] Not sure if I missed adding support anywhere?  
* [x] Unit test boundaries for unmanaged and managed values
* [x] Looks like boundaries and date representation might differ between Java and Realm Core. Figure out exactly what.
* [x] Determine if we need a RealmInstant "fake" constructor or not.


cmelchior:
Turns out Core has a different representation of Timestamp than Java Instant. Their boundaries are different and the time before epoch is represented differently 🙄 . So we will have to make some decisions about which semantics we want to expose to end-users:

```
    @Test
    fun timestamp_boundaries() {
        realm.writeBlocking {
            val sample = copyToRealm(Sample())
            var max = RealmInstant(Long.MAX_VALUE, 999_999_999)
            sample.timestampField = max
            max = sample.timestampField

            assertEquals(Long.MAX_VALUE, max.epochSeconds)
            assertEquals(999_999_999 , max.nanosecondsOfSecond)

            var min = RealmInstant(Long.MIN_VALUE, 0)
            sample.timestampField = min
            min = sample.timestampField
            assertEquals(Long.MIN_VALUE, min.epochSeconds)
            assertEquals(0 , min.nanosecondsOfSecond)

            // Nanoseconds are always positive, which makes negative time a bit akward.
            // Core uses a more sensible (0, -1) in this case.
            var zeroMinusOne = RealmInstant(0, -1)
            sample.timestampField = zeroMinusOne
            zeroMinusOne = sample.timestampField
            assertEquals(0, zeroMinusOne.epochSeconds)
            assertEquals(-1, zeroMinusOne.nanosecondsOfSecond)

            var zeroPlusOne = RealmInstant(0, 1)
            sample.timestampField = zeroPlusOne
            zeroPlusOne = sample.timestampField
            assertEquals(0, zeroPlusOne.epochSeconds)
            assertEquals(1, zeroPlusOne.nanosecondsOfSecond)
        }
    }

    @Test
    fun instant_boundaries() {
        // Some reasoning for Java Instant boundaries http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/time/Instant.java#l228

        val max = Instant.fromEpochSeconds(Long.MAX_VALUE, 999_999_999)
        assertEquals(31_556_889_864_403_199, max.epochSeconds)
        assertNotEquals(Long.MAX_VALUE, max.epochSeconds)
        assertEquals(999_999_999 , max.nanosecondsOfSecond)

        val min = Instant.fromEpochSeconds(Long.MIN_VALUE, 0)
        assertEquals(-31_557_014_167_219_200, min.epochSeconds)
        assertNotEquals(Long.MIN_VALUE, min.epochSeconds)
        assertEquals(0 , min.nanosecondsOfSecond)

        // Nanoseconds are always positive, which makes negative time a bit akward.
        // Core uses a more sensible (0, -1) in this case.
        val zeroMinusOne = Instant.fromEpochSeconds(0, -1)
        assertEquals(-1, zeroMinusOne.epochSeconds)
        assertEquals(999_999_999, zeroMinusOne.nanosecondsOfSecond)

        val zeroPlusOne = Instant.fromEpochSeconds(0, 1)
        assertEquals(0, zeroPlusOne.epochSeconds)
        assertEquals(1, zeroPlusOne.nanosecondsOfSecond)
    }
```

packages/library-base/src/commonMain/kotlin/io/realm/RealmInstant.kt
```diff
@@ -0,0 +1,37 @@
+package io.realm
+
+import io.realm.internal.interop.Timestamp
+
+/**
+ * TODO: Should type be Interface, Normal class or Data Class?
+ * TODO: Should name be RealmInstant, RealmDate or something else?
+ */
+class RealmInstant internal constructor(internal val data: Timestamp) {
```
cmelchior, line 9: Consider this making this an interface

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmInstantImpl.kt
```diff
@@ -0,0 +1,44 @@
+package io.realm.internal
+
+import io.realm.RealmInstant
+import io.realm.internal.interop.Timestamp
+
+class RealmInstantImpl(seconds: Long, nanoSeconds: Int) : Timestamp(seconds, nanoSeconds), RealmInstant {
```
rorbech, line 6: You could still just use data classes for implementation if you want to piggyback on default `equals` and `hashCode`. But might as well just do the full implementation ... Guess we still want to print the interface name in `toString` anyway, so 🤷

packages/library-base/src/commonMain/kotlin/io/realm/RealmInstant.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * The number of nanoseconds by which this instant is different than [epochSeconds].
+     *
+     * The value always lies in the range `-999_999_999..999_999_999`.
+     */
+    val nanosecondsOfSecond: Int
+}
+
+// TODO Do we want this type of constructor? How does it show up in Dokka?
```
rorbech, line 95: I am not sure that it add that much value in this case, since there is already an immediate public factory method. But it is rather convenient, as it is the first place you would look. But yeah, depends on whether it the requires an import, how it shows in Dokka, etc.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/Timestamp.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.internal.interop
+
+/**
+ * Wrapper around Core Timestamp values.
+ * See https://github.com/realm/realm-core/blob/master/src/realm/timestamp.hpp for more information
+ */
+open class Timestamp(open val seconds: Long, open val nanoSeconds: Int)
```
rorbech, line 22: I guess we can't just have this as an interface as I said since we also need to instantiate these in `cinterop`. 
But with an `cinterop`-interface you could let the `library` implementation implement that and skip one of the translations. But probably cleaner to keep them completely separate and follow the typemapping thoughts.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -133,6 +145,32 @@ object RealmObjectHelper {
         }
     }
 
+    // Consider inlining
+    @Suppress("unused") // Called from generated code
```
rorbech, line 149: Already refactoring as part of #596, so doesn't make sense to try to factor to avoid duplication in this PR.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmInstantTests.kt
```diff
@@ -0,0 +1,138 @@
...
+    private fun roundTrip(timestamp: RealmInstant, function: (RealmInstant) -> Unit) {
+
+        // Test unmanaged objects
+        function(timestamp)
+
+        // Test managed objects
+        realm.writeBlocking {
+            val sample = copyToRealm(Sample())
+            sample.timestampField = timestamp
+            function(sample.timestampField)
```
rorbech, line 102: I would prefer that we asserted on a value retrieved from the realm. If we mess up something in the accessors the value could technically survive in the backing field. Pulling the sample through a query would ensure that it has indeed been round tripping the realm. I guess the `SampleTest` ensures that the accessors are actually directing to the realm, but still more comforting if we also did it here.

rorbech (commented):
LGTM, but don't know if there is still some progress, so leaving it un-approved. There are also some minor details about actually ensuring realm round trip in some of the test.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/Timestamp.kt
cmelchior: Yeah, this was turning into a bit of a mess.....It felt a bit much having both `RealmInstant`, `RealmInstantImpl`, `Timestamp` and `TimestampImpl`....Having `RealmInstant` being a class with an inner Timestamp was the smallest I could make the type system, but that went a bit against our decision to keep the API as interfaces + made problems when moving the type to Cinterop. 

For now I would probably prefer to keep it like this, and then we can iterate once we get more types and/or start looking into type converters.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmInstantTests.kt
cmelchior: Sure, I can rewrite the test.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmInstantImpl.kt
cmelchior: data classes doesn't work here when we extend Timestamp...but maybe a reason to introduce that Timestamp as an interface after all 🤔

cmelchior: After thinking a bit more...i ended up converting Timestamp to an interface, this allows both `RealmInstantImpl` and `TimestampImpl` to be data classes.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/Timestamp.kt
cmelchior: After thinking a bit more...i ended up converting Timestamp to an interface, this allows both RealmInstantImpl and TimestampImpl to be data classes.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -44,6 +46,16 @@ object RealmObjectHelper {
...
     }
 
+    @Suppress("unused") // Called from generated code
+    internal fun <R> getTimestamp(obj: RealmObjectInternal, col: String): RealmInstant? {
+        obj.checkValid()
+        val realm = obj.`$realm$Owner` ?: throw IllegalStateException("Invalid/deleted object")
+        val o = obj.`$realm$ObjectPointer` ?: throw IllegalStateException("Invalid/deleted object")
+        val key = RealmInterop.realm_get_col_key(realm.dbPointer, obj.`$realm$TableName`!!, col)
+        val res = RealmInterop.realm_get_value<Timestamp>(o, key)
+        return if (res == null) null else RealmInstantImpl(res)
```
rorbech, line 56: We could consider utilizing typemaps here to be able to generalize the path. But for speed I guess it would be better specializing the path. 

When refactored for the cached key based look up I think it would be relatively clean to just have a specialized variant for each type and by inlining it we skip the tedious modifications of the compiler plugin. Unless there would be a benefit in keeping the logic in library implementation instead of as compile time generated parts of user models 🤔

test/base/src/commonMain/kotlin/io/realm/test/util/Utils.kt
```diff
@@ -49,3 +49,12 @@ suspend fun <T : RealmObject> T.update(block: T.() -> Unit): T {
         liveObject
     }
 }
+
+// Expose a try-with-resource pattern for Realms
+fun Realm.use(action: (Realm) -> Unit) {
```
rorbech, line 54: Should we consider providing this for the user? I guess users don't need it that often that it is worth adding it.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmInstantTests.kt
```diff
@@ -0,0 +1,143 @@
...
+            val managedTimestamp = objects(Sample::class).first().timestampField
+            function(managedTimestamp)
+            cancelWrite() // So we can use .first()
+        }
+    }
+
+    @Test
+    fun equals() {
+        assertTrue(RealmInstant.fromEpochSeconds(42, 42) == (RealmInstant.fromEpochSeconds(42, 42)))
+        assertFalse(RealmInstant.fromEpochSeconds(0, 0) == (RealmInstant.fromEpochSeconds(42, 42)))
```
rorbech, line 114: Maybe also test the two remaining cases when equal on one axis. Few lines of code but could lead to a major mess. Also for `hashCode`.

rorbech (approved):
Nice. 

I guess it is a bit weird with the asymmetry on the types when going in an out, but I think the simplicity in the implementation pays off.

clementetb (approved):
LGTM, good work!

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
cmelchior: Yes, I think there is room for improvements here, but we should probably do this as part of cleaning this up in general. So would probably just prefer to keep this here for now.

test/base/src/commonMain/kotlin/io/realm/test/util/Utils.kt
cmelchior: I would probably keep this out of the public API for now since we are pushing the idea that you don't really need to close a Realm again, so this method mostly seems useful for tests IMO.

PR #596, opened 2021-11-25
Optimize accessors using cached property keys

### LINKED ISSUE #600: Optimize accessors using cached property keys
We should use cache keys to lookup classes and properties.

rorbech:
This fetches schema information on updates and uses cached property keys (`SchemaMetadata` and `ClassMetadata`) to access properties. The old string based API is preserved for future reuse in a public string based API (part of #86). 

There are still room for optimizing this, but would be beneficial to have better benchmarking infrastructure to measure experiments before lazy loading schema and before heading into modifying the compiler plugin to generate metadata classes for each class with fixed members for the cached keys as it is done in realm-java.

Propagating the schema information through the `RealmReference`s revealed some fundamental flaws in the bookkeeping of versions, thus exploded a bit.

Closes #600


packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -43,22 +43,30 @@ internal class SuspendableNotifier(
         }
     }
 
-    // FIXME Work-around for the global Realm changed listener not working.
     // Adding extra buffer capacity as we are otherwise never able to emit anything
     // see https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt#L78
-    private val _realmChanged = MutableSharedFlow<RealmReference>(
-        onBufferOverflow = BufferOverflow.SUSPEND,
+    private val _realmChanged = MutableSharedFlow<FrozenRealmReference>(
+        onBufferOverflow = BufferOverflow.DROP_OLDEST,
```
cmelchior, line 49: I haven't traced the logic fully, but is it safe if events are dropped? I'm thinking both about memory leaks objects/Realm getting out of sync?

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableWriter.kt
```diff
@@ -53,7 +56,27 @@ internal class SuspendableWriter(private val owner: RealmImpl, val dispatcher: C
         tid = runBlocking(dispatcher) { threadId() }
     }
 
-    suspend fun <R> write(block: MutableRealm.() -> R): Pair<RealmReference, R> {
+    // Currently just for internal-only usage in test, thus API is not polished
+    suspend fun updateSchema(schema: RealmSchemaImpl): FrozenRealmReference {
+        return withContext(dispatcher) {
+            transactionMutex.withLock {
+                realm.log.debug("Updating schema: $schema")
```
cmelchior, line 63: Does the `toString()` of schema output anything useful? Maybe print path to Realm here?

packages/library-base/src/commonMain/kotlin/io/realm/internal/schema/CachedClassKeyMap.kt
```diff
@@ -0,0 +1,85 @@
...
+import io.realm.internal.interop.PropertyKey
+import io.realm.internal.interop.RealmInterop
+
+/**
+ * Schema metadata providing access to class metadata for the schema.
+ */
+interface SchemaMetadata {
+    operator fun get(className: String): ClassMetadata?
+    fun getOrThrow(className: String): ClassMetadata = get(className)
+        ?: throw IllegalArgumentException("Object of type '$className not found")
```
cmelchior, line 31:
```suggestion
        ?: throw IllegalArgumentException("Schema for object type '$className' could not be found.")
```

packages/library-base/src/commonMain/kotlin/io/realm/internal/schema/CachedClassKeyMap.kt
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * Class metadata providing access class and property keys.
+ */
+interface ClassMetadata {
+    val className: String
+    operator fun get(propertyName: String): PropertyKey?
+    fun getOrThrow(propertyName: String): PropertyKey = get(propertyName)
+        ?: throw IllegalArgumentException("Object of type '$className doesn't have a property named '$propertyName'")
```
cmelchior, line 41:
```suggestion
        ?: throw IllegalArgumentException("Schema for type '$className' doesn't have a property named '$propertyName'.")
```

packages/library-base/src/commonMain/kotlin/io/realm/internal/schema/CachedClassKeyMap.kt
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// FIXME We are actually not lazy loading these right now
```
cmelchior, line 44: Not 100% this is actually something we do want to fix...running the logic on the notifier thread will keep the Main thread faster at the close of slightly slower notifications when the schema actually does change.

packages/library-base/src/commonMain/kotlin/io/realm/internal/schema/CachedClassKeyMap.kt
```diff
@@ -0,0 +1,85 @@
...
+class CachedSchemaMetadata(private val dbPointer: NativePointer) : SchemaMetadata {
+    // TODO OPTIMIZE We should theoretically be able to lazy load these, but it requires locking
+    //  and 'by lazy' initializers can throw
+    //  kotlin.native.concurrent.InvalidMutabilityException: Frozen during lazy computation
+    val classMap: Map<String, CachedClassMetadata>
+
+    init {
+        classMap = RealmInterop.realm_get_class_keys(dbPointer).map<ClassKey, Pair<String, CachedClassMetadata>> {
+            val classInfo = RealmInterop.realm_get_class(dbPointer, it)
+            println("Looking up class info for $this ${classInfo.name}")
```
cmelchior, line 60: Remove? or use proper logs

packages/library-base/src/commonMain/kotlin/io/realm/internal/schema/CachedClassKeyMap.kt
```diff
@@ -0,0 +1,85 @@
...
+ */
+class CachedClassMetadata(dbPointer: NativePointer, override val className: String, val classKey: ClassKey) : ClassMetadata {
+    // TODO OPTIMIZE We should theoretically be able to lazy load these, but it requires locking
+    //  and 'by lazy' initializers can throw
+    //  kotlin.native.concurrent.InvalidMutabilityException: Frozen during lazy computation
+    val propertyMap: Map<String, PropertyKey>
+
+    init {
+        val classInfo = RealmInterop.realm_get_class(dbPointer, classKey)
+        println("Looking up property info for ${classInfo.name}")
```
cmelchior, line 80: Remove?

test/base/build.gradle.kts
```diff
@@ -60,7 +60,6 @@ kotlin {
         kotlinOptions.freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
         // @ExperimentalPathApi is only available for JVM, so cannot use the annotation in shared
```
cmelchior, line 61: Also remove this comment

cmelchior (commented):
I found the changes to reference tracking a bit hard to follow, so I probably missed some nuances. But generally looks fine. I mostly left minor comments.

packages/library-base/src/commonMain/kotlin/io/realm/internal/LiveRealm.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    override val realmReference: LiveRealmReference by lazy {
+        val dbPointer = RealmInterop.realm_open(configuration.nativeConfig, dispatcher)
+        LiveRealmReference(this, dbPointer)
+    }
+
+    private val _snapshot: AtomicRef<RealmReference?> = atomic(null)
+    internal val snapshot: RealmReference
+        get() {
+            if (_snapshot.value == null) {
```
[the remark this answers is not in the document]
rorbech: The `LiveRealm` is only accessed on a single thread, but it is still frozen in the old memory model, so we need the atomic reference to be able to update it. I will improve the docs.

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
rorbech: The versions are now tracked on the `LiveRealm`s issuing the actual `FrozenRealmReference`s, instead of relying on the various frozen versions to actually realm the user facing realm. So we are now actually NOT leaking them because we are now also closing them even though the were not emitted to the user facing realm. Also, dropping the older versions ensures that the user facing realm will get the latest version when it is advanced instead of throwing all newer version away with a log entry indication.

packages/library-base/src/commonMain/kotlin/io/realm/internal/VersionTracker.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    fun trackAndCloseExpiredReferences(realmReference: FrozenRealmReference? = null) {
+        val references = mutableSetOf<Pair<NativePointer, WeakReference<RealmReference>>>()
+        realmReference?.let {
+            references.add(Pair(realmReference.dbPointer, WeakReference(it)))
+        }
+        intermediateReferences.value.forEach { entry ->
+            val (pointer, ref) = entry
+            if (ref.get() == null) {
+                log.debug("Closing unreferenced version: ${RealmInterop.realm_get_version_id(pointer)}")
```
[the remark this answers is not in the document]
rorbech: We don't have access to anything sane info besides what we can retrieve from the raw native pointer. Only ending up here if the `RealmReference` is not referenced any longer. Holding on to it would keep it alive. But we could extend the map with some a link to the configuration if we really need it!?

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -36,6 +36,12 @@ value class PropertyKey(val key: Long)
 expect val INVALID_CLASS_KEY: ClassKey
 expect val INVALID_PROPERTY_KEY: PropertyKey
 
+// TODO Again it would be awesome with marker interfaces for the various realm types, so we could
+//  add it as generic parameters here ...
+// Registration token that represent realm and schema change callback registration
+@JvmInline
+value class RegistrationToken(val value: Long)
```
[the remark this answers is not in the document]
rorbech: The two _registration mechanisms_ for notifications and callbacks are completely differently implemented in core, so we already have a `NotificationToken` for notification registrations (as opposed to callback registrations 🤪). I have created an issue in core to align the concepts https://github.com/realm/realm-core/issues/5217.

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -433,7 +471,7 @@ actual object RealmInterop {
             return with(classInfo) {
                 ClassInfo(
                     name.safeKString("name"),
-                    primary_key?.toKString(),
+                    primary_key?.toKString() ?: "",
```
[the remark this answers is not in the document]
rorbech: The C-API also returns `""` when it is not set, but the native wrapper type is still nullable. If have updated it to use our `safeKString` that will throw an error if this happens.

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -143,13 +143,29 @@ actual object RealmInterop {
             realmc.open_realm_with_scheduler(
                 (config as LongPointerWrapper).ptr,
                 if (dispatcher != null) JVMScheduler(dispatcher) else null
-            )
+            ),
```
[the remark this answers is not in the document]
rorbech: Left over from some experiments and forgot to remove the trailing comma again.

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseRealmImpl.kt
```diff
@@ -63,15 +60,16 @@ abstract class BaseRealmImpl internal constructor(
     internal val log: RealmLog = RealmLog(configuration = configuration.log)
 
     init {
-        log.info("Realm opened: ${configuration.path}")
+        log.info("Realm opened: $this ${this.configuration} ${configuration.path}")
```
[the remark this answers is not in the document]
rorbech: No. I have stripped it back to normal after finished debugging leaking memory and realm references

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseRealmImpl.kt
```diff
@@ -114,9 +112,9 @@ abstract class BaseRealmImpl internal constructor(
 
     // Not all sub classes of `BaseRealm` can be closed by users.
     internal open fun close() {
-        val reference = realmReference
-        reference.checkClosed()
-        RealmInterop.realm_close(reference.dbPointer)
-        log.info("Realm closed: ${configuration.path}")
+        realmReference.close()
```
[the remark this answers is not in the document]
rorbech: Had some issue getting all references closed without being able to close some of them multiple times. After bookkeeping the references in the notifier/writer we are now back at closing reference once and enforcing this by calling `checkClose` in `RealmReference.close()`. Only minor tricky thing is that the user facing realm is now not responsible for closing its own `realmReference` as this is actually one of the references otherwise managed by the user facing realms version tracker, the writer or the notifier.

packages/library-base/src/commonMain/kotlin/io/realm/internal/LiveRealm.kt
```diff
@@ -0,0 +1,94 @@
...
+        schemaChangeRegistration = RealmInterop.realm_add_schema_changed_callback(realmReference.dbPointer, ::onSchemaChanged)
+    }
+
+    protected open fun onRealmChanged() {
+        log.debug("Realm changed: $this $configuration")
+        _snapshot.value = null
+    }
+
+    protected open fun onSchemaChanged(schema: NativePointer) {
+        if (log.logLevel >= LogLevel.DEBUG) {
```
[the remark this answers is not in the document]
rorbech: Removed after done debugging

packages/library-base/src/commonMain/kotlin/io/realm/internal/schema/CachedClassKeyMap.kt
rorbech: Rephrased into a `TODO OPTIMIZE`, so that we revisit this as part of optimizing when we have an appropriate benchmarking mechanism.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/Identifiers.kt
```diff
@@ -78,7 +80,8 @@ internal object FqNames {
     val REALM_REFERENCE = FqName("io.realm.internal.RealmReference")
     val REALM_MEDIATOR_INTERFACE = FqName("io.realm.internal.Mediator")
     val REALM_CLASS_IMPL = FqName("io.realm.internal.schema.RealmClassImpl")
-    val REALM_CONFIGURATION = FqName("io.realm.RealmConfiguration")
+    val CLASS_METADATA_CLASS = FqName("io.realm.internal.schema.ClassMetadata")
+    val REALM_LOCAL_CONFIGURATION = FqName("io.realm.LocalConfiguration")
```
nhachicha, line 84: Unused, same for `SYNC_CONFIGURATION`, `REALM_CONFIGURATION_BUILDER`, `SYNC_CONFIGURATION_BUILDER`  and other FqNames ...

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmSchemaTests.kt
```diff
@@ -53,6 +57,7 @@ public class RealmSchemaTests {
         tmpDir = PlatformUtils.createTempDir()
         val configuration =
             RealmConfiguration.Builder(schema = setOf(SchemaVariations::class, Sample::class))
+                .log(LogLevel.DEBUG)
```
nhachicha, line 60: left over?

packages/library-base/src/commonMain/kotlin/io/realm/internal/LiveRealm.kt
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2021 Realm Inc.
```
cmelchior, line 2:
```suggestion
 * Copyright 2022 Realm Inc.
```

packages/library-base/src/commonMain/kotlin/io/realm/internal/LiveRealm.kt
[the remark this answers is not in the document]
cmelchior: Ah okay, I thought it could be called from other threads. The comment clarifies it 👍

packages/library-base/src/commonMain/kotlin/io/realm/internal/VersionTracker.kt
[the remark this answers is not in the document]
cmelchior: Good point. I'm not sure it is worth tracking extra things just for debugging, so let's just see if this is good enough.

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
```diff
@@ -99,6 +99,7 @@ class TestApp(
         RealmInterop.realm_clear_cached_apps()
 
         // Delete metadata Realm files
+        // TODO This actually wipes metadata for all realms and not only the specific instance data
```
cmelchior, line 102: This is intentional, so not sure there is a TODO here?

cmelchior (approved):

packages/jni-swig-stub/realm.i
```diff
@@ -87,6 +87,30 @@ std::string rlm_stdstr(realm_string_t val)
         get_env(true)->DeleteGlobalRef(static_cast<jobject>(userdata));
     };
 }
+// Reuse void callback typemap as template for `realm_on_realm_change_func_t`
+%apply (realm_app_void_completion_func_t, void* userdata, realm_free_userdata_func_t) {
+(realm_on_realm_change_func_t, void* userdata, realm_free_userdata_func_t)
+};
+%typemap(in) (realm_on_realm_change_func_t, void* userdata, realm_free_userdata_func_t) {
+    auto jenv = get_env(true);
```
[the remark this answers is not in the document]
rorbech: It is used to create the global reference below

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -23,6 +23,46 @@
...
+
+void
+schema_changed_callback(void* userdata, const realm_schema_t* new_schema) {
+    auto env = get_env(true);
+    static JavaClass java_callback_class(env, "kotlin/jvm/functions/Function1");
+    static JavaMethod java_callback_method(env, java_callback_class, "invoke",
+                                           "(Ljava/lang/Object;)Ljava/lang/Object;");
+    jobject schema_pointer_wrapper = wrap_pointer(env,reinterpret_cast<jlong>(new_schema));
+    env->CallObjectMethod(static_cast<jobject>(userdata), java_callback_method, schema_pointer_wrapper);
+    jni_check_exception(env);
```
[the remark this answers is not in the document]
rorbech: I think it makes most sense to check after the call and I would that that should be the convention ... but now I also added the check before the `CallObjectMethod` along with a reference to #665 that tracks the task of stream lining it.

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/TestApp.kt
rorbech: I removed the TODO as it didn't have anything to do with this PR ... but to me it seems like we could potentially wipe files for another app if a test initialized two. Maybe @edualonso can elaborate on if it does the correct thing.

PR #599, opened 2021-11-30
Update IDEA formatting to correspond with Ktlint default rules

nhachicha:
- Adding `.idea` folder to VCS
- `.idea` has a generated code style from `ktlint` using the official Kotlin `KOTLIN_OFFICIAL` code style
- Now formatting with IDE should correspond to Ktlint rules (examples with compiler module, reformatted manually using the IDE)


edualonso (approved):
LGTM. We should keep an eye on how these settings are preserved after IDE updates though.

rorbech (approved):
Works fine on my side. 

Should we consider to do a separate PR that actually reformats all files with the `KOTLIN_OFFICIAL` just to get to the same page through the project? I would think we would get more benefit from this, especially when touching things in future PRs, than we would loose by messing up the few places where we have specific manually formatting that goes against it (if any). But maybe up for a team vote.

cmelchior (approved):

clementetb (approved):
works at my end!

PR #601, opened 2021-12-01
Fix erroneous SyncConfiguration schema mode

### LINKED ISSUE #604: Erroneous migration exceptions when opening synced realms
Opening a synced realm could trigger erroneous migration exceptions. 

The wrong schema mode could cause migration errors when calling `realm_open` for the `SuspendableWriter` or `SuspendableNotifier` if a previous call to `realm_open` already opened the realm and successfully downloaded the backend schema before the second call to `realm_open`.

Since this is a race condition it only surfaced when opening a synced realm that was already present on the backend for the first time.

rorbech:
Closes #604


packages/library-base/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -111,7 +105,7 @@ interface RealmConfiguration {
         // Should always follow Builder constructor arguments
         fun with(
             schema: Set<KClass<out RealmObject>>
-        ): RealmConfiguration {
+        ): LocalConfiguration {
```
cmelchior, line 108: There is something slightly annoying about returning a `LocalConfiguration` from a `RealmConfiguration.with(...)`, but I guess that is unavoidable. But shouldn't `build()` also return a `LocalConfiguration` then?

test/sync/src/androidTest/kotlin/io/realm/test/shared/SyncedRealmTests.kt
```diff
@@ -145,6 +147,47 @@ class SyncedRealmTests {
...
+        // There is no guarantee that this wouldn't succeed if all internal realms (user facing,
+        // writer and notifier) are opened before the schema is synced from the server, but
+        // empirically it has shown not to be the case and cause trouble if opening the second or
+        // third realm with the wrong sync-intended schema mode.
+        val dir3 = PlatformUtils.createTempDir()
+        val config3 = createSyncConfig(path = "$dir3/$DEFAULT_NAME", user = user)
+        val realm3 = Realm.open(config3)
+        assertNotNull(realm3)
+
+        PlatformUtils.deleteTempDir(dir1)
```
cmelchior, line 186: It looks like you are forgetting to close Realms again, I suspect that will leak memory even if we delete the files?

cmelchior (changes requested):

packages/library-base/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
rorbech: Yes. `build()` should also return a `LocalConfiguration` 👍 

Reason why I kept the builder in `RealmConfiguration` was just to keep it backward compatible. Further, there will be no scenarios (🙈) where we would like to have a builder on both `RealmConfiguration` and `LocalConfiguration` so I though reusing the `RealmConfiguration` namespace was OK. But maybe we would rather want to make a braking change now and do the clean split now?

cmelchior: Yes, I was entertaining the idea...it would also make it really clear about the feature set....but there might be a discoverability issue since it would still be `Realm.open(config: RealmConfiguration)` so it would require som digging to find the correct constructors 🤔

cmelchior: Maybe it is fine to keep it as it it now, and then we can think about a breaking change later?

rorbech: I also had a good feeling about the explicit separation of `Local` vs. `Sync`. Since we already have the indirection through the various builders and people cannot just type ahead and construct them anyway, it is maybe ok to just point to the various options of constructing the configurations in the `Realm.open` documentation 🤔

cmelchior: Yes, just putting it in the documentation might be enough.

rorbech: I think it would be a reasonable change, but I have left it to stay _non-breaking_ until the idea has settled and maybe piled up with other breaking changes. Added a TODO with link to these comments.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 ## 0.8.0 (YYYY-MM-DD)
 
 ### Breaking Changes
-* None.
+* Reworked configuration hierarchy:
+  * Separated common parts of `RealmConfiguraion` and `SyncConfiguration` into `io.realm.Configuration` to avoid poluting the base configuration with local-only options.
```
nhachicha, line 5:
```suggestion
  * Separated common parts of `RealmConfiguraion` and `SyncConfiguration` into `io.realm.Configuration` to avoid polluting the base configuration with local-only options.
```

cmelchior (approved):
:shipit:

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * 64 byte key used to encrypt and decrypt the Realm file.
+     *
+     * @return null on unencrypted Realms.
+     */
+    val encryptionKey: ByteArray?
+
+    /**
+     * This class contains shared properties across the two types of configuration builders.
```
nhachicha, line 86: This level of details doesn't belong to a Javadoc level IMO, we can keep it as a classic comment though

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
```diff
@@ -0,0 +1,264 @@
...
+         * Sets the absolute path of the realm file.
+         *
+         * If not set the realm will be stored at the default app storage location for the platform.
+         */
+        fun path(path: String?): S = apply { this.path = path } as S
+
+        /**
+         * Sets the filename of the realm file.
+         *
+         * If setting the full path of the realm this name is not taken into account.
```
nhachicha, line 132:
```suggestion
         * If setting the full path of the realm, this name is not taken into account.
```

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
```diff
@@ -0,0 +1,264 @@
...
+         *
+         * Realm is capable of concurrently handling many different versions of Realm objects, this can e.g. happen if
+         * a flow is slow to process data from the database while a fast writer is putting data into the Realm.
+         *
+         * Under normal circumstances this is not a problem, but if the number of active versions grow too large, it
+         * will have a negative effect on the file size on disk. Setting this parameters can therefore be used to
+         * prevent uses of Realm that can result in very large file sizes.
+         *
+         * @param number the maximum number of active versions before an exception is thrown.
+         * @see [FAQ](https://realm.io/docs/java/latest/.faq-large-realm-file-size)
```
nhachicha, line 167: The link is `403` ing

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
```diff
@@ -0,0 +1,264 @@
...
+         *
+         * It is important that this key is created and stored securely. See [this link](https://docs.mongodb.com/realm/sdk/android/advanced-guides/encryption/) for suggestions on how to do that.
+         *
+         * @param encryptionKey 64-byte encryption key.
+         */
+        fun encryptionKey(encryptionKey: ByteArray) =
+            apply { this.encryptionKey = validateEncryptionKey(encryptionKey) } as S
+
+        /**
+         * TODO Evaluate if this should be part of the public API. For now keep it internal.
```
nhachicha, line 247: TODO should be out of the Javadoc

packages/library-base/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
```diff
@@ -333,9 +76,23 @@ interface RealmConfiguration {
...
+                encryptionKey,
                 deleteRealmIfMigrationNeeded,
-                encryptionKey
             )
         }
     }
+
+    companion object {
+        /**
+         * Create a configuration using default values except for schema, path and name.
```
nhachicha, line 87:
```suggestion
         * Creates a configuration using default values except for schema, path and name.
```

nhachicha (approved):
minor comments 👍

packages/library-base/src/commonMain/kotlin/io/realm/RealmConfiguration.kt
rorbech: To avoid confusion with a new name, I have kept the `RealmConfiguration` for the local configuration but introduced an underlying `Configuration` base interface.

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
rorbech: Yeah. Seems like the docs don't have subsection anchors at their new location. Tried to phrase sub section details with words instead together with a link to the top level FAQ section.

PR #603, opened 2021-12-02
Query API - 1 - public API and basic functionality

edualonso:
Added `RealmQuery` interface plus documentation. See https://github.com/realm/realm-kotlin/pull/557 and https://github.com/realm/realm-kotlin/pull/594 for context and design decisions.

Also updated Core to https://github.com/realm/realm-core/commit/039f5b10049919f5bef8931ee8af54bee6408bd7 which includes a fix for appending queries..

Necessary changes to existing components:
* `RealmResults` are not upper-bound by `RealmObject` anymore - see below.
* `RealmResultsImpl` has been split into `ElementResults<E : RealmObject>` and `ScalarResults<E : Any>` for scalar/Realm primitive types. This is needed in order to support reactive versions of aggregation operators on queries that return scalar values.
* The `Observable` interface has been split into the `Observable`, `Flowable`, `Thawable` and `Freezable` interfaces. This is needed to allow queries to produce frozen `RealmResults` when calling `asFlow()` in a more systematic way - queries need not `registerForNotification` nor `emitFrozenUpdate` nor by extension `freeze` since these are carried out by the results themselves.

The `RealmQueryImpl` class implements the following functionality so far:
* `query` composition using logical `AND`.
* `find` results (blocking operation).
* query descriptors `sort`, `distinct` and `limit`.

`TypedRealm`s expose the `RealmQuery` API:
* queries produced by `Realm`s will return **frozen** `RealmResults`.
* queries produced by `MutableRealm`s will return **live** `RealmResults` - only accessible within write transactions.

`RealmQuery` Implementation details:
* Added interop implementations for all query API methods.
* Query strings are parsed via `realm_query_parse` as soon as they are created.
* Descriptors return a new compound query using a logical `AND` via `realm_query_append_query`. Doing it this way we avoid having to keep track of the query strings when composing and parsing the query itself.


packages/library-base/src/commonMain/kotlin/io/realm/RealmQuery.kt
```diff
@@ -0,0 +1,361 @@
...
+
+    /**
+     * Finds the first object that fulfills the query conditions and returns it in a blocking
+     * fashion.
+     *
+     * Launching heavy queries from the UI thread may result in a drop of frames or even ANRs. **We
+     * do not recommend doing so.** If you want to prevent these behaviors you can obtain the
+     * object asynchronously using [asFlow] instead.
+     *
+     * @return a [RealmObject] instance or `null` if no object matches the condition.
```
cmelchior, line 253: There is some corner cases around this that might turn out slightly problematic, especially in the type system.

1. We are going to allow queries of e.g. `RealmList<Integer>` so other things than RealmObjects can be returned.
2. When we add fine-grained notifications, we will have a problem as e.g. `ObjectChange<E>` requires `E` to be a a subclass of RealmObject... I'm still not 100% sure how we can resolve this.

packages/library-base/src/commonMain/kotlin/io/realm/RealmQuery.kt
```diff
@@ -0,0 +1,361 @@
...
+inline fun <reified T : Any> RealmQuery<*>.sum(property: String): RealmScalarQuery<T> =
+    sum(property, T::class)
+
+/**
+ * Similar to [RealmQuery.find] but it receives a block in which the [RealmResults] from the query
+ * are provided.
+ */
+fun <T> RealmQuery<T>.find(block: (RealmResults<T>) -> Unit): RealmResults<T> = find().let {
+    block.invoke(it)
+    it
```
cmelchior, line 342: Shouldn't we return whatever people return from the closure instead? Or am I missing a use case here? Same for the others.

packages/library-base/src/commonMain/kotlin/io/realm/TypedRealm.kt
```diff
@@ -10,14 +10,31 @@ interface TypedRealm : BaseRealm {
...
+     * the results obtained from the query will not change on updates to the Realm. For a
+     * [MutableRealm] the query will produce live results and will in fact reflect updates to the
+     * [MutableRealm].
+     *
+     * @param query the Realm Query Language predicate to append.
+     * @param args Realm values for the predicate.
+     */
+    fun <T : RealmObject> query(
+        clazz: KClass<T>,
+        query: String = "TRUEPREDICATE",
```
cmelchior, line 35: How does this look in the IDE...is it visible in the type hints that this is the default when using a normal Realm? If not, I would suggest moving this default value to the `Realm` and `MutableRealm` classes, even though it means we are duplicating it.

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseRealmImpl.kt
```diff
@@ -66,25 +68,29 @@ abstract class BaseRealmImpl internal constructor(
...
-        return RealmResultsImpl.fromQuery(
-            realmReference,
-            RealmInterop.realm_query_parse(
-                realmReference.dbPointer,
-                clazz.simpleName!!,
-                "TRUEPREDICATE"
-            ),
-            clazz,
-            configuration.mediator
+        val findAllQuery = RealmInterop.realm_query_parse(
```
cmelchior, line 71: Instead of doing all of this, I would call into the publicAPI of `query`. That way we only have one internal codepath.

packages/library-base/src/commonMain/kotlin/io/realm/RealmQuery.kt
```diff
@@ -0,0 +1,361 @@
+/*
```
cmelchior, line 1: Missing CHANGELOG entry about this new API...also about the Core upgrade.

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -67,6 +67,26 @@ interface MutableRealm : TypedRealm {
...
+    /**
+     * Returns a [RealmQuery] matching the predicate represented by [query].
+     *
+     * The results yielded by the query are live and thus also reflect any update to the
+     * [MutableRealm].
+     *
+     * Said results are only valid on the calling thread.
+     *
+     * The resulting query is lazily evaluated and will not perform any calculations until
+     * [RealmQuery.find] is called.
```
cmelchior, line 79: Should we mention that calling `asFlow()` is illegal here? Or should we mention it on each `asFlow()` method in the   subclasses?

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
```diff
@@ -0,0 +1,152 @@
...
+import io.realm.internal.interop.Link
+import io.realm.internal.interop.NativePointer
+import io.realm.internal.interop.RealmCoreException
+import io.realm.internal.interop.RealmInterop
+import kotlinx.coroutines.channels.ChannelResult
+import kotlinx.coroutines.channels.SendChannel
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+internal abstract class BaseResults<E : Any> constructor(
```
cmelchior, line 31: Maybe add a comment exactly what this is a base class for?

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    override val size: Int
+        get() = RealmInterop.realm_results_count(nativePointer).toInt()
+
+    @Suppress("SpreadOperator")
+    override fun query(query: String, vararg args: Any?): BaseResults<E> {
+        try {
+            val table = clazz.simpleName!!
+            val queryPointer = RealmInterop.realm_query_parse(nativePointer, table, query, *args)
+            val resultsPointer = RealmInterop.realm_query_find_all(queryPointer)
```
cmelchior, line 60: Not 100% sure if this method is just for the old API 🤔. If yes, just ignore this comment.

If not, calling `find_all` here should be avoided as it evaluates the query directly on the calling thread. Results in ObjectStore has a constructor that takes the Query object directly.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,561 @@
...
+            // Queries inside a write transaction are live and can be reused within the closure
+            val query = query(Sample::class)
+
+            assertEquals(0, query.find().size)
+            copyToRealm(Sample().apply { stringField = stringValue })
+            assertEquals(1, query.find().size)
+        }
+
+        // No filter
+        realm.query(Sample::class)
```
cmelchior, line 71: Nice tests 👏 ...but we should have similar ones for MutableRealm

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,561 @@
...
+            .find { results -> assertEquals(1, results.size) }
+
+        // Filter by string that doesn't match
+        realm.query(Sample::class, "stringField = $0", "invalid string")
+            .find { results -> assertEquals(0, results.size) }
+    }
+
+    @Test
+    fun find_malformedQueryThrows() {
+        assertFailsWith<IllegalArgumentException> {
```
cmelchior, line 85: Good tests, but should also be on MutableRealm

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,561 @@
...
+
+    @Test
+    @Ignore
+    fun asFlow_throwsInsideWrite() {
+        // TODO
+    }
+
+    @Test
+    fun composedQuery() {
+        val value1 = 1
```
cmelchior, line 118: These doesn't really seem needed as you only use them once when creating the data?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,561 @@
...
+    }
+
+    @Test
+    @Ignore
+    fun asFlow_throwsInsideWrite() {
+        // TODO
+    }
+
+    @Test
+    fun composedQuery() {
```
cmelchior, line 117: Missing a more advanced test, like the `.query("name = 'Jonn'").query("age = 42 OR ag = 43")` mentioned elsewhere.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,561 @@
...
+        realm.query(Sample::class, "intField > 1")
+            .query("intField > 2")
+            .find()
+            .let { results ->
+                assertEquals(2, results.size)
+            }
+    }
+
+    @Test
+    fun composedQuery_withDescriptor() {
```
cmelchior, line 139: Missing cases where the descriptors are part of the query string.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,561 @@
...
+        realm.writeBlocking {
+            copyToRealm(Sample().apply { intField = value1 })
+            copyToRealm(Sample().apply { intField = value2 })
+            copyToRealm(Sample().apply { intField = value3 })
+            copyToRealm(Sample().apply { intField = value4 })
+        }
+
+        realm.query(Sample::class, "intField > 1")
+            .query("intField > 2")
+            .sort(Sample::intField.name, QuerySort.DESCENDING)
```
cmelchior, line 189: Not 100% sure if this test is interesting...the sort doesn't do anything here... Maybe if you used limit instead you could write a test where this actually had an impact.

cmelchior: Generally, if this test was for testing query -> descriptor -> query, we should cover all the descriptors in the test.

cmelchior (changes requested):
Really nice start 🥇 .... I had a few comments about the API, especially SingleQueries that might turn into a larger headache. If it turns out too difficult to find a solution, just commenting out the functionality and getting back to it in another PR also seems fine.

packages/library-base/src/commonMain/kotlin/io/realm/RealmQuery.kt
```diff
@@ -0,0 +1,361 @@
...
+import io.realm.internal.Flowable
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+/**
+ * A `RealmQuery` encapsulates a query on a [Realm] or a [RealmResults] instance using the `Builder`
+ * pattern. The query is executed using either [find] or subscribing to the [Flow] returned by
+ * [asFlow].
+ *
+ * The input to many of the query functions take a field name as [String]. Note that **this is not
```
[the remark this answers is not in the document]
edualonso: Fair enough, I'll remove it.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectQuery.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    override fun query(filter: String, vararg arguments: Any?): RealmQuery<E> {
+        val appendedQuery = tryCatchCoreException {
+            RealmInterop.realm_query_append_query(queryPointer, filter, *arguments)
+        }
+        return RealmObjectQuery(appendedQuery, this)
+    }
+
+    override fun sort(property: String, sortOrder: Sort): RealmQuery<E> =
+        query("TRUEPREDICATE SORT($property ${sortOrder.name})")
```
cmelchior, line 77: I wonder how that makes the final query look like?

E.g.
```
val desc: String = realm.query<Example>().sort("name", Sort.Descending).getDescription()
// Does this return
// TRUEPREDICATE AND TRUEPREDICATE SORT(name DESC)?
```

That would be a bit annoying, especially since the Core API has methods that allow us to not do that....But It can be fixed in a followup PR.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -0,0 +1,371 @@
...
+import io.realm.internal.Flowable
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+/**
+ * A `RealmQuery` encapsulates a query on a [Realm], a [RealmResults] or a [RealmList] instance
+ * using the `Builder` pattern. The query is executed using either [find] or subscribing to the
+ * [Flow] returned by [asFlow].
+ *
+ * If a [RealmObject] class is refactored, care has to be taken not to break any query filters.
```
cmelchior, line 35: I would just delete this sentance.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -0,0 +1,371 @@
...
+ * objects in the order they where inserted. Use the [sort] functions if a specific order is
+ * required.
+ *
+ * Results are obtained quickly most of the times when using [find]. However, launching heavy
+ * queries from the UI thread may result in a drop of frames or even ANRs. If you want to prevent
+ * these behaviors, you can use [asFlow] and collect the results asynchronously.
+ *
+ * @param E the class of the objects to be queried.
+ */
+interface RealmQuery<E> : RealmElementQuery<E> {
```
cmelchior, line 47: After our discission, Isn't this always going to be a RealmObject?

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -0,0 +1,371 @@
...
+     *
+     * @return a [RealmScalarQuery] that counts the number of objects for a given query.
+     */
+    fun count(): RealmScalarQuery<Long>
+}
+
+/**
+ * Query returning [RealmResults].
+ */
+interface RealmElementQuery<E> : Flowable<RealmResults<E>> {
```
cmelchior, line 209: Same, only RealmObjects?

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
cmelchior, line 47: Missing a `getDescription()` method (at least that is what it is called in Java) and returns the query as a String representation. Useful for debugging and late Flexible Sync. Can come in a later PR.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,779 @@
...
+            .sort(Sample::stringField.name, Sort.ASCENDING)
+            .find()
+            .let { results ->
+                assertEquals(3, results.size)
+                assertEquals(bob, results[0].stringField)
+                assertEquals(ruth, results[1].stringField)
+                assertEquals(stacy, results[2].stringField)
+            }
+
+        // Conflicting descriptors, query string vs function - Bob, Ruth, Stacy
```
cmelchior, line 330: Missing case where you are comining string and function descriptors where they both take effect.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,779 @@
...
+        // One would expect this to be Stacy 3 but notice the last 'query', it moves the descriptors
+        // to the end - this is how it is also implemented in Java
+        realm.query(Sample::class)
+            .distinct(Sample::stringField.name) // normal execution: Sylvia 1, Stacy 3, Ruth 4
+            .sort(
+                Sample::stringField.name,
+                Sort.ASCENDING
+            ) // normal execution: Ruth 4, Stacy 4, Sylvia 1
+            .limit(2) // normal execution: Ruth 4, Stacy 3
+            .query("intField < $value4") // this puts all the previous descriptors at the end of this query!
```
cmelchior, line 407: We could get around this by introducing yet another layer with a RealmQueryDescriptor interface, but probably a bit overkill? At least I have not heard any complaints about this behaviour in Java.

cmelchior (commented):
Nice... have some questions about the limitations of Primitive queries. It doesn't look like it is supported nor on the roadmap. So do we want to complicate the API to make it somewhat possible?

E.g. given that we want to add support for fine-grained object changes, I suspect having `T: RealmObject` will make that a lot easier.

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@
 
 ### Enhancements
 * [Sync] Added support for `User.logOut()` ([#245](https://github.com/realm/realm-kotlin/issues/245)).
-* Added supported for dates through a new property type: `RealmInstant`.
+* Added support for dates through a new property type: `RealmInstant`.
+* Added support for `RealmQuery` ([#84](https://github.com/realm/realm-kotlin/issues/84)).
```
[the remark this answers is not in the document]
edualonso: I will add the `Realm.query(...)` part but not the rest. We can't offer support for queries on lists and results using the new API until we have the missing C-API functions - see https://github.com/realm/realm-core/issues/5099.

packages/library-base/src/commonMain/kotlin/io/realm/Queryable.kt
```diff
@@ -25,6 +25,6 @@ package io.realm
 /**
  * Interface holding common query methods.
  */
-interface Queryable<T : RealmObject> {
+interface Queryable<T> {
```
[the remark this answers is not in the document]
edualonso: I think this interface will be rendered unnecessary once we remove the old query/objects API - see proposed solution that restricts queries on collections: https://github.com/realm/realm-kotlin/pull/603#discussion_r764015262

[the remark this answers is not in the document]
edualonso: This is also linked to your comment here: https://github.com/realm/realm-kotlin/pull/603#discussion_r766654362

packages/library-base/src/commonMain/kotlin/io/realm/internal/BaseResults.kt
```diff
@@ -0,0 +1,161 @@
...
+import io.realm.internal.interop.RealmCoreException
+import io.realm.internal.interop.RealmInterop
+import kotlinx.coroutines.channels.ChannelResult
+import kotlinx.coroutines.channels.SendChannel
+import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
+
+/**
+ * This class gathers all type-agnostic logic for handling [RealmResults]. This class is needed as
+ * Core can produce both objects and primitive results.
```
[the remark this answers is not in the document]
edualonso: It appears to be true for dictionaries. However, Claus and I talked to Jørgen about it and he sounded as though Core returning results is a "misuse of the results API". It may be necessary to spark a conversation about this with the other SDKs to see what alternatives there are to avoid exposing scalar results.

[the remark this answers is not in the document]
edualonso: Furthermore, I think leaving scalar results out is perfectly reasonable for the time being. We can perfectly manage using ordinary `RealmResults<RealmObject>` for now to compute query aggregators.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
edualonso: We need to get that function added to the C-API as well. I've created https://github.com/realm/realm-core/issues/5106.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: My first implementation used a descriptor that kept track of the different elements in the query string. Then I found out that in Java we build queries with descriptors using raw predicates, which in essence does what we are doing here. I think we should avoid changing the semantics to preserve consistency across the two SDKs...

... though I personally find it slightly annoying that descriptors are moved to the end in queries (they do not do that in results - that's one of the reasons why James suggested we used `realm_results_t` and not `realm_query_t` for queries on `RealmResults`) but I see and understand why it's done like that. Again, I would leave it like this for consistency.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
edualonso: See other related discussions.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
edualonso: See other related discussions

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectQuery.kt
edualonso: It does 😕  I guess that's another function to be added to the C-API 😓

cmelchior: Can we add a TODO or create an issue, then it is fine to resolve  this.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -0,0 +1,333 @@
...
+     * @return a flow representing changes to the [RealmResults] resulting from running this query.
+     */
+    override fun asFlow(): Flow<RealmResults<T>>
+}
+
+/**
+ * Query returning a single [RealmObject].
+ *
+ * TODO : The interaction with primitive queries might be a bit awkward
+ * TODO : answer to above from C-API:
```
rorbech, line 207: Are they intended to be left here. Maybe move them out of the Dokka block.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectQuery.kt
```diff
@@ -0,0 +1,149 @@
...
+    private val filter: String,
+    private vararg val args: Any?
+) : RealmQuery<E>, Thawable<RealmResultsImpl<E>> {
+
+    private val queryPointer: NativePointer = when {
+        composedQueryPointer != null -> composedQueryPointer
+        else -> parseQuery()
+    }
+
+    private val resultsPointer: NativePointer by lazy {
```
rorbech, line 50: This doesn't seem properly frozen. I get that it will work in a lot of scenarios, but constructing a query on one thread and passing it to another to be processed (ex. with`find()`) would cause mutability issues on native.  

So I guess there isn't a test for that :)

rorbech (commented):
Very nice. It is a bit hard to judge if the various query types (objects, scalar, etc.) and especially the mix of them is playable from user side without trying it out in more detail. But at least seem like a clean cut. 

There are some details about
-  Break up of `Observable` that calls for a new combining interface, and 
-  Some weird things about the `found` argument being used for some aggregators, but not all
- The bound on `RealmResult<>`. To me it sounds like it should be `RealmObject`. Maybe it is actually enforced correctly through the query entry point API, but it is hard to follow. I would have preferred that we separated an implementation of the collection-ish part for dictionaries into its own interface/part, but if the underlying C-API result is the actual denominator, then I guess this is maybe the way to go.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectQuery.kt
edualonso: There is: https://github.com/realm/realm-kotlin/pull/603/files#diff-4e7ed4ec30aad17f1a9fb5c70a9035859e8eecf7b61728001177f403d5f7a7edR762 - but it's commented out because of the PR split (`count` is not implemented yet, so I left it there for later). I will uncomment it and remove the references to `count`, but as far as I can see it runs just fine on native.

Please check whether the logic makes sense or not, but I made sure to create the query on a thread and call `find()` on a different one and there aren't any crashes due to mutability.

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -126,3 +146,13 @@ interface Realm : TypedRealm {
      */
     fun close()
 }
+
+/**
+ * Returns a [RealmQuery] matching the predicate represented by [query].
+ *
+ * Reified convenience wrapper for [Realm.query].
+ */
+inline fun <reified T : RealmObject> Realm.query(
```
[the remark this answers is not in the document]
rorbech: I wouldn't expect there to be many (if any) use cases for the non-reified versions. So we could rename that and/or make it internal. Then I think the IDE will automatically import the extension method. It is just really annoying to manually add the import.

packages/library-base/src/commonMain/kotlin/io/realm/internal/Observable.kt
```diff
@@ -21,23 +21,25 @@ import kotlinx.coroutines.channels.ChannelResult
...
-    fun observe(): Flow<T>
+    fun emitFrozenUpdate(
+        frozenRealm: RealmReference,
+        change: NativePointer,
+        channel: SendChannel<T>
+    ): ChannelResult<Unit>?
+}
+
+interface Freezable<T> {
+    fun freeze(frozenRealm: RealmReference): Observable<T>?
```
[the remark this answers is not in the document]
rorbech: Maybe mis-phrased a bit and maybe also not critical enough to change, but I will just try again to clarify. I get why you made the split, but if you split out the `Freezable` wouldn't it then be sufficient to return `T` instead of `Observable<T>`:
```
 fun freeze(frozenRealm: RealmReference): T?
```
And then we could catch the connection to `Observable<T>` only in the aforementioned grouping interface? 

Anyway, just thoughts. As said I am ready to go with your `Freezable<T>`

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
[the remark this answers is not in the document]
cmelchior: I'm a bit torn on this....IMO the reified version is easier on the eye when reading code, but at least I find it harder to type in the IDE because the autocomplete is less helpful.

Having two different names for the same thing would be super annoying, but only having extension functions would also make the API less discoverable :(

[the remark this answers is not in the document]
rorbech: If we don't consider the extension function discoverable if it is the only option, how would people ever find it if we keep it but the IDE always defaults to the non-reified version? 

IMO this ends up being worthless if people need to add the import themselves. I don't know if it too drastic to make the inner method internal, but while having two names is really annoying, it is better than worthless. 

But we can leave it as is and then monitor the frustrations.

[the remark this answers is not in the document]
cmelchior: I should probably play around a little more with this in the IDE to get a more informed opinion, but right now I would be leaning towards:

1) Use the same name for both methods (mostly since I don't have any ideas on how to name them).
2) Reference the reified version from Dokka.
3) Always use the reified version in samples.
4) Find and upvote or create an issue for the issue with finding and importing the reified version.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectQuery.kt
```diff
@@ -0,0 +1,145 @@
...
+        throw when (exception) {
+            is RealmCoreInvalidQueryStringException ->
+                IllegalArgumentException("Wrong query string: ${exception.message}")
+            is RealmCoreInvalidQueryException ->
+                IllegalArgumentException("Wrong query field provided or malformed syntax for query '$filter': ${exception.message}")
+            is RealmCoreIndexOutOfBoundsException ->
+                IllegalArgumentException("Have you specified all parameters for query '$filter'?: ${exception.message}")
+            else ->
+                genericRealmCoreExceptionHandler(
+                    "Invalid syntax for query '$filter': ${exception.message}",
```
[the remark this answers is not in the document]
edualonso: Fair enough, I will remove the query string.

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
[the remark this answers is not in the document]
edualonso: Just a not on the usage of reified functions, not necessarily linked to this one in particular: when doing exhaustive testing where we need to iterate through different properties it may not be possible to use the reified version. But I'm all onboard with using it in the samples.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectQuery.kt
rorbech: Seems like the `lazy` is properly _freeze aware_ and implemented on top of atomic references 🎉

CHANGELOG.md
```diff
@@ -5,11 +5,14 @@
   * Separated common parts of `RealmConfiguraion` and `SyncConfiguration` into `io.realm.Configuration` to avoid polluting the base configuration with local-only options.
   * Changed `Realm.open(RealmConfiguration)` to accept new base configuration with `Realm.open(Configuration)`.
   * Removed option to build `SyncConfiguration`s with `deleteRealmIfMigrationNeeded` option.
+* `RealmResults.observe()` and `RealmList.observe()` have been renamed to `asFlow()`.
+* Querying via `Realm.objects(...)` is no longer supported. Use `Realm.query(...)` instead.
```
cmelchior, line 9:
```suggestion
* Querying via `Realm.objects(...)` is deprecated. Use `Realm.query(...)` instead.
```

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -69,12 +72,33 @@ interface Realm : TypedRealm {
      */
     override fun <T : RealmObject> objects(clazz: KClass<T>): RealmResults<T>
 
+    /**
+     * Returns a [RealmQuery] matching the predicate represented by [query].
+     *
+     * A reified version of this method is also available as an extension function,
+     * `realm.query<YourClass>(...)`. Import `io.realm.query` to access it.
+     *
+     * The results yielded by the query reflect the state of the realm at invocation time, so the
```
cmelchior, line 81: I would swap this section with the next as this section talks about Results, which isn't available until you call `find` or `asFlow`

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmResultsImpl.kt
```diff
@@ -27,74 +27,53 @@ import kotlinx.coroutines.channels.SendChannel
...
-    private val mode: Mode
-    private val realm: RealmReference
-    private val clazz: KClass<T>
-    private val schema: Mediator
-    internal val result: NativePointer
-
-    private enum class Mode {
+/**
+ * This class gathers all type-agnostic logic for handling [RealmResults]. This class is needed as
+ * Core can produce both objects and primitive results.
```
cmelchior, line 32: This is no longer true.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -0,0 +1,327 @@
...
+     */
+    fun limit(limit: Int): RealmQuery<T>
+
+    /**
+     * Returns a query that finds the first object that fulfills the query conditions.
+     */
+    fun first(): RealmSingleQuery<T>
+
+    /**
+     * Finds the minimum value of a property.
```
cmelchior, line 116: We should mention the reified variant in all of these similar to what you did for `Realm.query`

cmelchior (approved):
I had a few minor comments about docs. Otherwise looks ready 🚀  (But wait with merging until we released 0.8.0)

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmResultsImpl.kt
edualonso: That's true, forgot to remove

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -0,0 +1,799 @@
...
+            assertEquals(0, query.find().size)
+            copyToRealm(Sample().apply { stringField = stringValue })
+            assertEquals(1, query.find().size)
+        }
+
+        // No filter
+        realm.query(Sample::class)
+            .find { results -> assertEquals(1, results.size) }
+
+        // Filter by string
```
clementetb, line 80: Aren't we exhaustive testing all supported types, os is it left for another PR? At least a roundtrip to validate we can pass data over the bridge.

clementetb (approved):
Looks good to me 💎. Great work!

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: I don't think it makes much sense to test that the predicates actually work. I suspect it has been tested at a Core level, maybe...? 🤞 😅

PR #607, opened 2021-12-09
Replace companion map with reflection/associated objects

### LINKED ISSUE #389: Not able to build configuration non-fluently
Collection of schema is not as robust as it should be. Currently collection of schema from various `RealmConfiguration.builder` calls only works if the configuration is built as a fully fluent construct.  

```
// Works
val conf1 = RealmConfiguration.Builder().schema(A::class, B::class, C::class).build()
val conf2 = RealmConfiguration.Builder(schema = setOf(A::class, B::class, C::class)) .build()

// Fails
val builder = RealmConfiguration.Builder(schema = setOf(A::class, B::class, C::class))
val conf3 = builder.build()
```

See https://github.com/realm/realm-kotlin/blob/a8223e9fd1ca6e2d63ca8da983d448c86ca4fb7e/packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmSchemaLoweringExtension.kt#L171 for the details.

rorbech:
This PR replaces our internally generated `Map<KClass<out RealmObject>, RealmObjectCompanion>` with reflective(JVM)/associated objects(Native) mechanisms.

This removes the constraint that we must provide the schema classes as direct class literals. This eliminates the whole effort of constructing the map at compile time. ... uses experimental APIs though 🤔 

Closes #389


cmelchior (commented):
Without going into a deeper review. What is the performance impact of this? I know reflection has improved quite a lot on Android, but historically it has been pretty bad....It would be interesting to see what the impact was on doing this with 10...20...50....100 model classes. I believe the average is 10-20'ish.

rorbech:
> Without going into a deeper review. What is the performance impact of this? I know reflection has improved quite a lot on Android, but historically it has been pretty bad....It would be interesting to see what the impact was on doing this with 10...20...50....100 model classes. I believe the average is 10-20'ish.

I think we should favor the simplicity of this solution over optimizations at the moment. After all this removes a huge constraint on how the configurations can be built. I have added a `TODO OPTIMIZE` to investigate performance later.

packages/library-base/src/darwin/kotlin/io/realm/internal/platform/ModelObject.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.realm.internal.RealmObjectCompanion
+import kotlin.reflect.AssociatedObjectKey
+import kotlin.reflect.ExperimentalAssociatedObjects
+import kotlin.reflect.KClass
+
+@OptIn(ExperimentalAssociatedObjects::class)
+@AssociatedObjectKey
+@Retention(AnnotationRetention.BINARY)
+@PublishedApi
+internal annotation class ModelObject(public val companionClass: KClass<out RealmObjectCompanion>)
```
cmelchior, line 28: This class could use some docs.

packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt
```diff
@@ -92,25 +93,26 @@ class GenerationExtensionTest {
     }
 
     @Test
+    // FIXME Do we actually want to fail at compile time if there is not schema?
```
cmelchior, line 96: Since `schema` is a required parameter, it seems rather hard to get into this state. I'm not sure I see a use case for creating a Realm with no schema, but on the other hand, I'm not sure it is worth it to check for the case.

IMO shouldn't we just re-enable this test and maybe change the name to to `define schema in external variable` (or something like that).

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmConfigurationImpl.kt
```diff
@@ -75,7 +75,7 @@ open class RealmConfigurationImpl constructor(
         } else path
         this.name = name // FIXME Should read name from end of path
         this.schema = schema
-        this.mapOfKClassWithCompanion = companionMap
+        this.mapOfKClassWithCompanion = schema.map { it to it.realmObjectCompanion()!! }.toMap()
```
cmelchior, line 78: Wouldn't it be better to handle the null pointer exception in the `realmObjectCompanion()` methods? If, for some reason, there is no realmObjectCompanion we get a rather unhelpful NullPointerException here.

If we check inside the method we can also craft a better error message containing the name of the offending class.

cmelchior (changes requested):
Really nice approach 💯 ... I have a few minor comments + we should probably update the CHANGELOG to mention that the schema set can be defined in an external variable now.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmConfigurationImpl.kt
rorbech: 👍  Makes sense

cmelchior (approved):
Nice 🚀 .... Just need a CHANGELOG entry ;)

packages/library-base/src/darwin/kotlin/io/realm/internal/platform/ModelObject.kt
```diff
@@ -0,0 +1,36 @@
...
+package io.realm.internal.platform
+
+import io.realm.internal.RealmObjectCompanion
+import kotlin.reflect.AssociatedObjectKey
+import kotlin.reflect.ExperimentalAssociatedObjects
+import kotlin.reflect.KClass
+
+/**
+ * Native-only internal annotation that is used by the compiler plugin to mark the user model
+ * classes (implementing [RealmObject]) with it's associated companion object (implementing
```
nhachicha, line 26:
```suggestion
 * classes (implementing [RealmObject]) with its associated companion object (implementing
```

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/RealmModelLoweringExtension.kt
```diff
@@ -53,8 +60,39 @@ class RealmModelLoweringExtension : IrGenerationExtension {
...
+    val modelObjectAnnotationClass by lazy {
+        pluginContext.lookupClassOrThrow(MODEL_OBJECT_ANNOTATION)
+    }
+
     override fun lower(irClass: IrClass) {
         if (irClass.hasRealmModelInterface) {
+            // For native we add @ModelObject(irClass.Companion::class) as associated object to be
+            // able to resolve the companion object during runtime due to absence of
+            // kotlin.reflect.full.companionObjectInstance
+            if (pluginContext.platform?.componentPlatforms?.first()?.platformName == "Native") {
```
nhachicha, line 74: `pluginContext.platform.isNative()` should also work

nhachicha (approved):
Very clever! 👍

Issue #397, opened 2021-08-31
Fix up stale NotificationTests


rorbech:
There are some issues around default implementation of `@Test` annotated methods in interfaces. 

When executing `./gradlew jvmTest --info` in `test` it yields
```
io.realm.NotificationTests$DefaultImpls[jvm] > io.realm.NotificationTests.DefaultImpls.initializationError FAILED
    java.lang.Exception: Method useTestDispatchers() should not be static
```
Running a similar test in a pure Android project yields a little more elaborate error message: 
```
org.junit.runners.model.InvalidTestClassError: Invalid test class 'io.realm.examples.realmkotlinandroid.CommonTests$DefaultImpls':      
    1. Test class should have exactly one public constructor
    2. Method common() should not be static
    3. Method common should have no parameters
```
so maybe the approach is not valid at all, as it seems like the test are actually also executed on the interface. The `jvmTest` only targets tests in `io.realm.test.compiler` (Jenkinsfile: `testAndCollect("test", 'jvmTest --tests "io.realm.test.compiler*"')`, but don't know why this is not a problem for `connectedAndroidTest` when the `NotificationTests` interface is part of `commonTest` 🤷 

Anyway the current empty default implementation of methods in `NotificationTests` are just causing that we forget to implement it in the implementing class, so I guess we should only allow _empty_ default implementations that are marked with `@Ignore` and a `FIXME` referencing the issue for implementing the tests. 

Alternatively we should just never have default implementations and require the `@Ignore`/`FIXME` marker in the implementing class.

PR #608, opened 2021-12-09
Query API - 2 - scalar queries

edualonso:
~~Blocked by https://github.com/realm/realm-kotlin/pull/603~~

Added scalar queries/aggregators: `min`, `max`, `sum`, `count`
Fixed bug in error handling in generated swig code.


packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
```diff
@@ -0,0 +1,131 @@
...
+
+        val numberResult: Number = result?.let { it as Number } ?: return null
+
+        // TODO Expand to support other numeric types, e.g. Decimal128
+        return when (type) {
+            Int::class -> numberResult.toInt()
+            Long::class -> numberResult.toLong()
+            Float::class -> numberResult.toFloat()
+            Double::class -> numberResult.toDouble()
+            else -> throw IllegalArgumentException("Invalid numeric type for '$property', it is not a '${type.simpleName}'.")
```
cmelchior, line 128: Should we also allow conversion to strings or does that just raise too many questions about format so better left up to developers?

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
```diff
@@ -0,0 +1,131 @@
...
+                RealmInterop.realm_results_max(resultsPointer, colKey)
+            AggregatorQueryType.SUM ->
+                RealmInterop.realm_results_sum(resultsPointer, colKey)
+        }
+
+        val numberResult: Number = result?.let { it as Number } ?: return null
+
+        // TODO Expand to support other numeric types, e.g. Decimal128
+        return when (type) {
+            Int::class -> numberResult.toInt()
```
cmelchior, line 124: Missing Short?

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
cmelchior: We should probably give people a list of supported types here.

cmelchior: We also need to support `RealmInstant` for min and max...`sum` probably doesn't make sense.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -347,7 +348,56 @@ class QueryTests {
     }
 
     @Test
-    fun composedQuery_withDescriptorAndQueryAgain() {
+    fun composedQuery_withDescriptors() {
```
cmelchior, line 351: Seems to be missing quite a lot of tests here. I would expect tests for each supported type pr. aggregate. Boundary tests, what happens on overflows and so on.

cmelchior (commented):
I assume this PR is still in progress, so just leaving a few initial remarks.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: Yeah, I was still devising the exhaustive testing strategy. There should be quite a lot more tests now.

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
edualonso: I suspect adding string conversion would make it a bit more confusing, so I would leave it as is.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -676,9 +807,131 @@ class QueryTests {
     }
 
     @Test
-    @Ignore
+    fun sum_find_emptyTable() {
+        val assertions = { propertyDescriptor: PropertyDescriptor ->
+            realm.query(QuerySample::class)
+                .sum(propertyDescriptor.property.name, propertyDescriptor.clazz)
+                .find { sumValue: Any? ->
```
cmelchior, line 814: Something doesn't look right if this returns `Any?` Or is it because `propertyDescriptior.clazz` is a runtime variable?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -796,4 +1225,270 @@ class QueryTests {
...
+
+    // ------------------------
+    // General-purpose helpers
+    // ------------------------
+
+    // Deletes all objects after assertions to avoid "null vs 0" results when testing aggregators
+    private fun cleanUpBetweenProperties() = realm.writeBlocking {
+        query(QuerySample::class)
+            .find { results ->
+                results.toList() // We cannot iterate over the results and delete directly!
```
rorbech, line 1404: Couldn't you just call `.delete()` on the result or have you removed that?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -676,9 +807,131 @@ class QueryTests {
...
+                }
+        }
+
+        // Iterate over non-nullable properties only - exclude RealmInstant
+        for (propertyDescriptor in propertyDescriptorsForSum) {
+            assertions(propertyDescriptor)
+        }
+
+        // Iterate over nullable properties containing both null and non-null values - exclude RealmInstant
+        for (nullablePropertyDescriptor in nullablePropertyDescriptorsForSum) {
```
cmelchior, line 827: I would create an `allPropertyDescriptorsForSum` since that would make this an other tests a bit more readable IMO.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -796,4 +1225,270 @@ class QueryTests {
...
+                    }
+            }
+    }
+
+    // ----------------------------------------
+    // Descriptors used for exhaustive testing
+    // ----------------------------------------
+
+    private val basePropertyDescriptors = listOf(
+        PropertyDescriptor(QuerySample::intField, Int::class, INT_VALUES),
```
rorbech, line 1418: Seems weird that we test `Short` explicitly. All integral types map to Core's INTEGER, so if we test all Kotlin integral types shouldn't we then also test `Char` and `Byte`?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -796,4 +1225,270 @@ class QueryTests {
...
+                        delete(latest)
+                    }
+            }
+    }
+
+    // ----------------------------------------
+    // Descriptors used for exhaustive testing
+    // ----------------------------------------
+
+    private val basePropertyDescriptors = listOf(
```
rorbech, line 1417: Maybe somehow use the `TypeDescriptor.classifier` as root of making the test exhaustive, so that when introducing a new type, we would have a failing tests that required to add the new type to a set of unsupported aggregator types. Don't know if it makes sense, but would be nice if we somehow ensured that we had to evaluate if a new type was applicable for a certain feature. Otherwise we forget to add it somewhere.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -796,4 +1225,270 @@ class QueryTests {
...
+        expectedAggregator(clazz, kType, AggregatorQueryType.SUM)
+
+    private fun expectedMax(clazz: KClass<*>, kType: KType): Any =
+        expectedAggregator(clazz, kType, AggregatorQueryType.MAX)
+
+    private fun expectedMin(clazz: KClass<*>, kType: KType): Any =
+        expectedAggregator(clazz, kType, AggregatorQueryType.MIN)
+
+    @Suppress("LongMethod", "ComplexMethod")
+    private fun expectedAggregator(
```
rorbech, line 1305: Seems very bloated. Can't you just assign an aggregator to a variable by referencing `::minOrNull()`, `::maxOrNull`, etc. and then apply it on a set looked up from a map of `Map<Pair<KClass<*>, Boolean>. Collection<*>`. Pretty hard to catch errors here, because you loose the overview.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -676,9 +807,131 @@ class QueryTests {
...
 
     @Test
-    @Ignore
+    fun sum_find_emptyTable() {
+        val assertions = { propertyDescriptor: PropertyDescriptor ->
+            realm.query(QuerySample::class)
+                .sum(propertyDescriptor.property.name, propertyDescriptor.clazz)
+                .find { sumValue: Any? ->
+                    if (sumValue is Number) {
+                        assertEquals(0, sumValue.toInt())
```
cmelchior, line 816: Shouldn't `sum` also return `null` if there are no entries? I would think the only aggregate that always have a value is `count` since that is the only one that doesn't depend on the values of the entries?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -676,9 +807,131 @@ class QueryTests {
...
+    @Test
+    fun sum_find_allNullValues() {
+        val assertions = { propertyDescriptor: PropertyDescriptor ->
+            realm.writeBlocking {
+                // Queries inside a write transaction produce live results which means they can be
+                // reused within the closure
+                val sumQuery = query(QuerySample::class)
+                    .sum(propertyDescriptor.property.name, propertyDescriptor.clazz)
+
+                // The sum of all null values is 0
```
cmelchior, line 841: See above. I'm not sure that is the correct semantics?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -689,22 +942,109 @@ class QueryTests {
...
     }
 
     @Test
     @Ignore
-    fun max_find() {
+    fun sum_asFlow_throwsIfRealmInstant() {
         // TODO
     }
 
+    @Test
```
rorbech, line 961: Don't know if there is infrastructure to collect multiple errors in a test, but would maybe make sense to do an grouping of tests with something like:
```
@Test
fun aggregators() {
    for (aggregator in setOf(MIN, MAX, COUNT, ...) {
        find_emptyTable(...)
        find_allNullValues(....)
       ...
    }
}
```
To ensure that we catch all the uses cases for each aggregator type. Don't know if the implementations vary too much for it to make sense though.

rorbech (changes requested):
Nice. Some comments/suggestions to the test structure, but nothing really critical. 

But, the error check in the typemap does require an update to only return when actually throwing.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -676,9 +807,131 @@ class QueryTests {
...
+            realm.query(QuerySample::class)
+                .sum(timestampDescriptor.property.name, timestampDescriptor.clazz)
+                .find() // The sum is only evaluated after obtaining results!
+        }
+
+        assertFailsWith<IllegalArgumentException> {
+            realm.query(QuerySample::class)
+                .sum(nullableTimestampDescriptor.property.name, nullableTimestampDescriptor.clazz)
+                .find() // The sum is only evaluated after obtaining results!
+        }
```
cmelchior, line 934: Missing some tests here IMO:

* Error when trying to sum a list or object
* What happens on overflow, e.g. If you want a sum as Short but it overflows, should we just overflow or return Short.MAX_VALUE... this needs to be documented in unit tests and documentation.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -689,22 +942,109 @@ class QueryTests {
...
+        // Iterate over non-nullable properties only
+        for (propertyDescriptor in propertyDescriptors) {
+            assertions(propertyDescriptor)
+        }
+
+        // Iterate over nullable properties containing both null and non-null values
+        for (nullablePropertyDescriptor in nullablePropertyDescriptors) {
+            assertions(nullablePropertyDescriptor)
+        }
+    }
```
cmelchior, line 1046: Seems to be missing tests for unsupported field types like objects and lists

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -796,4 +1225,270 @@ class QueryTests {
...
+            }
+            AggregatorQueryType.MAX -> when {
+                returnType.isMarkedNullable -> NULLABLE_INT_VALUES.mapNotNull { it }.maxOrNull()
+                else -> INT_VALUES.maxOrNull()
+            }
+            AggregatorQueryType.SUM -> when {
+                returnType.isMarkedNullable -> NULLABLE_INT_VALUES.mapNotNull { it }.sum()
+                else -> INT_VALUES.sum()
+            }
+        }?.toInt() ?: throw IllegalArgumentException("Aggregate result cannot be null.")
```
cmelchior, line 1323: Not sure I understand this....they are allowed to be null? 🤔

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
cmelchior, line 1417: Missing support for Byte as well?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
cmelchior, line 1305: These helper functions are rather hard to follow at a glance, some kind of Dokka for them would be nice.

cmelchior (changes requested):
Nice.... There seem to be missing some corner cases in tests though... I also had a few other minor comments.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
rorbech: It sounds rather dangerous to automatic truncation based on the type. I would probably prefer to always return the long-value from core and have the user to do the truncation explicitly with `toShort` if they want. This way we get pure Kotlin behavior. Alternatively throw and overflow exception if we do the type conversion and the value doesn't fit.

cmelchior: The problem with overflow exists for all types and we cannot only return a long... e.g. there is also double, float and Decimal128 that can be summed.

Personally, I would lean towards respecting the type the user wants, but let it overflow as normal...Throwing seems very heavy-handed and unexpected.

But maybe we should discuss later today?

edualonso: > Personally, I would lean towards respecting the type the user wants, but let it overflow as normal...Throwing seems very heavy-handed and unexpected.

I'm also inclined to take that approach, which is how I implemented it in this PR. The more I think about it the more I believe it's the least complicated one for both us and end-users. All the other options I've tried feel pretty awkward.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: I have split `TypeDescriptor.classifiers` into two categories now:
```
    // Classifiers for types that can be used in aggregated queries
    val aggregateClassifiers: Map<KClassifier, CoreFieldType> = mapOf(
        Byte::class to CoreFieldType.INT,
        Char::class to CoreFieldType.INT,
        Short::class to CoreFieldType.INT,
        Int::class to CoreFieldType.INT,
        Long::class to CoreFieldType.INT,
        Float::class to CoreFieldType.FLOAT,
        Double::class to CoreFieldType.DOUBLE
    )

    // Kotlin classifier to Core field type mappings
    val classifiers: Map<KClassifier, CoreFieldType> = aggregateClassifiers + mapOf(
        Boolean::class to CoreFieldType.BOOL,
        String::class to CoreFieldType.STRING,
        RealmInstant::class to CoreFieldType.TIMESTAMP,
        RealmObject::class to CoreFieldType.OBJECT
    )
```
and now I'm using the `aggregateClassifiers` in here instead, mapping them to what I need.

This makes it more failproof. However, I would like to give it another spin at some other point. I think we could make it even more granular.

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
```diff
@@ -0,0 +1,146 @@
...
+        // TODO Expand to support other numeric types, e.g. Decimal128
+        return when (result) {
+            null -> null
+            is Timestamp -> RealmInstant.fromEpochSeconds(result.seconds, result.nanoSeconds)
+            is Number -> when (type) {
+                Int::class -> result.toInt()
+                Short::class -> result.toShort()
+                Long::class -> result.toLong()
+                Float::class -> result.toFloat()
+                Double::class -> result.toDouble()
```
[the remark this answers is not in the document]
rorbech: I missed that it was `when(type)` and not `result.class`, sorry.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
rorbech: I see. I guess we could then use the pattern with a `AggregatorTest` interface with the specific tests and have a test class for each aggregator. This haven't worked well on the `NotificationTests` though. Part of the issues with that is:
- that default methods in the interface will trigger running the actual test on the interface (probably the underlying `DefaultImpl` class) but the test executor fails
- running the `jvmTest`

I previously tried summarizing the issues in https://github.com/realm/realm-kotlin/issues/397. Some of the it could be avoided by eliminating default implementations and gradle/test-runner infrastructure might have changes, so maybe worth a very quick shot again 🤔

packages/jni-swig-stub/realm.i
```diff
@@ -129,13 +129,17 @@ void throw_as_java_exception(JNIEnv *jenv) {
...
+        // Check if it works if throw_as_java_exception doesn't throw
+        return $null;
     }
     *($1_type*)&jresult = result;
 }
 
 %typemap(out) bool {
     if (!result) {
         throw_as_java_exception(jenv);
+        // Check if it works if throw_as_java_exception doesn't throw
```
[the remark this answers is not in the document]
rorbech: Same ;)

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
rorbech: I guess that works. At least they are somehow linked now. 

Just input for future experimens/modeling: Maybe it should even be a boolean attribute on the CoreFieldType so that we could just iterate the classifiers and test those where the corresponding core-type supports aggregation.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
rorbech: Way better, but I still think that I would have done it with separate mappings for:
- _classifier/KClass_ to test collection, and 
- `AggregatorQueryType` to the Collection aggregator
And then apply the aggregator on the collection in the end. Maybe your proposal is better handling the special case that sum is not supported for RealmInstant

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: That's a very good point, but implementing a one-size-fits-all solution is tricky because not all aggregators work for all types, e.g. it's possible to compute the `max`/`min` of a `Timestamp` field but not its `sum` 😭 - that's why I used `propertyDescriptorsForSum` and `propertyDescriptors` in the tests.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: Yes, it's a bummer that `sum` and `RealmInstant` break the general case. I will leave it like this for now though, we can revisit it once we put time aside to standardise our exhaustive testing strategy.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -14,26 +14,42 @@
  * limitations under the License.
  */
 
+@file:Suppress("invisible_reference", "invisible_member")
```
rorbech, line 17: I think we should strive to avoid having these at file level? What is the purpose?

I have used utility methods or extension properties elsewhere to encapsulate the internal access, to ensure that we don't leak all internal symbols into the full test.

edualonso: I agree but here I'm using it to access `AggregatorQueryType` and switch over its values so I need the import which cannot be found unless I put the annotation at a file level. Writing the full path to the enum would work too but it's used quite a lot of times so I don't think that scales up.

rorbech: Then I think we should consider making the class public. As long as it is in an `internal` sub-package then the IDE will not autocomplete and automatically import it for the public users.

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
```diff
@@ -0,0 +1,144 @@
...
+    realmReference: RealmReference,
+    queryPointer: NativePointer,
+    mediator: Mediator,
+    private val clazz: KClass<*>,
+    private val property: String,
+    private val type: KClass<T>,
+    private val queryType: AggregatorQueryType
+) : BaseScalarQuery<E, T>(realmReference, queryPointer, mediator) {
+
+    override fun find(): T? = findInternal(queryPointer)
```
rorbech, line 90: Any reason for not just inlining `findInternal` here. Doesn't seem to be used elsewhere.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
rorbech: Are these tests added somewhere now? I didn't find them ... unless it is some of the still pending TODOs. Is that for future a PR?

rorbech (approved):
Looks really good. As usual it is extremely difficult to keep the overview with these incremental PRs. I assume that you have iterated the JAVA tests and ensured that we have same use-case coverage?

Besides that, I agree that it would be nicer returning `null` when there are no elements. But I know that I sometimes looses these battles and if realm-java has a different semantics then it is maybe ok ... but ... 🤔 (if this is really core semantics, then we should probably just proceed with this, but we could have a final vote before we merge the integration branch)

I also still would prefer to avoid the file level suppressing of internals (see separate comment about options).

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
edualonso: Yes, I noticed that later in the process. It should be there in https://github.com/realm/realm-kotlin/pull/609

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: Yes, all TODOs in this test file are to be added later when the corresponding functionality is added in upcoming PRs.

PR #609, opened 2021-12-09
Query API - 3 - added `asFlow` variants to ordinary and scalar queries

edualonso:
Blocked by https://github.com/realm/realm-kotlin/pull/608

- Added `asFlow` for ordinary queries - `thaw` creates `ElementResults` and `asFlow` uses them via the notifier.
- Added `asFlow` for scalar queries - `thaw` creates `ScalarResults` and `asFlow` uses them via the notifier. The aggregator values are computed using the C-API functions exposed through `realm_results_t` - it could be improved if `realm_query_t` exposed these through the C-API, but since we still need results for a subscription, I think it's fair to just leave it like this.
- Fixed bug in swig generated code when there is an exception being thrown but not returned the correct value which caused the exception to be swallowed.


packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
```diff
@@ -142,3 +166,31 @@ internal class AggregatorQuery<E : RealmObject, T : Any> constructor(
         } as T?
     }
 }
+
+/**
+ * Convenience class to encapsulate the logic behind whether repeated elements should be emitted by
+ * [RealmScalarQuery.asFlow]. The [firstUpdate] and [previousValue] properties need to be
+ * [AtomicRef]s due to Kotlin Native's memory model.
+ */
+internal class ValueChangeManager<T> {
```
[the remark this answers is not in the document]
edualonso: 🤦

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -903,29 +991,63 @@ class QueryTests {
...
+        }
+
+        runBlocking {
+            assertFailsWith<IllegalArgumentException> {
+                realm.query<QuerySample>()
+                    .sum<String>(QuerySample::intField.name)
+                    .asFlow()
+                    .collect { /* No-op */ }
+            }.let {
+                assertTrue(it.message!!.contains("Invalid numeric type"))
```
rorbech, line 1017: Maybe we should somehow hint that it is the aggregators `T` that is of the invalid type. It could also be the actual field. Or is there more context in the full message?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -1043,23 +1165,71 @@ class QueryTests {
...
+        }
+
+        runBlocking {
+            assertFailsWith<IllegalArgumentException> {
+                realm.query<QuerySample>()
+                    .max<String>(QuerySample::intField.name)
+                    .asFlow()
+                    .collect { /* No-op */ }
+            }.let {
+                assertTrue(it.message!!.contains("Invalid numeric type"))
```
rorbech, line 1191: Same

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -903,29 +991,63 @@ class QueryTests {
...
     fun sum_asFlow_throwsIfInvalidProperty() {
-        // TODO
+        runBlocking {
+            assertFailsWith<IllegalArgumentException> {
+                realm.query<QuerySample>()
+                    .sum<String>(QuerySample::stringField.name)
+                    .asFlow()
+                    .collect { /* No-op */ }
+            }.let {
+                assertTrue(it.message!!.contains("Invalid query formulation"))
```
rorbech, line 1031: Again, is the error test explanatory enough for people to understand it. Or is there more context in the full message?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -154,35 +157,59 @@ class QueryTests {
...
+                query<QuerySample>("nonExistingField = 13")
             }.let {
                 assertTrue(it.message!!.contains("Wrong query field"))
             }
         }
     }
 
     @Test
-    @Ignore
     fun asFlow() {
```
rorbech, line 180: Aren't there more corner cases that we should test? Empty results, deleting parent object, etc.?

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -669,34 +722,67 @@ class QueryTests {
...
 
-        realm.query(QuerySample::class)
+        realm.query<QuerySample>()
             .count()
             .find { countValue -> assertEquals(1, countValue) }
     }
 
     @Test
-    @Ignore
     fun count_asFlow() {
```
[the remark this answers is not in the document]
rorbech: Were these tests added. Don't think I stumbled upon them, but maybe it is because it is opaque when the data set is not visible in the individual test.

rorbech (commented):
Maybe I am missing it or maybe it is in another of the PRs, but wouldn't we also need:
- Aggregator flow tests for the empty set
- Ordinary flow test for empty set
And what about flow-termination? I might have asked this before, but I fail to recall it and why it maybe wouldn't be required.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: The exception message is
```
else -> throw IllegalArgumentException("Invalid numeric type for '$property', it is not a '${type.simpleName}'.")
```
But I will change it to `"Invalid numeric type for '$property', it is not a '${type.simpleName}' or cannot be represented by it."` to better reflect what is said in the documentation.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: "Invalid query formulation" is thrown when we receive a `RealmCoreLogicException` or we are processing `genericRealmCoreExceptionHandler`. The message is delivered in full like so: `"Invalid query formulation: ${exception.message}"` being `exception.message` the actual message from core, which in most cases (as far as I've tested) should be clear enough.

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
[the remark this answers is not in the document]
edualonso: They were. They can be found in `aggregatorAssertions`. The execution flow is as follows:
1. subscribe to the flow and receive initial aggregated value (0 or null)
2. update the realm
3. receive emission and assert
4. attempt to trick `distinctUntilChanged` by deleting the previously stored objects and save the same ones, resulting into an unchanged aggregate
5. ensure we receive no emissions
6. delete stored values to trigger an emission

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: See above

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
edualonso: Empty results are being tested here, but I will split the test in two for better visibility. I forgot about deleting the observable object 🙈

edualonso:
> * Aggregator flow tests for the empty set

It might not be the best in terms of visibility but they are already there, they can be found in `asFlowAggregatorAssertions`. I have added some documentation to the method for clarity.

> * Ordinary flow test for empty set

Same as before, they were there, but now I've split the test in two for better visibility.

>   And what about flow-termination? I might have asked this before, but I fail to recall it and why it maybe wouldn't be required.

I have added those in the latest commit 👍

rorbech:
> * And what about flow-termination? I might have asked this before, but I fail to recall it and why it maybe wouldn't be required.

Now I recall. I guess I confused it with object notifications where we actually complete the flow if the observer object is delete. But here the query is not tied to an object, so should be good.

cmelchior (approved):

test/base/src/androidTest/kotlin/io/realm/test/shared/QueryTests.kt
```diff
@@ -668,34 +775,169 @@ class QueryTests {
...
-        // TODO
+        val channel = Channel<Long>(1)
+
+        runBlocking {
+            val observer = async {
+                realm.query<QuerySample>()
+                    .count()
+                    .asFlow()
+                    .collect { countValue ->
+                        assertNotNull(countValue)
```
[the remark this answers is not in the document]
edualonso: We discussed this at some point and it is hard to have all aggregate queries under `RealmScalarQuery` since count isn't nullable per definition but max and min can return null if a table is empty.

I will quickly prototype a nullable interface and a non-nullable one instead for better type safety and see how it feels.

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/ScalarQuery.kt
```diff
@@ -66,32 +74,35 @@ internal abstract class BaseScalarQuery<E : RealmObject, T : Any> constructor(
...
+    mediator: Mediator,
+    clazz: KClass<E>
+) : BaseScalarQuery<E, Long>(realmReference, queryPointer, mediator, clazz) {
 
     override fun find(): Long = RealmInterop.realm_query_count(queryPointer)
 
-    override fun Flow<RealmResultsImpl<E>?>.queryMapper(): Flow<Long?> = TODO()
-
-    override fun discardRepeated(latestValue: Long?): Boolean = TODO()
+    override fun Flow<RealmResultsImpl<E>?>.queryMapper(): Flow<Long?> = this.map {
```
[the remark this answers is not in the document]
edualonso: See comment below.

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -256,21 +257,54 @@ interface RealmSingleQuery<T> : Flowable<T> {
 
 /**
  * Queries that return scalar values. This type of query is used to more accurately represent the
- * results provided by some query operations, e.g. [RealmQuery.count] or [RealmQuery.min].
+ * results provided by some query operations, e.g. [RealmQuery.count] or [RealmQuery.sum].
```
cmelchior, line 260: Isn't it only count that nonnull, sum also returns `null` if no elements are found?

edualonso: No, core returns 0 if the table is empty.

cmelchior: I would probably question if that was a good idea, similar to how `average` shouldn't return 0, but probably not something we should fix in this PR

PR #615, opened 2021-12-21
DynamicRealms API

clementetb:
With this PR we add a draw a proposal for `DynamicRealm` and `DynamicObject`, two data structures that allow accessing a `Realm` with type unsafe methods. The main use of this feature would be Migrations, but it has other uses where we could not use a type safe environment, like implementing a database explorer.

This API does not cover implementing a mutable schema.


rorbech:
Superseded by #674

PR #620, opened 2022-01-04
Remove libc++_shared after reworking Android build

rorbech:
This remove the `libc++_shared.so` files from the `cinterop` AAR as it is no longer required after cleaning up the Android build and consuming core through their cmake infrastructure.


packages/cinterop/build.gradle.kts
```diff
@@ -295,7 +295,6 @@ android {
         // Out externalNativeBuild (outside defaultConfig) does not seem to have correct type for setting cmake arguments
         externalNativeBuild {
             cmake {
-                arguments("-DANDROID_STL=c++_shared")
```
cmelchior, line 298: Nice.

But I'm trying to figure out the impact of this....maybe I still have vacation brain, but I cannot see Core or us setting an STL anywhere 🤔 , which I assume means it will use the default one? And the default one is static, which will blow up our library size https://developer.android.com/ndk/guides/cpp-support

rorbech: Since we are building a single library I guess it is stripped to the minimal so don't know the actual impact of it. But seems like we would probably hit issues if apps depends on multiple libraries anyway: https://developer.android.com/ndk/guides/cpp-support#sr

So I guess we are left with using the `c++_shared` one and require that other dependencies use a binary compatible one https://developer.android.com/ndk/guides/cpp-support#one_stl_per_app

nhachicha (approved):

cmelchior (approved):
Looks good to me. We just need a changelog entry that this fixes https://github.com/realm/realm-kotlin/issues/694

Issue #327, opened 2021-07-02
[Frozen Architecture] Native leak memory when closing Realms


cmelchior:
Native leak memory in `MemoryTests.garbageCollectorShouldFreeNativeResources` 

It looks like it could be related to circles in our object-graph created through `RealmReference`: https://github.com/realm/realm-kotlin/pull/314

rorbech:
The garbage collector for the new memory model should be able to detect those cycles now, so should probably be able to reenable the test.

Issue #147, opened 2021-03-17
Investigate why releasing realm_collection_changes_t* and realm_object_changes_t * Seg Fault


nhachicha:
Currently, we use unmanaged Native Pointers (`managed = false`) to disable GC'ing (and calling `realm_release`) otherwise we'll get a segmentation fault when trying to release the pointer

PR #623, opened 2022-01-05
Support for Kotlin's new memory model

rorbech:
Fixes a couple of minor issues with using Kotlin's new memory model.

Adds a separate duplicate Macos test stage to verify our test suite with the new memory model. The test report currently overlaps, but as we wouldn't reach that stage unless previous stages succeeded, so assume that is ok for now. 

The new memory model still doesn't seem to catch the memory leak of #327, but investigations of this shouldn't necessarily hold this update back. There are also changes to the `AtomicRef` implementation with Kotlin 1.6.20 so maybe worth reiteration at that time along with Coroutines 1.6.0.

Relates to #635


packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -1127,7 +1127,7 @@ actual object RealmInterop {
                     SyncException(message)
                 }
                 val errorCallback = safeUserData<SyncErrorCallback>(userData)
-                val session = CPointerWrapper(syncSession)
+                val session = CPointerWrapper(realm_clone(syncSession))
```
cmelchior, line 1130: Any reason why this needs to change?

test/base/src/macosTest/kotlin/io/realm/test/CoroutineTests.kt
```diff
@@ -49,8 +49,13 @@ class CoroutineTests {
     // Fails on non native-mt as dispatching from background thread to main does change actual
     // thread, thus failing to assert main thread id
     @Test
+    // Ignored for new memory model as (from https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md):
```
cmelchior, line 52: Is there an intent to do something about this? With the new test changes in 1.6.0, I think it becomes possible to override the Main dispatcher on all platforms.

test/base/src/macosTest/kotlin/io/realm/test/MemoryTests.kt
```diff
@@ -36,6 +36,8 @@ import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
+// New memory model doesn't seem to fix or memory leak, or maybe we have to reevaluate how we do
```
cmelchior, line 39:
```suggestion
// New memory model doesn't seem to fix our memory leak, or maybe we have to reevaluate how we do
```

cmelchior (approved):

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
rorbech: Think it is just a miss in the original implementation. @fealebenpae internally commented on https://github.com/realm/realm-core/pull/4951:
```
I understand in pretty much all cases you want to prolong the lifetime of the user object, but I also want to keep the ownership rules clear and consistent: objects returned by calling functions belong to the caller, and objects passed to callbacks are released by default after the callback ends so the implementor of the callback needs to clone them and save them for later.
```
Don't know why it hasn't been crashing before now 🤷

test/base/src/macosTest/kotlin/io/realm/test/CoroutineTests.kt
rorbech: I added a comment to #634 on addressing this

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
rorbech: Probably the same issue causing #147

nhachicha (approved):

PR #633, opened 2022-01-17
Use custom module name to avoid buildtype mismatch 

### LINKED ISSUE #621: java.lang.NoSuchMethodError: No virtual method setValue
When trying out the example of creating an object for realm creating a Task class, i am getting this error **ONLY** on android, the iOS target works fine:

```
java.lang.NoSuchMethodError: No virtual method setValue$library_base_release(Lio/realm/internal/RealmObjectInternal;Ljava/lang/String;Ljava/lang/Object;)V in class Lio/realm/internal/RealmObjectHelper; or its super classes (declaration of 'io.realm.internal.RealmObjectHelper' appears in /data/data/com.abc.android.searchFlowPrototype/code_cache/.overlay/base.apk/classes12.dex)
        at Task.setName(Main.kt:59)
```

Just for reference Main.kt: 59 is var  name.
```
class Task : RealmObject {
    var name: String = "new task"
    var status: String = "Open"
}
```
```
realm.writeBlocking {
            copyToRealm(Task())
        }
```

Using version 0.8.0 of the plugin and library.

nhachicha:
fixes #621 
Kotlin compiler uses name mangling for certain language aspects not available in Java (like [Inline Classes and Internal methods](https://ncorti.com/blog/name-mangling-in-kotlin)). By default, the mangling for an internal method for instance uses the module name which contains also the build type variant 
classes
This is problematic when the build type of the client doesn't match the dependency we pull (example: the User is pulling release version of the library `implementation("io.realm.kotlin:library-base-debug:0.8.0` while working with a debug build type) 

This fix will set a custom (build type independent) module name to be used to mangle internal method names


CHANGELOG.md
```diff
@@ -1,3 +1,39 @@
+## 0.8.0 (2022-01-17)
```
edualonso, line 1:
```suggestion
## 0.8.1 (2022-01-17)
```

edualonso (changes requested):

PR #641, opened 2022-01-19
Running monkey on release variant to detect any Proguard misconfiguration

nhachicha:
Note: The PR is expected to fail on CI, until we fix the Proguard configuration needed to cache JClass referenced from JNI https://github.com/realm/realm-kotlin/blob/master/packages/cinterop/src/jvm/jni/java_class_global_def.hpp#L46-L52 

```
01-18 16:37:22.722 10545 10545 E REALM   : /Users/realm/workspace-realm-kotlin/releases/packages/cinterop/src/jvm/jni/java_class.cpp:49: [realm-core-11.6.1] Assertion failed: cls with (class_name) =  ["io/realm/internal/interop/sync/NetworkTransport"]
```


edualonso:
Replaced by https://github.com/realm/realm-kotlin/pull/644 due to having to release a hotfix.

PR #644, opened 2022-01-19
Running monkey on release variant to detect any Proguard misconfiguration

### LINKED ISSUE #643: Sync-related classes are not excluded from Proguard correctly
Having exceptions for sync-related classes in the `library-sync` Proguard consumer doesn't work. `JavaClassGlobalDef` has references to said classes when JNI is initialised, causing it to crash.

Said classes are all located under the `cinterop` module, which is shared across `library-base` and `library-sync`. Given that the JNI initialiser is only present for `library-base`, said exceptions must also be in the `library-base` Proguard consumer.

Additionally, `RealmObjectCompanion` had not been included from being stripped, resulting in crashes too as trying to retrieve companion objects from `RealmObjects` would cause crashes in release builds too.

edualonso:
Fixes #643

Moved Proguard sync-related exceptions located under `cinterop` to the `library-base` Proguard consumer file as they are needed when instantiating JVM's `JavaClassGlobalDef`.

Added missing exceptions for `RealmObjectCompanion` and `JVMScheduler` as well since it was also causing similar crashes.

Added a step to CI where the monkey runner is executed on the release sample app instead of the debug one to catch these types of Proguard issues.


nhachicha (commented):
Nice catch 👏, I think you forgot to either rebase on top of my branch or add the commit that modifies the Jenkins file to run monkey against the release variant

edualonso:
> Nice catch 👏, I think you forgot to either rebase on top of my branch or add the commit that modifies the Jenkins file to run monkey against the release variant

🤦 that's right. I've already changed it. It will hopefully be fine on CI, since I ran it like 10 times on my machine yesterday.

cmelchior (approved):

Issue #631, opened 2022-01-14
Flaky: Android Sample App test in CI


rorbech:
Crash when installing the sample app project https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-623/14/pipeline 🤔

cmelchior:
Stacktrace:
```
:Sending Touch (ACTION_DOWN): 0:(404.0,1500.0)
:Sending Touch (ACTION_UP): 0:(340.42627,1458.785)
// NOT RESPONDING: com.google.android.googlequicksearchbox:search (pid 20118)
ANR in com.google.android.googlequicksearchbox:search
PID: 20118
Reason: Input dispatching timed out (e6b612d VoiceInteractionSession (server) is not responding. Waited 5003ms for MotionEvent(deviceId=-1, source=0x00010004, displayId=0, action=MOVE, actionButton=0x00000000, flags=0x00000000, metaState=0x00000000, buttonState=0x00000000, classification=NONE, edgeFlags=0x00000000, xPrecision=1.0, yPrecision=1.0, xCursorPosition=nan, yCursorPosition=nan, pointers=[0: (-3.0, -3.0)]), policyFlags=0x6b000000)
Load: 0.0 / 0.0 / 0.0
----- Output from /proc/pressure/memory -----
some avg10=15.16 avg60=4.40 avg300=0.99 total=1009575631
full avg10=2.06 avg60=0.59 avg300=0.13 total=304333570
----- End output from /proc/pressure/memory -----
```


It happens because the Monkey issues events outside the app, which can cause all kinds of weird issues. Apparently there is a parameter that can prevent this.

PR #647, opened 2022-01-21
Removed system key events from monkey runner

edualonso:
Fixes https://github.com/realm/realm-kotlin/issues/631

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -553,7 +553,7 @@ def runMonkey() {
             sh """
                 cd examples/kmm-sample
                 ./gradlew uninstallAll installRelease --stacktrace --no-daemon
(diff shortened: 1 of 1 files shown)


nhachicha (approved):

cmelchior (commented):
Have you tried running this a few times? I believe I tried something similar in Java, and still it sometimes go outside the app.

I was wondering if we also need to set the package name for the  events? 🤔 https://stackoverflow.com/questions/20911714/disable-system-level-events-on-android-monkey

cmelchior:
Doesn't hurt, so even if my comment is true, we can always go back and revisit

edualonso:
I also saw that Stackoverflow post, though I didn't see that modifier in the official docs. The post is quite old (2015) so it may be some legacy stuff. I did try running the monkey a number of times and didn't see anything out of the ordinary. I didn't see it crash either, so I assume it will be fine. And if not, as you say, we can come back to it and look at it more in depth.

Issue #638, opened 2022-01-18
Flaky test: [ios] io.realm.test.shared.SyncedRealmTests.canSync


rorbech:
iOS test stage failed in https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/master/297/pipeline with the below stack trace. Haven't been able to reproduce locally and subsequent build succeeded. The succeeding build ran on the same machine and since the server is rebooted pr. test stage it shouldn't be leaking data from other stages ... but I guess the primary key default could be an issue 🤔  
```
[ RUN      ] io.realm.test.shared.SyncedRealmTests.canSync
2022-01-18 03:32:59.180 test.kexe[66019:50717365] INFO: [SYNC] Connection[1]: Session[1]: client_reset_config = false, Realm exists = true, client reset = false(null)
2022-01-18 03:32:59.180 test.kexe[66019:50717365] INFO: [SYNC] Connected to endpoint '::1:9090' (from '::1:61012')(null)
2022-01-18 03:32:59.240 test.kexe[66019:50717365] INFO: [SYNC] Connection[2]: Session[2]: client_reset_config = false, Realm exists = true, client reset = false(null)
2022-01-18 03:32:59.241 test.kexe[66019:50717365] INFO: [SYNC] Connected to endpoint '::1:9090' (from '::1:61013')(null)
2022-01-18 03:32:59.305 test.kexe[66019:50717365] INFO: [SYNC] Connection[1]: Disconnected(null)
2022-01-18 03:32:59.305 test.kexe[66019:50717365] INFO: [SYNC] Connection[2]: Disconnected(null)
kotlin.IllegalArgumentException: Cannot create object with existing primary key
Failed to read dyld info for process 66019 (5)
    at 0   test.kexe                           0x000000010014ad91 kfun:kotlin.Throwable#<init>(kotlin.String?){} + 97
    at 1   test.kexe                           0x0000000100143c9d kfun:kotlin.Exception#<init>(kotlin.String?){} + 93
    at 2   test.kexe                           0x0000000100143e3d kfun:kotlin.RuntimeException#<init>(kotlin.String?){} + 93
    at 3   test.kexe                           0x00000001001441dd kfun:kotlin.IllegalArgumentException#<init>(kotlin.String?){} + 93
[...]

PR #650, opened 2022-01-24
Fix flaky canSync test

rorbech:
This PR addresses the issue with flaky `SyncedRealmTest.canSync` test. The PR just ensures that the data for `canSync` is completely separated from other tests by specifying a unique partition value. This is an appropriate measure that we should have done anyway and this seems to fix the issue, but I will keep the original issue (#638) open as we probably need to understand exactly why this sometimes failed on iOS and isn't reproducible locally or on any of the other platform. Understanding the root of that would help ensuring that our test tear setup/teardown mechanisms are adequate. 

The platform specific test stages are each spinning up their own instance of the server so nothing leaks between those, so maybe just a matter of the order of the iOS test that leaves some `ChildPK` instances from a previous test hanging around on the server and not being correctly separated due to the missing partition value, but again ... that should most likely also be reproducible on local tests 🤔


rorbech:
Tried retriggering the build five times on CI and didn't see the issue

edualonso (approved):

nhachicha (approved):

PR #664, opened 2022-02-03
Support coroutine 1.6.0's freeze transparency

rorbech:
This PR makes our internal `freeze` calls a no-op when running with Kotlin's new experimental memory model. This way we avoid freezing coroutine primitive to support Coroutine 1.6.0's new _freeze transparency_.

NOTE: Since `ktor` 1.6.5 doesn't support that yet the `kotlin.native.binary.freezing=disabled` is still required for `library-sync`

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/internal/util/PlatformUtils.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/internal/util/PlatformUtils.kt
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2022 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal.util
+
+/**
+ * Method to freeze state.
+ * Calls the platform implementation of 'freeze' on native, and is a noop on other platforms.
+ *
+ * Note, this method refers to Kotlin Natives notion of frozen objects, and not Realms variant
+ * of frozen objects.
+ */
+expect fun <T> T.freezeOnOldMM(): T
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
@@ -141,7 +141,7 @@ interface AppConfiguration {
                         appLogger.debug(message)
                     }
                 }
-            ).freeze() // Kotlin network client needs to be frozen before passed to the C-API
+            ).freeze() // MM: Catches userLoggers // Kotlin network client needs to be frozen before passed to the C-API
 
             return AppConfigurationImpl(
                 appId = appId,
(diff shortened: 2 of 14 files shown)


packages/cinterop/src/commonMain/kotlin/io/realm/internal/util/PlatformUtils.kt
```diff
@@ -0,0 +1,26 @@
...
+package io.realm.internal.util
+
+/**
+ * Method to freeze state.
+ * Calls the platform implementation of 'freeze' on native, and is a noop on other platforms.
+ *
+ * Note, this method refers to Kotlin Natives notion of frozen objects, and not Realms variant
+ * of frozen objects.
+ */
+expect fun <T> T.freezeOnOldMM(): T
```
cmelchior, line 26: We already have `io.realm.internal.platform.SystemUtils` which has these kind of methods. I would move it there instead of creating another file.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
```diff
@@ -141,7 +141,7 @@ interface AppConfiguration {
                         appLogger.debug(message)
                     }
                 }
-            ).freeze() // Kotlin network client needs to be frozen before passed to the C-API
+            ).freeze() // MM: Catches userLoggers // Kotlin network client needs to be frozen before passed to the C-API
```
cmelchior, line 144: Not sure I understand this comment?

cmelchior: Should this use the new freeze method?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/EmailPasswordAuth.kt
```diff
@@ -33,7 +33,7 @@ class EmailPasswordAuth(
                 Validation.checkEmpty(password, "password"),
                 channelResultCallback<Unit, Unit>(channel) {
                     // No-op
-                }.freeze()
+                }.freeze() // MM: Doesn't catch user references
```
cmelchior, line 36: Not sure I understand the comment?

cmelchior (changes requested):
I have a few comments and questions:

- Is this being tested somehow? I assume not (outside manual testing)? But I guess there is a question about how many different memory models we want to run tests for.
- Some of the comments could be expanded a bit. At least I don't understand some of the MM comments.

cmelchior:
Ah sorry. Missed this was still a draft 🙈 ... Use the comments for what you want.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/util/PlatformUtils.kt
rorbech: The `SystemUtils` live in `library-base` but we need the `freezeOnOldMM` to be available in `cinterop`. I could move all the `SystemUtils` to `cinterop` if you'd like?

rorbech:
> Ah sorry. Missed this was still a draft 🙈 ... Use the comments for what you want.

Yeah. I didn't settle on whether we should use the new conditional freeze on all our `freeze` calls. Some of them are actually catching/freezing user provided data. If we used the new variant that only freezes data on the old memory model people could just opt into the new one if it caused troubles for them. So I guess best thing would just be to use the new freeze-variant all over, but wasn't completely ready to do that when I did the PR. Any thoughts?

rorbech:
> * Is this being tested somehow? I assume not (outside manual testing)? But I guess there is a question about how many different memory models we want to run tests for.

We do a full test suite run with the new memory model, but the interesting parameter is actually the coroutine version. This PR enables us to run with Coroutines 1.6.0's _freeze transparency_ without adding the `kotlin.native.binary.freezing=disabled` runtime flag (still required for sync as ktor also needs it) because we _no-op_ the freeze outselves.

cmelchior:
Superseded by https://github.com/realm/realm-kotlin/pull/1059

Issue #619, opened 2021-12-25
Failed to compile on Kotlin 1.6.20


pduy99:
My project was built and run successfully with Kotlin 1.6.10 (Realm 0.9.0-SNAPSHOT). However, when I upgrade Kotlin to version 1.6.20-dev-7317, Realm cannot be compiled. I have opened an [issue](https://youtrack.jetbrains.com/issue/KT-50506) on Kotlin Youtrack and got an answer that this problem seems to be Realm's compiler. 

stack trace:

> e: org.jetbrains.kotlin.backend.common.CompilationException: Back-end: Please report this problem https://kotl.in/issue
/Users/141339/Helios/BLE_SDK/sdk/src/commonMain/kotlin/com/fossil/sdk/ble/database/entity/FileItem.kt:-1:-1
Details: Internal error in file lowering: java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.ir.expressions.impl.IrSetFieldImpl org.jetbrains.kotlin.ir.builders.ExpressionHelpersKt.irSetField(org.jetbrains.kotlin.ir.builders.IrBuilderWithScope, org.jetbrains.kotlin.ir.expressions.IrExpression, org.jetbrains.kotlin.ir.declarations.IrField, org.jetbrains.kotlin.ir.expressions.IrExpression)'
	at io.realm.compiler.RealmModelSyntheticPropertiesGeneration.addVariableProperty(RealmModelSyntheticPropertiesGeneration.kt:586)
	at io.realm.compiler.RealmModelSyntheticPropertiesGeneration.addProperties(RealmModelSyntheticPropertiesGeneration.kt:139)
	at io.realm.compiler.RealmModelLowering.lower(RealmModelLoweringExtension.kt:104)
	at org.jetbrains.kotlin.backend.common.ClassLoweringVisitor.visitClass(Lower.kt:100)
	at org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid$DefaultImpls.visitClass(IrElementVisitorVoid.kt:44)
	at org.jetbrains.kotlin.backend.common.ClassLoweringVisitor.visitClass(Lower.kt:91)
	at org.jetbrains.kotlin.backend.common.ClassLoweringVisitor.visitClass(Lower.kt:91)
	at org.jetbrains.kotlin.ir.declarations.IrClass.accept(IrClass.kt:57)
[...]

PR #666, opened 2022-02-04
Refactor the compiler plugin to avoid breaking API in 1.6.20

nhachicha:
Some compiler API have changed in 1.6.20 (example https://github.com/realm/realm-kotlin/issues/619#issuecomment-1029962450) we can avoid bumping to `1.6.20` and still be compatible with `1.6.20` with the following refactors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,6 @@
 * Minimum Android SDK: 16.
 
 ### Internal
-* Updated to Gradle 7.3.3.
-* Updated to Android Gradle Plugin 7.1.0.
-* Updated to AndroidX JUnit 1.1.3.
-* Updated to AndroidX Test 1.4.0.
 
 
 ## 0.9.0 (2022-01-28)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Enhancements
 
 ### Fixed
-* Refactoring the compiler plugin to use API available in Kotlin `1.6.20` fixes ([#619](https://github.com/realm/realm-kotlin/issues/619)).
+* Refactor the compiler plugin to use API's compatible with Kotlin `1.6.20`. (Issue ([#619](https://github.com/realm/realm-kotlin/issues/619)).
 
 
 ### Compatibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,29 @@
+## 0.10.0-SNAPSHOT (2022-MM-DD)
+
+### Breaking Changes
+
+### Enhancements
+
+### Fixed
+* Refactoring the compiler plugin to use API available in Kotlin `1.6.20` fixes ([#619](https://github.com/realm/realm-kotlin/issues/619)).
+
+
+### Compatibility
+* This release is compatible with:
+  * Kotlin 1.6.10.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
+## 0.10.0-SNAPSHOT (2022-MM-DD)
+
+### Breaking Changes
+
+### Enhancements
+
+### Fixed
+* Refactoring the compiler plugin to use API available in Kotlin `1.6.20` fixes ([#619](https://github.com/realm/realm-kotlin/issues/619)).
```
```suggestion
* Refactor the compiler plugin to use API's compatible with Kotlin `1.6.20`. (Issue ([#619](https://github.com/realm/realm-kotlin/issues/619)).
```

CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
...
+* This release is compatible with:
+  * Kotlin 1.6.10.
+  * Coroutines 1.6.0-native-mt. Also compatible with Coroutines 1.6.0 but requires enabling of the new memory model and disabling of freezing, see https://github.com/realm/realm-kotlin#kotlin-memory-model-and-coroutine-compatibility for details on that.
+  * AtomicFu 0.17.0.
+* Minimum Gradle version: 6.1.1.  
+* Minimum Android Gradle Plugin version: 4.0.0.
+* Minimum Android SDK: 16.
+
+### Internal
+* Updated to Gradle 7.3.3.
```
cmelchior, line 21: These 4 lines are for the 0.9.0 release and should be deleted.

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -479,10 +481,13 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
                                     // For managed property call C-Interop function
                                     cinteropCall,
                                     // For unmanaged property set backing field
-                                    irSetField(
-                                        irGet(receiver),
-                                        backingField,
-                                        irGet(setter.valueParameters.first())
+                                    IrSetFieldImpl(
```
cmelchior, line 484: Do we need to use the `IrSetFieldImpl` here instead of the `irSetField`. It feels like we are accessing something we perhaps shouldn't? 🤔

cmelchior (approved):

packages/plugin-compiler/src/main/kotlin/io/realm/compiler/AccessorModifierIrGeneration.kt
nhachicha: `irSetField` [signature](https://github.com/JetBrains/kotlin/blob/dump-model-compiler-key-1.6.10/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt#L144) changed in [1.6.20](https://github.com/JetBrains/kotlin/blob/1.6.20/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/builders/ExpressionHelpers.kt#L146) The only workaround that will avoid us to bump to `1.6.20` to use the new signature is to call `IrSetFieldImpl` directly

PR #667, opened 2022-02-04
Enable Explicit API mode

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/22

I found few TODO that should be fixed in other PR's.

* `Flowable` interface is placed in internal but exposed in the public API. We should move it rework how it is used.
* `WeakReference` seems to trigger a bug between strict mode detection, expect/actual, and type aliases. It is marked public for now, but placed in the `internal` package.


packages/library-base/src/commonMain/kotlin/io/realm/internal/MutableRealmImpl.kt
```diff
@@ -122,24 +124,34 @@ internal class MutableRealmImpl : BaseRealmImpl, MutableRealm {
...
+    override fun <T : RealmObject> registerObjectChangeListener(
         obj: T,
         callback: Callback<T?>
     ): Cancellable {
         throw IllegalStateException("Changes to RealmResults cannot be observed during a write.")
     }
+
+    // Required as Kotlin otherwise gets confused about the visibility and reports
+    // "Cannot infer visibility for '...'. Please specify it explicitly"
+    override fun <T : RealmObject> query(
```
nhachicha, line 150: Is there any test using/covering this method?

nhachicha (approved):
👍 minor stuff

packages/library-base/build.gradle.kts
```diff
@@ -135,6 +135,12 @@ kotlin {
 //                .all { onlyIf { findProperty("isMainHost") == "true" } }
 //        }
 //    }
+
+    // Require that all methods in the API have visibility modifiers and return types.
+    // Anything inside `io.realm.internal.*` is considered internal regardless of their
+    // visibility modifier and will be stripped from Dokka, but will unfortunately still
+    // leak into auto-complete in the IDE.
```
rorbech, line 142: That is weird because I specifically tested this and it wasn't the case. Maybe this has changed in the past year or maybe my experiments was contaminated by some other KMP/gradle issues (not that we haven't seen red lines, etc. over the past year). We can maybe `@Suppress("invisible_reference", "invisible_member")` our way through internalizing more symbols from `library-base` even though they are used in `library-sync`, but I think there are issues if the type is extended.

packages/library-base/src/commonMain/kotlin/io/realm/internal/platform/WeakReference.kt
```diff
@@ -5,7 +5,8 @@ package io.realm.internal.platform
  *
  * Basically just wrapping underlying platform implementations.
  */
-expect class WeakReference<T : Any>(referred: T) {
-    fun clear()
-    fun get(): T?
+// TODO This is currently public because making it public breaks something in Kotlins type check for Strict mode
```
rorbech, line 8:
```suggestion
// TODO This is currently public because making it internal breaks something in Kotlins type check for Strict mode
```

packages/library-base/src/commonMain/kotlin/io/realm/internal/platform/CoroutineUtils.kt
```diff
@@ -8,13 +8,13 @@ import kotlin.coroutines.EmptyCoroutineContext
 /**
  * Returns a default Realm write dispatcher for a Realm opened on the calling thread.
  */
-expect fun singleThreadDispatcher(id: String): CoroutineDispatcher
+public expect fun singleThreadDispatcher(id: String): CoroutineDispatcher
```
rorbech, line 11: Why are these public? Just due to test usage?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncConfigurationImpl.kt
```diff
@@ -26,7 +26,8 @@ import io.realm.mongodb.SyncConfiguration
 import io.realm.mongodb.SyncException
 import io.realm.mongodb.SyncSession
 
-internal class SyncConfigurationImpl(
+// TODO Public as it is used from the io.realm.mongodb.sync.SyncConfiguration interface
+public class SyncConfigurationImpl(
```
rorbech, line 30: But is it considered public when we are just calling the constructor? This sounds a bit weird, because then everything would need to be public as there is a public entry point for all internal things.

rorbech (approved):
LGTM I guess, though there are some single publicly exposed things that I cannot understand. But hard to question without experimenting myself. 

There have been some effort in exposing internal symbols to other compilation units (https://youtrack.jetbrains.com/issue/KTIJ-20112), but don't think they have a way to do it across project dependencies. 

I have sometimes wrapped some of our internal stuff in helper methods with `@Suppress("invisible_member", "invisible_reference")` to avoid exposing the symbols. This can be done at file level. I have tried to avoid that in tests, to ensure that we don't accidentally start using the whole internal API in the test, but maybe that would be a solution in `library-sync` when accessing some of the `library-base` symbols.

packages/library-base/build.gradle.kts
cmelchior: Not sure there are problems with extensions as such, but interface delegation across module boundaries is definitely broken. I should probably try to create a minimal project we can throw at JetBrains.

Alternatively we should rethink our current module setup in order to work around these issues. I'll create an issue that capture this.

packages/library-base/src/commonMain/kotlin/io/realm/internal/platform/CoroutineUtils.kt
cmelchior: Yes, they are used in enough places that using suppress everywhere seems wrong.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncConfigurationImpl.kt
cmelchior: Good point, fixed

PR #673, opened 2022-02-10
Support multiple constructors in model definition 

nhachicha:


packages/plugin-compiler/src/test/kotlin/io/realm/compiler/GenerationExtensionTest.kt
```diff
@@ -120,6 +120,16 @@ class GenerationExtensionTest {
         inputs.assertGeneratedIR()
     }
 
+    @Test
+    fun `multiple constructors in model definition`() {
```
rorbech, line 124: Since we are not modifying the IR it is maybe overkill to have a test comparing the IR dump. It is annoying to have to update too many files on other changes, so I wouldn't mind removing this.

rorbech (changes requested):
Nice, but lacking a test that shows that compilation fails when there is no no-arg constructor. Maybe put it in https://github.com/realm/realm-kotlin/tree/master/test/base/src/jvmTest/kotlin/io/realm/test/compiler rather than in the actual compiler plugin tests

rorbech (approved):

Issue #95, opened 2020-12-07
Support for deleting Realms


rorbech:
We need to be able to delete the actual files of a Realm. 

As of now the is no C-API for ensuring exclusive access to the files like the old 
```
DB::call_with_lock(realm_path, [&](std::string path)
```
from Java.

cmelchior:
Added https://github.com/realm/realm-core/issues/5058

fealebenpae:
Please see https://github.com/realm/realm-core/pull/5127.

PR #675, opened 2022-02-11
Added `deleteRealm` function

edualonso:
Fixes https://github.com/realm/realm-kotlin/issues/95.

### DIFF
--- a/packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
@@ -162,6 +162,19 @@ actual object RealmInterop {
         realmc.realm_close((realm as LongPointerWrapper).ptr)
     }
 
+    actual fun realm_delete_files(path: String): Boolean {
+        val deleted = booleanArrayOf(false)
+        realmc.realm_delete_files(path, deleted)
+
+        // TODO error handling
+        if (!deleted[0]) {
+            throw IllegalStateException("It's not allowed to delete the file associated with an open Realm. " +
+                "Remember to call 'close()' on all the instances of the Realm before deleting its file: "
+                + path);
+        }
+        return deleted[0]
+    }
+
     actual fun realm_schema_validate(schema: NativePointer, mode: SchemaValidationMode): Boolean {
         return realmc.realm_schema_validate((schema as LongPointerWrapper).ptr, mode.nativeValue.toLong())
     }
--- a/packages/jni-swig-stub/realm.i
+++ b/packages/jni-swig-stub/realm.i
@@ -184,7 +184,7 @@ bool throw_as_java_exception(JNIEnv *jenv) {
 %apply int64_t[] { size_t* };
 
 // bool output parameter
-%apply bool* OUTPUT { bool* out_found };
+%apply bool* OUTPUT { bool* out_found, bool* did_delete_realm };
 
 // uint64_t output parameter for realm_get_num_versions
 %apply int64_t* OUTPUT { uint64_t* out_versions_count };
--- a/packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
@@ -18,6 +18,7 @@ package io.realm
 import io.realm.internal.InternalConfiguration
 import io.realm.internal.RealmImpl
 import io.realm.internal.interop.Constants
+import io.realm.internal.interop.RealmInterop
 import io.realm.query.RealmQuery
 import kotlinx.coroutines.flow.Flow
 import kotlin.reflect.KClass
@@ -56,9 +57,32 @@ interface Realm : TypedRealm {
          *
          * @throws IllegalArgumentException on invalid Realm configurations.
          */
-        fun open(configuration: Configuration): Realm {
-            return RealmImpl(configuration as InternalConfiguration)
-        }
+        fun open(configuration: Configuration): Realm =
+            RealmImpl(configuration as InternalConfiguration)
+
+        /**
+         * TODO revisit docs
+         *
+         * Deletes the Realm file along with the related temporary files specified by the given
+         * [RealmConfiguration] from the filesystem. The temporary file with ".lock" extension won't be
+         * deleted.
+         *
+         * All Realm instances must be closed before calling this method.
+         *
+         * TODO does this still apply?
+         * WARNING: For synchronized Realm, there is a chance that an internal Realm instance on the
+         * background thread is not closed even all the user controlled Realm instances are closed.
+         * This will result an `IllegalStateException`. See issue
+         * https://github.com/realm/realm-java/issues/5416 for more details.
+         *
+         * @param configuration a [RealmConfiguration].
+         * @return `false` if the Realm file could not be deleted. Temporary files deletion failure
+         * won't impact the return value. All of the failing file deletions will be logged.
+         * @throws IllegalStateException if there are open Realm instances on other threads or other
+         * processes.
+         */
+        fun deleteRealm(configuration: RealmConfiguration): Boolean =
+            RealmInterop.realm_delete_files(configuration.path)
     }
 
     /**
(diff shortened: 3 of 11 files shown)


test/base/src/androidTest/kotlin/io/realm/test/shared/RealmTests.kt
```diff
@@ -478,6 +482,114 @@ class RealmTests {
...
+            .path("$tmpDir/default.realm")
+            .build()
+        val tempDirB = PlatformUtils.createTempDir()
+        val pathB = "$tempDirB/yetAnotherRealm.realm"
+        val configB = RealmConfiguration.Builder(schema = setOf(Parent::class, Child::class))
+            .path(pathB)
+            .build()
+
+        // This instance is already cached because of the setUp() method so this deletion should throw.
+        try {
```
cmelchior, line 566: There is `assertFailsWith<IllegalStateException>` which you can use instead of the try-catch

cmelchior (changes requested):
Nice, I had few comments here and there, but nothing too major

packages/jni-swig-stub/realm.i
```diff
@@ -184,7 +184,7 @@ bool throw_as_java_exception(JNIEnv *jenv) {
 %apply int64_t[] { size_t* };
 
 // bool output parameter
-%apply bool* OUTPUT { bool* out_found };
+%apply bool* OUTPUT { bool* out_found, bool* did_delete_realm };
```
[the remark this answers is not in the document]
edualonso: I think what this typemap does is to convert all specified variable names to `bool*`. I haven't seen issues other places after introducing this change so I assume it is the correct way - see http://www.swig.org/Doc4.0/Typemaps.html#Typemaps_nn48

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -56,9 +59,36 @@ interface Realm : TypedRealm {
...
+         *
+         * **WARNING**: For synchronized realms there is a chance that an internal Realm instance on
+         * the background thread is not closed even though the user controlled Realm instances are
+         * closed. This will result an `IllegalStateException`. See issue
+         * https://github.com/realm/realm-java/issues/5416 for more details.
+         *
+         * @param configuration a [RealmConfiguration].
+         * @return `false` if the Realm file could not be deleted. Temporary files deletion failure
+         * won't impact the return value. All of the failing file deletions will be logged.
+         * @throws IllegalStateException if there are open Realm instances on other threads or other
```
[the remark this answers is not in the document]
edualonso: Sure. I've also removed the `return` part as the C-API throws in case there has been an error, returning `true` if all went well.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmTests.kt
```diff
@@ -478,6 +482,114 @@ class RealmTests {
...
+            }
+
+            testRealm.writeBlocking {
+                copyToRealm(Child())
+            }
+
+            // Waits for bg thread's opening the same Realm.
+            bgThreadReadyChannel.receive()
+
+            // A core upgrade might change the location of the files
```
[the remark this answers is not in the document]
edualonso: Legacy stuff. I've removed it.

packages/jni-swig-stub/realm.i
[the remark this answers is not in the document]
rorbech: This is not a multi argument typemap, but just applying the typemap to multiple different _selectors_. When you specify a variable name as part of the type you are matching the typemap will only be applied to variables of that name instead of all usages of the specific type. ... so seems right to me.

packages/jni-swig-stub/realm.i
```diff
@@ -239,6 +239,7 @@ bool throw_as_java_exception(JNIEnv *jenv) {
 
 // bool output parameter
 %apply bool* OUTPUT { bool* out_found, bool* did_create };
+%apply bool* OUTPUT { bool* out_found, bool* did_delete_realm };
```
rorbech: You could collapse this with the line before
```
%apply bool* OUTPUT { bool* out_found, bool* did_create, bool* did_delete_realm };
```

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -190,6 +190,15 @@ actual object RealmInterop {
         realmc.realm_close((realm as LongPointerWrapper).ptr)
     }
 
+    actual fun realm_delete_files(path: String) {
+        val deleted = booleanArrayOf(false)
+        realmc.realm_delete_files(path, deleted)
+
+        if (!deleted[0]) {
```
rorbech, line 197: Seems like the implementation differs on the platforms. If we throw on `!deleted[0]` we should probably also checked `deleted` on Kotlin Native!

packages/library-base/src/commonMain/kotlin/io/realm/Realm.kt
```diff
@@ -75,6 +78,32 @@ public interface Realm : TypedRealm {
         public fun open(configuration: Configuration): Realm {
             return RealmImpl(configuration as InternalConfiguration)
         }
+
+        /**
+         * Deletes the Realm file along with the related temporary files specified by the given
```
rorbech, line 83:
```suggestion
         * Deletes the realm file along with other related temporary files specified by the given
```

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmTests.kt
```diff
@@ -402,6 +406,93 @@ class RealmTests {
...
+    @Test
+    @Suppress("LongMethod")
+    fun deleteRealm() {
+        val fileSystem = FileSystem.SYSTEM
+        val testDir = PlatformUtils.createTempDir("test_dir")
+        val testDirPath = testDir.toPath()
+        assertTrue(fileSystem.exists(testDirPath))
+
+        val configuration = RealmConfiguration.Builder(schema = setOf(Parent::class, Child::class))
+            .path("$testDir/default.realm")
```
cmelchior, line 418: This shouldn't compile anymore? 🤔 
```suggestion
            .directory(testDir)
```

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmTests.kt
```diff
@@ -402,6 +406,93 @@ class RealmTests {
...
+            closedChannel.close()
+        }
+    }
+
+    @Test
+    fun deleteRealm_failures() {
+        val tempDirA = PlatformUtils.createTempDir()
+
+        val configA = RealmConfiguration.Builder(schema = setOf(Parent::class, Child::class))
+            .path("$tempDirA/anotherRealm.realm")
```
cmelchior, line 476:
```suggestion
            .directory(tempDirA).
            .name("anotherRealm.realm")
```

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmTests.kt
```diff
@@ -402,6 +406,93 @@ class RealmTests {
...
+            .build()
+
+        // Creates a new Realm file.
+        val anotherRealm = Realm.open(configA)
+
+        // Deleting it without having closed it should fail.
+        assertFailsWith<IllegalStateException> {
```
rorbech, line 483: Maybe assert on the actual message too with `assertFailsWith` because the implementation seems to differ on the two platforms and would actually yield different exceptions.

rorbech (commented):
Really good catch with `readdir$INODE64` 🎉 

There is differing implementation across platform that I think should be fixed.

cmelchior (approved):
If CI is happy

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 * Add support for data migration as part of an automatic schema upgrade through `RealmConfiguration.Builder.migration(RealmMigration)` (Issue [#87](https://github.com/realm/realm-kotlin/issues/87))
 * Added ability to delete objects specified by a `RealmQuery` or `RealmResults` through `MutableRealm.delete(...)` and `DynamicMutableRealm.delete(...).
 * Add support for updating existing objects through `copyToRealm`. This requires them having a primary key. (Issue [#564](https://github.com/realm/realm-kotlin/issues/564))
+* Added `Realm.deleteRealm(RealmConfiguration)` function (Issue [#95](https://github.com/realm/realm-kotlin/issues/95)).
```
clementetb, line 25:
```suggestion
* Added `Realm.deleteRealm(RealmConfiguration)` function that deletes the Realm files from the filesystem (Issue [#95](https://github.com/realm/realm-kotlin/issues/95)).
```

clementetb (approved):
LGTM 💎

rorbech (approved):

PR #679, opened 2022-02-12
Add Android x86 and armeabi-v7a

nhachicha:


packages/jni-swig-stub/realm.i
```diff
@@ -222,18 +222,22 @@ bool throw_as_java_exception(JNIEnv *jenv) {
 // Type map for int64_t has an erroneous cast, don't know how to fix it except with this
 %typemap(in) void** ( jlong *jarr ){
     // Original
-    %#if defined(__ANDROID__)
+    %#if defined(__ANDROID__) && defined(__aarch64__) // Android arm64-v8a
```
cmelchior, line 225: This is also x86_64, right?

packages/jni-swig-stub/realm.i
```diff
@@ -222,18 +222,22 @@ bool throw_as_java_exception(JNIEnv *jenv) {
...
         if (!SWIG_JavaArrayInLonglong(jenv, &jarr, (long **)&$1, $input)) return $null;
-    %#elif defined(__aarch64__)
+    %#elif defined(__ANDROID__) // Android armeabi-v7a
```
cmelchior, line 227: Also x86?

cmelchior (approved):
Nice 💯 . Only thing seems to be a missing changelog about 32-bit support for Android (only).

packages/jni-swig-stub/realm.i
nhachicha: no `armeabi-v7a, x86_64 and x86` is handled in the next branch

packages/jni-swig-stub/realm.i
nhachicha: yes as well as `x86_64`

PR #681, opened 2022-02-15
Downgrade to Gradle 7.2

cmelchior:
This should allow us to run macOS tests from Android Studio again. Also created https://youtrack.jetbrains.com/issue/KT-51325 to track this.


cmelchior:
Root cause seems to be fixed by this: https://github.com/gradle/gradle/pull/19908

nhachicha (approved):

PR #682, opened 2022-02-15
Fine grained notifications on single object queries.

clementetb:
With this PR we allow to observe `ObjectChange` events on flows from single object queries.

It is achieved by mapping the query `RealmResults` flow with the fine grained flows of the head of that list on that state.

We have added support for cancelling a flow, a flow can be casted into a `Cancellable`. It is required to stop observing object events when the head list changes because of a replacement, object flows only cancels automatically when the object gets deleted.

Realm flows are now cancellable because object flows only get cancel automatically when the object gets deleted and we needed a way to stop the flow when the query head gets replaced.

The new flow lifecycle is of the following:
- If subscribed on an empty query the flow will not yield any event until an element is added, then it would yield an `InitialObject` event for the first element. On a non-empty list it would start with an `InitialObject` event for its first element.
- Once subscribed and the `InitialObject` event is observed, sequential `UpdatedObject` instances would be observed if the first element is modified. If the element is deleted a `DeletedObject` would be yield.
- If the first element is replaced with a new value, an `InitialObject` would be yield for the new head, and would be follow with `UpdatedObject` on all its changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add support for fine grained notification on Realm objects. `RealmObject.asFlow()` yields `ObjectChange` that represent the `ObjectInitial`, `ObjectUpdated` or `ObjectDeleted` states.
 * Add support for fine grained notification on Realm lists. `RealmList.asFlow()` yields `ListChange` that represent the `InitialList`, `UpdatedList` or `DeletedList` states.
 * Add support for fine grained notification on Realm results. `RealmResults.asFlow()` yields `ListChange` that represent the `InitialList`, `UpdatedList` or `DeletedList` states.
+* Add support for fine grained notification on RealmSingleQuery. `RealmSingleQuery.asFlow()` yields `ObjectChange` that represent the events might happen on the first element of the query.
 
 ### Fixed
 * Refactor the compiler plugin to use API's compatible with Kotlin `1.6.20`. (Issue ([#619](https://github.com/realm/realm-kotlin/issues/619)).
--- a/packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
@@ -9,6 +9,7 @@ import io.realm.notifications.Callback
 import io.realm.notifications.Cancellable
 import kotlinx.atomicfu.AtomicRef
 import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.channels.ChannelResult
 import kotlinx.coroutines.channels.awaitClose
@@ -92,46 +93,58 @@ internal class SuspendableNotifier(
     }
 
     internal fun <T, C> registerObserver(thawableObservable: Thawable<Observable<T, C>>): Flow<C> {
-        return callbackFlow {
-            val token: AtomicRef<Cancellable> = kotlinx.atomicfu.atomic(NO_OP_NOTIFICATION_TOKEN)
-            withContext(dispatcher) {
-                ensureActive()
-                val liveRef: Observable<T, C> = thawableObservable.thaw(realm.realmReference)
-                    ?: error("Cannot listen for changes on a deleted reference")
-                val interopCallback: io.realm.internal.interop.Callback =
-                    object : io.realm.internal.interop.Callback {
-                        override fun onChange(change: NativePointer) {
-                            // FIXME How to make sure the Realm isn't closed when handling this?
+        var cancelCallback: () -> Unit = {}
 
-                            // FIXME The Realm should have been frozen in `realmChanged`, but this isn't supported yet.
-                            //  Instead we create the frozen version ourselves (which is correct, but pretty inefficient)
-                            //  We also send it to the owner Realm, so it can keep track of its lifecycle
-                            val frozenRealm = RealmReference(
-                                owner,
-                                RealmInterop.realm_freeze(realm.realmReference.dbPointer)
-                            )
-                            notifyRealmChanged(frozenRealm)
-                            // Notifications need to be delivered with the version they where created on, otherwise
-                            // the fine-grained notification data might be out of sync.
-                            liveRef.emitFrozenUpdate(
-                                frozenRealm,
-                                change,
-                                this@callbackFlow
-                            )?.let {
-                                checkResult(it)
+        return object :
+            Cancellable,
+            Flow<C> by callbackFlow({
+                cancelCallback = {
+                    cancel()
+                }
+                val token: AtomicRef<Cancellable> =
+                    kotlinx.atomicfu.atomic(NO_OP_NOTIFICATION_TOKEN)
+                withContext(dispatcher) {
+                    ensureActive()
+                    val liveRef: Observable<T, C> = thawableObservable.thaw(realm.realmReference)
+                        ?: error("Cannot listen for changes on a deleted reference")
+                    val interopCallback: io.realm.internal.interop.Callback =
+                        object : io.realm.internal.interop.Callback {
+                            override fun onChange(change: NativePointer) {
+                                // FIXME How to make sure the Realm isn't closed when handling this?
+
+                                // FIXME The Realm should have been frozen in `realmChanged`, but this isn't supported yet.
+                                //  Instead we create the frozen version ourselves (which is correct, but pretty inefficient)
+                                //  We also send it to the owner Realm, so it can keep track of its lifecycle
+                                val frozenRealm = RealmReference(
+                                    owner,
+                                    RealmInterop.realm_freeze(realm.realmReference.dbPointer)
+                                )
+                                notifyRealmChanged(frozenRealm)
+                                // Notifications need to be delivered with the version they where created on, otherwise
+                                // the fine-grained notification data might be out of sync.
+                                liveRef.emitFrozenUpdate(
+                                    frozenRealm,
+                                    change,
+                                    this@callbackFlow
+                                )?.let {
+                                    checkResult(it)
+                                }
                             }
-                        }
-                    }.freeze<io.realm.internal.interop.Callback>() // Freeze to allow cleaning up on another thread
-                val newToken =
-                    NotificationToken<Callback<T>>(
-                        // FIXME What is this callback for anyway?
-                        callback = Callback { _, _ -> /* Do nothing */ },
-                        token = liveRef.registerForNotification(interopCallback)
-                    )
-                token.value = newToken
-            }
-            awaitClose {
-                token.value.cancel()
+                        }.freeze<io.realm.internal.interop.Callback>() // Freeze to allow cleaning up on another thread
+                    val newToken =
+                        NotificationToken<Callback<T>>(
+                            // FIXME What is this callback for anyway?
+                            callback = Callback { _, _ -> /* Do nothing */ },
+                            token = liveRef.registerForNotification(interopCallback)
+                        )
+                    token.value = newToken
+                }
+                awaitClose {
+                    token.value.cancel()
+                }
+            }) {
+            override fun cancel() {
+                cancelCallback()
             }
         }
     }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/internal/query/SingleQuery.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/internal/query/SingleQuery.kt
@@ -1,6 +1,8 @@
 package io.realm.internal.query
 
 import io.realm.RealmObject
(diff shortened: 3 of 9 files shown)


packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -92,46 +93,58 @@ internal class SuspendableNotifier(
...
+            Flow<C> by callbackFlow({
+                cancelCallback = {
+                    cancel()
+                }
+                val token: AtomicRef<Cancellable> =
+                    kotlinx.atomicfu.atomic(NO_OP_NOTIFICATION_TOKEN)
+                withContext(dispatcher) {
+                    ensureActive()
+                    val liveRef: Observable<T, C> = thawableObservable.thaw(realm.realmReference)
+                        ?: error("Cannot listen for changes on a deleted reference")
```
cmelchior, line 109:
```suggestion
                        ?: error("Cannot listen for changes on a deleted Realm reference")
```

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -92,46 +93,58 @@ internal class SuspendableNotifier(
...
+                    val interopCallback: io.realm.internal.interop.Callback =
+                        object : io.realm.internal.interop.Callback {
+                            override fun onChange(change: NativePointer) {
+                                // FIXME How to make sure the Realm isn't closed when handling this?
+
+                                // FIXME The Realm should have been frozen in `realmChanged`, but this isn't supported yet.
```
cmelchior, line 115: I thought this was merged as part of the schema PR? https://github.com/realm/realm-kotlin/pull/539

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/SingleQuery.kt
```diff
@@ -30,11 +36,32 @@ internal class SingleQuery<E : RealmObject> constructor(
...
+                // Denotes when to subscribe to the next object:
+                // A change on the head of the list, and it is not the same as the previous head
+                val newHead: E? = listChange.list.firstOrNull()
+
+                (newHead != null && !newHead.hasSameObjectKey(head)).also {
+                    head = newHead
+                }
+            }.flatMapMerge { listChange ->
+                // Don't close the flow if the head was removed
+                if (listChange !is UpdatedList<*> || !listChange.deletions.contains(0))
```
cmelchior, line 56: Given the logic above, we only get here if the event is a `UpdatedList` event, correct? So perhaps it would be cleaner to just turn the first if into an assert `val change = listChange as UpdatedList` (or something).

Since this is a query result, we should never see a `DeletedList` I think?

cmelchior: This actually reminds me, that at some point I was considering if we should have separate type hierarchies for `List` and `Query` since `DeleteList` doesn't really make sense for queries 🤔

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
cmelchior: Although, I guess you just moved code around, so this should probably be addressed in another PR?

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -246,15 +250,27 @@ interface RealmSingleQuery<T : RealmObject> {
      * Finds the first object that fulfills the query conditions and returns it asynchronously as a
      * [Flow].
      *
-     * If there is any changes to the object represented by the query, the flow will emit the
-     * updated object. The flow will continue running indefinitely until canceled.
+     * If there is any changes to the first object represented by the query, the flow will emit an
+     * [ObjectChange] event depending on the query state. The flow will continue running indefinitely until
+     * canceled.
+     *
+     * If subscribed on an empty query the flow will not yield any event until an element is added, then
```
cmelchior, line 257: Hmm, I'm starting to think that we might have to rethink semantics around query types and data model types....I should probably have thought of this sooner 🤕 

We have two types of flows on queries:

1) On RealmResults
2) On RealmSingleQuery

Both of these differ from the states we assigned to objects, lists and eventually map and set.

* For RealmResults, the difference is that the deleted state doesn't really make sense
* For SingleQuery, the difference is that the initial state might be empty.

We can probably reuse the types we already have (which is what you are doing), but I'm beginning to wonder if that is too annoying to consume or we should introduce two new Change types: `QueryChange` and `ObjectQueryChange` (or something like that)....not sure if that is something you have thought about?

cmelchior (commented):
While reviewing this I started to think about the overall design...left some comments we should probably talk about.

packages/library-base/src/commonMain/kotlin/io/realm/internal/query/SingleQuery.kt
```diff
@@ -30,11 +39,49 @@ internal class SingleQuery<E : RealmObject> constructor(
         return model as E
     }
 
-    override fun asFlow(): Flow<E?> {
+    /**
+     * Because Core does not support subscribing for the head element of a query this feature
```
cmelchior, line 43:
```suggestion
     * Because Core does not support subscribing to the head element of a query this feature
```

packages/library-base/src/commonMain/kotlin/io/realm/notifications/ObjectChange.kt
```diff
@@ -17,6 +17,23 @@
 package io.realm.notifications
 
 import io.realm.RealmObject
+import io.realm.query.RealmSingleQuery
+
+/**
+ * This sealed class describe the possible changes that can be observed on a [RealmSingleQuery] flow.
+ */
+sealed interface QueryObjectChange<O : RealmObject> {
```
cmelchior, line 25: `QueryObjectChange` sounds a bit strange to me, but not sure that `ObjectQueryChange` is any better and prefixing `ObjectChange` with `Query<X>` might make sense 🤔

packages/library-base/src/commonMain/kotlin/io/realm/notifications/ObjectChange.kt
```diff
@@ -47,12 +64,12 @@ import io.realm.RealmObject
  *
  * For state of update changes, a list with the updated field names from the previous version is provided.
  */
-sealed interface ObjectChange<O : RealmObject> {
+sealed interface ObjectChange<O : RealmObject> : QueryObjectChange<O> {
```
cmelchior, line 67: This seems problematic? Doesn't this force people using `ObjectChange` to also reason about `PendingObject` now?

packages/library-base/src/commonMain/kotlin/io/realm/query/RealmQuery.kt
```diff
@@ -248,15 +253,42 @@ interface RealmSingleQuery<T : RealmObject> {
...
+     *                │ start ├───────────────────────────────┐
+     *                └───┬───┘       [InitialObject]         │
+     *                    │                                   │
+     *    [PendingObject] │                                   │
+     *                    │                                   │
+     *              ┌─────▼─────┐     [InitialObject]    ┌────▼───┐
+     *              │           ├────────────────────────►        ├─────┐[InitialObject]
+     *              │  No Head  │                        │  Head  │     │
+     *              │           ◄────────────────────────┤        │◄────┘[UpdatedObject]
+     *              └───────────┘     [DeletedObject]    └────────┘
```
cmelchior, line 283:
```suggestion
     * If subscribed on an empty query the flow will emit a [PendingObject] event to signal the query
     * is empty, it would then yield an [InitialObject] event for the first element. On a non-empty
     * query result it would start emitting an [InitialObject] event for its first element.
     *
     * Once subscribed and the [InitialObject] event is observed, sequential [UpdatedObject] instances
     * would be observed if the first element is modified. If the element is deleted a [DeletedObject]
     * would be observed.
     *
     * If the first element is replaced with a new object, an [InitialObject] would be sent for the new
     * head, which would be followed by [UpdatedObject] on all subsequent changes to that object.
     *
     * The state machine looks like this:
     *
     *                ┌───────┐
     *                │ start ├───────────────────────────────┐
     *                └───┬───┘       [InitialObject]         │
     *                    │                                   │
     *    [PendingObject] │                                   │
     *                    │                                   │
     *              ┌─────▼─────┐     [InitialObject]    ┌────▼───┐
     *              │           ├────────────────────────►        ├─────┐[InitialObject]
     *              │  No Head  │                        │  Head  │     │
     *              │           ◄────────────────────────┤        │◄────┘[UpdatedObject]
     *              └───────────┘     [DeletedObject]    └────────┘
```

cmelchior: Having the state machine in the doc is really awesome, but I'm getting very confused that you have the states as transitions in the diagram instead of them being states... IMO we shouldn't emphasize the underlying implementation that much but focus on the visible user-facing behavior.

cmelchior (commented):
I think the seperation into `ObjectChange` and `QueryObjectChange` is good, but I think this implementation affect the Realm Object notifications the wrong way?

Apart from that, I only had minor comments.

packages/library-base/src/commonMain/kotlin/io/realm/notifications/ObjectChange.kt
cmelchior: I missed some nuances around how this works. This seems fine 👍

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
clementetb: You couldn't see the changes from #539 because they weren't merged in here yet.

cmelchior (approved):
Yay 🚀 ..only had a few minor doc comments

CHANGELOG.md
```diff
@@ -8,7 +8,10 @@
 * Add support for fine grained notification on Realm instances. `Realm.asFlow()` yields `RealmChange` that represent the `RealmInitial` or `RealmUpdated` states.
 * Add support for fine grained notification on Realm objects. `RealmObject.asFlow()` yields `ObjectChange` that represent the `ObjectInitial`, `ObjectUpdated` or `ObjectDeleted` states.
 * Add support for fine grained notification on Realm lists. `RealmList.asFlow()` yields `ListChange` that represent the `InitialList`, `UpdatedList` or `DeletedList` states.
-* Add support for fine-grained notifications on Realm query results. `RealmResults.asFlow()` yields `ResultsChange` that represent the `InitialResults`, `UpdatedResults` states.
+* Add support for fine grained notification on Realm results. `RealmResults.asFlow()` yields `ListChange` that represent the `InitialList`, `UpdatedList` or `DeletedList` states.
+* Add support for fine-grained notifications on `RealmSingleQuery`. `RealmSingleQuery.asFlow()` yields `ObjectChange` events that represent changes that happen to the first result of the query. Unlike flows directly on Realm objects, deleting this object does not terminate the flow.
```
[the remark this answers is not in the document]
clementetb: It is a merge issue, they got mixed.

CHANGELOG.md
```diff
@@ -8,7 +8,10 @@
...
+* Add support for fine-grained notifications on Realm query results. `RealmResults.asFlow()` yields `ResultsChange` that represent the `InitialResults` or `UpdatedResults` states.
+* Add support for fine-grained notifications on `RealmSingleQuery`. `RealmSingleQuery.asFlow()` yields `SingleQueryChange` that represent the `PendingObject`, `ObjectInitial`, `ObjectUpdated` or `ObjectDeleted` states.
```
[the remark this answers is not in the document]
clementetb: Yes, it is a typo.

packages/library-base/src/commonMain/kotlin/io/realm/notifications/ObjectChange.kt
```diff
@@ -17,6 +17,39 @@
...
+ *                                   └─────────┬─────────┘
+ *                                ┌────────────┴───────────┐
+ *                         ┌──────▼───────┐        ┌───────▼───────┐
+ *                         │ ObjectChange │        │ PendingObject │
+ *                         └──────┬───────┘        └───────────────┘
+ *               ┌────────────────┼────────────────────┐
+ *      ┌────────▼──────┐  ┌──────▼────────┐  ┌────────▼──────┐
+ *      │ InitialObject │  │ UpdatedObject │  │ DeletedObject │
+ *      └───────────────┘  └───────────────┘  └───────────────┘
+ */
```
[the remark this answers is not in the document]
clementetb: You are right, it should be in a code block.

PR #689, opened 2022-02-22
Allow JVM lamdas and functional interfaces to be called from JNI

cmelchior:
Not 100% sure why this wasn't caught by our PR builds? 🤔


rorbech (approved):

nhachicha (approved):
We do have a step that tests a release APK, maybe we can update it to exercise similar code paths ...

Issue #691, opened 2022-02-22
IOS query with boolean as parameter crashes - no conversion for Boolean


promanowicz:
realm 0.9.0
kotlin 1.6.10
device: Iphone 6s
The following query crashesh app:
 query(clazz,"ObjectGuid = $0 AND Pending = $1",ObjectGuid!!,true).count().find()
 
 fix:
 query(clazz,"ObjectGuid = $0 AND Pending = TRUE",ObjectGuid!!).count().find()

cmelchior:
It looks like our conversion of input parameters is going wrong somewhere...I suspect it gets turned into a String when it shouldn't. We will investigate

This should also work btw:

```
query(clazz,"ObjectGuid = ${ObjectGuid!!} AND Pending = ${true}").count().find()
``

PR #695, opened 2022-02-24
Improve query testing

cmelchior:
Attempt to reproduce #691 

Also add support for a new `assertFailsWithMessage` which is a helper method that assert that a function throws with both a defined exception type AND message.

### DIFF
--- a/test/base/src/commonTest/kotlin/io/realm/test/Utils.kt
+++ b/test/base/src/commonTest/kotlin/io/realm/test/Utils.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2022 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


rorbech:
https://github.com/realm/realm-kotlin/pull/603#discussion_r771622012 😬

test/base/src/commonTest/kotlin/io/realm/test/Utils.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import kotlin.reflect.KClass
+import kotlin.test.assertFailsWith
+
+/**
+ * Assert that a statement fails with a specific Exception type AND message. The message match is
+ * fuzzy, i.e. we only check that the provided message is contained within the whole exception
+ * message. The match is case sensitive.
+ */
+fun <T : Throwable> assertFailsWithMessage(exceptionClass: KClass<T>, exceptionMessage: String, block: () -> Unit): T {
```
rorbech, line 26: Nice

rorbech (approved):

PR #696, opened 2022-02-24
Support for updates through copyToRealm

### LINKED ISSUE #564: Add `MutableRealm.copyToRealmOrUpdate()`
API similar to Realm Java.

We currently miss the ability to easily upsert objects. Which results in a lot of annoying code you have to write.  We also need to consider the `ImportFlag`  that allows for not re-setting values with the same value.

rorbech:
This PR adds ability to update existing objects through `copyToRealm`:

```
interface MutableRealm {
    /**
     * Update policy for imports with [copyToRealm].
     *
     * @see copyToRealm
     */
    // FIXME #naming
    enum class UpdatePolicy {
        /**
         * Update policy that causes import of an object with an existing primary key to fail.
         */
        ERROR,

        /**
         * Update policy that will update any existing objects identified with the same primary key.
         */
        ALL,
    }

    /**
     * Creates a copy or update existing objects in the Realm.
     *
     * This will recursively copy non-primary key objects and non-existing primary key objects into
     * the realm. The behavior of copying existing primary key objects will depend on the specified
     * update policy. Calling with [UpdatePolicy.ERROR] will cause creating of objects with existing
     * primary key to throw, while calling with [UpdatePolicy.ALL] will update existing primary key
     * object and all it's properties.
     *
     * Already managed update-to-date objects will not be copied but just return the instance
     * itself. Trying to copy outdated objects will throw an exception. To get hold of an updated
     * reference for an object use * [findLatest].
     *
     * @param instance the object to create a copy from.
     * @param updatePolicy update policy for the import.
     * @return the managed version of the `instance`.
     *
     * @throws IllegalArgumentException if the object graph of `instance` either contains a primary
     * key object that already exists and the update policy is [UpdatePolicy.ERROR] or if the object
     * graph contains an outdated object.
     */
    public fun <T : RealmObject> copyToRealm(instance: T, updatePolicy: UpdatePolicy = UpdatePolicy.ERROR): T

}
```

TODOs
- [x] Test of more advanced object graphs
- [x] Import flags

Closes #564


test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
```diff
@@ -61,13 +67,256 @@ class MutableRealmTests {
...
+        assertEquals(1, realm.query<SampleWithPrimaryKey>().find().size)
+    }
+
+    @Test
+    fun copyToRealm_updatePolicy_all() {
+        realm.writeBlocking {
+            val obj = StringPropertyWithPrimaryKey()
+            copyToRealm(obj.apply { id = "PRIMARY_KEY" })
+
+            obj.apply { this.value = "UPDATED_VALUE" }
```
cmelchior, line 105:
```suggestion
            obj.apply { value = "UPDATED_VALUE" }
```

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
```diff
@@ -61,13 +67,256 @@ class MutableRealmTests {
...
+
+    @Test
+    @Suppress("LongMethod")
+    fun copyToRealm_updatePolicy_all_allTypes() {
+        val sample = SampleWithPrimaryKey().apply {
+            primaryKey = 1
+        }
+        realm.writeBlocking { copyToRealm(sample) }
+
+        // TODO Vefiry that we cover all types
```
cmelchior, line 133:
```suggestion
        // TODO Verify that we cover all types
```

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
```diff
@@ -61,13 +67,256 @@ class MutableRealmTests {
...
+                assertEquals(2, primaryKey)
+                assertEquals("Four", stringField)
+            }
+        }
+    }
+
+    @Test
+    fun copyToRealm_updatePolicy_all_nonPrimaryKeyObject() {
+        realm.writeBlocking {
+            copyToRealm(Parent(), MutableRealm.UpdatePolicy.ALL)
```
cmelchior, line 249: Shouldn't we assert that the object is actually inserted?

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
```diff
@@ -61,13 +67,256 @@ class MutableRealmTests {
...
+        }
+    }
+
+    // The cache maintained during import doesn't recognize previously imported object
```
cmelchior, line 253: Shouldn't we track this with an issue? Sounds like bug we want to fix?

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
```diff
@@ -61,13 +67,256 @@ class MutableRealmTests {
...
+            }
+        }
+    }
+
+    @Test
+    fun copyToRealm_throwsWithDeletedObject() {
+        val frozenParent = realm.writeBlocking { copyToRealm(Parent()) }
+        realm.writeBlocking {
+            assertFailsWith<IllegalArgumentException> {
+                copyToRealm(frozenParent)
```
cmelchior, line 292: Object isn't deleted?

test/base/src/commonMain/kotlin/io/realm/entities/SampleWithPrimaryKey.kt
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2020 Realm Inc.
```
cmelchior, line 2:
```suggestion
 * Copyright 2022 Realm Inc.
```

cmelchior (changes requested):
Nice 👍 ... I had a few comments about the docs and a few of the tests...but otherwise I think this looks fine.

Personally, I think the naming used in the PR is fine, but we should probably get input from the rest of the team.

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -44,19 +44,45 @@ public interface MutableRealm : TypedRealm {
...
      * @param instance the object to create a copy from.
+     * @param updatePolicy update policy for the import.
      * @return the managed version of the `instance`.
      *
-     * @throws IllegalArgumentException if the class has a primary key field and an object with the same
-     * primary key already exists.
+     * @throws IllegalArgumentException if the object graph of `instance` either contains a primary
+     * key object that already exists and the update policy is [UpdatePolicy.ERROR] or if the object
+     * graph contains an outdated object.
      */
```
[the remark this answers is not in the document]
rorbech: I think that is a concern that we need to address generically, so don't think we can fix it here. I think users are bound to start reasoning about versions in general, so I would add anything to your suggestion.

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
rorbech: Test is actually overlapping with `copyToRealm_updatePolicy_all_noPrimaryKeyField` that checks insertion.

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -44,19 +44,45 @@ public interface MutableRealm : TypedRealm {
...
+     */
+    // FIXME #naming
+    public enum class UpdatePolicy {
+        /**
+         * Update policy that causes import of an object with an existing primary key to fail.
+         */
+        ERROR,
+
+        /**
+         * Update policy that will update any existing objects identified with the same primary key.
```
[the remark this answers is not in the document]
cmelchior: Ups....forgot to add my change 🙈 ... I added a note about what actually gets updated and semantics on changelisteners.

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
cmelchior: True, maybe just add a comment about that?

rorbech: I just removed this redundant one 😄

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
rorbech: I created and referenced #708, but maybe sounds more like a thing we should prevent than try to fix. Isn't it impossible for the user to reason about which of the instances with a given primary key that would be added first and what the end result will be ... anyway, we can drive the discussion in the issue.

test/base/src/androidTest/kotlin/io/realm/test/shared/MutableRealmTests.kt
rorbech: Good catch. Update the test.

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -44,19 +44,45 @@ public interface MutableRealm : TypedRealm {
...
-     * Creates a copy of an object in the Realm.
+     * Update policy for imports with [copyToRealm].
      *
-     * This will create a copy of an object and all it's children. Any already managed objects will
-     * not be copied, including the root `instance`. So invoking this with an already managed
-     * object is a no-operation.
+     * @see copyToRealm
+     */
+    // FIXME #naming
+    public enum class UpdatePolicy {
```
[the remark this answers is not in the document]
rorbech: I think the current names work out quite well when reading through the docs. And if staying with that there would at least be two SDKs with the same wording.

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -44,19 +44,50 @@ public interface MutableRealm : TypedRealm {
...
-     * Creates a copy of an object in the Realm.
+     * Update policy that sets the behavior when importing objects with [copyToRealm] that
+     * has the same primary key as objects already in the realm.
      *
-     * This will create a copy of an object and all it's children. Any already managed objects will
-     * not be copied, including the root `instance`. So invoking this with an already managed
-     * object is a no-operation.
+     * @see copyToRealm
+     */
+    // FIXME #naming
```
cmelchior, line 52:
```suggestion
```

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmUtils.kt
```diff
@@ -109,63 +110,75 @@ internal fun <T : RealmObject> create(
...
+            throw IllegalArgumentException("Cannot copy an invalid managed object to Realm.")
         }
 
-        // Copy object if it is not managed
-        if (!elementToCopy.isManaged()) {
+        if (element.isManaged()) {
+            if (element.`$realm$Owner` == realmReference) {
+                element
+            } else {
+                throw IllegalArgumentException("Cannot set/copyToRealm an outdated object. User findLatest(object) to resolve the latest version in the given context.")
```
cmelchior, line 178:
```suggestion
                throw IllegalArgumentException("Cannot set/copyToRealm an outdated object. User findLatest(object) to find the version of the object required in the given context.")
```

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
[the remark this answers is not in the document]
cmelchior: Something I forgot to mention in my last review. Do we want this to be under `MutableRealm`? I know it is only available there, but at least to me, reading `MutableRealm.UpdatePolicy` in the code feels a bit long and annoying? 🤔 

Not sure if there is a chance it will eventually be available elsewhere, if so, we should probably make it a standalone enum.

cmelchior (commented):
Looks great. I only have a small question about whether or not this should be `MutableRealm.UpdatePolicy` or just `UpdatePolicy`

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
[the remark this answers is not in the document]
cmelchior: Thinking a bit more.... I guess `MutableRealm.UpdatePolicy` does a better job at scoping the concept, so I can probably live with the longer name.

cmelchior (approved):

PR #698, opened 2022-02-25
Improving compiler/runtime errors

nhachicha:
Data class with a default ctor should work, if none is provided it will fail with an error indicating the need to define one.


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### Enhancements
 * Improved exception message when attempting to delete frozen objects. (Issue [#616](https://github.com/realm/realm-kotlin/issues/616))
 * Added `RealmConfiguration.Builder.compactOnLaunch()`, which can be used to control if a Realm file should be compacted when opened.
+* Improving runtime/compiler errors, (Issue [#684](https://github.com/realm/realm-kotlin/issues/684)), (Issue [#676](https://github.com/realm/realm-kotlin/issues/676)) and , (Issue [#680](https://github.com/realm/realm-kotlin/issues/680)).
```
cmelchior, line 9:
```suggestion
* Data classes can now be used as model classes. Note, it requires default no-arg constructor and properties must still be `var`'s. (Issue [#684](https://github.com/realm/realm-kotlin/issues/684))
* A better error message if Realm plugin was not applied to the module containing model classes. (Issue [#676](https://github.com/realm/realm-kotlin/issues/676))
* A better error message if a class is used that is not part of the schema. (Issue [#680](https://github.com/realm/realm-kotlin/issues/680))
```

This felt a bit too vague...would suggest splitting into multiple lines.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmSchemaTests.kt
```diff
@@ -228,6 +229,47 @@ public class RealmSchemaTests {
         realm.close()
     }
 
+    @Test
+    fun dataClass() {
```
cmelchior, line 233: Nice 👍 

However, I'm slightly concerned about some of the semantics of data classes and I think we need to have tests for these. Also so we can document if data classes behave differently than normal objects:

- `DataClass.copy()`: What state is the new copy... if our internal properties are copied, it means that the copy is also managed and you are actually modifying both objects. If not, this is effectively a `copyFromRealm()` method.

- What does `toString()` output? I'm not sure if we are providing a toString method currently, but since we are doing it in Java, I suspect we might end up doing it in Kotlin as well.

- What semantics does `equals()` have? In Java we have a custom implementation, I cannot remember what we do in Kotlin, but data class semantics differ from normal object equals.

cmelchior (commented):
Really nice to get all of these in 💯 .... However I'm slightly concerned about us just adding data class support like this. I have a nagging feeling that we are missing quite some testing and knowledge around how our plugin works together with data classes.

If we don't think there is time to verify these behaviors now, maybe it is just better to disallow data classes for now and then cycle back to it when we have better time?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 ### Enhancements
 * Improved exception message when attempting to delete frozen objects. (Issue [#616](https://github.com/realm/realm-kotlin/issues/616))
 * Added `RealmConfiguration.Builder.compactOnLaunch()`, which can be used to control if a Realm file should be compacted when opened.
+* A better error message if a Data classes was used as model classes. (Issue [#684](https://github.com/realm/realm-kotlin/issues/684))
```
cmelchior, line 9:
```suggestion
* A better error message if a data class was used as model classes. (Issue [#684](https://github.com/realm/realm-kotlin/issues/684))
```

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
...
+* A better error message if Realm plugin was not applied to the module containing model classes. (Issue [#676](https://github.com/realm/realm-kotlin/issues/676))
```
cmelchior, line 10:
```suggestion
* A better error message if the Realm plugin was not applied to the module containing model classes. (Issue [#676](https://github.com/realm/realm-kotlin/issues/676))
```

cmelchior (approved):

packages/library-base/src/darwin/kotlin/io/realm/internal/platform/RealmObject.kt
```diff
@@ -26,5 +26,5 @@ internal actual fun <T : RealmObject> realmObjectCompanion(clazz: KClass<T>): Re
     @OptIn(ExperimentalAssociatedObjects::class)
     when (val associatedObject = clazz.findAssociatedObject<ModelObject>()) {
         is RealmObjectCompanion -> associatedObject
-        else -> error("Couldn't find companion object of class ${clazz.simpleName}")
+        else -> error("Couldn't find companion object of class '${clazz.simpleName}'.\nA common cause for this is when the `io.realm.plugin` is not applied to the Gradle module that contains the '${clazz.simpleName}' class.")
```
rorbech, line 29: Our plugin is named `io.realm.kotlin`!

packages/library-base/src/jvm/kotlin/io/realm/internal/platform/RealmObject.kt
```diff
@@ -25,4 +25,4 @@ import kotlin.reflect.full.companionObjectInstance
 //  through the compiler plugin or something similar to the Native findAssociatedObject
 internal actual fun <T : RealmObject> realmObjectCompanion(clazz: KClass<T>): RealmObjectCompanion =
     clazz.companionObjectInstance?.let { it as RealmObjectCompanion }
-        ?: error("Couldn't find companion object of class ${clazz.simpleName}")
+        ?: error("Couldn't find companion object of class '${clazz.simpleName}'.\nA common cause for this is when the `io.realm.plugin` is not applied to the Gradle module that contains the '${clazz.simpleName}' class.")
```
rorbech, line 28: Our plugin is named `io.realm.kotlin`!

rorbech (commented):
Seems like the plugin name is wrong. I will fix it as part of #674.

PR #701, opened 2022-02-26
Prevent updates of primary key properties 

### LINKED ISSUE #353: Prevent updates of primary key fields
It should not be possible to update a primary key field. Probably by enforcing it to be a `val`, but also throwing errors for dynamic APIs. 

Right now it is even possible to update with a value that is overlapping. 

Underlying core issue of not raising errors is tracked by https://github.com/realm/realm-core/issues/4808

rorbech:
Closes #353


nirinchev:
We should still allow primary key changes during migration - is there some other mechanism that Kotlin is employing to support that?

rorbech:
> We should still allow primary key changes during migration - is there some other mechanism that Kotlin is employing to support that?

I am not sure that we support that in realm-java, so will have to discuss with the team what we actually want.

cmelchior:
I suspect that with automatic migrations where you create a new primary key column, all objects have the default value as their primary key...but that is not user-initiated behavior. Not sure if there is another use case? 🤔

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -221,6 +221,13 @@ internal object RealmObjectHelper {
         value: R
     ) {
         val o = obj.`$realm$ObjectPointer` ?: throw IllegalStateException("Invalid/deleted object")
+        obj.`$realm$metadata`!!.let { classMetaData ->
```
cmelchior, line 224: Unfortunately, we cannot prevent updates altogether since a migration might move the primary key column to a new one, which will have all default values which must be set to valid values.

We work around that in Java since people can populate a column before marking it the new primary key

cmelchior (changes requested):

rorbech:
Updated it to allow updates on primary key for `DynamicRealm`. I guess this is not exactly what we want, but as we currently only expose `DynamicRealm` as part of migrations, it works out. There are different schemes for enforcing this and we should consider if it should be enforced in core directly before changing the compiler plugin, etc.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -221,7 +214,26 @@ internal object RealmObjectHelper {
...
+            //  only. This relates to the overall concern of having a generic path for getter/setter
+            //  instead of generating a typed path for each type.
+            RealmInterop.realm_set_value(o, key, value, false)
+            // The catch block should catch specific Core exceptions and rethrow them as Kotlin exceptions.
+            // Core exceptions meaning might differ depending on the context, by rethrowing we can add some context related
+            // info that might help users to understand the exception.
+        } catch (exception: RealmCorePropertyNotNullableException) {
+            throw IllegalArgumentException("Required property `${obj.`$realm$ClassName`}.${obj.`$realm$metadata`!!.get(key)!!.name}` cannot be null")
+        } catch (exception: RealmCorePropertyTypeMismatchException) {
+            throw IllegalArgumentException("Property `${obj.`$realm$ClassName`}.${obj.`$realm$metadata`!!.get(key)!!.name}` cannot be assigned with value '$value' of wrong type")
```
cmelchior, line 230: We should probably have the Core type in the message somewhere?

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -221,7 +214,26 @@ internal object RealmObjectHelper {
...
+        } catch (exception: RealmCoreException) {
+            throw IllegalStateException(
```
cmelchior, line 232: Are we sure this is the only exception possible here? Or should we match on the Core message? (ugly and error-prone).

test/base/src/androidTest/kotlin/io/realm/test/shared/dynamic/DynamicMutableRealmObjectTests.kt
```diff
@@ -328,9 +328,9 @@ class DynamicMutableRealmObjectTests {
         }
     }
 
-    // This tests the current behavior of actually being able to update a primary key attribute.
-    // TODO Implement guard for updating primary keys
-    //  https://github.com/realm/realm-kotlin/issues/353
+    // This tests the current behavior of actually being able to update a primary key attribute on
+    // a dynamic realm as it is required for migrations and that is the only place we actuall
```
cmelchior, line 332:
```suggestion
    // a dynamic realm as it is required for migrations and that is the only place we actually
```

cmelchior (approved):
👍 ...had a few comments, but nothing blocking

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
rorbech: These are just moved from the `setValue` to here in `setValueByKey` to ensure that we throw errors uniformly for all access. But I can try to improve it if we need.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
rorbech: Again just moved these. But I can try to improve it if we need.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
cmelchior: If they where just moved, we can probably just cycle back to this as part of future improvements 👍

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
cmelchior: If they where just moved, we can probably just cycle back to this as part of future improvements 👍

PR #704, opened 2022-02-27
Clean up deletion API

### LINKED ISSUE #181: Evaluate how to best scope delete method on RealmObject interface
Currently we have `RealmObject.delete()` as an extension method. This allows for a nice shorthand when writing

```
realm.writeBlocking {
  val john = copyToRealm(Person())
  john.delete()
}
```

Unfortunately, this is also possible outside the write transaction

```
val john = Person()
john.delete() // Will throw an exception
```

We should evaluate if the fluency of the API trumps making it slightly less safe or if we can find a better way to control the scope of the `delete` method.

We already expose a `delete` method on the `MutableRealm`, so you can do something like this:

```
realm.writeBlocking {
  val john = copyToRealm(Person())
  delete(john)
}
```

rorbech:
Removed all `X.delete()` methods in favor or 

```
interface Deleteable

interface RealmObject : Deleteable
interface RealmList : Deleteable
interface RealmQuery : Deleteable
interface RealmSingleQuery : Deleteable
interface RealmResult : Deleteable

interface MutableRealm {
    fun delete(deleable: Deleteable)
}
interface DynamicMutableRealm {
    fun delete(deleable: Deleteable)
}
```

Closes #181


packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -79,9 +79,10 @@ public interface MutableRealm : TypedRealm {
     ): RealmQuery<T>
 
     /**
-     * Deletes the object from the underlying Realm. Only live objects can be deleted.
+     * Deletes object or objects of a query or result.
```
cmelchior, line 82:
```suggestion
     * Delete objects from the underlying Realm. [RealmObject], [RealmList], [RealmQuery] and [RealmResults] can be deleted this way.
```

The old description was a bit hard to parse for me...especially `objects of a query or result.`. Tried to reformulate.

cmelchior: In Realm Java we didn't have a `deleteAllFromRealm` method on `RealmQuery`, but I don't think it is harmful and given how our query API is structured it probably makes sense 👍

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -90,9 +91,10 @@ public interface MutableRealm : TypedRealm {
      * }
      * ```
      *
+     * @param the [RealmObject], [RealmQuery] or [RealmResults] to delete.
```
cmelchior, line 94: `RealmList` and `RealmSingleQuery` should also be Deletable.

packages/library-base/src/commonMain/kotlin/io/realm/dynamic/DynamicMutableRealm.kt
```diff
@@ -78,7 +80,12 @@ public interface DynamicMutableRealm : DynamicRealm {
      */
     public fun findLatest(obj: RealmObject): DynamicMutableRealmObject?
 
-    // FIXME Align delete behavior with MutableRealm
-    //  https://github.com/realm/realm-kotlin/issues/181
-    // fun delete(obj: RealmObject)
+    /**
+     * Deletes object or objects of a query or result.
```
cmelchior, line 84: The description and param should probably match the one in `MutableRealm`

packages/library-base/src/commonMain/kotlin/io/realm/internal/InternalMutableRealm.kt
```diff
@@ -37,7 +36,7 @@ internal interface InternalMutableRealm : MutableRealm {
                     "they can be queried this way. Use `MutableRealm.copyToRealm()` to turn it into " +
                     "a managed object."
             )
-        } else if (!obj.isFrozen()) {
+        } else if ((obj as RealmObjectInternal).`$realm$Owner` == realmReference) {
```
cmelchior, line 39: Isn't this a different check than `isFrozen`? But not sure I fully grasp the implications.

cmelchior (changes requested):
Nice...I think we are missing a few classes that should also be Deletable, but otherwise it looks great.

packages/library-base/src/commonMain/kotlin/io/realm/internal/InternalMutableRealm.kt
rorbech: It isn't sufficient to check if it is live because it could theoretically be from another live realm or a completely different realm.

packages/library-base/src/commonMain/kotlin/io/realm/MutableRealm.kt
```diff
@@ -79,9 +79,13 @@ public interface MutableRealm : TypedRealm {
...
     /**
-     * Deletes the object from the underlying Realm. Only live objects can be deleted.
+     * Delete objects from the underlying Realm.
      *
-     * Frozen objects can be converted using [MutableRealm.findLatest]:
+     * [RealmObject], [RealmList], [RealmQuery], [RealmSingleQuery] and [RealmResults] can be
+     * deleted this way.
+     *
+     * *NOTE:* Only live objects can be deleted. Frozen objects must be resolved in the current
+     * context [MutableRealm.findLatest]:
```
cmelchior, line 88:
```suggestion
     * context by using [MutableRealm.findLatest]:
```

packages/library-base/src/commonMain/kotlin/io/realm/dynamic/DynamicMutableRealm.kt
```diff
@@ -78,7 +82,24 @@ public interface DynamicMutableRealm : DynamicRealm {
...
-    //  https://github.com/realm/realm-kotlin/issues/181
-    // fun delete(obj: RealmObject)
+    /**
+     * Delete objects from the underlying Realm.
+     *
+     * [RealmObject], [RealmList], [RealmQuery], [RealmSingleQuery] and [RealmResults] can be
+     * deleted this way.
+     *
+     * *NOTE:* Only live objects can be deleted. Frozen objects must be resolved in the current
+     * context [MutableRealm.findLatest]:
```
cmelchior, line 92:
```suggestion
     * context using [MutableRealm.findLatest]:
```

cmelchior (approved):
Nice :shipit:

PR #706, opened 2022-02-28
RealmConfiguration.Builder.directory as replacement for RealmConfiguration.Builder.path

### LINKED ISSUE #346: Name of RealmConfiguration.Builder is not taken into account
The `RealmConfiguration.Builder` is not taken into account and `RealmConfiguration.name` is not right when full path is specified with `RealmConfiguration.Builder.path(...)`.

cmelchior:
Closes #346 

This PR replaces `RealmConfiguration.Builder.path` with `RealmConfiguration.Builder.directory`. This fixes the problem that specifying both `path` and `name` was counter-intuitive and didn't work together as well as aligning the API with Realm Java.

Having both `name` and `directory` is also beneficial on multiplatform where specifying paths always have some platform dependant component. So now users only have to reason about `directory` as the platform dependant part of setting up a Realm. 

**Note for reviews**
Most of this PR is just updating tests. The most relevant files to review are `SystemUtils` for Jvm and Darwin + associated tests in `RealmConfigurationTests`

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/internal/platform/SystemUtils.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/internal/platform/SystemUtils.kt
@@ -27,6 +27,18 @@ public expect val OS_VERSION: String
 public expect fun appFilesDirectory(): String
 
 /**
+ * Normalize and prepare a platform dependant path to a realm file.
+ *
+ * This method will normalize the path to a standard format on the platform, validate the filename
+ * and create any intermediate directories required to store the the file.
+ *
+ * @throws IllegalArgumentException if the directory path is somehow not valid or the required
+ * directories cannot be created.
+ * @see https://github.com/realm/realm-kotlin/issues/699
+ */
+public expect fun prepareRealmFilePath(directoryPath: String, filename: String): String
+
+/**
  * Returns the default logger for the platform.
  */
 public expect fun createDefaultSystemLogger(tag: String, logLevel: LogLevel = LogLevel.NONE): RealmLogger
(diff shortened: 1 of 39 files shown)


test/base/src/androidTest/kotlin/io/realm/test/shared/RealmConfigurationTests.kt
```diff
@@ -155,12 +197,22 @@ class RealmConfigurationTests {
     @Test
     fun name() {
         val realmName = "my.realm"
-
         val config = RealmConfiguration.Builder(schema = setOf(Sample::class)).name(realmName).build()
         assertEquals(realmName, config.name)
         assertTrue(config.path.endsWith(realmName))
     }
 
+    // TODO This should probably throw an IllegalArgumentException, for now just document behaviour.
```
rorbech, line 205: Maybe we should just not allow separators in the name at all?

rorbech (approved):
LGTM ... remember to verify it on Windows somehow

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmConfigurationTests.kt
cmelchior: Yes, that would probably make sense...It should be easy to check for in the platform specific code, but it does mean the check is deferred to when the user is calling build() rather at the callsite...but that is probably acceptable until we get better file system support.

packages/library-base/src/commonMain/kotlin/io/realm/internal/platform/SystemUtils.kt
```diff
@@ -21,11 +21,28 @@ public expect val OS_NAME: String
...
 /**
  * Returns the root directory of the platform's App data.
  */
 public expect fun appFilesDirectory(): String
 
+/**
+ * Normalize and prepare a platform dependant path to a realm file.
+ *
+ * This method will normalize the path to a standard format on the platform, validate the filename
+ * and create any intermediate directories required to store the the file.
```
nhachicha, line 38:
```suggestion
 * and create any intermediate directories required to store the file.
```

clementetb (approved):
LGTM

PR #707, opened 2022-02-28
Fix fine grained notification doc examples

clementetb:
Fixes some wrong references and unifies the style. It also adds to the readme some examples on notifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ person.addresses.asFlow()
 Query results are also observable, and like `RealmList` on each update the inserted, changed and deleted indices are also provided.
 
 ```Kotlin
-realm.query<Person>().find().asFlow()
+realm.query<Person>().asFlow()
   .collect { resultsChange: ResultsChange<Person> ->
       when(resultsChange) {
          is InitialResults -> setResultsUI(resultsChange.list)
--- a/README.md
+++ b/README.md
@@ -169,23 +169,13 @@ Realm support asynchronous observers on all its data structures.
 A Realm can be observed globally for changes on its data.
 
 ```Kotlin
-// Subscribe for change notifications on a Realm instance
-CoroutineScope(context).async {
-    realm.asFlow()
-        .collect { realmChange: RealmChange<Realm> ->
-            when (realmChange) {
-                is InitialRealm<*> -> println("Initial Realm")
-                is UpdatedRealm<*> -> println("Realm updated")
-            }
+realm.asFlow()
+    .collect { realmChange: RealmChange<Realm> ->
+        when (realmChange) {
+            is InitialRealm<*> -> println("Initial Realm")
+            is UpdatedRealm<*> -> println("Realm updated")
         }
-}
-// out: "Initial Realm"
-
-// Write data
-realm.writeBlocking {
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -160,6 +160,109 @@ realm.writeBlocking {
...
+```
+
+### RealmObject
+
+Realm objects can be observed individually. A list of the changed field names is provided on each update.
+
+```Kotlin
+val person = realm.query<Person>("name = $0", "Carlo").find().first()
+
+person.collect { objectChange: ObjectChange<Person> ->
```
nhachicha, line 188:
```suggestion
person.asFlow().collect { objectChange: ObjectChange<Person> ->
```

nhachicha (commented):
Nice update, there's one fix for a snippet, also https://github.com/realm/realm-kotlin#query needs to be updated since the code snippet using `collect` is outdated 👍

README.md
```diff
@@ -160,6 +160,109 @@ realm.writeBlocking {
...
+      }
+  }
+```
+
+### RealmResults
+
+Query results are also observable, and like `RealmList` on each update the inserted, changed and deleted indices are also provided.
+
+```Kotlin
+realm.query<Person>().find().asFlow()
```
cmelchior, line 229:
```suggestion
realm.query<Person>().asFlow()
```

I would rename this section to `RealmQuery` ... while it is true that you can observe changes to RealmResults as well, that shouldn't be the majority of use cases...calling `find()` executes the query on the main thread the first time, which is something we shouldn't promote.

cmelchior (approved):
Nice...I have a small suggestion for updating the RealmResults section, which right now promotes the wrong pattern and we should add the samples to our CI as well. Otherwise it looks great 🎉

README.md
clementetb: You are right, I was thinking on using a `RealmResults` not using the right pattern.

README.md
```diff
@@ -160,6 +160,167 @@ realm.writeBlocking {
...
+Realm data structures can be observed too. On `RealmList` on each update you receive what positions were inserted, changed or deleted.
+
+```Kotlin
+// Person named Carlo
+val person = realm.writeBlocking {
+    copyToRealm(Person().apply { name = "Carlo" })
+}
+
+// Subscribe for a RealmList change notifications
+CoroutineScope(context).async {
```
cmelchior, line 237: Maybe nitpicking a bit here....I really like this part of the docs, but it seems a bit overkill to go through all the states in the README...We should defer to the website docs for that IMO...

It isn't that I'm against having this here, but we should be mindful about the length of the README and this is adding quite a bit to that, and I'm not sure the tradeoff is worth it when we can just show this snippet and otherwise point to the docs?

cmelchior (approved):
I have some reservations about adding so much to the README when we can just refer to the proper docs...but not sure what the rest of the team thinks?

README.md
clementetb: I wasn't sure about how much detail to put in then though on putting the whole test cases. Yeah, it is quite a bit too much for the README. I am going to put in the Readme a snippet of the ReadmeTests.

cmelchior (approved):

PR #714, opened 2022-03-03
Remove default empty schema argument on all configuration builders

rorbech:


cmelchior (approved):
If CI is happy.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * `RealmObject.asFlow` will throw `UnsupportedOperationException` instead of `IllegalStateException` if called on a live or dynamic object in a write transaction or in a migration.
 * `RealmObject.asFlow` will throw `UnsupportedOperationException` instead of `IllegalStateException` if called on a live or dynamic object in a write transaction or in a migration.
 * Removed `RealmObject.delete()` and `RealmResults.delete()`. All objects, objects specified by queries and results must be delete through `MutableRealm.delete(...)` and `DynamicMutableRealm.delete(...).
+* Removed default empty schema argument for `RealmConfiguration.Builder(schema = ... )` and `SyncConfiguration.Builder(..., schema= ... )` as all configuraitons require a non-empty schema.
```
clementetb, line 9: Do we actually validate that the schemas are not empty?

clementetb (approved):
LGTM added a minor comment

CHANGELOG.md
rorbech: No, but we decided to leave it as is for now: https://mongodb.slack.com/archives/CSXQ43YFK/p1646321007587149?thread_ts=1646319193.809879&cid=CSXQ43YFK

packages/plugin-compiler/src/test/resources/schema/input/Schema.kt
```diff
@@ -23,7 +23,7 @@ class A : RealmObject
 class B : RealmObject
 class C : RealmObject
 
-val conf1 = RealmConfiguration.Builder()
+val conf1 = RealmConfiguration.Builder(setOf())
     .schema(A::class, B::class, C::class)
```
nhachicha, line 27: I wonder why we should keep `schema` if we already set it via the builder argument. One could also set some class on the builder and some on the `schema` method which could lead to confusion (more error-prone). Can't we just remove the `schema` since it's redundant?

nhachicha (approved):

PR #719, opened 2022-03-07
Fix assignments to RealmList properties on managed objects

### LINKED ISSUE #718: Setting a new list fails to modify list on managed object
It looks like the underlying issue is that lists just aren't updated on managed objects, no matter if the elements of the list are managed or not.
```

    @Test
    fun updateListOnManagedObjectTest() {
        realm.writeBlocking {
            val thingHolder = ThingHolder().apply { listOfThings = realmListOf() }
            copyToRealm(thingHolder)
        }

        realm.writeBlocking {

            val thing1 = Thing().apply { id = "thing1" }
            val thing1Managed = copyToRealm(thing1)
            val thingHolder = realm.query<ThingHolder>()
                .first()
                .find()!!
                .let { findLatest(it)!! }

            thingHolder.apply {
                name = "updatedName"
                listOfThings = realmListOf(thing1Managed)
            }

            copyToRealm(thingHolder)
        }

        val thingHolder = realm.query<ThingHolder>()
            .first()
            .find()!!

        assertEquals(thingHolder.name, "updatedName") // Passes
        assertNotNull(thingHolder.listOfThings.firstOrNull()) // Fails
    }
```

rorbech:
This adds support for assigning and importing unmanaged objects by assigning `RealmList` property on an already managed object.

Closes #718

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
@@ -329,7 +329,10 @@ internal object RealmObjectHelper {
     }
 
     @Suppress("UNUSED_PARAMETER")
-    internal fun setList(obj: RealmObjectInternal, col: String, list: RealmList<Any?>) {
-        TODO()
+    inline fun <reified T : Any> setList(obj: RealmObjectInternal, col: String, list: RealmList<Any?>) {
+        getList<T>(obj, col).also {
+            it.clear()
+            it.addAll(list)
+        }
     }
 }
--- a/test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
+++ b/test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
@@ -190,6 +190,13 @@ class RealmListTests {
     }
 
     @Test
+    fun add() {
+        for (tester in managedTesters) {
+            tester.add()
+        }
+    }
+
+    @Test
     fun addWithIndex() {
         for (tester in managedTesters) {
             tester.addWithIndex()
@@ -257,6 +264,13 @@ class RealmListTests {
     fun setFailsIfClosed() {
         // No need to be exhaustive
         managedTesters[0].setFailsIfClosed(getCloseableRealm())
+    }
+
+    @Test
+    fun reassign() {
+        for (tester in managedTesters) {
+            tester.reassign()
+        }
     }
 
     @Test
@@ -343,6 +357,7 @@ internal interface ListApiTester {
     fun copyToRealm()
     fun get()
     fun getFailsIfClosed(realm: Realm)
+    fun add()
     fun addWithIndex()
     fun addWithIndexFailsIfClosed(realm: Realm)
     fun addAllWithIndex()
@@ -353,6 +368,7 @@ internal interface ListApiTester {
     fun removeAtFailsIfClosed(realm: Realm)
     fun set()
     fun setFailsIfClosed(realm: Realm)
+    fun reassign()
 
     // All the other functions are not tested since we rely on implementations from parent classes.
 
@@ -375,7 +391,7 @@ internal interface ListApiTester {
         try {
             block()
         } catch (e: AssertionError) {
-            throw AssertionError("'${toString()}' failed - ${e.message}")
+            throw AssertionError("'${toString()}' failed - ${e.message}", e)
         }
     }
 }
@@ -518,6 +534,29 @@ internal abstract class ManagedListTester<T>(
             }
         }
 
+        assertContainerAndCleanup { container -> assertions(container) }
+    }
+
+    override fun add() {
+        val dataSet: List<T> = typeSafetyManager.getInitialDataSet()
+
+        val assertions = { container: RealmListContainer ->
+            val list = typeSafetyManager.getList(container)
+            dataSet.forEachIndexed { index, t ->
+                assertElementsAreEqual(t, list[index])
+            }
+        }
+
+        errorCatcher {
+            realm.writeBlocking {
+                val list = typeSafetyManager.createContainerAndGetList(this)
+                dataSet.forEachIndexed { index, e ->
+                    assertEquals(index, list.size)
+                    list.add(e)
+                    assertEquals(index + 1, list.size)
+                }
+            }
+        }
         assertContainerAndCleanup { container -> assertions(container) }
     }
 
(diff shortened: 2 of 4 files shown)


packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -329,7 +329,10 @@ internal object RealmObjectHelper {
     }
 
     @Suppress("UNUSED_PARAMETER")
```
clementetb, line 331: Is this annotation still relevant?

clementetb, line 331: Is this annotation still relevant?

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
```diff
@@ -819,6 +858,31 @@ internal abstract class ManagedListTester<T>(
...
 
+    override fun reassign() {
+        val dataSet = typeSafetyManager.getInitialDataSet()
+        val reassignedDataSet = listOf(dataSet[1])
+
+        val assertions = { list: RealmList<T> ->
+            assertEquals(1, list.size)
+            // We cannot assert equality on RealmObject lists as the object isn't equals to the
+            // unmanaged object from before the assignment
+            if (list[0] !is RealmObject) {
```
cmelchior, line 869: Would it be possible to check some property on the object instead, just as a smoke test? Right now RealmObjects are kinda slipping through this test.

cmelchior (approved):
Nice....I'm not 100% sure `reassign` should be in the ListTests, but we don't have another more obvious place right now, and probably speak to a broader discussion.

clementetb (approved):
LGTM, happy if CI is happy

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
rorbech: We already assert that the list is actually updated since the initial lists have two elements (guaranteed by accessing index 1) and the resulting list only having one (`list.size == 1` from above assert). But I can probably find something to assert.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
rorbech: Yes. It is only invoked from compiler plugin generated code. I will add a comment.

rorbech:
> Nice....I'm not 100% sure `reassign` should be in the ListTests, but we don't have another more obvious place right now, and probably speak to a broader discussion.

Yeah. I though of it, but didn't want to put up the whole _exhaustive type_ infrastructure somewhere else. I guess we should get the infrastructure migrated to use the schema API and reassess where to put these.

rorbech:
The existing tests showed some issues with assignments from already managed list and especially self assignments resulting in empty lists. Added some `RealmInterop.realm_equals` checks and some additional test of specific assignment scenarios. 

Haven't touched `DynamicMutableRealmObject.set` so that still doesn't support updating lists. Requires a bit more effort to split paths based on type (or name `Value/Object/List` variants?) along with all the tests, so I would suggest filing a separate issue on that to avoid blocking this fix. All list operations are also still doable on the list retrieved by `getList` so we have an immediate workaround for that.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
```diff
@@ -328,8 +328,14 @@ internal object RealmObjectHelper {
         }
     }
 
-    @Suppress("UNUSED_PARAMETER")
-    internal fun setList(obj: RealmObjectInternal, col: String, list: RealmList<Any?>) {
-        TODO()
+    @Suppress("unused") // Called from generated code
+    inline fun <reified T : Any> setList(obj: RealmObjectInternal, col: String, list: RealmList<Any?>) {
+        val existingList = getList<T>(obj, col)
+        if (list !is ManagedRealmList || !RealmInterop.realm_equals(existingList.nativePointer, list.nativePointer)) {
```
cmelchior, line 334: I don't think this check is correct..imagine this usecase, which I would expect to work:

```
realm.write {
  val obj1 = copyToRealm(Person("Mom"))
  val obj2 = copyToRealm(Person("Data))
  obj1.children.add(Person("Mia"))
  obj2.children = obj1.children
}
```

Looks like we are also missing a test for this.

test/base/src/androidTest/kotlin/io/realm/test/shared/SampleTests.kt
```diff
@@ -257,4 +259,47 @@ class SampleTests {
...
+    // @Test
+    // fun list_assign_allTypes() {}
+
+    @Test
+    fun list_assign_unmanaged() {
+        realm.writeBlocking {
+            val objectList: List<Sample> = (0..9).map { Sample().apply { intField = it } }
+            val sample = copyToRealm(Sample()).apply { objectListField = objectList.toRealmList() }
+        }
+        assertEquals(11, realm.query<Sample>().count().find())
```
cmelchior, line 274: Ahh, good old inclusive `..` in both ends 🙄

cmelchior (changes requested):
Nice...but I think I identified a corner case that isn't supported. But apart from that everything looks good.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
rorbech: I am pretty sure that it is right, and that it is indirectly covered by the test in https://github.com/realm/realm-kotlin/pull/719/files#diff-a9009c4bd0a6af3222c32304364a65fe380a3f5ecdd1e84db0f5356ada531193R278 ... but lets see when I test it explicitly 🙈

test/base/src/androidTest/kotlin/io/realm/test/shared/SampleTests.kt
rorbech: Yeah. _All inclusive_ seems to be the new standard for next generations 👴

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmObjectHelper.kt
rorbech: Reinspecting the test in https://github.com/realm/realm-kotlin/pull/719/files#diff-a9009c4bd0a6af3222c32304364a65fe380a3f5ecdd1e84db0f5356ada531193R278 and it already seem to update a managed object assigning a managed list from object to another list. Isn't it the linked test exactly what you are looking for?

cmelchior: Not 100%, there you assign a managed list to an unmanaged object here https://github.com/realm/realm-kotlin/pull/719/files#diff-a9009c4bd0a6af3222c32304364a65fe380a3f5ecdd1e84db0f5356ada531193R287... I was looking for a test that assigned a managed list to an managed object

rorbech: But the object is managed. The `copyToRealm(Sample())` returns a managed object and the assignment is inside the `apply` block on that.

cmelchior: Ah yes 🙈 ...I missed a brace there ... Yes, that means the case I was thinking about is tested. Sorry for the confusion.

rorbech:
> Nice...but I think I identified a corner case that isn't supported. But apart from that everything looks good.

@cmelchior As per https://github.com/realm/realm-kotlin/pull/719#discussion_r824728455 I think this should be good to go?

cmelchior (approved):
:shipit:

PR #723, opened 2022-03-14
Adding support for Configuration.InitialData

### LINKED ISSUE #302: Add support for RealmConfiguration.initialData()
In Realm Java, we have a concept called `RealmConfiguration.initialData()` which is a  way to populate a Realm file the first time it is opened. This is pretty helpful in a lot of scenarios and should be added to Kotlin as well.

It isn't supported in ObjectStore however, so should be added there, and the current behavior of resetting the Realm file will also cause problems with such a block. See #301 

Proposed API
```
val config = RealmConfiguration.Builder(schema = setOf(...))
  .initialData { realm: MutableRealm ->
     /* Write data */
  }
  .build()

### LINKED ISSUE #422: SyncConfiguration - InitialData
See [https://github.com/realm/realm-kotlin/issues/195](https://github.com/realm/realm-kotlin/issues/195)

Similar to Realm Java we should have support for `initialData`:

```java
SyncConfiguration.Builder(user, partition)
  .initialData { this: MutableRealm ->
    // Write initial data
  }
  .build() 
```

cmelchior:
Closes #302
Closes #422 

Add support `Configuration.initialData()`, so it works for both `RealmConfiguration` and `SyncConfiguration`. Similar to how it works in Realm Java.

**TODO**
- [x] Figure out how to get a MutableRealm during initialization? Right now I have two options
           - 1. Refactor how MutableRealms are created. Right now it is pretty heavily intertwined with WriterRealm. Unclear how difficult this is.
           - 2. Only use the callback to set a flag we can react to after opening the Realm. This appears to be easier, but might also be problematic if the same configuration is used to open multiple Realms at the same time.
 - [x] Change to SchemaMode::HardReset
 - [x] Tests
 - [x] This approach means it will not work for `DynamicRealm`...This is also the case in Java where no-one has complained about the missing support...could be worked around by having a `MutableRealm.asDynamicRealm()` extension method (or something like that)

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/Callback.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/Callback.kt
@@ -62,3 +62,7 @@ fun interface CompactOnLaunchCallback {
 fun interface MigrationCallback {
     fun migrate(oldRealm: NativePointer, newRealm: NativePointer, schema: NativePointer): Boolean
 }
+
+fun interface DataInitializationCallback {
+    fun invoke(realm: NativePointer): Boolean
+}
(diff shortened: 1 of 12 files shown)


cmelchior:
Replaced by https://github.com/realm/realm-kotlin/pull/839

Issue #710, opened 2022-03-01
Evaluate use of C++ runtime


cmelchior:
Right now it looks like we are bundling `libc++_shared.so` in our cinterop package. This is causing problems with other libraries if they do not agree on the NDK version. Surfacing as errors such as these: https://github.com/realm/realm-kotlin/issues/694.

Looking at Realm Java, it doesn't look like we are bundling `libc++_shared` which probably means we are using `libc++_static` instead. At least according to the documentation: https://developer.android.com/ndk/guides/cpp-support#selecting_a_c_runtime

We should investigate why there is such a difference between Realm Java and Realm Kotlin and what would be the most appropriate choice for Realm Kotlin.

cmelchior:
Probably also relevant to #216

rorbech:
I revived an old PR to use the static runtime in #620

PR #732, opened 2022-03-18
Reducing binary size for Andorid dependency

nhachicha:
- There's a difference of ~1MB on the overall AAR file when applying the linker `gc` flags
- `arm64` was `7.5MB` and now `7MB`
- This is also way below the size of a Java shared object with Sync which is around `9.2MB` for `arm64`
  
Most of the other flags set in Java are not necessary since they're set by the Android toolchain file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
-* None.
+* Reducing the binary size for Android dependency . (Issue [#616](https://github.com/realm/realm-kotlin/issues/216)).
 
 ### Fixed
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
-* Reducing the binary size for Android dependency . (Issue [#616](https://github.com/realm/realm-kotlin/issues/216)).
+* Reducing the binary size for Android dependency . (Issue [#216](https://github.com/realm/realm-kotlin/issues/216)).
 
 ### Fixed
 * None.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
-* None.
+* Reducing the binary size for Android dependency . (Issue [#616](https://github.com/realm/realm-kotlin/issues/216)).
```
clementetb, line 7:
```suggestion
* Reducing the binary size for Android dependency . (Issue [#216](https://github.com/realm/realm-kotlin/issues/216)).
```

clementetb (approved):
Happy if CI 🤖 is happy

packages/cinterop/src/jvm/CMakeLists.txt
```diff
@@ -36,6 +36,12 @@ set(REALM_ENABLE_SYNC ON)
 set(REALM_BUILD_LIB_ONLY ON)
 add_subdirectory("${CMAKE_SOURCE_DIR}/../../../external/core" core EXCLUDE_FROM_ALL)
 
+if(ANDROID)
+    if (CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+        set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-gc-sections -Wl,--exclude-libs,ALL")
```
cmelchior, line 41: It would be nice to see which flags are actually set? E.g. are we actually enabling LTO anywhere? I don't think that is enabled by default by the Android toolchain

CHANGELOG.md
cmelchior: Would be nice to see how the Kotlin `.so` compare to the one we have in Android. That might show if we are forgetting something?

cmelchior (approved):

CHANGELOG.md
nhachicha: let's find out when we link against [STL](https://github.com/realm/realm-kotlin/issues/710)

Issue #733, opened 2022-03-20
Investigate if Realm.close actually closes all ressources correctly.


cmelchior:
Adding JMH to the project and running it resulted in an error like the below, also after calling `Realm.close()`:

```
<JMH had finished, but forked VM did not exit, are there stray running threads? Waiting 24 seconds more...>

Non-finished threads:

Thread[Thread-3,5,main]

Thread[DestroyJavaVM,5,main]

Thread[pool-3-thread-1,5,main]
  at java.base@11.0.10/jdk.internal.misc.Unsafe.park(Native Method)
  at java.base@11.0.10/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
  at java.base@11.0.10/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2081)
  at java.base@11.0.10/java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:433)
  at java.base@11.0.10/java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1054)
  at java.base@11.0.10/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1114)
  at java.base@11.0.10/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
  at java.base@11.0.10/java.lang.Thread.run(Thread.java:834)

Thread[pool-2-thread-1,5,main]
  at java.base@11.0.10/jdk.internal.misc.Unsafe.park(Native Method)
  at java.base@11.0.10/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
  at java.base@11.0.10/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2081)
  at java.base@11.0.10/java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:433)
  at java.base@11.0.10/java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1054)
  at java.base@11.0.10/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1114)
  at java.base@11.0.10/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[...]

PR #734, opened 2022-03-21
Add preliminary support for benchmarks

### LINKED ISSUE #726: Internal refactor: Benchmarking
From #713

(Christian - 1) Add Benchmarking setup. The first step should be to setup benchmarks for Android (since that is our slowest platform), then later for iOS.

- Keep in mind that it should eventually work in CI context
- And be able to compare results against previous results.

cmelchior:
Closes #726 

This PR adds support for benchmarking.

Initially, I thought I could add this as part of the `test` project, but due to project constraints required by Jetpack Benchmark, we would have ended up with a number of modules which would be annoying. So I moved everything into its own project `/benchmarks` that need to be opened by itself. It uses the same composite build infrastructure that `/test` do.

The PR does the following:
- Infrastructure to run Android tests using Jetpack Microbenchmarks.
- Infrastructure to run JVM tests using JMH.
- No attempt has been made to support benchmarking on iOS/macOS yet.
- No attempt has been made to add tooling for comparing/analyzing data. This should be in a future PR.
- I added a README with detailed explanations on how to run and gather data.
- Benchmark results are output as JSON files on both platforms, which will allow for further analysis in the future.

The benchmarks added are not intended to be the full set, but should work as a starting point. Currently the following things are benchmarked:

* Managed accessors (read)
* Unmanaged accessors (read)
* Open Realm with different schema sizes
* How fast inserting data is at various sizes and with or without primary keys

TODO:
- [x] Align Gradle versions with rest of project
- [x] Add benchmarks for Bulk insert (100, 1000, 10000) objects (JVM)
- [x] Add Benchmarks for opening a Realm (JVM)
- [x] Add benchmarks for Bulk insert (100, 1000, 10000) objects (Android)
- [x] Add Benchmarks for opening a Realm (Android)
- [x] Cleanup sample data
- [x] Cleanup JVM test state
- [x] Add support for detect/ktlint


benchmarks/androidApp/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+        testInstrumentationRunner = "androidx.benchmark.junit4.AndroidBenchmarkRunner"
+        testInstrumentationRunnerArguments["androidx.benchmark.suppressErrors"] = "EMULATOR,UNLOCKED"
+        // Enable profiling. See https://developer.android.com/studio/profile/microbenchmark-profile
+        // testInstrumentationRunnerArguments["androidx.benchmark.profiling.mode"] = "StackSampling"
+    }
+
+    testBuildType = "release"
+    buildTypes {
+        debug {
+            // Since isDebuggable can"t be modified by gradle for library modules,
```
rorbech, line 31:
```suggestion
            // Since isDebuggable can't be modified by gradle for library modules,
```

benchmarks/shared/src/commonMain/kotlin/io/realm/kotlin/benchmarks/Entities.kt
```diff
@@ -0,0 +1,215 @@
...
+import io.realm.RealmList
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import io.realm.realmListOf
+
+enum class SchemaSize {
+    SINGLE, SMALL, LARGE
+}
+
+val SINGLE_SCHEMA = setOf(Entity1::class)
```
rorbech, line 28: I would put these in a map by the above keys, so that we can easily lookup the correct schema from the `SchemaSize` programatically.

benchmarks/jvmApp/src/jmh/kotlin/io/realm/kotlin/benchmark/OpenRealmTests.kt
```diff
@@ -0,0 +1,83 @@
...
+open class OpenRealmTests {
+
+    @Param("SINGLE", "SMALL", "LARGE") // Must match enum names
+    var schemaSize: String = SchemaSize.SINGLE.name
+    var realm: Realm? = null
+    lateinit var config: RealmConfiguration
+
+    @Setup(Level.Invocation)
+    fun setUp() {
+        val schema = when (schemaSize) {
```
rorbech, line 60: If the schemas was in a map (`SchemaSize->Set<KClass...>`) we could just look them up instead of this cumbersome mapping.

benchmarks/androidApp/src/androidTest/java/io/realm/kotlin/benchmarks/android/OpenRealmTests.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    @get:Rule
+    val benchmarkRule = BenchmarkRule()
+
+    private lateinit var config: RealmConfiguration
+    private var realm: Realm? = null
+
+    @Before
+    fun setUp() {
+        val schema = when (schemaSize) {
```
rorbech, line 54: Again, cumbersome mapping. Maybe use a map for the schemas.

rorbech (approved):
Awesome. I guess it leaves a bit for the inspection part as is, but really great base infrastructure 🌮 

Maybe we should add a test for the setters, seems to be the only very basic test that lacks. But could also just be in an upcoming PR for whomever needs it. 

Had some issues running it on an emulator (also when rooted), but works on a real devices
```
Perfetto unexpected exit code, output = 5491
        EXITCODE=2
```
Also some warnings about the forks not existing, so forcing closing the forked VM on JVM. But seems to run, so I guess we can improve over the next PRs 🏗️

clementetb (approved):
LGTM 🧙🏻‍♂️ Great work!

benchmarks/shared/src/commonMain/kotlin/io/realm/kotlin/benchmarks/Entities.kt
cmelchior: Good idea 👍

cmelchior:
I created some issues to follow up and improve this PR: #733 #738 #740

The only thing I'm unsure about is why the Benchmarks might not run on some emulators. I noticed that profiling was enabled as a default. So instead I disabled that.

PR #760, opened 2022-03-25
Type safe native pointer wrappers

rorbech:
This PR adds type information to the native pointes. Ideally the native struct should have a corresponding Kotlin object and encapsulate operations on that type, but this would require reworking our current native memory management routines and would potentially overlap with any efforts in automatic generation of such classes from the core api. Without that rework we would introduce yet another level of indirection to get hold of the pointers. Thus, this implementation is a minimal implementation of a typed cinterop API that gives a type safe pointer system (which cannot be achieved directly through `typealiases`) that still allows taking advantage of inheritance relationship between pointer types (which isn't possible by `value class`es).


rorbech:
The current implementation doesn't differentiate `RealmChangesPointer` into `RealmObjectChangesPointer` and `RealmCollectionChangesPointer`. Doing that would introduce even more generic parameters to the whole notification registration infrastructure which in my opinion will make it too cluttered and not worth it.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -46,26 +43,80 @@ expect val INVALID_PROPERTY_KEY: PropertyKey
...
+typealias RealmConfigurationPointer = NativePointer<RealmConfigT>
+typealias RealmSchemaPointer = NativePointer<RealmSchemaT>
+typealias RealmPointer = NativePointer<out RealmT>
+typealias LiveRealmPointer = NativePointer<LiveRealmT>
+typealias FrozenRealmPointer = NativePointer<FrozenRealmT>
+typealias RealmObjectPointer = NativePointer<RealmObjectT>
+typealias RealmListPointer = NativePointer<RealmListT>
+typealias RealmResultsPointer = NativePointer<RealmResultsT>
+typealias RealmQueryPointer = NativePointer<RealmQueryT>
+// FIXME Should we differentiate this by generics
```
cmelchior, line 73: IMO no... typealiasing is for making the types easier to reason about. Adding generics goes in the opposite direction

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -46,26 +43,80 @@ expect val INVALID_PROPERTY_KEY: PropertyKey
...
+typealias RealmNotificationTokenPointer = NativePointer<RealmNotificationTokenT>
+typealias RealmChangesPointer = NativePointer<RealmChangesT>
+typealias RealmObjectChangesPointer = NativePointer<RealmObjectChangesT>
+typealias RealmCollectionChangesPointer = NativePointer<RealmCollectionChangesT>
+
+// Sync types
+// Pure marker interfaces corresponding to the C-API realm_x_t struct types
+interface RealmAppT : CapiT
+interface RealmAppConfigT : CapiT
+interface RealmSyncConfigT : CapiT
```
cmelchior, line 83: Is it possible to keep `RealmConfigT and `RealmSyncConfigT` completely separate? I would assume it would cause problems when opening the Realm?

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -46,26 +43,80 @@ expect val INVALID_PROPERTY_KEY: PropertyKey
...
-    fun realm_config_get_encryption_key(config: NativePointer): ByteArray?
-    fun realm_config_set_should_compact_on_launch_function(config: NativePointer, callback: CompactOnLaunchCallback)
-    fun realm_config_set_migration_function(config: NativePointer, callback: MigrationCallback)
+    fun realm_config_new(): RealmConfigurationPointer
+    fun realm_config_set_path(config: RealmConfigurationPointer, path: String)
+    fun realm_config_set_schema_mode(config: RealmConfigurationPointer, mode: SchemaMode)
+    fun realm_config_set_schema_version(config: RealmConfigurationPointer, version: Long)
+    fun realm_config_set_schema(config: RealmConfigurationPointer, schema: RealmSchemaPointer)
+    fun realm_config_set_max_number_of_active_versions(config: RealmConfigurationPointer, maxNumberOfVersions: Long)
+    fun realm_config_set_encryption_key(config: RealmConfigurationPointer, encryptionKey: ByteArray)
```
cmelchior, line 114: These are also available on synchronized Realms.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmImpl.kt
```diff
@@ -50,7 +50,9 @@ import kotlin.reflect.KClass
 // TODO API-PUBLIC Document platform specific internals (RealmInitializer, etc.)
 internal class RealmImpl private constructor(
     configuration: InternalConfiguration,
-    dbPointer: NativePointer
+    // TODO Should actually be a frozen pointer, but since we cannot directly obtain one we expect
+    //  a live reference and grab the frozen version of that in the init-block
+    dbPointer: LiveRealmPointer,
```
cmelchior, line 55: Great that this is now captured by the type system 👍

cmelchior (commented):
Nice 💯 . I focused mostly on the changes in RealmInterop and assume that the types in all other files will match.

Generally, it looks good to me. I just think there is something odd about the config pointers. We have some builder methods that work on both. And in core, it is also the same struct (except for the sync-struct attached to the local realm one).

So I wonder what is the reason for keeping the completely split here?

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
cmelchior: Turns out I misunderstood what this actually was. Just disregard the above comment.

rorbech: Yeah. These pointers doesn't correspond 1-1 to the user facing `RealmConfiguration` and `SyncConfiguration`. 

A `RealmConfiguration` has a `RealmConfigurationPointer` and all sync related configurations are done through the `RealmConfigurationPointer` by updating its `RealmSyncConfigurationPointer` entry.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
rorbech: Common properties are set on the `RealmConfigurationPointer` with the `realm_config_x`. All sync related things are separated into the sync configuration parts pointed to by `RealmSyncConfigurationPointer`.

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmImpl.kt
rorbech: Yep, should help avoiding calling all the update methods on a frozen realm pointer, etc. ... if only the IDE was actually able to recognize type symbols and types 😞

cmelchior (approved):
🚀

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmImpl.kt
```diff
@@ -209,5 +211,9 @@ internal class RealmImpl private constructor(
 
 // Returns a DynamicRealm of the current version of the Realm. Only used to be able to test the
 // DynamicRealm API outside of a migration.
-internal fun Realm.asDynamicRealm(): DynamicRealm =
-    DynamicRealmImpl(this@asDynamicRealm.configuration as InternalConfiguration, (this as RealmImpl).realmReference.dbPointer)
+internal fun Realm.asDynamicRealm(): DynamicRealm {
+    // The RealmImpl.realmReference should be a FrozenRealmReference, but since we cannot
+    // initialize it as such we need to case it here
```
nhachicha, line 216:
```suggestion
    // initialize it as such we need to cast it here
```

nhachicha (approved):
Nice effort 🥇

Issue #524, opened 2021-10-27
Provide fine grained AppException details in API


rorbech:
With #521 the C-API app errors are correctly propagated to the user with details serialized into the exception message. 

We should expose the details in a manner where users could handle the various error scenarios programatically. 

Essential differentiation would be:
- Recoverable or not
- Error codes for to drive specific use cases, i.e. invalid username/password. But https://github.com/realm/realm-core/blob/master/src/realm.h#L2022 error categories and category specific error codes are not necessarily useful for users.

Same considerations probably apply to sync errors in #493

PR #762, opened 2022-03-28
Initial implementation for supporting uploading and downloading changes

cmelchior:
Initial stab at surfacing `SyncSession.downloadAllServerChanges()` and `SyncSession.uploadAllLocalChanges()`

I converted the Java implementation to use coroutines instead, which massively simplified it 🥳 . However there are also some missing questions that need to be solved:

TODO
- [x] Testing depends on us finding a way to expose SyncSession. Should probably be in another PR: https://github.com/realm/realm-kotlin/issues/246
- [x] In Java, the SyncSession is largely decoupled from the Realm due to the thread-confinement, this e.g. means you have to manually call `Realm.refresh()` after this is to access the latest state. If Kotlin, since we only have one Realm, it would make sense if we did this on the users behalf, but it is unclear exactly how to implement this.
- [x] Constructing the `SyncErrorCode` object is fairly annoying, but not sure there is a better way?


cmelchior:
Ready for review

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/Callback.kt
```diff
@@ -50,6 +51,20 @@ interface SyncErrorCallback {
     fun onSyncError(pointer: RealmSyncSessionPointer, throwable: SyncException)
 }
 
+// Interface used internally as a bridge between Kotlin (JVM) and JNI.
+// We pass all required primitive parameters to JVM and construct the objects there, rather than
+// having to do this on the JNI side, which is both a ton of boilerplate, but also expensive in
+// terms of the number of JNI traversals.
+internal interface JVMSyncSessionTransferCompletionCallback {
```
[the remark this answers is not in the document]
cmelchior: Good idea 👍

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncSession.kt
```diff
@@ -38,6 +39,34 @@ import io.realm.Realm
...
+    /**
+     * Calling this method will block until all known remote changes have been downloaded and
+     * applied to the Realm or the specified timeout is hit. This will involve network access, so
+     * calling this method should only be done from a non-UI thread.
+     *
+     * @param timeout Maximum amount of time before this method should return.
+     * @throws IllegalArgumentException if `timeout` is <= 0.
+     * @throws SyncException if a problem was encountered with the connection during the download.
+     * @throws IllegalStateException if called from inside a [SyncSession.ErrorHandler].
+     * @return `true` if the data was downloaded. `false` if the download timed out before it
```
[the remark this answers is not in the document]
cmelchior: Yes, fixed.

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
```diff
@@ -152,4 +153,18 @@ internal class SuspendableNotifier(
...
+    /**
+     * Manually force a refresh of the Realm, moving it to the latest version.
+     * This will also trigger the evaluation of all change listeners, which will
+     * be triggered as normal if anything changed.
+     *
+     * @return a frozen reference to the version of the Realm after the refresh.
+     */
+    suspend fun refresh(): FrozenRealmReference {
+        return withContext(dispatcher) {
+            RealmInterop.realm_refresh(realm.realmReference.dbPointer)
```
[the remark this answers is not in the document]
cmelchior: Hmm, no. But the `realm` reference here is a Live Reference, and it is the one we refresh and get the snapshot from, so it should be the latest version, no?

This is also the reason for my comment in RealmImpl. `NotifierRealm.onRealmChanged()` will be triggered as well _eventually_  and there is a race condition with regards to who gets to update the RealmImpl first.

But if I understand things correctly, that should be fine?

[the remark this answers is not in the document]
rorbech: The `snapshot` is currently only updated on `onRealmChanged` (actually it is cleared in `onRealmChanged` and lazily refreshed when needed). The reason for this was that I didn't want to create new frozen reference each time we requested a snapshot. We are potentially issuing many different object notifications from different C-API callbacks on the same versions, so I tried to reuse the same frozen reference across all notifications from the same version (our logic/clean-up-strategy might also break if we have multiple `RealmReference` objects pointing to the same underlying native shared realm 🤔). By only updating it on the `onRealmChanged` callback we ensure that we only have one reference per version (and per the guarantee that `onRealmChanged` is issued before any other callback it should be safe to use that version for any other notification). And by only creating the snapshot when needed we ensure that we don't bother bookkeeping references for when there is no object notification listeners (since we are now using new references to `RealmImpl` on each `onRealmChanged` this optimization is maybe worthless). 

Since this code is running in the single threaded dispatcher you could maybe just clear the `snapshot` after the refresh, but could it be that the `onRealmChanged` was actually guaranteed to be triggered synchronously when already on the right thread? If not, then we should maybe do a pattern around this that ensures we remember to wipe the `snapshot` on each explicit update like this in the future.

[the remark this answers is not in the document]
cmelchior: Hmm, sounds like I need to think a bit more about this implementation.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/SyncErrorCodeCategory.kt
```diff
@@ -0,0 +1,9 @@
+package io.realm.internal.interop.sync
+
+expect enum class SyncErrorCodeCategory {
```
nhachicha, line 3: Where are we indicating the recoverability aspect of the error?

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -1481,6 +1490,54 @@ actual object RealmInterop {
...
+            staticCFunction { userdata ->
+                disposeUserData<(RealmSyncSessionPointer, SyncSessionTransferCompletionCallback) -> Unit>(userdata)
+            }
+        )
+    }
+
+    private fun handleCompletionCallback(
+        userData: CPointer<out CPointed>?,
+        error: CPointer<realm_sync_error_code_t>?
+    ): Boolean {
```
nhachicha, line 1528: The returned boolean is not used?

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/sync/SyncErrorCodeCategory.kt
```diff
@@ -0,0 +1,22 @@
...
+    UNKNOWN(realm_sync_error_category.RLM_SYNC_ERROR_CATEGORY_UNKNOWN);
+
+    companion object {
+        fun of(nativeValue: realm_sync_error_category): SyncErrorCodeCategory {
+            for (value in values()) {
+                if (value.nativeValue == nativeValue) {
+                    return value
+                }
+            }
+            error("Unknown sync error code category: $nativeValue")
```
nhachicha, line 19: Won't this reintroduce the breaking change we're trying to avoid when the protocol changes or introduce a new error code?

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/JVMSyncSessionTransferCompletionCallback.kt
```diff
@@ -0,0 +1 @@
+package io.realm.internal.interop
```
nhachicha, line 1: why empty?

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/sync/SyncErrorCodeCategory.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    companion object {
+        @JvmStatic
+        fun of(nativeValue: Int): SyncErrorCodeCategory {
+            for (value in values()) {
+                if (value.nativeValue == nativeValue) {
+                    return value
+                }
+            }
+            error("Unknown sync error code category value: $nativeValue")
```
nhachicha, line 20: ditto

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmImpl.kt
```diff
@@ -118,6 +118,20 @@ public class RealmImpl private constructor(
             }
         )
 
+    /**
+     * Manually force this Realm to update to the latest version.
+     * The refresh will also trigger any relevant notifications.
+     * TODO Public because it is called from `SyncSessionImpl`.
+     */
+    public suspend fun refresh() {
```
nhachicha, line 126: This concept is a leftover from the old architecture and I feel is alien to the new unidirectional frozen architecture, we should strive to keep this method internal/out of the public API

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncSession.kt
```diff
@@ -38,6 +39,34 @@ import io.realm.Realm
  */
 public interface SyncSession {
 
+    /**
+     * Calling this method will block until all known remote changes have been downloaded and
+     * applied to the Realm or the specified timeout is hit. This will involve network access, so
+     * calling this method should only be done from a non-UI thread.
+     *
+     * @param timeout Maximum amount of time before this method should return.
```
nhachicha, line 47: you should specify the unit?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncSessionImpl.kt
```diff
@@ -15,10 +15,103 @@
...
+     * or downloaded.
+     *
+     * When this method returns. The user facing Realm has been updated to the latest state.
+     *
+     * @param direction whether data is being uploaded or downloaded.
+     * @param timeout timeout parameter.
+     * @return `true` if the job completed before the timeout was hit, `false` otherwise.
+     */
+    private suspend fun waitForChanges(direction: TransferDirection, timeout: Duration): Boolean {
+        if (realm == null) {
```
nhachicha, line 69: Is this the only way we assume we're in a `SyncSession.ErrorHandler`?

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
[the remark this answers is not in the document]
cmelchior: Instead of calling `realm.snapshot` we could call `realmChanged().first()` as a way to hook into the change mechanism, but then we probably run into a race condition where we don't know if a change event is a buffered one (and thus older than the one we need).

I suspect we are forced to do some kind of synchronous thing here, and then verify that cleanup logic doesn't break if we have multiple versions of the same Realm 🤔

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -110,6 +122,199 @@ class SyncSessionTests {
...
+        }
+        Unit
+    }
+
+    @Ignore // TODO https://github.com/realm/realm-core/issues/5365
+    @Test
+    fun uploadDownload_throwsUnderlyingSyncError() {
+        openSyncRealm { realm ->
+            val session = realm.syncSession
+            runBlocking {
```
nhachicha, line 287: Maybe wrap this in a method or add a comment to highlight the intention (i.e Restart Sync to trigger a session error)

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncSession.kt
cmelchior: `timeOut` is a `Duration`, which is a new Kotlin API that captures both value and duration in a quite succinct way, i.e. you can write `10.seconds`

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncSessionImpl.kt
cmelchior: Yes, it is a bit of a hack. But currently the only place we construct a session without a Realm is inside the error handler.  I can add a comment here to make that more clear?

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/sync/SyncErrorCodeCategory.kt
cmelchior: Currently the category is not exposed to end users, but gets serialized to a String, so we can change it without breaking people. But I expect we are going to revisit this once we have a better idea about how to expose errors. Right now, the shortest path was just to expose what the C-API has.

But we serialize all of it to a String, which is also pretty bad, but is what we have done so far.

This will be fixed in https://github.com/realm/realm-kotlin/issues/524

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/JVMSyncSessionTransferCompletionCallback.kt
cmelchior: Added by accident. Will remove

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -110,6 +122,199 @@ class SyncSessionTests {
...
+                session.downloadAllServerChanges()
+            }.also {
+                assertTrue(it.message!!.contains("Uploading and downloading changes is not allowed"))
+            }
+        } finally {
+            wrongSchemaRealm?.close()
+            job.cancel()
+            channel.close()
+        }
+        Unit
```
nhachicha, line 279:
```suggestion
```

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/SyncErrorCodeCategory.kt
cmelchior: We don't right now. See https://github.com/realm/realm-kotlin/pull/762#discussion_r845405228

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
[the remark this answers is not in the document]
cmelchior: I think I found a way to hook into the notification mechanism so we still only have one frozen instance...It does introduce a race to update `RealmImpl`, but since RealmImpl already supports this and I don't expect these API's to be called frequently, that is probably fine.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
cmelchior: We need to return Unit here as test methods are not allowed to return anything and with this pattern `fun method() = runBlocking { }`, it will return whatever is returned from runBlocking.

packages/library-base/src/commonMain/kotlin/io/realm/internal/SuspendableNotifier.kt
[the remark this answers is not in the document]
cmelchior: Also see https://mongodb.slack.com/archives/C010R3CMP3N/p1649411267690299

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -110,6 +122,192 @@ class SyncSessionTests {
...
+            assertTrue(realm1.syncSession.uploadAllLocalChanges())
+
+            // Due to the Server Translator, there is a small delay between data
+            // being uploaded and it not being immediately ready for download
+            // on another Realm. In order to reduce the flakyness, we are
+            // re-evaluating the assertion multiple times.
+            for (i in 4 downTo 0) {
+                assertTrue(realm2.syncSession.downloadAllServerChanges())
+                val size = realm2.query<ParentPk>().count().find()
+                when (size == 10L) {
```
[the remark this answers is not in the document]
rorbech: I guess now you only enter this when `size==10` so you would never throw below if we haven't downloaded the data.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncSessionImpl.kt
```diff
@@ -15,10 +15,111 @@
...
+        // out in the error message.
+        if (realm == null) {
+            throw IllegalStateException(
+                """
+                Uploading and downloading changes is not allowed when inside 
+                a `SyncSession.ErrorHandler`.
+                """.trimIndent()
+            )
+        }
+        require(timeout.isPositive()) {
```
rorbech, line 80: Maybe consider adding our own `Validation.require` just to ensure that we can report errors and update it consistently throughout the SDK.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncSessionImpl.kt
```diff
@@ -15,10 +15,111 @@
...
+                    }
+                    channel.receive()
+                }
+            }
+            if (direction == TransferDirection.DOWNLOAD) {
+                realm.refresh()
+            }
+            when (result) {
+                is Boolean -> return result
+                is Throwable -> throw result
```
rorbech, line 115: Should this throw a new exception with `cause = result` 🤔

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -110,6 +121,185 @@ class SyncSessionTests {
...
+                .name("new_realm.realm")
+                .errorHandler(object : SyncSession.ErrorHandler {
+                    override fun onError(session: SyncSession, error: SyncException) {
+                        channel.trySend(session)
+                    }
+                })
+                .build()
+            wrongSchemaRealm = Realm.open(config)
+        }
+        val session = channel.receive()
```
rorbech, line 249: Is this actually executed inside the error callback? I guess it seems so if the below asserts succeeds, but I wouldn't  expect that sending something on a buffered channel would mean the that the receiving end will handle it synchronously in receive 🤔

rorbech (changes requested):
There is still an issue with at least one of the tests ... also had a clarifying question for one of the others.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncSessionImpl.kt
cmelchior: Hmm, I would think that propagating the underlying exception would be the correct thing to do? What exception would otherwise make sense here?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
cmelchior: It isn't executed inside the callback as such. But the callback is the only place where the SyncSession is constructed in a special manner, so this test just ensure that if _that_ instance is called, some functionality is not working. I can add a comment about this or find a better name for the test if that would help?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/SyncSessionImpl.kt
rorbech: It was mostly to got the full trace, but rethinking it might be the right thing just to have the original.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
rorbech: Yeah. I guess the test should just have another name them.

rorbech (approved):

PR #773, opened 2022-03-31
Allow update of existing objects in lists with UpdatePolicy.ALL

### LINKED ISSUE #771: UpdatePolicy.ALL ignored for internal objects
Error is received if some object from internal RealmList already exists in the database. Haven't tried with just object instead of list. 
Received error: Object with this primary key already exists.

rorbech:
Closes #771


cmelchior (approved):
🚀

PR #774, opened 2022-03-31
Update docs upload path

rorbech:


cmelchior (approved):
According to https://mongodb.slack.com/archives/C018UTGP23T/p1648648952615329?thread_ts=1648636619.426099&cid=C018UTGP23T this looks correct 👍 

(I will probably regret saying that)

PR #776, opened 2022-04-01
Running Kmm Sample with latest Kotlin version

nhachicha:
The Kotlin SDK should be built with the lowest Kotlin version. On the other hand, consumer apps should be compatible with the latest Kotlin version.

Since `min-android-sample` tests the minimum version of various dependencies, this PR runs `kmm-sample` with the latest version of Kotlin (usually an EAP upcoming release like `1.6.20-RC2`) to make sure we catch any breaking changes and guarantees compatibility with the upcoming releases.


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 
 ### Internal
 * Updated to Ktor 1.6.8.
+* Renaming internal synthetic variables to avoid deprecated prefix `$realm$` by `io_realm_kotlin_`.
```
cmelchior, line 25:
```suggestion
* Rename internal synthetic variables prefix to `io_realm_kotlin_`, so deprecated prefix `$realm$` is avoided.
```

cmelchior (approved):

rorbech (approved):

Issue #671, opened 2022-02-08
Unregister realm and schema change callbacks when realm is GC'ed


rorbech:
As part of #596 we have started using the C-API's `realm_add_realm_changed_callback` and `realm_add_schema_changed_callback` but these callbacks are not unregistered when the realm is garbage collected. We need to hook this up with our native reference queue/cleaner API. 

Until this is done the realm will not be closed unless explicitly closing the realm with `realm.close()`. To overcome this limitation in the memory test I have added an internal method to explicit release the callbacks https://github.com/realm/realm-kotlin/pull/596/files#diff-4d0528b8584520f9db14ff4031d83ca6cb837bcd889022c4abcc08b6309235baR188

We should either generalize the `NativeObjectReference`/`CPointerWrapper` to allow managing other things than native pointer and allow overriding the _clean up_ action to unregister the callback tokens or rely on core to align the callback registration mechanisms as filed in https://github.com/realm/realm-core/issues/5217.

PR #779, opened 2022-04-06
Upgrade to Core 11.13

cmelchior:
This brings in a lot of changes:

- Apple builds have been reworked in Core, which impacted some of our builds. A close review of `cinterop` would be needed as I was mostly guessing on what to do. /cc @nhachicha 
- Schema/RealmChanged callbacks now return tokens that can be released similar to other notifications. I tried to modify their call site according to what looked needed, but @rorbech should probably take a closer look.
- A new reset mode was added, which also trigger the initial data block on migrations (which we need), but it looks like semantics changed when adding new schemas. I didn't document it at as breaking change, but just modified the unit test.
- Google Auth Code is now supported (will be fixed in another PR)
- Credentials can now be json serialized, which we needed for some unit tests. This will also be added in another PR.


packages/library-base/src/commonMain/kotlin/io/realm/internal/LiveRealm.kt
```diff
@@ -39,8 +39,8 @@ import kotlinx.coroutines.CoroutineDispatcher
  */
 internal abstract class LiveRealm(val owner: RealmImpl, configuration: InternalConfiguration, dispatcher: CoroutineDispatcher? = null) : BaseRealmImpl(configuration) {
 
-    private val realmChangeRegistration: RegistrationToken
-    private val schemaChangeRegistration: RegistrationToken
+    private val realmChangeRegistration: RealmCallbackTokenPointer
```
rorbech, line 42: The `RealmCallbackTokenPointer` should probably somehow be wrapped in a `NotificationToken` so we are able to both `cancel` them when the realm is closed but also release it if the objects are GC'ed when the realm itself is GC'ed. As it is right here we still cannot eliminate the explicit [`unregisterCallback`](https://github.com/realm/realm-kotlin/pull/596/files#diff-4d0528b8584520f9db14ff4031d83ca6cb837bcd889022c4abcc08b6309235baR188) to make our `garbageCollectorShouldFreeNativeResources` memory test pass.

rorbech (approved):
We are still not fully leveraging the new releasable callback registration tokens from core as we are still not unregistering the callbacks if the new tokens are GC'ed. Thus, we are still not able to close #671 with this PR, but since users can just `close` the realm explicitly I think we can leave it as is.

nhachicha (changes requested):

nhachicha (approved):
👍

PR #780, opened 2022-04-07
Re-enable Google Auth and JSON tests.

cmelchior:
Google Auth Code and JSON is now available in the C-API, allowing us to re-enable them.


test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/CredentialsTests.kt
```diff
@@ -88,29 +88,31 @@ class CredentialsTests {
         assertFailsWith<IllegalArgumentException> { Credentials.emailPassword("foo@bar.com", "") }
 
         // Test Google as a special case as two types of Google login exists
-        // TODO https://github.com/realm/realm-core/issues/5347
-        // assertFailsWith<IllegalArgumentException> { Credentials.google("", GoogleAuthType.AUTH_CODE) }
+        assertFailsWith<IllegalArgumentException> { Credentials.google("", GoogleAuthType.AUTH_CODE) }
         assertFailsWith<IllegalArgumentException> { Credentials.google("", GoogleAuthType.ID_TOKEN) }
     }
 
+    @Suppress("invisible_reference", "invisible_member")
```
rorbech, line 95: Not that you need to rework this, but I usually try to encapsulate usage of the internal API into a separate method, so that we don't accidentally unintentionally uses internal stuff elsewhere in the tests. Here you could do an `assertJson(....)` and do the cast and `asJson` in there.

rorbech (approved):
Nothing critical but left some comments/questions?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/internal/CredentialsImpl.kt
```diff
@@ -32,6 +32,10 @@ internal open class CredentialImpl constructor(
             RealmInterop.realm_auth_credentials_get_provider(nativePointer)
         )
 
+    internal fun asJson(): String {
```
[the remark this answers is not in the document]
cmelchior: It is mostly public in Java because there was no good way to hide it....it might be marginally useful, but would prefer to keep it hidden until a clear use case surfaces.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/CredentialsTests.kt
cmelchior: Good point 👍

PR #781, opened 2022-04-08
Add compatible Kotlin version

nhachicha:
image


cmelchior (commented):
Nice, but Isn't it confusing that is shows an earlier version than 1.6.20? Similar to the release tags, shouldn't it show the latest version we are compatible with? 🤔

nhachicha:
@cmelchior I was trying to following the convention of specifying the minimum Kotlin version to be compatible with the SDK like https://github.com/ktorio/ktor

cmelchior (approved):
Ah okay...I haven't noticed they started doing that. This sounds fine then 👍

PR #784, opened 2022-04-11
Improved App/Sync exceptions

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/524

The purpose of this PR is to improve how errors are being reported for Sync/App Services.

High-level exception hierarchy is as follows (feel free to bikeshed naming):

- `AppException: RealmException`
    - `ServiceException: AppException`
        -  `BadRequestException: ServiceException`
        - `ConnectionException: ServiceException`
        - `AuthException: ServiceException`
            -  `UserNotFoundException : AuthException`  
            - `UserAlreadyConfirmedException: AuthException`
            - `UserAlreadyExistsException: AuthException`
            - `InvalidCredentialsException: AuthException`  
    - `SyncException: AppException`
        - `UnrecoverableSyncException: SyncException` 
        - `WrongSyncTypeException: SyncException`

Primary classes to review: `AppExceptions` and `RealmSyncUtils`

- Exceptions have all been moved from internal cinterop to public library modules in  `io.realm.exceptions` and `io.realm.mongodb.exceptions`.

- All Sync/App errors are not being propagated as data classes and then converted into exceptions on the library side (so we can use the public API exceptions). This is happening in `RealmSyncUtils`

- Many enums have a static `of` method for fetching the Kotlin enum value from the cinterop value. This is intended to be used internally. Because we need to map these values on the library side, I have added a public `fromInt()` method to all enums that need it. In some cases, it mirrors the `of` method completely (on JVM)`, which is a bit silly. For now, that seemed like the best way to split concerns. But we should probably discuss it.

- All exception message attempt to follow the same format and will now look like this `[<category>][<errorCodeDescription>(<errorCode>)] <Message>. Server log: <server link>`. Example `[Service][UserAlreadyConfirmed(44)] This user was already confirmed. Server log: https://link.to.server/logs`

TODO
- [x] Settle on an approach
- [x] Define high-level idea driving the API and the rough categories.
- [x] Define the exact set of initial exceptions we want to expose.


ERRORS.md
```diff
@@ -0,0 +1,368 @@
...
+	    AUTH_PROVIDER_DUPLICATE_NAME(Type.SERVICE, 38),
+	    RESTRICTED_HOST(Type.SERVICE, 39),
+	    API_KEY_ALREADY_EXISTS(Type.SERVICE, 40),
+	    INCOMING_WEBHOOK_AUTH_FAILED(Type.SERVICE, 41),
+	    EXECUTION_TIME_LIMIT_EXCEEDED(Type.SERVICE, 42),
+	    NOT_CALLABLE(Type.SERVICE, 43),
+	    BAD_REQUEST(Type.SERVICE, 48),
+
+
+		- AuthException (Most things involving users should probably go here?)
```
cmelchior, line 230: What happens if the refresh token expires?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/ExceptionPrototype.kt
```diff
@@ -0,0 +1,64 @@
...
+}
+
+public open class RealmError : Error {
+    internal constructor(message: String): super(message)
+    internal constructor(message: String, cause: Throwable): super(message, cause)
+}
+
+// Top-level exception for all MongoDB/Sync API's
+public open class AppException: RealmException {
+    public val category: String // Is there any value in making this an enum?
```
cmelchior, line 17: We should put all of these into the message.

ERRORS.md
cmelchior: `InvalidCredentialsException` will be thrown

cmelchior:
This PR should be ready enough to start reviewing, while I cleanup the last few remaining things.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ * further analysis. But in most cases a more specific subtype will be thrown, which will be easier
+ * to react to.
+ *
+ * Subclasses of this class fall into two broad categories: [ServiceException] and [SyncException].
+ *
+ * 1. [ServiceException]'s are being thrown by all API's that talk directly to Atlas App Services
+ *    through HTTP requests. These API's are found in the [io.realm.mongodb.App],
+ *    [io.realm.mongodb.User] and [[io.realm.mongodb.EmailPasswordAuth] classes.
+ *
+ * 2. [SyncException]'s are thrown by errors caused when using Device Sync API's, i.e
```
nathan-contino-mongo, line 18:
```suggestion
 * 2. [SyncException]'s are thrown by errors caused when using Device Sync APIs, i.e
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ *   - [SyncException]
+ *     - [UnrecoverableSyncException]
+ *     - [WrongSyncTypeException]
+ *
+ * This hierarchy is intended to model errors in a way so exceptions at the bottom of the hierarchy
+ * are _actionable_, i.e. it should be clear from the exception which action can be taken to
+ * resolve it.
+ *
+ * Exceptions further up the hierarchy are used as a way to categorize the errors and are
+ * harder to react to in a single way that will make sense to end users of an app, but they should
```
nathan-contino-mongo, line 42:
```suggestion
 * harder to react to in a single way that will make sense to end users of an app, but should
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ * @see BadRequestException
+ * @see AuthException
+ */
+public open class ServiceException : AppException {
+    internal constructor(message: String) : super(message)
+}
+
+/**
+ * Exception indicating that something went wrong with the underlying HTTP request towards one of
+ * the Atlas App Services. The exact cause is in the exception message.
```
nathan-contino-mongo, line 105:
```suggestion
 * Exception indicating that something went wrong with the underlying HTTP request to 
 * Atlas App Services. The exact cause is in the exception message.
```
We're still settling our language around "Atlas App Services" for the documentation site, but I suspect we won't refer to "Services" as individual items much if ever (since they're all part of one product).  This should be a little more consistent with the language we'll use when we complete that refactor.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * Errors resulting in this exception is something outside the apps control and can be considered
```
nathan-contino-mongo, line 107:
```suggestion
 * Errors resulting in this exception are outside the apps control and can be considered
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ * temporary. Retrying the action some time in the future should generally be safe, but since
+ * corrective measures potentially needs to taken in either the apps network environment
```
nathan-contino-mongo, line 109:
```suggestion
 * corrective measures potentially need to taken in either the apps network environment
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ * or on the server, this error should be logged for further analysis.
```
nathan-contino-mongo, line 110:
```suggestion
 * or on the server, you should log this error for further analysis.
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+}
+
+/**
+ * Thrown when something has gone wrong with Device Sync in a way that is not recoverable.
+ *
+ * Generally, errors of this kind are due to incompatible versions of Realm and Atlas App Services
+ * being used or bugs in the library or on the server, and the only fix would be installing a new
+ * version of the app with a new version of Realm.
+ *
+ * It is still possible to use the Realm locally after this error occurred. However, this must be
```
nathan-contino-mongo, line 219:
```suggestion
 * It is still possible to use the Realm locally after this error occurs. However, this must be
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ * done with caution as data written to the Realm after this point risk getting lost as
```
nathan-contino-mongo, line 220:
```suggestion
 * done with caution as data written to the realm after this point risk getting lost as
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,237 @@
...
+ * many errors of this category will result in a Client Reset once the connection to the server
+ * is re-established.
```
nathan-contino-mongo, line 222:
```suggestion
 * many errors of this category will result in a Client Reset once the client
 * re-connects to the server.
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/App.kt
```diff
@@ -86,7 +89,15 @@ public interface App {
...
      * @return the logged in [User].
-     * @throws AppException if the user could not be logged in.
+     * @throws InvalidCredentialsException if the provided credentials were not correct. Note, only
+     * [AuthenticationProvider.EMAIL_PASSWORD], [AuthenticationProvider.API_KEY] and
+     * [AuthenticationProvider.JWT] can throw this exception. Other authentication providers will
+     * throw an [AuthException] instead
+     * @throws AuthException if a problem occurred when logging in. See the exception message for
+     * further details.
+     * @throws io.realm.mongodb.exceptions.ServiceException All API's that talk to Atlas App
+     * Services through a HTTP request can fail in a variety of ways. See [AppException] for details
```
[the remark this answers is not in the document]
cmelchior: Discussed offline.

packages/library-base/src/commonMain/kotlin/io/realm/exceptions/RealmException.kt
```diff
@@ -0,0 +1,15 @@
+package io.realm.exceptions
+
+/**
+ * Top level class for all exceptions that are Realm specific and doesn't have a better
+ * existing platform exception like [IllegalArgumentException] or [IllegalStateException].
+ *
+ * In most cases, a subclass of this type will be thrown. Exactly which, will will be documented by
+ * the individual API entry points.
+ */
+public open class RealmException : RuntimeException {
```
rorbech, line 10: We should probably wrap the core exceptions in this on in https://github.com/realm/realm-kotlin/blob/master/packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmUtils.kt#L307 instead

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/AppErrorCategory.kt
```diff
@@ -0,0 +1,36 @@
...
+    RLM_APP_ERROR_CATEGORY_JSON,
+    RLM_APP_ERROR_CATEGORY_CLIENT,
+    RLM_APP_ERROR_CATEGORY_SERVICE,
+    RLM_APP_ERROR_CATEGORY_CUSTOM;
+
+    // Public visible description of the enum value
+    public val description: String
+
+    companion object {
+        fun fromInt(nativeValue: Int): AppErrorCategory
```
rorbech, line 34: We could define an `FromInt`-interface (#naming ;)) for this method and just extend all the companion object with it with `companion objects : FromInt<AppErrorCategory>` ... maybe not that important but you would maybe be able to supply the implementation with a default implementation instead of repeating it 20 times 😓 😉

packages/cinterop/src/androidTest/java/io/realm/test/sync/SyncEnumTests.kt
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Test
+    fun jsonErrorCode() {
+        checkEnum(realm_app_errno_json_e::class) { nativeValue ->
+            JsonErrorCode.fromInt(nativeValue)
+        }
+    }
+
+    private inline fun <T : Any> checkEnum(enumClass: KClass<out Any>, mapNativeValue: (Int) -> T) {
```
rorbech, line 135: Always nice with a test that verifies that such an assisting assertion method is actually catching an error. Don't know if you can put in some invalid definition though 🤔

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/AppTests.kt
```diff
@@ -90,12 +90,34 @@ class AppTests {
...
     @Test
-    fun login_InvalidUserThrows() = runBlocking {
-        assertFailsWith<AppException> {
-            app.login(Credentials.emailPassword("foo", "bar"))
-        }.let { exception: AppException ->
-            assertTrue(exception.message!!.startsWith("invalid username/password [error_category=3, error_code=50, link_to_server_logs="))
+    fun login_invalidCredentialsThrows() = runBlocking {
+        for (provider in AuthenticationProvider.values()) {
+            try {
+                when (provider) {
```
rorbech, line 100: Maybe rework into a `when`-block that returns a credentials and then 
```
   ?.let { 
      assertFailsWithMessage(...) {
         app.login(it)
      }
   }
```
to also assert the message 🤷

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,253 @@
...
+package io.realm.mongodb.exceptions
+
+import io.realm.exceptions.RealmException
+
+/**
+ * Top level class for all exceptions that are specific to using Atlas App Services and Device Sync.
+ *
+ * Generally, reacting to this exception will be hard at runtime, except to log the error for
+ * further analysis. But in most cases a more specific subtype will be thrown, which will be easier
+ * to react to.
```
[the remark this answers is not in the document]
cmelchior: Reordered the sections, this comes further below.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,253 @@
...
+ * This exception is considered the top-level or "catch-all" for problems related to user account
+ * actions. The exact reason for the error can be found in [Throwable.message].
+ *
+ * Generally, this exception does not need to be caught as more specific subtypes are available.
+ * These will be documented for the relevant API methods.
+ *
+ * @see UserAlreadyConfirmedException
+ * @see UserNotFoundException
+ * @see UserAlreadyExistsException
+ * @see InvalidCredentialsException
```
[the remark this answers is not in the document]
cmelchior: To some degree you are correct, but it is rather hard to find:

image

Compared to:
image

image

So I would probably prefer to keep the mentions for now.

packages/library-base/src/commonMain/kotlin/io/realm/exceptions/RealmException.kt
cmelchior: Yup, that was the plan. It will happen as part of #745

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/AppErrorCategory.kt
cmelchior: I tried to play around with this, but given that this is expect/actual functions and that their implementations depend on the constructor value (which is not known by the interface), it doesn't look possible to do something easily. So I would prefer to punt this down the field

packages/cinterop/src/androidTest/java/io/realm/test/sync/SyncEnumTests.kt
cmelchior: It is possible, but the chance is fairly slim as you have to mess up both the common code and you usually see both "our" name and the autogenerated one side by side. It also looks like something that is very hard to check, except through some clever use of reflection, so It didn't seem worth it.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/exceptions/AppExceptions.kt
```diff
@@ -0,0 +1,253 @@
...
+ *    handle HTTP requests for the SDK.
+ *
+ * 2. [SyncException]'s are thrown by errors caused when using Device Sync APIs, i.e
+ *    realms opened using a [io.realm.mongodb.SyncConfiguration]. These errors propagate
+ *    through the [io.realm.mongodb.SyncConfiguration.Builder.errorHandler].
+ *
+ * Each of these categories are divided further:
+ *
+ * - [AppException]
+ *   - [ServiceException]
```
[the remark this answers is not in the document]
cmelchior: I guess the file could get rather large. I split it into `AppException` with just the high top-level class and then `ServiceExceptions.kt` and `SyncExceptions.kt.` with all exceptions for that subgroup.

PR #793, opened 2022-04-19
Don't share partition value between tests.

cmelchior:
Re-using the same partition key between tests seems to cause issues in some cases. The root cause is unclear, but two hypothesizes  could either be that some tests doesn't clean up correctly in case of other failures or that the Randomizer sometimes reuse a value. 

To work around this, this PR creates separate partition values for all sync-related tests. This finding the root cause would be ideal, `master` is currently broken so fixing that has the highest priority.


rorbech (approved):
LGTM. We could consider to make the partition value default to a random value with `createSyncConfig(..., partitionValue = TestHelper.randowPartitionValue, ...)` to save some lines, but maybe better to require an explicit value 🤷

Issue #504, opened 2021-10-14
Sync - add `with` method to Sync config builder


edualonso:
Recipe: if the builder constructor has default values leave those properties out of the `with` function.

rorbech:
This will greatly simplified with #607 as that eliminates compiler plugin re-wirering.

Issue #795, opened 2022-04-20
Bug: Paths for Synced Realms appear broken


cmelchior:
While writing tests for `SyncConfiguration.with` I noticed that Sync paths appear broken in surprising ways.

It probably happened when we moved support for `name()` and `directory()` to `Configuration` so it is now shared between synced and normal Realms.

cmelchior:
This probably pulls in #412 as well

cmelchior:
Intended semantics:

* Synced Realms should be placed in `<syncRoot>/<appId>/<userId>/<filename>`. The filename is based on the partition value.
* Setting `AppConfiguration.syncRootDirectory` will the `<syncRoot>` to whatever is provided
* Setting `SyncConfiguration.directory()` will replace the `<syncRoot>/<appId>/<userId>/` part.
* Setting `SyncConfiguration.name()` will replace the `<filename>`

`realm_sync_client_config_set_base_file_path` should set the `syncRoot`
`realm_app_sync_client_get_default_file_path_for_realm` should create the entire path including the realm file name.

Given that there are various ways we can customize the path, we need to split the path returned from  `realm_app_sync_client_get_default_file_path_for_realm`. This could be done using `SystemUtils.PATH_SEPERATOR`

cmelchior:
I do remember some issues with Core setting the root dir name, so if we set the `syncroot` to `/app/files/mongodb-realm/...`...the path ends up being `/app/files/mongodb-realm/mongodb-realm/...`.

cmelchior:
After discussing, we don't see a use case for having `directory` for SyncConfigurations. So we should move `directory()` so it only exists on `RealmConfiguration`.

PR #796, opened 2022-04-20
Add support for SyncConfiguration.with

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/504
Blocked by #795 

- Add support for Integer and Null partition values.
- Temporarily remove the partition value from the public API since it exposes internal details we want to hide.
- Add support for shorthand `SyncConfiguration.with()` similar to how `RealmConfiguration.with()` works.


cmelchior:
I'm trying to hunt down some flaky tests that this PR seems to expose, but the reset of the PR should be ready for review.

rorbech (commented):
We could also just have made a public `PartitionValue` with factory methods `of(...)` for the various types ... then we wouldn't have to add all these specific constructor/`with`-variants.

But LGTM ... but I will await approving until I see how many tests you'll have to ignore 🤣

edualonso (approved):
It looks good to me 👍

cmelchior:
Exposing a public `PartitionValue` just seems to add another level of indirection, ie. a PartitionValue would then need to have 3 similar constructors, so we would need to document them there instead.

Also it would make all callsites like this `SyncConfiguration.with(user, PartititionValue("my-partition"), setOf())` which also feels slightly annoying.

We agreed to just ship hiding it for now, and can then re-evaluate. We can easily add a wrapper later without breaking the API.

PR #798, opened 2022-04-21
Support ObjectId

nhachicha:
- [x] Merge https://github.com/realm/realm-kotlin/pull/790 and refactor this PR to use the new storage conversion mechanism 
- [x] Using signed bytes for internal representation to avoid the experimental API annotation 
- [x] Add edge case tests https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/test/resources/bson/oid.json 
- [x] Add round trip (Sync) tests to make sure MongoDB represents correctly the OID primary keys


rorbech:
I have updated this to use the new converter infrastructure from #790 after it was merged to master.

packages/library-base/src/commonMain/kotlin/io/realm/ObjectId.kt
```diff
@@ -0,0 +1,35 @@
...
+public interface ObjectId : Comparable<ObjectId> {
+    public companion object {
+        public fun from(hexString: String): ObjectId = ObjectIdImpl(hexString)
+
+        public fun from(date: RealmInstant): ObjectId = ObjectIdImpl(date)
+
+        @OptIn(ExperimentalUnsignedTypes::class)
+        public fun from(bytes: UByteArray): ObjectId = ObjectIdImpl(bytes)
+
+        public fun get(): ObjectId = ObjectIdImpl()
```
cmelchior, line 33: Should this be `create()` similar to our discussion around `DynamicRealmObject`?

`get()` kinda implies that the same is being returned every time?

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -287,6 +291,89 @@ class SyncSessionTests {
...
+            val json: JsonObject = adminApi.insertDocument(
+                ObjectIdPk::class.simpleName!!,
+                """
+                    {
+                        "name": "$partitionValue",
+                        "realm_id" : "$partitionValue"
+                    }
+                """.trimIndent()
+            )
+            val oid = json["insertedId"]?.toString()?.removeSurrounding("\"")
```
cmelchior, line 317: I believe you can use `json["insertedId"].jsonPrimitive.asString` or something like that instead of this work-around.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -287,6 +291,89 @@ class SyncSessionTests {
...
+                }
+
+                realm.syncSession.uploadAllLocalChanges()
+                realm.close()
+            }
+
+            delay(1_000) // let Sync integrate the changes
+
+            val syncedDocumentJson = adminApi.queryDocumentById(ObjectIdPk::class.simpleName!!, oid)
+            val oidAsString = syncedDocumentJson["_id"]?.toString()?.removeSurrounding("\"")
```
cmelchior, line 370: See above. This hack shouldn't be needed.

cmelchior (commented):
Nice 💯 

I have a bunch of minor comments + one major (the datetime dependency)
Getting a setup for injecting documents directly into MongoDB could definitely also be useful elsewhere.

Also missing Changelog update.

What is the intended strategy when getting a proper BSON library? From the structure here, I assume we are going to deprecate `io.realm.ObjectId` and use `org.bson.kotlin.ObjectId` (or whatever it ends up being)? (And then support both of them for while)

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -1147,6 +1160,15 @@ actual object RealmInterop {
         return TimestampImpl(this.timestamp.seconds, this.timestamp.nanoseconds)
     }
 
+    private fun realm_value_t.asObjectId(): ObjectIdWrapper {
+        if (this.type != realm_value_type_e.RLM_TYPE_OBJECT_ID) {
+            error("Value is not of type ObjectId: $this.type")
+        }
+        val byteArray = ByteArray(OBJECT_ID_BYTES_SIZE)
+        this.object_id.bytes.mapIndexed { index, sh -> byteArray[index] = sh.toByte() }
```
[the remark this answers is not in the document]
nhachicha: typo

test/base/src/jvmTest/kotlin/io/realm/test/compiler/IndexTests.kt
```diff
@@ -90,7 +93,6 @@ class IndexTests {
             source = SourceFile.kotlin(
                 "indexing_collections.kt",
                 """
-                        import io.realm.RealmInstant
```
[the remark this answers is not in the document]
nhachicha: it's an unused import

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/util/AdminApi.kt
```diff
@@ -96,6 +98,16 @@ interface AdminApi {
      */
     suspend fun getAuthConfigData(): String
 
+    /**
+     * Insert a MongoDB document which will be eventually synced as RealmObject.
+     */
+    suspend fun insertDocument(clazz: String, json: String): JsonObject
+
+    /**
+     * Query the specified database and collection
```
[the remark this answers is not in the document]
nhachicha: The private method works with the collection, whereas the public API uses the Realm types ...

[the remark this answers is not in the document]
cmelchior: Yup, I know. It isn't a big deal, it just felt weird to use different wordings here.

cmelchior (approved):
Nice, only had minor comments. Then :shipit:

packages/library-base/src/commonMain/kotlin/io/realm/internal/ObjectIdImpl.kt
```diff
@@ -0,0 +1,295 @@
...
+    public override val bytes: ByteArray
+        get() = _bytes
+
+    /**
+     * Represents an ObjectID from an array of 12 bytes.
+     */
+    private val _bytes: ByteArray
+
+    /**
+     * Time in milliseconds.
```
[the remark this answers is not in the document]
nhachicha: actually, they're seconds from epoch.

test/base/src/androidTest/kotlin/io/realm/test/shared/ObjectIdTests.kt
```diff
@@ -0,0 +1,220 @@
...
+    @AfterTest
+    fun tearDown() {
+        if (!realm.isClosed()) {
+            realm.close()
+        }
+        PlatformUtils.deleteTempDir(tmpDir)
+    }
+
+    @Test
+    fun invalid_arguments() {
```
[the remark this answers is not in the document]
nhachicha: RealmInstants MAX & MIN values are tested in `boundaries` test

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
   * Besides the changes of base class of `DynamicRealmObject`, this should not require and code changes.
 
 ### Enhancements
-* None.
+* [Sync] Added support for `ObjectId` to be used as a primary key ([#652](https://github.com/realm/realm-kotlin/issues/652)).
```
rorbech, line 10:
```suggestion
* [Sync] Added support for `ObjectId` ([#652](https://github.com/realm/realm-kotlin/issues/652)).
```

test/base/src/androidTest/kotlin/io/realm/test/shared/ObjectIdTests.kt
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        assertFailsWithMessage<IllegalArgumentException>("byte array size must be 12") {
+            ObjectId.from(byteArrayOf(0xCA.toByte(), 0xFE.toByte(), 0xBA.toByte(), 0xBE.toByte()))
+        }
+    }
+
+    @Test
+    fun boundaries() {
+        roundTrip(ObjectId.from("000000000000000000000000")) { value ->
```
rorbech, line 67: I am reluctant with just testing against objects create with the constructor 🤔  

Isn't there some way to verify that the actual input is also represented in the underlying bytes. Especially since you don't have negative tests for `equals`. Maybe a test against the `toString` representation for each of the constructors. Basically all of these tests would pass if you didn't store the input argument in the internal bytes.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -287,6 +293,95 @@ class SyncSessionTests {
...
+        runBlocking {
+            val config = SyncConfiguration.Builder(user, partitionValue, schema = setOf(ObjectIdPk::class))
+                .build()
+            val realm = Realm.open(config)
+
+            val json: JsonObject = adminApi.insertDocument(
+                ObjectIdPk::class.simpleName!!,
+                """
+                    {
+                        "name": "$partitionValue",
```
rorbech, line 314: Again no test against visible validatable data, so this would also pass even though our serialization was off. We need some minimal proof that the round trip yields what we expect.

rorbech (changes requested):
Awesome, but I am a bit reluctant with the tests. I am pretty confident that the implementation is actually right, but think we need some tests against visible recognizable data. Both the `ObjectId` tests and the sync round trip. 

We should also add `ObjectId` variant option for sync partition value, right?

test/base/src/androidTest/kotlin/io/realm/test/shared/ObjectIdTests.kt
rorbech: I guess the `toString` test is actually doing exactly this kind of test, sorry. Probably too difficult to do for the non-hexstring variants, so if we just include some negative `equals` test between the objects created with the variants we should probably be good to go.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
rorbech: Should be ok after discovering that the `toString`-test is actually verifying that the bytes are serialized correctly ... at least for the hex-string variant.

rorbech:
> We should also add `ObjectId` variant option for sync partition value, right?

It probably doesn't make sense to add support for `ObjectId` partition value now as #824 reworks the api for setting the partition value.

rorbech:
> Awesome, but I am a bit reluctant with the tests. I am pretty confident that the implementation is actually right, but think we need some tests against visible recognizable data. Both the `ObjectId` tests and the sync round trip.

I guess `toString` test is probably sufficient ... but still lacks negative tests for `equals`

nhachicha:
@rorbech I've added inequality test in https://github.com/realm/realm-kotlin/pull/798/commits/86c3008b632da4c2bfe244d836b00d680d185241#diff-cf276e81f3a9533dca9c16d4222905996022cbaffdc9254475b4f2111b23e8feR153-R154

PR #800, opened 2022-04-25
Bugfix for paths for Synced Realms

edualonso:


packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -234,6 +234,7 @@ expect object RealmInterop {
     fun realm_app_log_out(app: RealmAppPointer, user: RealmUserPointer, callback: AppCallback<Unit>)
     fun realm_app_remove_user(app: RealmAppPointer, user: RealmUserPointer, callback: AppCallback<Unit>)
     fun realm_clear_cached_apps()
+    fun realm_app_sync_client_get_default_file_path_for_realm(app: RealmAppPointer, syncConfig: RealmSyncConfigurationPointer, name: String): String
```
cmelchior, line 237: Isn't `name` optional here?

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
```diff
@@ -142,54 +140,20 @@ public interface Configuration {
...
-         *      NSUserDomainMask,
-         *      null,
-         *      true,
-         *      null
-         * )
-         * ```
+         * If setting the full path of the realm, this name is not taken into account.
          *
-         * @param directoryPath either the canonical absolute path or a relative path from the current directory ('./').
+         * @throws IllegalAttributeException if the name includes a path separator.
```
cmelchior, line 147: What is `IllegalAttributeException`? Shouldn't this be `IllegalArgumentException`?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncConfiguration.kt
```diff
@@ -31,6 +33,8 @@ import io.realm.mongodb.internal.SyncConfigurationImpl
 import io.realm.mongodb.internal.UserImpl
 import kotlin.reflect.KClass
 
+private const val REALM_EXTENSION = ".realm"
```
cmelchior, line 36: Perhaps rename to `REALM_FILE_EXTENSION`?

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncConfiguration.kt
```diff
@@ -99,13 +111,22 @@ public interface SyncConfiguration : Configuration {
...
-         *
-         * @param errorHandler lambda to handle the error.
-         */
-        public fun errorHandler(errorHandler: SyncSession.ErrorHandler): Builder =
-            apply { this.errorHandler = errorHandler }
+        override fun name(name: String): Builder = apply {
+            if (name.contains(PATH_SEPARATOR)) {
+                throw IllegalArgumentException("Name cannot contain path separator '$PATH_SEPARATOR': '$name'")
+            }
+            if (name.isEmpty()) {
```
cmelchior, line 118: This check is missing in `RealmConfiguration`

cmelchior (commented):
I know this is still draft, so just leaving a few drive-by comments.

packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/RealmInterop.kt
edualonso: Yes, I missed it. I will change it.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -242,6 +243,26 @@ class SyncConfigTests {
         assertEquals(config.partitionValue.asString(), partitionValue)
     }
 
+    @Test
+    fun name() {
```
cmelchior, line 247: I think we lack some corner case tests here:

- E.g. partition-based default names. They should be something like `s_mypartition.realm` or `i_42.realm`.
- What happens if you bring a filename with no `.realm` suffix. 

Since our default name is `default.realm` and if ObjectStore doe append `.realm` to all names, then I could be tempted to also enforce it for local Realms, but that would be a breaking change. Not sure what your thoughts are?

It does have the advantage that you can always double-click on a file and get it to open in Realm Studio though.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -65,20 +63,18 @@ class SyncSessionTests {
         user = runBlocking {
             app.createUserAndLogIn(email, password)
         }
-        tmpDir = PlatformUtils.createTempDir()
```
cmelchior, line 68: Are you sure it is safe to just not use the temporary dir? How is cleanup done then? I would have assumed that the `TestApp` then needs to define this? Or we have some other logic that deletes the default folder after each test?

cmelchior (changes requested):
Lots of nice improvements in here 💯 .... I left some minor comments, but also a few bigger ones:

- Naming between SyncConfiguration and RealmConfiguration seems to be inconsistent with regard to the `.realm` suffix.
- We lack tests for corner cases around naming.
- Missing CHANGELOG. This PR is breaking paths for synched builds, so we should make sure to mention it. Also directory is no longer available for SyncConfiguration and `AppConfiguration.Builder.syncRootDirectory` has been added.

packages/cinterop/src/darwin/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -1440,6 +1436,20 @@ actual object RealmInterop {
...
+        syncConfig: RealmSyncConfigurationPointer,
+        name: String?
+    ): String {
+        val cPath = realm_wrapper.realm_app_sync_client_get_default_file_path_for_realm(
+            app.cptr(),
+            syncConfig.cptr(),
+            name
+        )
+        return cPath.safeKString()
+            .also { realm_wrapper.realm_free(cPath) }
```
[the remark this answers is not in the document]
edualonso: Not that I'm aware of. I did it because the C-API docs explicitly state the following:
```
/**
 * Get the default realm file path based on the user and partition value in the config.
 *
 * @param custom_filename custom name for the realm file itself. Can be null,
 *                        in which case a default name based on the config will be used.
 *
 * Return value must be manually released with realm_free().
 */
```
There are other C-API functions that require it but we haven't implemented them yet.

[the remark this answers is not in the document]
cmelchior: :+1: Sounds like we need to go through all of them then 🙈

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
```diff
@@ -106,6 +113,36 @@ public interface AppConfiguration {
                 this.userLoggers = customLoggers
             }
 
+        /**
+         * Configures the root folder containing all files and realms used when synchronizing data
+         * between the device and MongoDB Realm.
+         *
+         * The default root directory is platform-dependent:
+         * - Android: `Context.getFilesDir()/mongodb-realm`
```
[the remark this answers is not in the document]
edualonso: Yeah, I forgot about adding the extra sync part.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
```diff
@@ -106,6 +113,36 @@ public interface AppConfiguration {
...
+        public fun syncRootDirectory(rootDir: String): Builder = apply {
+            val directoryExists = directoryExists(rootDir)
+            if (!directoryExists && fileExists(rootDir)) {
+                throw IllegalArgumentException("'rootDir' is a file, not a directory: $rootDir.")
+            }
+            if (!directoryExists) {
+                try {
+                    prepareRealmDirectoryPath(rootDir)
+                } catch (e: IllegalArgumentException) {
+                    throw IllegalArgumentException("Could not create the specified directory: $rootDir.")
```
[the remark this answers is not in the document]
edualonso: `prepareRealmDirectoryPath` already handles the cases of `!dir.mkdirs()` and `dir.isFile` internally. This catch will only be triggered if `File(directoryPath, filename).absolutePath` fails for some reason (not related to Realm) so I guess it's OK to not throw this exception manually.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/SyncConfiguration.kt
```diff
@@ -99,13 +114,17 @@ public interface SyncConfiguration : Configuration {
...
-         *
-         * @param errorHandler lambda to handle the error.
-         */
-        public fun errorHandler(errorHandler: SyncSession.ErrorHandler): Builder =
-            apply { this.errorHandler = errorHandler }
+        override fun name(name: String): Builder = apply {
+            nameCheck(name)
+
+            // Strip `.realm` suffix as it will be appended by Object Store later
+            this.name = if (name.endsWith(REALM_FILE_EXTENSION)) {
```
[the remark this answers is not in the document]
edualonso: Yeps, `realm_app_sync_client_get_default_file_path_for_realm` does append `.realm` to whatever `name` we use.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncConfigTests.kt
edualonso: I think it would be fair to add the constraint for local Realms for the sake of uniformity.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
edualonso: Yeah, I noticed that too. I sort of left it in limbo while I was working on implementing the rest. I will sort it out.

edualonso: Turns out cleanup is done correctly when we call `TestApp.close()`. Also and just for fun I changed the deletion logic so that we use `PlatformUtils.deleteTempDir` instead of OKIO's `FileSystem.deleteRecursively` and it also works fine. If I'm honest, I don't recall why we reached the conclusion that we needed to use OKIO 😑  although after this PR it might make sense to use it if we keep having the need for handling more IO scenarios.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * [Sync] `SyncConfiguration` and `SyncSession` has been moved to `io.realm.mongodb.sync`.
 * [Sync] `EmailPasswordAuth` has been movedto `io.realm.mongodb.auth`.
 * [Sync] Improved exception hiearchy for App and Sync exceptions. All sync/app exceptions now use `io.realm.mongodb.exceptions.AppException` as their top-level exception type. Many methods have more specialized exceptions for common errors that can be caught and reacted to. See `AppException` documentation for more details.
+* [Sync] `SyncConfiguration.directory` is no longer available and `AppConfiguration.syncRootDirectory` has been added instead to allow users to set the root folder containing all files used for data synchronization between the device and MongoDB Realm. (Issue [#795](https://github.com/realm/realm-kotlin/issues/795))
```
cmelchior, line 7: ``SyncConfiguration.directory` is no longer available` this should move to `Breaking Changes`

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
```diff
@@ -106,6 +113,47 @@ public interface AppConfiguration {
...
             }
 
+        /**
+         * Configures the root folder containing all files and realms used when synchronizing data
+         * between the device and MongoDB Realm.
+         *
+         * The default root directory is platform-dependent:
+         * ```
+         * // For Android the default directory is obtained using
+         * val dir = ${Context.getFilesDir()/mongodb-realm}
```
cmelchior, line 123: Nitpick...but this isn't actually valid Kotlin code...these should be encapsulated in `""` for that.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -242,25 +241,74 @@ class SyncConfigTests {
...
-//    @Test
-//    fun name_illegalValuesThrows() {
-//        val user: User = createTestUser(app)
-//        val builder = SyncConfiguration.Builder(user, DEFAULT_PARTITION)
-//
-//        assertFailsWith<IllegalArgumentException> { builder.name(TestHelper.getNull()) }
-//        assertFailsWith<IllegalArgumentException> { builder.name(".realm") }
-//    }
+    @Test
+    fun name_withoutFileExtension() {
```
cmelchior, line 245: We are missing a test for what the default value would look like.

cmelchior (commented):
Nice...just left a few comments. Will do a final review once the name changes around the file extension we discussed has landed

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncConfigTests.kt
edualonso: I added `name_similarToDefaultObjectStoreName` below, if that's what you mean?

CHANGELOG.md
edualonso: I'm not sure what you mean by moving this to breaking changes. This is the breaking changes section.

I will move the part about `AppConfiguration.syncRootDirectory` to the `Fixed` section.

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/AppConfiguration.kt
edualonso: 🙈

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncConfigTests.kt
cmelchior: Yes, but that also passes in a user-defined name here: https://github.com/realm/realm-kotlin/pull/800/files/2c8370b257770e55e75b92ee8e731b6afb380939#diff-f3172b56e038676ca6061b7181e7eaae31885fe86018ccc332e00d2274b3524dR297

We are missing a test for what happens if no name is given.

CHANGELOG.md
cmelchior: Ups 🙈 ... for some reason I thought this was under Enhancements.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncConfigTests.kt
edualonso: All the other tests in the suite that are not name-related don't specify a name so that should already be covered, no? On top of that, if Object Store assigns somewhat "random" names internally, to which don't have access until we instantiate the config object itself, what is it that we should test? I don't see how to compare the resulting name to an expected value.

cmelchior: But we do know....partition values would be encoded with something like `s_<value>.realm` for Strings `i_<value>.realm` for Ints

The reason we need would to be ensure it was mostly for testing that the suffix `.realm` appear for these as well.

cmelchior (approved):
Awesome :shipit: ... I had a few minor comments to tests and documentation but addressed them in a commit instead.

This can be merged if CI is happy.

PR #817, opened 2022-04-29
Cleanup HttpEngine in tests

### LINKED ISSUE #814: Broken tests on CI: io.realm.test.shared.SyncedRealmTests
This test seems to fail consistently on CI only: 
https://ci.realm.io/job/realm/job/realm-kotlin/job/PR-800/19/consoleFull

Attempting to reproduce locally has failed so far.

In order to prevent us from being completely blocked, I have disabled the test, but we should do a more in-depth investigation.

Right now my only hypothesis is either Ktor acting up on the server or some weirdness with coroutines and channels

cmelchior:
Closes #814 

This PR introduces a `close()` method for the NetworkTransport. This allows us to close the underlying HttpEngine, which seems to be required on macOS, as creating many engines eventually causes requests to timeout. Probably because some underlying resources is shared and not closed correctly.

`TestApp` has been reworked to correctly close the engines it uses but it is not doing anything for the public API.

In the public API, the NetworkTransport is tied to `AppConfiguration` which is a problem since we cannot close it from the App/Realm instances as it might be shared between many Apps/Realms. We need to figure out a pattern that allows people to do this, perhaps by lazily creating the network transport when needed.

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/NetworkTransport.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/internal/interop/sync/NetworkTransport.kt
@@ -36,6 +36,12 @@ interface NetworkTransport {
         body: String,
         callback: ResponseCallback
     )
+
+    /**
+     * Close any native resources associated with running a NetworkClient.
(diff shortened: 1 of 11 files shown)


edualonso (approved):
LGTM

PR #822, opened 2022-05-02
Added `@SharedImmutable` annotation to freeze values

### LINKED ISSUE #799: IncorrectDereferenceException: Trying to access top level value not marked as @ThreadLocal or @SharedImmutable from non-main thread
Hello guys . I have an exception and crash only for iOs when I config my Real from non-main thread :( . Take a look pls to source code:
```
@ThreadLocal
object RealmManager {

    val realmDb by lazy { initRealmAsync() }

    private fun initRealmAsync(): Deferred<Realm> {
        return CoroutineScope(Dispatchers.Default).async {
            val config = RealmConfiguration.Builder(
                schema = setOf(
                    UserEntity::class,
                    EquipmentEntity::class,
                    UserActionEntity::class))
                .log(LogLevel.ALL)
                .schemaVersion(1)
                .build()
            Realm.open(config)
        }
    }

    fun forceInit() {
        CoroutineScope(Dispatchers.Default).launch {
            realmDb.await()
        }
    }
}
```

if I calling `forceInit()` from iOs it crashes . This code in common module . For Android works fine .
With `Dispatchers.Main` all good.
Thanks .

edualonso:
Closes #799

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * None.
 
 ### Fixed
-* None.
+* Creating a `RealmConfiguration` off the main thread on Kotlin Native could crash with `IncorrectDereferenceException`. (Issue [#799](https://github.com/realm/realm-kotlin/issues/799))
 
 ### Compatibility
 * This release is compatible with:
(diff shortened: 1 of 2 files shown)


cmelchior (approved):
We should probably add unit tests for this, but I hit the error in this PR https://github.com/realm/realm-kotlin/pull/833, and this fixed it. So I guess that is test enough :D 

But we do need a CHANGELOG update.

PR #825, opened 2022-05-04
Add support for Flexible Sync to test apps and on CI

cmelchior:
Splitting #824 into smaller pieces.

This PR adds a test app configured with Flexible Sync that can be used for testing, locally and on CI. It mirrors the setup in Realm Java.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -479,7 +479,8 @@ def testWithServer(tasks) {
             }
             def commandServerEnv = docker.build 'mongodb-realm-command-server', "tools/sync_test_server"
             def tempDir = runCommand('mktemp -d -t app_config.XXXXXXXXXX')
-            sh "tools/sync_test_server/app_config_generator.sh ${tempDir} tools/sync_test_server/app_template testapp1 testapp2"
+            sh "tools/sync_test_server/app_config_generator.sh ${tempDir} tools/sync_test_server/app_template partition testapp1 testapp2"
+            sh "tools/sync_test_server/app_config_generator.sh ${tempDir} tools/sync_test_server/app_template flex testapp3"
 
             sh "docker network create ${dockerNetworkId}"
             mongoDbRealmContainer = mdbRealmImage.run("--rm -i -t -d --network ${dockerNetworkId} -v$tempDir:/apps -p9090:9090 -p8888:8888 -p26000:26000 -e AWS_ACCESS_KEY_ID='$BAAS_AWS_ACCESS_KEY_ID' -e AWS_SECRET_ACCESS_KEY='$BAAS_AWS_SECRET_ACCESS_KEY'")
--- a/dependencies.list
+++ b/dependencies.list
@@ -1,3 +1,3 @@
 # Version of MongoDB Realm used by integration tests
 # See https://github.com/realm/ci/packages/147854 for available versions
-MONGODB_REALM_SERVER=2021-03-30
+MONGODB_REALM_SERVER=2022-05-04
--- a/test/sync/src/commonMain/kotlin/io/realm/entities/sync/flx/FlexChildObject.kt
+++ b/test/sync/src/commonMain/kotlin/io/realm/entities/sync/flx/FlexChildObject.kt
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
- package io.realm.entities.sync.flx
+package io.realm.entities.sync.flx
 
 import io.realm.RealmObject
 
 /**
  * Object used when testing Flexible Sync.
  */
-class FlexChildObject: RealmObject {
+class FlexChildObject : RealmObject {
     var section: Int = 0
     var name: String = ""
 }
--- a/test/sync/src/commonMain/kotlin/io/realm/entities/sync/flx/FlexChildObject.kt
+++ b/test/sync/src/commonMain/kotlin/io/realm/entities/sync/flx/FlexChildObject.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2022 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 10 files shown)


cmelchior:
I think I hacked my way around all the race conditions now. At least to a degree where this can be reviewed. Will be rerunning the tests a few times to see if anything is flaky.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -181,55 +193,56 @@ class EmailPasswordAuthTests {
...
             try {
-                provider.retryCustomConfirmation("foo")
+                provider.retryCustomConfirmation("foo@gen.com")
                 fail()
             } catch (error: UserNotFoundException) {
                 assertTrue(error.message!!.contains("user not found"), error.message)
             } finally {
                 adminApi.setCustomConfirmation(false)
+                adminApi.setAutomaticConfirmation(true)
             }
```
edualonso, line 226:
```suggestion
            assertFailsWith<UserNotFoundException> {
                provider.retryCustomConfirmation("foo@gen.com")
            }.also { error ->
                assertTrue(error.message!!.contains("user not found"), error.message)
            }
            adminApi.setCustomConfirmation(false)
            adminApi.setAutomaticConfirmation(true)
```

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -181,55 +193,56 @@ class EmailPasswordAuthTests {
         // Only emails containing realm_tests_do_autoverify will be confirmed
         val email = "do_not_confirm@10gen.com"
         app.setAutomaticConfirmation(false)
+        app.setCustomConfirmation(true)
         try {
```
edualonso, line 197: You shouldn't need this `try`-`finally` if you are using `assertFailsWith`. Clean up after the assertion normally.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -181,55 +193,56 @@ class EmailPasswordAuthTests {
...
     }
 
     @Test
     fun retryCustomConfirmation_alreadyConfirmedThrows() {
-        val email = "test_realm_tests_do_autoverify@10gen.com"
+        val email = "realm_verify_${Random.nextULong()}@10gen.com"
         val adminApi = app.asTestApp
         runBlocking {
-            adminApi.setAutomaticConfirmation(false)
             try {
```
edualonso, line 235: Same

edualonso (approved):
Looks good, though the race condition hacks are pretty... hacky 🙈

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/util/AdminApi.kt
```diff
@@ -305,16 +305,27 @@ open class AdminApiImpl internal constructor(
...
-        }.let {
-            if (!it.status.isSuccess()) {
-                throw IllegalStateException("PATCH request failed: $it")
+
+        // It is unclear exactly what is happening, but if we only send the request once
+        // it appears as the server accepts it, but is delayed deploying the changes,
+        // i.e. the change will appear correct in the UI, but later requests against
+        // the server will fail in a way that suggest the change wasn't applied after all.
+        // Sending these requests twice seems to fix most race conditions.
+        repeat(2) {
```
nhachicha, line 314: Couldn't this be solved instead by waiting for the change to be applied? sending the request twice may introduce further flakiness in the future, as we don't know exactly the root cause of the delay ...

tools/sync_test_server/app_template/functions/confirmFunc/source.js
```diff
@@ -38,12 +38,26 @@
...
     // process the confirm token, tokenId and username
-    if (username.includes("realm_tests_do_autoverify")) {
+
+    if (username.includes("realm_verify")) {
+      // Automatically confirm users with `realm_verify` in their email.
       return { status: 'success' }
+    } else if (username.includes("realm_pending")) {
+      // Emails with `realm_pending` in their email will be placed in Pending
+      // the first time they register and will then be fully confirmed when
+      // they retry their their confirmation logic.
```
nhachicha, line 50:
```suggestion
      // they retry their confirmation logic.
```

test/sync/src/commonMain/kotlin/io/realm/test/mongodb/util/AdminApi.kt
cmelchior: That was actually what I originally did, but even waiting for 5-10 seconds didn't improve things + it makes the tests quite a lot slower. For some reason sending the request twice works better :/

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
cmelchior: Actually, this will do the wrong thing if an assertion is thrown, then the state in 224/225 will not be reset correctly which will impact other tests.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
cmelchior: Same as above, we need the finally to be able to reset the state if this test fails.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/EmailPasswordAuthTests.kt
edualonso: If an exception is thrown within the `assertFailsWith` block the test won't stop and whatever comes after that will be executed. Am I missing something else?

edualonso: Oh, I completely missed the call to `fail()` - I see it now 🙈

PR #828, opened 2022-05-05
Fix list notifications operation on the wrong realm reference

### LINKED ISSUE #827: Crash when collecting RealmObjects
Hello,

I'm using the `0.11.0-SNAPSHOT` version of the SDK.

I'm trying to collect some changes in a RealmList.
It's working when data are removed from the list, but if some items are updated or added, it crashes with this exception:
```
java.lang.RuntimeException: Could not get element at list index 1: RealmCoreException([20]: No object with key '66' in 'class_Thread')
```

As you will see in the logs, the number of items in the list is correct: 1 item the 1st time the collect triggers, 3 the 2nd time.
I'm able to log the 1st item both time, as expected.
But when I try to log the other items, that's when it crashes.

I wrote a small sample to reproduce the situation:

```kotlin
import android.os.Bundle
import android.util.Log
import androidx.appcompat.app.AppCompatActivity
import androidx.lifecycle.lifecycleScope
import io.realm.*
import io.realm.MutableRealm.UpdatePolicy
import io.realm.annotations.PrimaryKey
import kotlinx.coroutines.launch
import java.util.*

    private class Folder : RealmObject {
        @PrimaryKey
        var id: String? = null
        var threads: RealmList<Thread> = realmListOf()
    }

    private class Thread : RealmObject {
        @PrimaryKey
        var id: String? = null
        var subject: String? = null
    }

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)

        val config = RealmConfiguration.Builder(
            setOf(Folder::class, Thread::class)
        ).build()

        val realm = Realm.open(config)

        // Delete everything to have a clean start
        deleteRealmContent(realm)

        // Create folder
        val folder = createFolder(realm)

        // Get Realm folder
        val realmFolder = realm.query<Folder>("${Folder::id.name} == '${folder.id}'").first().find() ?: return

        // Listen to Threads changes
        collectThreads(realmFolder)

        // Populate folder
        addThreadsToFolder(realm, folder)
    }

    private fun deleteRealmContent(realm: Realm) {
        with(realm) {
            val threadsToDelete = query<Thread>().find()
            val foldersToDelete = query<Folder>().find()
            writeBlocking {
                threadsToDelete.mapNotNull(::findLatest).forEach(::delete)
                foldersToDelete.mapNotNull(::findLatest).forEach(::delete)
            }
        }
    }

    private fun createFolder(realm: Realm): Folder {
        val folder = Folder().apply { id = "Folder_1"; threads = getThreads(1) }
        realm.writeBlocking { copyToRealm(folder, UpdatePolicy.ALL) }
        return folder
    }

    private fun collectThreads(folder: Folder) {
        lifecycleScope.launch {
            folder.threads.asFlow().collect { listChange ->
                Log.i("Threads", "Size: ${listChange.list.size}")
                listChange.list.forEachIndexed { index, thread ->
                    Log.d("Threads", "Index: $index - Subject: ${thread.subject}")
                }
            }
        }
    }

    private fun addThreadsToFolder(realm: Realm, folder: Folder) {
        folder.threads = getThreads(3)
        realm.writeBlocking { copyToRealm(folder, UpdatePolicy.ALL) }
    }

    private fun getThreads(nbThreads: Int): RealmList<Thread> {
        val thread1 = Thread().apply { id = "Thread_1"; subject = "Subject of 1st thread_${UUID.randomUUID()}" }
        val thread2 = Thread().apply { id = "Thread_2"; subject = "Subject of 2nd thread_${UUID.randomUUID()}" }
        val thread3 = Thread().apply { id = "Thread_3"; subject = "Subject of 3rd thread_${UUID.randomUUID()}" }
        val thread4 = Thread().apply { id = "Thread_4"; subject = "Subject of 4th thread_${UUID.randomUUID()}" }
        val thread5 = Thread().apply { id = "Thread_5"; subject = "Subject of 5th thread_${UUID.randomUUID()}" }

        return when (nbThreads) {
            1 -> listOf(thread1)
            2 -> listOf(thread1, thread2)
            3 -> listOf(thread1, thread2, thread3)
            4 -> listOf(thread1, thread2, thread3, thread4)
            else -> listOf(thread1, thread2, thread3, thread4, thread5)
        }.toRealmList()
    }
}
```

rorbech:
Closes #827

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmListInternal.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmListInternal.kt
@@ -31,6 +31,7 @@ import io.realm.notifications.internal.UpdatedListImpl
 import kotlinx.coroutines.channels.ChannelResult
 import kotlinx.coroutines.channels.SendChannel
 import kotlinx.coroutines.flow.Flow
+import kotlin.reflect.KClass
 
 /**
  * Implementation for unmanaged lists, backed by a [MutableList].
@@ -49,7 +50,7 @@ internal class UnmanagedRealmList<E> : RealmList<E>, InternalDeleteable, Mutable
  */
 internal class ManagedRealmList<E>(
     internal val nativePointer: RealmListPointer,
-    private val metadata: ListOperatorMetadata<E>
+    private val metadata: ListOperatorMetadata<E>,
 ) : AbstractMutableList<E>(), RealmList<E>, InternalDeleteable, Observable<ManagedRealmList<E>, ListChange<E>>, Flowable<ListChange<E>> {
     override val size: Int
         get() {
@@ -133,13 +134,15 @@ internal class ManagedRealmList<E>(
 
     override fun freeze(frozenRealm: RealmReference): ManagedRealmList<E>? {
         return RealmInterop.realm_list_resolve_in(nativePointer, frozenRealm.dbPointer)?.let {
-            ManagedRealmList(it, metadata.copy(realm = frozenRealm))
+            val converter: RealmValueConverter<E> = converter<Any>(metadata.clazz, metadata.mediator, frozenRealm) as CompositeConverter<E, *>
+            ManagedRealmList(it, metadata.copy(realm = frozenRealm, converter = converter))
         }
     }
 
     override fun thaw(liveRealm: RealmReference): ManagedRealmList<E>? {
         return RealmInterop.realm_list_resolve_in(nativePointer, liveRealm.dbPointer)?.let {
-            ManagedRealmList(it, metadata.copy(realm = liveRealm))
+            val converter: RealmValueConverter<E> = converter<Any>(metadata.clazz, metadata.mediator, liveRealm) as CompositeConverter<E, *>
+            ManagedRealmList(it, metadata.copy(realm = liveRealm, converter = converter))
         }
     }
 
@@ -185,6 +188,7 @@ internal class ManagedRealmList<E>(
 internal data class ListOperatorMetadata<E>(
     val mediator: Mediator,
     val realm: RealmReference,
+    val clazz: KClass<*>,
     val converter: RealmValueConverter<E>
 )
 
--- a/test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
+++ b/test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
@@ -304,7 +304,7 @@ class RealmListTests {
                 .takeWhile { it.list.size < 5 }
                 .collect {
                     it.list.forEach {
-                        // No-op ... just verifying that we can access each element
+                        // No-op ... just verifying that we can access each element. See https://github.com/realm/realm-kotlin/issues/827
                     }
                 }
         }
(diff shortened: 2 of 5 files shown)


test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
```diff
@@ -293,6 +296,25 @@ class RealmListTests {
         assertEquals(2, realm.query<RealmListContainer>().find().size)
     }
 
+    @Test
+    fun listNotifications() = runBlocking {
+        val container = realm.writeBlocking { copyToRealm(RealmListContainer()) }
+        val collect = async {
+            container.objectListField.asFlow()
+                .takeWhile { it.list.size < 5 }
```
cmelchior, line 304: This looks wrong? It will take the empty initial list and then complete? Shouldn't this be `it.list.size == 1` ?

cmelchior (changes requested):
Looks good, but we should probably retarget releases + I had a few questions about the test.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
rorbech: `takeWhile` collects until the condition is no longer met at which time the flow will be completed!? 

So we will keep on adding elements until there are five elements in the list and stop the test. ??

packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmListInternal.kt
```diff
@@ -133,13 +134,15 @@ internal class ManagedRealmList<E>(
...
+            val converter: RealmValueConverter<E> = converter<Any>(metadata.clazz, metadata.mediator, frozenRealm) as CompositeConverter<E, *>
+            ManagedRealmList(it, metadata.copy(realm = frozenRealm, converter = converter))
         }
     }
 
     override fun thaw(liveRealm: RealmReference): ManagedRealmList<E>? {
         return RealmInterop.realm_list_resolve_in(nativePointer, liveRealm.dbPointer)?.let {
-            ManagedRealmList(it, metadata.copy(realm = liveRealm))
+            val converter: RealmValueConverter<E> = converter<Any>(metadata.clazz, metadata.mediator, liveRealm) as CompositeConverter<E, *>
+            ManagedRealmList(it, metadata.copy(realm = liveRealm, converter = converter))
```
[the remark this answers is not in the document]
rorbech: Yeah. I have already changed it to an explicit interface method in https://github.com/realm/realm-kotlin/pull/797/files#diff-6ae484901f6301d8cce7b3b571876dbdefc491334d69fe44ec1f77ce69a96acfR212, but since this is a hot-fix I just added it to the existing data class's copy method. We are only copying the metadata when thawing/freezing so think it should be ok for the time being and until #797 is merged.

rorbech:
> Looks good, but we should probably retarget releases + I had a few questions about the test.

I was already doing the fix in `releases`, just forgot to target that. I have tried to put some clarifying comments on the concerns.

test/base/src/androidTest/kotlin/io/realm/test/shared/RealmListTests.kt
cmelchior: Ah yeah, I was reading the test wrong 👍

cmelchior (approved):
:shipit:

PR #839, opened 2022-05-21
Add support for writing initial data when opening a Realm file.

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/302
Closes https://github.com/realm/realm-kotlin/issues/422

Add support Configuration.initialData(), which works for both RealmConfiguration and SyncConfiguration. Similar to how it works in Realm Java.

This also sets a pattern for adding other functionality required to run at startup, like `initialSubscriptions` which was the reason for adding this now.

**Alternative designs**

The current design extends `realm_open`, which works because a previous PR changed our native configs, so they are created just before being needed, thus ensuring they are tied to a single Realm.

At first, I did attempt a different approach where the `DataInitialization` callback was passed into `InternalConfiguration.createNativeConfiguration()`, but it just made call sites really awkward since you needed a way to pass the boolean across thread boundaries on Native. 

While the `DataInitialization` callback, in theory, can provide a `SharedRealm` pointer, the rest of our code is not constructed in a way that makes it possible to use our API's until _after_ `realm_open` returns, so instead we just use the fact that `DataInitialization` is triggered as a signal that the file was recreated, allowing us to run any code we want _after_ the Realm is opened. This will also enable API's like `SyncConfiguration.waitForInitialRemoteData()` and `SyncConfiguration.initialSubscriptions()`

### DIFF
--- a/packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
@@ -157,9 +157,17 @@ actual object RealmInterop {
         realmc.realm_config_set_data_initialization_function(config.cptr(), callback)
     }
 
-    actual fun realm_open(config: RealmConfigurationPointer, dispatcher: CoroutineDispatcher?): LiveRealmPointer {
-        // create a custom Scheduler for JVM if a Coroutine Dispatcher is provided other wise pass null to use the generic one
+    actual fun realm_open(config: RealmConfigurationPointer, dispatcher: CoroutineDispatcher?): Pair<LiveRealmPointer, Boolean> {
+        // Configure callback to track if the file was created as part of opening
+        var fileCreated = false
+        val callback = DataInitializationCallback {
+            fileCreated = true
+            true
+        }
+        realm_config_set_data_initialization_function(config, callback)
 
+        // create a custom Scheduler for JVM if a Coroutine Dispatcher is provided other wise
+        // pass null to use the generic one
         val realmPtr = LongPointerWrapper<LiveRealmT>(
             realmc.open_realm_with_scheduler(
                 (config as LongPointerWrapper).ptr,
@@ -168,7 +176,7 @@ actual object RealmInterop {
         )
         // Ensure that we can read version information, etc.
         realm_begin_read(realmPtr)
-        return realmPtr
+        return Pair(realmPtr, fileCreated)
     }
 
     actual fun realm_add_realm_changed_callback(realm: LiveRealmPointer, block: () -> Unit): RealmCallbackTokenPointer {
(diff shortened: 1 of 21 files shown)


packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
```diff
@@ -157,9 +157,17 @@ actual object RealmInterop {
         realmc.realm_config_set_data_initialization_function(config.cptr(), callback)
     }
 
-    actual fun realm_open(config: RealmConfigurationPointer, dispatcher: CoroutineDispatcher?): LiveRealmPointer {
-        // create a custom Scheduler for JVM if a Coroutine Dispatcher is provided other wise pass null to use the generic one
+    actual fun realm_open(config: RealmConfigurationPointer, dispatcher: CoroutineDispatcher?): Pair<LiveRealmPointer, Boolean> {
+        // Configure callback to track if the file was created as part of opening
+        var fileCreated = false
```
rorbech, line 162: Should access to this be synchronized? I would have expected all of this to happen on the same thread, but from the Darwin implementation it looks like it isn't 🤔

rorbech (commented):
LGTM - with the exception of the test verifying that `initialData` is running on the `writerThread`. Also had a comment around synchronization on JVM ... maybe related as it is about what thread the callback is running on.

test/base/src/androidTest/kotlin/io/realm/test/shared/InitialDataTests.kt
```diff
@@ -0,0 +1,153 @@
...
+        val startThread: ULong = threadId()
+        val initialDataThread: AtomicRef<ULong?> = atomic(null)
+        val config = configBuilder.initialData {
+            initialDataThread.value = threadId()
+        }.build()
+
+        val writerThread: AtomicRef<ULong?> = atomic(null)
+        Realm.open(config).use {
+            it.writeBlocking {
+                writerThread.value = threadId()
```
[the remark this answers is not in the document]
rorbech: I must have misread this completely as if this was the `initialData`-block. Sorry. Just disregard the comment 😞

packages/library-base/src/commonMain/kotlin/io/realm/Configuration.kt
```diff
@@ -44,6 +44,17 @@ public fun interface CompactOnLaunchCallback {
...
 
+/**
+ * This interface is used to write data to a Realm file when the file is first created.
+ * It will be used in a way similar to using [Realm.writeBlocking].
+ *
+ * Note that writing data to a Realm file will involve IO, so it should generally only be done as
+ * part of opening a Realm on a background thread.
+ */
+public fun interface InitialDataCallback {
+    public fun MutableRealm.write()
```
rorbech, line 55: I think we need some explicit documentation of the actual method in the interface too. I like how this looks when using it, but it is a bit tricky to see how it is used it you are not very familiarized with our API. Maybe even more explicit if it included a small example to make it obvious how to use the `MutableRealm`-receiver.

packages/cinterop/src/jvm/kotlin/io/realm/internal/interop/RealmInterop.kt
cmelchior: It is happening on the same thread. but on Native Kotlin had a compile check around capturing lambdas + I needed to freeze the callback on Native so the GC would work.

rorbech (approved):

PR #844, opened 2022-05-23
Upgrade to latest core

rorbech:
Fixing Swig multi argument typemaps so that they match arguments according to latest core updates.


cmelchior (approved):
If CI is happy

PR #848, opened 2022-05-26
Add schema roundtrip test

clementetb:
Tests that we can roundtrip a schema with all the supported types.


cmelchior:
This should be ready for review.

I'm a bit unsure if there is a better way of making sure we test all the field types for Sync, but it should throw if we add more types, but we might consider adding something that fails if we add more collection types? 🤔

cmelchior:
@clementetb I cannot assign you as reviewer since you started the PR, but you should probably also take a look.

clementetb (commented):
Great solution to test all types. A core data structures list like the types descriptor one could help extending this pattern to such data structures.

Cannot approve it because I am the original author.

test/sync/src/androidTest/kotlin/io/realm/test/shared/SyncedRealmTests.kt
```diff
@@ -272,6 +279,49 @@ class SyncedRealmTests {
...
+        }
+        createSyncConfig(
+            user = user2,
+            partitionValue = partitionValue,
+            schema = setOf(SyncObjectWithAllTypes::class)
+        ).let { config ->
+            Realm.open(config).use { realm ->
+                val obj: SyncObjectWithAllTypes = realm.query<SyncObjectWithAllTypes>("_id = $0", id)
+                    .asFlow()
+                    .filter {
```
rorbech, line 313: Just use the predicate variant of `first` instead ;)

test/sync/src/commonMain/kotlin/io/realm/entities/sync/SyncObjectWithAllTypes.kt
```diff
@@ -0,0 +1,260 @@
...
+         */
+        @Suppress("LongMethod")
+        private fun createMapper(): Map<TypeDescriptor.CoreFieldType, Pair<FieldDataFactory, FieldValidator>> {
+            val map: MutableMap<TypeDescriptor.CoreFieldType, Pair<FieldDataFactory, FieldValidator>> = mutableMapOf()
+            // TODO Problems with accessing TypeDescriptor enums before class is fully initialized
+            //  work-around for now by accessing a property
+            TypeDescriptor.classifiers
+            TypeDescriptor.CoreFieldType.values().forEach { type: TypeDescriptor.CoreFieldType ->
+                map[type] = when (type) {
+                    TypeDescriptor.CoreFieldType.INT -> {
```
rorbech, line 98: Not important, but could have put all this in a map with `to` constructors to separate definition from implementation and then just do a generic implementation looking things up ... just sometimes easier to grasp as it is easier to keep an overview on the logic 🤷  
```
TypeDescriptor.CoreFieldType.Int to (
    { obj: SyncObjectWithAllTypes ->
                                obj.intField = 42
                                obj.intNullableField = 42
                                obj.intRealmList = realmListOf(42)
    } to
    { obj: SyncObjectWithAllTypes ->
                                assertEquals(42, obj.intField)
                                assertEquals(42, obj.intNullableField)
                                assertEquals(42, obj.intRealmList.first())
     }
)
```

rorbech (approved):
Only some minor style comments

test/sync/src/commonMain/kotlin/io/realm/entities/sync/SyncObjectWithAllTypes.kt
```diff
@@ -0,0 +1,260 @@
...
+
+    // Nullable RealmLists of primitive values, not currently supported by Sync
+    // Nullable Object lists, not currently supported by Core
+
+    companion object {
+
+        // Mapping between each Core Field type and functions that can insert data for that type
+        // and also verify the value. This can be used to test objects that has been roundtripped
+        // through Sync.
+        val mapper: Map<TypeDescriptor.CoreFieldType, Pair<FieldDataFactory, FieldValidator>> = createMapper()
```
rorbech, line 85: We now have the user facing `RealmStorageType` so if you don't need anything internal things from `CoreFieldType` you could just use that.

test/sync/src/commonMain/kotlin/io/realm/entities/sync/SyncObjectWithAllTypes.kt
```diff
@@ -0,0 +1,260 @@
...
+
+        /**
+         * Create mapping for both creating and validating sample data
+         */
+        @Suppress("LongMethod")
+        private fun createMapper(): Map<TypeDescriptor.CoreFieldType, Pair<FieldDataFactory, FieldValidator>> {
+            val map: MutableMap<TypeDescriptor.CoreFieldType, Pair<FieldDataFactory, FieldValidator>> = mutableMapOf()
+            // TODO Problems with accessing TypeDescriptor enums before class is fully initialized
+            //  work-around for now by accessing a property
+            TypeDescriptor.classifiers
```
nhachicha, line 95: left over?

nhachicha (approved):
Very nice approach with `createWithSampleData` & `compareAgainstSampleData`. I wonder if it's worth doing the same round trip for the embedded objects? cc @rorbech  may just add a field in `SyncObjectWithAllTypes` enclosing the embedded object

test/sync/src/commonMain/kotlin/io/realm/entities/sync/SyncObjectWithAllTypes.kt
```diff
@@ -0,0 +1,260 @@
...
+    var intField: Int = 0
+    var longField: Long = 0
+    var booleanField: Boolean = true
+    var doubleField: Double = 0.0
+    var floatField: Float = 0.0.toFloat()
+    var realmInstantField: RealmInstant = RealmInstant.MIN
+    var objectIdField: ObjectId = ObjectId.create()
+    var objectField: SyncObjectWithAllTypes? = null
+
+    // Nullable types
```
nhachicha, line 49: should we include `RealmList` of nullable as well for each type?

cmelchior: They are not supported by Sync: https://github.com/realm/realm-core/issues/5545... I'll add a comment here about it.

PR #850, opened 2022-05-30
Implement pausing and resuming sessions

clementetb:
This PR allows to control a SyncSession lifecycle by implementing the methods `pause` and `resume`.

They are required to be able to simulate a client reset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [Sync] Added support for `ObjectId` ([#652](https://github.com/realm/realm-kotlin/issues/652)). `ObjectId` can be used as a primary key in model definition.
 * [Sync] Support for `SyncConfiguration.Builder.InitialData()`. (Issue [#422](https://github.com/realm/realm-kotlin/issues/422))
 * Support for `RealmConfiguration.Builder.initialData()`. (Issue [#579](https://github.com/realm/realm-kotlin/issues/579))
+* [Sync] Support for accessing and controlling the session state through `SyncSession.state`, `SyncSession.pause()` and `SyncSession.resume()`.
 
 ### Fixed
 * Creating a `RealmConfiguration` off the main thread on Kotlin Native could crash with `IncorrectDereferenceException`. (Issue [#799](https://github.com/realm/realm-kotlin/issues/799))


packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/SyncSession.kt
```diff
@@ -68,6 +68,34 @@ public interface SyncSession {
      */
     public suspend fun uploadAllLocalChanges(timeout: Duration = Duration.INFINITE): Boolean
 
+    /**
+     * The current SyncSession state. See [SyncSessionState] for more details about each state.
```
cmelchior, line 72:
```suggestion
     * The current session state. See [SyncSessionState] for more details about each state.
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/SyncSession.kt
```diff
@@ -68,6 +68,34 @@ public interface SyncSession {
...
+     */
+    public val state: SyncSessionState
+
+    /**
+     * Pauses any synchronization with the Realm Object Server until the Realm is re-opened again
+     * after fully closing it.
+     * <p>
+     * Synchronization can be re-activated by calling [resume] again.
+     * <p>
+     * If the session is already [SyncSessionState.INACTIVE], calling this method will do nothing.
```
cmelchior, line 82:
```suggestion
     * Pauses synchronization with Atlas until the Realm is closed and re-opened again.
     * 
     * Synchronization can also be re-activated by calling [resume].
     *
     * If the session is already [SyncSessionState.INACTIVE], calling this method will do nothing.
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/SyncSession.kt
```diff
@@ -68,6 +68,34 @@ public interface SyncSession {
...
+    /**
+     * Attempts to resume the session and activate synchronization with the Realm Object Server.
+     * <p>
+     * This happens automatically when opening the Realm instance, so doing it manually should only
+     * be needed if the session was paused using [pause].
+     * <p>
+     * If the session was already [SyncSessionState.ACTIVE], calling this method will do nothing.
+     * <p>
+     * If the session state is [SyncSessionState.DYING], the session will be moved back to
+     * [SyncSessionState.ACTIVE].
```
cmelchior, line 95:
```suggestion
     * Attempts to resume the session and activate synchronization with Atlas.
     *
     * This happens automatically when opening the Realm, so doing it manually should only
     * be needed if the session was paused using [pause].
     * 
     * If the session was already [SyncSessionState.ACTIVE], calling this method will do nothing.
     * 
     * If the session state is [SyncSessionState.DYING], the session will be moved back to
     * [SyncSessionState.ACTIVE].
```

packages/library-sync/src/commonMain/kotlin/io/realm/mongodb/sync/SyncSession.kt
```diff
@@ -86,3 +114,13 @@ public interface SyncSession {
         public fun onError(session: SyncSession, error: SyncException)
     }
 }
+
+/**
+ * The possible states for [SyncSession] to be.
+ */
+public enum class SyncSessionState {
```
cmelchior, line 121: There is a lot of documentation in Realm Java we can move over for these.

cmelchior: I would probably move into`SyncSession.State`, similar to `User.State` and how it is done in Realm Java.

test/sync/src/androidTest/kotlin/io/realm/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -87,6 +88,25 @@ class SyncSessionTests {
         }
     }
 
+    @Test
+    fun sessionPauseAndResume() {
```
cmelchior, line 92: There are a few more tests in https://github.com/realm/realm-java/blob/f0eca830527cd8255ce8fb9bcc83a545a940f6d3/realm/realm-library/src/syncIntegrationTest/kotlin/io/realm/SyncSessionTests.kt#L729 we should probably move as well (i.e. calling pause/resume multiple times)

cmelchior (changes requested):
Nice to get this in. I added some comments for some smaller adjustments I think we should make and tests from Java we should port...and also the CHANGELOG ;)

cmelchior (approved):
Looks good to me. If CI is happy 👍

Still need a changelog update though ;)

clementetb:
🙈

PR #854, opened 2022-05-31
Preparing the compiler plugin to be compatible with Kotlin 1.7.0-RC

nhachicha:
The usage of `IrConstructorCallImpl` has changed in Kotlin 1.7.0-RC. To be compatible with the upcoming release we switch to using the `IrConstructorCallImpl.fromSymbolOwner` builder.


cmelchior (approved):
I don't really know what that code is doing 🙈, and the Compiler API doesn't seem to have any docs as usual. But if the tests pass and our samples compile I assume it is solid.

Great find 💯

rorbech (approved):
Difficult to reason about ... but if it works!

PR #856, opened 2022-06-01
Fixing a compiler plugin crash when using RealmList on Kotlin 1.7.0-RC

nhachicha:
While using the latest snapshot and updating the samples to use Kotlin `1.7.0-RC`, the compiler plugin crashes on apps using a model with `RealmList`. It turns out that getting the enclosing type of the `RealmList` i.e 
```Kotlin
CollectionType.LIST -> ((backingField.type as IrSimpleType).arguments[0] as IrTypeBase).type
```
Causes the compiler to throw 
```
java.lang.IncompatibleClassChangeError: Found class org.jetbrains.kotlin.ir.types.IrSimpleType, but interface was expected
```
This is because `IrSimpleType` became an abstract class in Kotlin 1.7 (see https://github.com/JetBrains/kotlin/commit/53210770a6877c5c08735070f8eff3e33573f0f5). 

- In this PR we use reflection to access the `arguments` field containing the list of the enclosing `IrType`

- This PR also add the `Alltype` classes to the `kmm-sample` used with `latest_kotlin_version` so we can catch similar issues in the future


packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/IrUtils.kt
```diff
@@ -434,6 +438,26 @@ fun IrBlockBuilder.createSafeCallConstruction(
...
 }
 
+/**
+ * Using reflection to invoke the `arguments` attribute of the `IrSimpleType` to determine the enclosing type
+ * of the `RealmList`.
+ * This work around is needed since `IrSimpleType` became an abstract class in Kotlin 1.7 (see https://github.com/JetBrains/kotlin/commit/53210770a6877c5c08735070f8eff3e33573f0f5)
+ * which causes the compiler to throw: "java.lang.IncompatibleClassChangeError: Found class org.jetbrains.kotlin.ir.types.IrSimpleType, but interface was expected"
+ * when the compiler plugin is compiled with Kotlin 1.6.10.
+ */
+fun getCollectionElementType(backingFieldType: IrType): IrType? {
```
cmelchior, line 448: Should we create an issue for tracking these kinds of hacks? Then we can reference them and use that as a guide when we eventually upgrade our minimum supported version of the Kotlin Compiler.

cmelchior (approved):

rorbech (commented):
Can't  we align/reuse our existing code from https://github.com/realm/realm-kotlin/blob/master/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt#L621 here?

rorbech:
> @rorbech https://github.com/realm/realm-kotlin/blob/master/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt#L621 uses `KotlinType` whereas the workaround deals with `IrType`, There's no direct conversion between them to my knowledge since they're used in a different stage (backend/frontend) see https://kotlinlang.slack.com/archives/C7L3JB43G/p1599512927082500?thread_ts=1599501968.080600&cid=C7L3JB43G
> 
> It's also relying on `@OptIn(ObsoleteDescriptorBasedAPI::class)` to access to type via the descriptor which is deprecated `declaration.symbol.descriptor.type`

Ok. One of these uncomfortable workarounds, but as long as we track it ... and can get rid of it relatively soon.

rorbech (approved):

PR #857, opened 2022-06-01
Align configuration factory method naming

### LINKED ISSUE #835: Align factory constructor naming
Across the Realm API there is a wide variety of constructors being used:

```
App.create("appId")
RealmInstant.fromEpochSeconds(seconds, nanoseconds)
RealmConfiguration.with(schema)
ObjectId.from("hexString")
ObjectId.get()
DynamicRealmObject.create(class, map)
```

It is potentially confusing and lowers API discoverability that we are not consistent with this naming. 
After internal discussion we agreed on the following design principle:

* Objects, where the constructor describes the entire state, should use `<Class>.create(args)`. Prefer named constructors or `from` instead of having multiple overloads of `create`
* Instances that are wholly created by mapping another data structure should use the `<Class>.from(arg)` or `<Class>.fromX(args)` name, i.e. `RealmInstant.fromEpochSeconds(sec,nano)`, `ObjectId.from(hexString)`, `RealmUUID.from(bytes)`.
* While the no-arg `create` method is allowed, a more descriptive name should be used if applicable, i.e. `RealmInstant.now()` or `RealmUUID.random()` 

Specifically, this means that we rename our `<X>.with()` constructors. While we do agree that these are probably more descriptive and better communicate that you are getting an object with default values + modifications, we do not feel this outweigh the counter-argument that having `create()` as the go-to name is easier for beginners and people new to the Realm SDK, i.e. the mental model for the API becomes simpler.

Regarding `create` vs `with` there was also the argument that `create` is an explicit action with a clear outcome where `with` is more passive and vague.

So for the current API it means the following changes:

* `RealmConfiguration.with` gets renamed to `RealmConfiguration.create(schema)`
* `SyncConfiguration.with` gets renamed to `SyncConfiguration.create(user, partition, schema)`
* `AppConfiguration.with` gets renamed to `AppConfiguration.create(appId)`
* `RealmInstant` should have a `RealmInstant.now()`
* We forgot to discuss: should we rename `RealmInstant.fromEpochSeconds(seconds, nano)` to `RealmInstant.from()`, with the argument that the arguments should be descriptive enough.
* `ObjectId.get()` gets renamed to`ObjectId.create()`
* `DynamicRealmObject` should use `DynamicRealmObject.create(class, props)`

rorbech:
This PR aligns configuration factory methods according to #835 

Closes #835


cmelchior (approved):
The compiler plugin still have a few broken tests by the looks of it, but once that is fixed, this looks pretty good 👍

nhachicha (approved):
👍 :shipit:

PR #859, opened 2022-06-01
Move modeling types to io.realm.kotlin.types package

### LINKED ISSUE #837: Refactor packages / types in preparation for GA
* Move "modeling" types to `io.realm.kotlin.types`, i.e. ObjectId, RealmList, RealmObject, RealmInstant, RealmUUID, EmbeddedObject, LinkingObjects (property delegate)
* Move RealmResults to `io.realm.kotlin.query`
* Split "fat" files into smaller isolated files, `RealmObject`
* Consider moving extension packages to their own packages, so we don't pollute main packages + make it easier to see where things are coming from
* Figure out what to do with a potential public DynamicRealm API as it impacts the current types.
* Figure out how to import "old" (managed) objects. This probably requires new updateFlags, but any support should not break whatever API we release. This might affect the behavior of setters as well.

rorbech:
This PR cleans up top level package. Primarily by moving all modeling related types to `io.realm.kotlin.types`

Closes #837


nhachicha (approved):
When CI is happy 👍

cmelchior (approved):
Nice...another PR that breaks all other PR's 😆 💯 

Looks good to me. Only concern I have is re-using the filenames in  `io.realm.kotlin.ext` like `Realm`, since that would probably make it slightly confusing if consumed from Java

Maybe rename them to `<X>Ext` or something like that? 

But it isn't a big deal and we can always come back to it.

rorbech:
> Looks good to me. Only concern I have is re-using the filenames in `io.realm.kotlin.ext` like `Realm`, since that would probably make it slightly confusing if consumed from Java
> 
> Maybe rename them to `<X>Ext` or something like that?

I moved them to `<X>Ext.kt`. Was also nicer to work with in the IDE

PR #860, opened 2022-06-01
Ensure that all Core exceptions use RealmException as the base class

### LINKED ISSUE #855: All RealmCore exceptions should have RealmException as their parent class
Currently, we are leaking cinterop exception types into the public API. Instead, all Core exceptions should extend from our top-level `RealmException` class. 

This will also allow us to subclass them later as needed without breaking the public API.

cmelchior:
Closes #855 

This PR reworks how our exceptions are being presented to end users. 

Before this PR, Core errors would result in `RealmCoreException` which is an internal `cinterop` type. With this PR, this has been reworked so all errors now have the public `RealmException` as their base class. This will allow us to add more fine-grained public exception types later without breaking the API.

The logic for doing this is a bit annoying since `cinterop` cannot call into `library-base`, we can only go the other direction. So I have added initializer code at our API entry points that set up a communication path between `cinterop` and `library-base`, which allows `cinterop` to delegate error mapping to the public side.

With our current split between `cinterop` and `library-base` this is probably the best we can do and doing better will probably require combining `cinterop` and `library-base` into one module.

This PR does not improve on the errors being thrown, for that we still need Cores unified error handling to be merged https://github.com/realm/realm-core/pull/5441.


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/RealmConfiguration.kt
```diff
@@ -49,6 +50,10 @@ public interface RealmConfiguration : Configuration {
         schema: Set<KClass<out BaseRealmObject>>
     ) : Configuration.SharedBuilder<RealmConfiguration, Builder>(schema) {
 
+        init {
```
rorbech, line 53: Anything wrong with putting this in `Configuration.SharedBuilder`'s `init` block?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncConfiguration.kt
```diff
@@ -187,6 +188,10 @@ public interface SyncConfiguration : Configuration {
         schema: Set<KClass<out RealmObject>>,
     ) : Configuration.SharedBuilder<SyncConfiguration, Builder>(schema) {
 
+        init {
```
rorbech, line 191: Same. Just put it in base class init method!

rorbech (approved):
LGTM with the exceptions of the concerns that you already mentioned in the PR test and as we already have discuess 😞  ... you could argue that we might as well just have to guard all `cinterop` calls in `library-base` as we are already dependent of remembering this on Native, where there is no generic error handling like injected through Swig on JVM ... but at least now the public exceptions are at least controlled and up for later reimplementation 🤷 

Besides that only some minor comments.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
```diff
@@ -84,6 +85,11 @@ public interface AppConfiguration {
     public class Builder(
         private val appId: String
     ) {
+
+        init {
+            CoreExceptionConverter.initialize()
```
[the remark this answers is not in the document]
cmelchior: Moved this into a `RealmInteropBridge` class with `RealmInteropBridge.initialize()` class instead.

PR #863, opened 2022-06-02
Adjust documentation to new product naming

cmelchior:
Adjust to new product naming on Atlas.

### DIFF
--- a/packages/library-sync/overview.md
+++ b/packages/library-sync/overview.md
@@ -1,8 +1,11 @@
 # Module Realm Kotlin Multiplatform Sync SDK
 
-Module containing documentation on how to setup synchronization with MongoDB Realm Apps.
+Module containing documentation on how to setup and work with Atlas App Services and Device Sync.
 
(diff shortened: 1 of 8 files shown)


cmelchior:
Note: I didn't run CI on this as it is only doc changes so should be safe.

rorbech (approved):
LGTM - only a single comment - Is it actually still called an _"App Service **App**"_? I couldn't find that term in the naming reference?

packages/library-sync/overview.md
```diff
@@ -1,8 +1,11 @@
 # Module Realm Kotlin Multiplatform Sync SDK
 
-Module containing documentation on how to setup synchronization with MongoDB Realm Apps.
+Module containing documentation on how to setup and work with Atlas App Services and Device Sync.
```
nhachicha, line 3: Can we add links to these new service definitions for more context ?

nhachicha (approved):

PR #864, opened 2022-06-02
Change implicit import policy to update existing primary key objects

### LINKED ISSUE #849: Support UpdatePolicy.ALL for RealmList.add
It is currently not possible with RealmList to add multiple items with the same primary key (or for different lists to add an item with the same primary key)
RealmListInternal uses UpdatePolicy.ERROR for its insert operators (https://github.com/realm/realm-kotlin/blob/master/packages/library-base/src/commonMain/kotlin/io/realm/internal/RealmListInternal.kt#L197) which differs from the behaviour in realm-java (which  uses copyToRealmOrUpdate when adding unmanaged objects to a managed RealmList https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/RealmList.java#L214)

```
class TestList: RealmObject {
    var items: RealmList<TestItem> = realmListOf()
}

class TestItem : RealmObject {
    @PrimaryKey var id: String = "1"
}
```

```
realm.write {
    val myList = copyToRealm(TestList())
    myList.items.add(TestItem())
    myList.items.add(TestItem())
}
```

```
realm.write {
    val testList1 = copyToRealm(TestList())
    val testList2 = copyToRealm(TestList())

    testList1.items.add(TestItem())
    testList2.items.add(TestItem())
}
```
java.lang.IllegalArgumentException: Failed to create object of type 'TestItem': RealmCoreException([33]: Object with this primary key already exists)

rorbech:
This PR changes the update policy for implicit imports (assignment to realm object members or through conventional list operations) to update existing primary key objects instead of throwing an `IllegalArgumentException.

Closes #849

### DIFF
--- a/test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
+++ b/test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
@@ -93,15 +93,14 @@ class MutableRealmTests {
 
     @Test
     fun set_updatesExistingObjectInTree() {
-        val child = SampleWithPrimaryKey().apply {
-            primaryKey = 1
-            stringField = "INIT"
-        }
         val parent = realm.writeBlocking {
             copyToRealm(
                 SampleWithPrimaryKey().apply {
                     primaryKey = 2
-                    nullableObject = child
+                    nullableObject = SampleWithPrimaryKey().apply {
+                        primaryKey = 1
+                        stringField = "INIT"
+                    }
                 }
             )
         }
@@ -109,16 +108,16 @@ class MutableRealmTests {
             assertEquals("INIT", stringField)
         }
 
-        child.apply {
-            stringField = "UPDATE"
-        }
         realm.writeBlocking {
             findLatest(parent)!!.apply {
-                nullableObject = child
+                nullableObject = SampleWithPrimaryKey().apply {
+                    primaryKey = 1
+                    stringField = "UPDATED"
+                }
             }
         }
         realm.query<SampleWithPrimaryKey>("primaryKey = '1'").find().single().run {
-            assertEquals("UPDATE", stringField)
+            assertEquals("UPDATED", stringField)
         }
     }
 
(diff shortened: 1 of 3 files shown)


test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -92,14 +92,34 @@ class MutableRealmTests {
...
+
+        child.apply {
+            stringField = "UPDATE"
+        }
         realm.writeBlocking {
-            val sample = copyToRealm(SampleWithPrimaryKey())
-            assertFailsWith<IllegalArgumentException> {
-                sample.nullableObject = SampleWithPrimaryKey()
+            findLatest(parent)!!.apply {
+                nullableObject = child
```
cmelchior, line 117: The code changes seems to have a few more code paths, e.g. realmObject, embeddedobject, embeddedWithKey... We should probably also test these. But since we didn't do that in original code, we can probably just do that in another PR.

cmelchior (approved):
Looks good to me. If CI is happy 👍

nhachicha (approved):
LGTM, when CI is fixed 👍

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
rorbech: There are test for the three types of assignments that we can do:
- RealmObject is this test
- EmbeddedRealmObject is in https://github.com/realm/realm-kotlin/pull/864/files#diff-e211bb2185cacfa97f056a11aaf87c881363ea9add72952fe2305adaafe6afc8R244
- RealmList<RealmObject> is in https://github.com/realm/realm-kotlin/pull/864/files#diff-bb43ea504213fcf23891f060506ce3837d2812808acb02286a49185f54883755R335

I guess it isn't tested for `RealmList<EmbeddedRealmObject>` but the codepath for that is not different from `RealmList<RealmObject>` so the above test should already verify that we get the UpdatePolicy propagated correctly.

PR #873, opened 2022-06-05
Fix erroneous SingleQuery notification events

rorbech:
This PR is for investigations of #871. Currently just adding debug statements to be able to trace and shed some light on the observations from the "Sync Demo" app. A similar branch exist for the actual app in (https://github.com/realm/realm-kotlin-samples/tree/cr/fix-single-query-state), that is updated to use a Maven-local snapshot that should be published from this branch with `./gradlew pTML` from `packages`


edualonso (approved):
LGTM

PR #874, opened 2022-06-06
Dokka walkthrough for dead links, outdated stuff, etc.

edualonso:
Fixed see notations for external http links, removed dead internal links plus some formatting. Nothing serious.


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Realm.kt
```diff
@@ -140,7 +141,7 @@ public interface Realm : TypedRealm {
      * @param block function that should be run within the context of a write transaction.
      * @return any value returned from the provided write block. If this is a [RealmObject] it is
      * frozen before being returned.
-     * @see [Configuration.writeDispatcher]
+     * @see [InternalConfiguration.writeDispatcher]
```
cmelchior, line 144: We should remove the reference to this. The write dispatcher isn't part of the public API yet.

cmelchior (approved):

PR #883, opened 2022-06-09
RealmUUID datatype

clementetb:
Because there is no official support for UUID on KMM we have decided to create a custom wrapper around the core data-structure, as we did previously with `RealmInstant`.

As the core data-structure this type does not validity checks against the RFC4122 spec, we leave it to the user to enforce any pattern if required.

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmUUID.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmUUID.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2022 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.kotlin.types
+
+/**
+ * A class that represents an immutable universally unique identifier (UUID). A UUID represents a 128-bit value.
+ *
+ * It does not enforce the standard [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122)
+ */
+public interface RealmUUID : Comparable<RealmUUID> {
+    public companion object {
+        /**
+         * Generates a new [RealmUUID] with a random UUID value.
+         */
+        public fun random(): RealmUUID = TODO()
+
+        /**
+         * Generates a new [RealmUUID] from an UUID string representation.
+         */
+        public fun from(uuidString: String): RealmUUID = TODO("validates uuid validity")
+
+        /**
+         * Generates a new [RealmUUID] from an UUID byte array representation.
+         */
+        public fun from(bytes: ByteArray): RealmUUID = TODO("validates uuid validity")
+    }
+
+    /**
+     * The UUID represented as a 16 byte array
+     */
+    public val bytes: ByteArray
+}
--- a/test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
+++ b/test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2022 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.kotlin.test.shared
+
+import io.realm.kotlin.test.assertFailsWithMessage
+import io.realm.kotlin.types.RealmUUID
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+
+class RealmUUIDTests {
+    @Test
+    fun from_uuidString() {
+        // Invalid arguments
+        assertFailsWithMessage<IllegalArgumentException>("invalid string representation of an UUID: ''") {
+            RealmUUID.from("") // empty string
+        }
+
+        assertFailsWithMessage<IllegalArgumentException>("invalid string representation of an UUID: 'ffffffff-ffff-xxxx-ffff-ffffffffffff'") {
+            RealmUUID.from("ffffffff-ffff-xxxx-ffff-ffffffffffff") // invalid uuid value
+        }
+
+        // Boundaries
+        assertContentEquals(
+            ByteArray(16) { 0x00.toByte() },
+            RealmUUID.from("00000000-0000-0000-0000-000000000000").bytes
+        )
+
+        assertContentEquals(
+            ByteArray(16) { 0xFF.toByte() },
+            RealmUUID.from("ffffffff-ffff-ffff-ffff-ffffffffffff").bytes
+        )
+
+        // Values
+    }
+
+    @Test
+    fun from_bytes() {
+        // Invalid arguments
+        assertFailsWithMessage<IllegalArgumentException>("byte array size must be 16") {
+            RealmUUID.from(byteArrayOf()) // 16 char needed
+        }
+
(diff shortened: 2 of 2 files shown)


cmelchior (changes requested):
Nice start 💯 . I think the public API looks solid...had some minor comments about Dokka and implementation

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUUIDImpl.kt
```diff
@@ -0,0 +1,113 @@
...
+                for (index in HYPHEN_INDEXES) insert(index, '-')
+            }
+            .toString()
+    }
+
+    companion object {
+        private const val UUID_BYTE_SIZE = 16
+        private val HYPHEN_INDEXES = listOf(8, 13, 18, 23)
+        private val VALUE_INDEXES = (0 until 36) - HYPHEN_INDEXES
+        private val VALID_CHARS = ('0'..'9') + ('a'..'f') + ('A'..'F')
```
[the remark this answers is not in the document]
clementetb: I did some quick tests and regex matching is slightly faster. I have ported the logic to use a regex pattern on `UUID`.

I have replaced the validation logic on `ObjectId` to use a regex pattern too.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUUIDImpl.kt
```diff
@@ -0,0 +1,113 @@
...
+import io.realm.kotlin.internal.util.toHexString
+import io.realm.kotlin.types.RealmUUID
+import kotlin.experimental.and
+import kotlin.experimental.or
+import kotlin.random.Random
+
+@Suppress("MagicNumber")
+internal class RealmUUIDImpl : RealmUUID, UUIDWrapper {
+
+    private val _bytes: ByteArray
```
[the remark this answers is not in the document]
clementetb: Yes, we could directly manipulate the index but the string parser yields 'ByteArray' instances that are easier to assign than to copy.

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/UUIDWrapper.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.kotlin.internal.interop
+
+/**
+ * Wrapper around Core UUID values.
+ * See https://github.com/realm/realm-core/blob/master/src/realm/uuid.hpp for more information
+ */
+interface UUIDWrapper {
```
rorbech, line 22: Any specific reason to have a full interface here and not just a typealias.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/util/Hexadecimal.kt
```diff
@@ -0,0 +1,56 @@
...
+ * @return a string representation of [ByteArray] in hexadecimal format.
+ */
+@Suppress("MagicNumber")
+internal fun ByteArray.toHexString(startIndex: Int = 0, endIndex: Int = this.size): String {
+    val chars = CharArray((endIndex - startIndex) * 2)
+
+    var i = 0
+    for (index in startIndex until endIndex) {
+        val byte = this[index]
+        chars[i++] = HEX_CHARS[byte.toInt() shr 4 and 0xF]
```
rorbech, line 33: `byte.toInt()` is called multiple times. Put in a local variable.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUUIDImpl.kt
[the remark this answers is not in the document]
rorbech: But it seems like `_bytes` is only assigned in the constructor, so why not just assign it directly to `bytes`?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUUIDImpl.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    override val bytes: ByteArray
+        get() = _bytes
+
+    constructor() {
+        _bytes = Random.nextBytes(UUID_BYTE_SIZE).apply {
+            // Set uuid to version 4, 6th byte must be 0x40
```
rorbech, line 38: Is this right? Or is the comment just wrong? Maybe it should just be
```suggestion
            // Set uuid to version 4, 6th byte must be 0x4x
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmUUID.kt
```diff
@@ -0,0 +1,46 @@
...
+        /**
+         * Generates a new [RealmUUID] from an UUID byte array representation.
+         */
+        public fun from(bytes: ByteArray): RealmUUID = TODO("validates uuid validity")
+    }
+
+    /**
+     * The UUID represented as a 16 byte array
+     */
+    public val bytes: ByteArray
```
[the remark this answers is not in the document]
rorbech: Public API shouldn't be defined by internal implementation details ... but I would also prefer the `ByteArray` over two longs anyway 🤷

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
```diff
@@ -0,0 +1,143 @@
...
+            ByteArray(16) { 0x00.toByte() },
+            RealmUUID.from("00000000-0000-0000-0000-000000000000").bytes
+        )
+
+        assertContentEquals(
+            ByteArray(16) { 0xFF.toByte() },
+            RealmUUID.from("ffffffff-ffff-ffff-ffff-ffffffffffff").bytes
+        )
+
+        assertContentEquals(
```
rorbech, line 72: Duplicate!?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUUIDImpl.kt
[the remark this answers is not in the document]
clementetb: Oh, yes I missed it. I am removing the `_bytes`

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/UUIDWrapper.kt
clementetb: A type alias would not be distinguishable from a regular `ByteArray`, as the type system resolves them as the same type.

rorbech: I forgot that we need this to differentiate the incoming values 👍

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmUUID.kt
```diff
@@ -0,0 +1,63 @@
...
+         * @throws IllegalArgumentException if [bytes] does not match the required byte array size 16.
+         */
+        public fun from(bytes: ByteArray): RealmUUID = RealmUUIDImpl(bytes)
+    }
+
+    /**
+     * The UUID represented as a 16 byte array.
+     */
+    public val bytes: ByteArray
+}
```
rorbech, line 63: Should we consider adding an explicit override of `toString` to document the string format? And maybe same for `equals`?

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
```diff
@@ -0,0 +1,138 @@
...
+            val uuid1 = RealmUUID.random()
+            val uuid2 = RealmUUID.random()
+
+            assertFalse(uuid1.bytes.contentEquals(uuid2.bytes))
+            assertNotEquals(uuid1, uuid2)
+        }
+    }
+
+    @Test
+    fun to_String() {
```
rorbech, line 130:
```suggestion
    fun toString() {
```

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
```diff
@@ -0,0 +1,138 @@
...
+        repeat(10) {
+            val uuid = RealmUUID.random()
+            val copy = RealmUUID.from(uuid.toString())
+
+            assertEquals(uuid, copy)
+        }
+    }
+}
```
rorbech, line 138: I think we should also add explicit `equals` and `notEquals` tests.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
clementetb: It cannot be named toString as it clashes with the toString method

nhachicha (approved):
💯

PR #888, opened 2022-06-13
Clean up strings stored in realm_value_t managed by the JVM-cinterop layer

### LINKED ISSUE #890: Leaking buffer when passing strings to realm_value_t on JVM
The buffer of the native representation of strings parsed to `realm_value_t` on JVM is not properly released.

rorbech:
This PR add explicit copying of string data passed to `realm_value_t.string`, thus when the `realm_value_t` is managed by the cinterop layer.

Questions:
- [x] We could introduce a `memScope`-method with specific `scope.to_realm_value` to automatically collect and release these
- [x] How to verify deallocation ... for now just manually verified by profiler inspection during our conventional memory test

Closes #890

### DIFF
--- a/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -17,6 +17,7 @@
 package io.realm.kotlin.internal.interop
 
 import io.realm.kotlin.internal.interop.Constants.ENCRYPTION_KEY_LENGTH
+import io.realm.kotlin.internal.interop.RealmInterop.cptr
 import io.realm.kotlin.internal.interop.sync.AuthProvider
 import io.realm.kotlin.internal.interop.sync.CoreSubscriptionSetState
 import io.realm.kotlin.internal.interop.sync.CoreSyncSessionState
@@ -329,11 +330,26 @@ actual object RealmInterop {
     }
 
     actual fun realm_object_create_with_primary_key(realm: LiveRealmPointer, classKey: ClassKey, primaryKey: RealmValue): RealmObjectPointer {
-        return LongPointerWrapper(realmc.realm_object_create_with_primary_key(realm.cptr(), classKey.key, to_realm_value(primaryKey)))
+        return memScope {
+            LongPointerWrapper<RealmObjectT>(
+                realmc.realm_object_create_with_primary_key(
+                    realm.cptr(),
+                    classKey.key,
+                    managedRealmValue(primaryKey)
+                )
+            )
+        }
     }
     actual fun realm_object_get_or_create_with_primary_key(realm: LiveRealmPointer, classKey: ClassKey, primaryKey: RealmValue): RealmObjectPointer {
         val created = booleanArrayOf(false)
-        return LongPointerWrapper(realmc.realm_object_get_or_create_with_primary_key(realm.cptr(), classKey.key, to_realm_value(primaryKey), created))
+        return memScope {
+            LongPointerWrapper(
+                realmc.realm_object_get_or_create_with_primary_key(
+                    realm.cptr(), classKey.key,
+                    managedRealmValue(primaryKey), created
+                )
+            )
+        }
     }
 
     actual fun realm_object_is_valid(obj: RealmObjectPointer): Boolean {
@@ -399,6 +415,8 @@ actual object RealmInterop {
                     value.asLink()
                 realm_value_type_e.RLM_TYPE_NULL ->
                     null
+                realm_value_type_e.RLM_TYPE_BINARY ->
+                    value.binary
                 else ->
                     TODO("Unsupported type for from_realm_value ${value.type}")
             }
@@ -406,8 +424,9 @@ actual object RealmInterop {
     }
 
     actual fun realm_set_value(obj: RealmObjectPointer, key: PropertyKey, value: RealmValue, isDefault: Boolean) {
-        val cvalue = to_realm_value(value)
-        realmc.realm_set_value(obj.cptr(), key.key, cvalue, isDefault)
+        memScope {
+            realmc.realm_set_value(obj.cptr(), key.key, managedRealmValue(value), isDefault)
+        }
     }
 
     actual fun realm_set_embedded(obj: RealmObjectPointer, key: PropertyKey): RealmObjectPointer {
@@ -436,8 +455,9 @@ actual object RealmInterop {
     }
 
     actual fun realm_list_add(list: RealmListPointer, index: Long, value: RealmValue) {
-        val cvalue = to_realm_value(value)
-        realmc.realm_list_insert(list.cptr(), index, cvalue)
+        memScope {
+            realmc.realm_list_insert(list.cptr(), index, managedRealmValue(value))
+        }
     }
 
     actual fun realm_list_insert_embedded(list: RealmListPointer, index: Long): RealmObjectPointer {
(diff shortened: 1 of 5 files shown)


clementetb (commented):
Looking good. 

We could automatically release the resources by tying the realm_value_t to the execution of `to_realm_value`. We could add a block to the signature, and any resource would be allocated before calling the block, and then would be freed once executed.

```
to_realm_value(value: RealmValue, block: (realm_value_t) -> Unit)
```

We would need something similar for lists of arguments on queries, but it would be safer than deallocating the resources manually.

rorbech:
@clementetb I already suggested something like that in the PR description. I went with the `memScope`-outline as on Kotlin Native as it made it easier to handle the list arguments and seems easier to extend to other resources if needed.

clementetb:
Yes,  I read it as a suggestion and that it might not be added. I pointed to the function block as an intermediate step if you didn't want to implement the `memscope`. I see that it is now implemented, looks great!

packages/jni-swig-stub/realm.i
```diff
@@ -196,7 +196,33 @@ typedef jstring realm_string_t;
...
+    $1.size = strlen(buf);
+    $1.data = buf;
+}
+%typemap(freearg) realm_string_t {
+    if ($1.data) jenv->ReleaseStringUTFChars($arg, (const char *)$1.data);
+}
+
+// Typemap used for passing realm_string_t into the C-API in situations where the string buffer
+// needs to be kept alive after returning from C-API call. This will copy the string buffer to the
+// heap and this have to be explicitly freed at a later point.
```
nhachicha, line 213:
```suggestion
// heap and this has to be explicitly freed at a later point.
```

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1443,6 +1414,113 @@ actual object RealmInterop {
...
+}
+
+/**
+ * A scope providing a [MemScope] that collects the created resources and automatically frees them
+ * upon exiting the scope.
+ */
+private fun <R> memScope(block: MemScope.() -> R): R {
+    val scope = MemScope()
+    val result: R = block(scope)
+    scope.free()
```
cmelchior, line 1443: Shouldn't this be in a `try/finally` block so we always free data even if there is an exception?

packages/jni-swig-stub/realm.i
```diff
@@ -196,7 +196,33 @@ typedef jstring realm_string_t;
...
+// Currently just matching 'realm_string_t string' arguments to match realm_value_t.string = $input
+%typemap(in) realm_string_t string (char* buf) {
+    buf = (char*)jenv->GetStringUTFChars($arg,0);
+    $1.size = strlen(buf);
+    $1.data = (char const *) (new char[$1.size]);
```
nhachicha, line 218: Who is freeing this heap-based string? I doubt C-API/Core is doing so, they mostly rely on stack-based allocation and move/copy assignment to deal with passed resources ...

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1443,6 +1414,113 @@ actual object RealmInterop {
     }
 }
 
+/**
+ * A factory and container for various resources that can be freed when calling [free].
```
cmelchior, line 1418: Should we specify that this is required for any RealmValue arguments sent to the C-API?

Also, the README only mentions Strings, but the code calling this are wrapping all types?

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -689,3 +689,22 @@ sync_after_client_reset_handler(realm_sync_config_t* config, jobject after_handl
     };
     realm_sync_config_set_after_client_reset_handler(config, after_func, user_data, free_func);
 }
+
+// Explicit clean up method for releasing heap allocated data of a realm_value_t instance
+void
+realm_value_t_cleanup(realm_value_t* value) {
```
nhachicha, line 695: To my point above, if we're passing a string that is going to outlive the C-API call why not build a stack-based string and pass it to Core? this will align also with the upcoming changes in the C-API where we plan to manipulate `string_data` instead of `char*` ....

cmelchior (approved):

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1443,6 +1414,113 @@ actual object RealmInterop {
...
+    val value = realmValue.value
+    if (value == null) {
+        cvalue.type = realm_value_type_e.RLM_TYPE_NULL
+    } else {
+        when (value) {
+            is String -> {
+                cvalue.type = realm_value_type_e.RLM_TYPE_STRING
+                cvalue.string = value
+            }
+            /*is Byte -> {
```
nhachicha, line 1461: Why this is still commented? there's no use case for these managed types yet?

packages/jni-swig-stub/realm.i
```diff
@@ -196,7 +196,33 @@ typedef jstring realm_string_t;
...
+}
+%typemap(freearg) realm_string_t {
+    if ($1.data) jenv->ReleaseStringUTFChars($arg, (const char *)$1.data);
+}
+
+// Typemap used for passing realm_string_t into the C-API in situations where the string buffer
+// needs to be kept alive after returning from C-API call. This will copy the string buffer to the
+// heap and this have to be explicitly freed at a later point.
+// Currently just matching 'realm_string_t string' arguments to match realm_value_t.string = $input
+%typemap(in) realm_string_t string (char* buf) {
```
nhachicha, line 215: I don't see these generated in the `realmc.cpp` ?

packages/jni-swig-stub/realm.i
rorbech: The problem is that we create the `realm_value_t` struct and assign the `string` field in different operations in `RealmInterop` which means that the c-stack goes out of scope before calling the actual C-API function with the `realm_value_t` so needs to be _alive_ from assignment and across the actual C-API call ... in which the C-API will correctly copy it.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
rorbech: I will update the comments. All other datatypes are transferred by value inside the `realm_value_t` so it was an issue for strings which required external management of the buffer pointed to by `realm_value_t.string.data`. ... and also the upcoming bytearray-support in #584.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
rorbech: I just moved the `to_realm_value` method around, so didn't really update anything. We are not passing other types than the `RealmStorageValue` types in/out after the #790. Will just remove them.

rorbech:
@nhachicha @cmelchior As discussed I haven't set up automatic verification of the fix, but you can see the memory profiling graphs of running our `Memorytest.garbageCollectorShouldFreeNativeResources` before and after the fix in https://mongodb.slack.com/archives/CSXQ43YFK/p1654862598688209

nhachicha (approved):

PR #892, opened 2022-06-14
Add input dependencies to cinterop native tasks

rorbech:
This adds Core source code as input for the various `cinterop` native compilation tasks and. This should ensure that we rebuild the C-API when the core sources is updated.

### DIFF
--- a/packages/cinterop/build.gradle.kts
+++ b/packages/cinterop/build.gradle.kts
@@ -637,13 +637,3 @@ realmPublish {
             "'io.realm.kotlin:gradle-plugin:${Realm.version}' instead."
     }
 }
-
-tasks.named("clean") {
-    doLast {
-        delete(
-            // FIXME Is is too dangerous to delete something outside build/ ... should these libs
-            //  go to build/libs or something like that
-            fileTree(project.file("src/jvmMain/resources/jni"))
-        )
-    }
-}
--- a/packages/cinterop/build.gradle.kts
+++ b/packages/cinterop/build.gradle.kts
@@ -55,6 +55,7 @@ val isReleaseBuild: Boolean = (System.getenv("CONFIGURATION") ?: "RELEASE").equa
 
 val corePath = "external/core"
 val absoluteCorePath = "$rootDir/$corePath"
+val jvmJniPath = "src/jvmMain/resources/jni"
 
 fun includeBinaries(binaries: List<String>): List<String> {
     return binaries.flatMap { listOf("-include-binary", it) }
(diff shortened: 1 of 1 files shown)


packages/cinterop/build.gradle.kts
```diff
@@ -378,30 +379,31 @@ fun Task.buildSharedLibrariesForJVM() {
...
             project.file("src/jvmMain/windows-build-dir/Release/realmc.dll")
-                .copyTo(project.file("src/jvmMain/resources/jni/windows/realmc.dll"), overwrite = true)
+                .copyTo(project.file("$jvmJniPath/windows/realmc.dll"), overwrite = true)
             genHashFile(platform = "windows", prefix = "", suffix = ".dll")
         }
     }
 
-    outputs.file(project.file("src/jvmMain/resources/jni/macos/librealmc.dylib"))
-    outputs.file(project.file("src/jvmMain/resources/jni/macos/dynamic_libraries.properties"))
+    inputs.dir(project.file("$absoluteCorePath/src"))
```
cmelchior, line 392: Do we need to use the entire folder as input or would just using `realm.h` be enough? 🤔

packages/cinterop/build.gradle.kts
```diff
@@ -632,3 +637,13 @@ realmPublish {
             "'io.realm.kotlin:gradle-plugin:${Realm.version}' instead."
     }
 }
+
+tasks.named("clean") {
+    doLast {
+        delete(
+            // FIXME Is is too dangerous to delete something outside build/ ... should these libs
```
cmelchior, line 644: It would be nice if we could move these files into `build` IMO

cmelchior (approved):
I haven't tested this myself, but my assumption was also that we just needed to set the `input` correctly.

packages/cinterop/build.gradle.kts
rorbech: If we don't set the whole tree then it won't trigger cmake's incremental reduild when the capi is stable but other sources have changed

packages/cinterop/build.gradle.kts
nhachicha, line 644: Linux & Windows are built on CI https://github.com/realm/realm-kotlin/blob/master/Jenkinsfile#L695-L721 so only macOS is already being built under the `build` directory https://github.com/realm/realm-kotlin/blob/master/packages/cinterop/build.gradle.kts#L344-L347 only the artifacts are copied into the `jni/resources` for them to be packaged automatically by the `Jar` task.
What didn't work previously is deleting the build directory won't trigger the recompilation of the  JVM/C-API since the gradle outputs were tied to the `jni/resources` folder

nhachicha, line 644:
```suggestion
            // FIXME Is it too dangerous to delete something outside build/ ... should these libs
```

nhachicha (approved):

rorbech:
@nhachicha @cmelchior I have removed the `clean` task completely as it isn't a blocker for proper incremental build and would requires significant work and test to get right compared to the benefit. If the incremental parts works then we don't need to clean the libs.

PR #894, opened 2022-06-15
Add support for `ByteArray`

### LINKED ISSUE #584: Datatype support: ByteArray
We should support all data types available in Core. Right now we are missing `ByteArray` (byte[ ] on JVM) support.

```kotlin
class Example: RealmObject {
  var data: ByteArray = ByteArray(0)
}
```

edualonso:
Closes #584 

Added support for `ByteArray` attributes
- [x] compiler support
- [x] support for `RealmList<ByteArray>`
- [x] check support through dynamic objects
- [ ] ~~check object notifications when modifying binary fields~~
- [x] check sync works
- [x] add freeing logic


test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/ByteArrayTests.kt
```diff
@@ -0,0 +1,122 @@
...
+    // Store value and retrieve it again
+    private fun <T> roundTrip(
+        sample: Sample,
+        property: KMutableProperty1<Sample, T>,
+        function: (T) -> Unit
+    ) {
+        // Test managed objects
+        realm.writeBlocking {
+            copyToRealm(sample)
+            val managedByteArray = query<Sample>()
```
rorbech, line 102: I think we should do the query on the `realm` after committing the transaction, just to be sure that we are actually able to write and commit the bytearrays. You can just do a `delete(query)` afterwards.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -961,7 +973,16 @@ actual object RealmInterop {
                     }
                 }
             }
-            //    RLM_TYPE_BINARY,
+            is ByteArray -> {
+                cvalue.type = realm_value_type.RLM_TYPE_BINARY
+                cvalue.binary.apply {
+                    data = alloc<UByteVarOf<UByte>>().ptr
```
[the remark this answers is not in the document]
edualonso: Incrementing the index and setting a new value per every byte seems to work so I assume it's working.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -208,6 +212,12 @@ fun realm_value_t.set(memScope: MemScope, realmValue: RealmValue): realm_value_t
                 }
             }
         }
+        is ByteArray -> {
+            type = realm_value_type.RLM_TYPE_BINARY
+            (0 until value.size).map {
+                binary.data?.set(it, requireNotNull(this.binary.data)[it])
+            }
```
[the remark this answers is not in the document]
edualonso: It was wrong 🙈 it should be setting the value from `value[it]`.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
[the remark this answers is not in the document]
edualonso: The boundary test would seemingly fail at random with this code. It didn't show a consistent pattern. Turns out changing the call to use `allocArray` fixes the issue, so it looks like it was after all a memory issue 🙊

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -208,6 +211,12 @@ fun realm_value_t.set(memScope: MemScope, realmValue: RealmValue): realm_value_t
                 }
             }
         }
+        is ByteArray -> {
+            type = realm_value_type.RLM_TYPE_BINARY
+            (0 until value.size).map {
+                binary.data?.set(it, value[it].toUByte())
```
clementetb, line 217: Seems that a alloc array is missing here.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -961,7 +972,16 @@ actual object RealmInterop {
                     }
                 }
             }
-            //    RLM_TYPE_BINARY,
+            is ByteArray -> {
+                cvalue.type = realm_value_type.RLM_TYPE_BINARY
+                cvalue.binary.apply {
```
clementetb, line 977: This logic can be replaced with cvalue.binary = memScope.allocArrayOf(value)

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
clementetb, line 217: Seems that a alloc array is missing here.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
edualonso: Ahhh... It would be nice, but `allocArrayOf` returns a pointer of type `ByteVarOf<Byte>` and we need a `UByteVarOf<UByte>`. Perhaps there is something else in the utils but I haven't found it :(

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
edualonso: Double checked parameterised queries and they failed hard 🙈 I fixed it and wrote a test

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/ByteArrayTests.kt
```diff
@@ -0,0 +1,161 @@
...
+            assertContentEquals(sample.binaryField, value)
+        }
+        roundTrip(sample, Sample::nullableBinaryField) { value ->
+            assertContentEquals(sample.nullableBinaryField, value)
+        }
+    }
+
+    @Test
+    fun equals_assignedValues() {
+        val byteArray = byteArrayOf(22, 44, 66)
```
nhachicha, line 82: Is there a round trip test for signed bytes (from `-128 -> 127`) ?

packages/jni-swig-stub/realm.i
```diff
@@ -326,6 +326,12 @@ bool throw_as_java_exception(JNIEnv *jenv) {
 // Enable passing uint8_t* parameters for realm_config_get_encryption_key and realm_config_set_encryption_key as Byte[]
 %apply int8_t[] {uint8_t *key};
 %apply int8_t[] {uint8_t *out_key};
+%apply int8_t[] {const uint8_t* data};
+
+%typemap(freearg) const uint8_t* data;
+%typemap(out) const uint8_t* data %{
+    $result = SWIG_JavaArrayOutSchar(jenv, (signed char *)result, arg1->size);
+%}
```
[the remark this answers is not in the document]
rorbech: The `data` correctly assures that the typemap is only applied when the member name is `data`, hence for `realm_binary_t`. But the current clean up of freeing data in https://github.com/realm/realm-kotlin/pull/894/files#diff-c71ff0791782da90014db6f7000bdeea6c883ca9729f17bd1b11041d1e7711faR703 is only done when `realm_binary_t` is embedded in a `realm_value_t`. This means that if we use a `realm_binary_t` outside the scope of a `realm_binary_t` that is properly clean up, then we would leak. For `realm_string_t` I specifically target assignments to `realm_value_t.string` with the typemap, so that it only copied the buffer in that cases and also ment that we could skip the JNI-roundtrip for setting the individual properties of `realm_string_t` by setting the members in C++. 

Seems like the only other usage of `realm_binary_t` is in `realm_convert_with_path` which we don't even use at the moment, so maybe just add an `%ignore realm_convert_with_path` to `realm.i` for now with a comment around this, as there is probably no need to block this PR further just to circle around this. There might be other updates on how we handle this in the near future anyway (https://mongodb.slack.com/archives/C01LVQ0K404/p1656680435874499?thread_ts=1656673194.461559&cid=C01LVQ0K404)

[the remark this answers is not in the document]
edualonso: I see, thanks for the explanation 🙇 I will add the comment you suggested then.

nhachicha (approved):

PR #897, opened 2022-06-16
Static analysis shouldn't trigger building native libs

cmelchior:
Running detekt and ktlint was causing the compiler plugin to trigger a build of the native code due to the way the dependency was set up.

Adding an option that ignores the native code allows us to run the static analysis before doing a full build.


cmelchior:
Outdated

PR #904, opened 2022-06-21
`RealmSet` - add base set interface and compiler plugin support

edualonso:
Created base `RealmSet` interface.
Added support for `RealmSet` attributes in objects: no embedded objects, no nullable objects

TODO:

- [x] compiler tests à la `io.realm.kotlin.test.compiler.list.ListTests`


packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -415,16 +431,26 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
                 //      return backing_field
                 //  }
 
-                // getList/setList gets/sets raw lists soi bypass any converters in accessors
+                // getCollection/setCollection gets/sets raw collections so it bypasses any converters in accessors
                 modifyAccessor(
                     property = declaration,
-                    getFunction = getList,
+                    getFunction = when (collectionType) {
+                        CollectionType.NONE -> TODO()
```
cmelchior, line 438: Shouldn't this fail with some kind of `IllegalStateException` instead?

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -415,16 +431,26 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
+                        CollectionType.LIST -> getList
+                        CollectionType.SET -> getSet
+                        CollectionType.DICTIONARY -> TODO()
+                    },
                     fromRealmValue = null,
                     toPublic = null,
-                    setFunction = setList,
+                    setFunction = when (collectionType) {
+                        CollectionType.NONE -> TODO()
```
cmelchior, line 446: Same?

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -616,16 +649,29 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
         val descriptorType = declaration.symbol.descriptor.type
-        val listGenericType = descriptorType.arguments[0].type
-        if (inheritsFromRealmObject(listGenericType.constructor.supertypes)) {
+        val collectionGenericType = descriptorType.arguments[0].type
+
+        // No embedded objects for sets
+        val supertypes = collectionGenericType.constructor.supertypes
+        if (collectionType == CollectionType.SET && inheritsFromEmbeddedRealmObject(supertypes)) {
+            logError(
+                "Error in field ${declaration.name} - ${collectionType.name}s do not support embedded realm objects element types.",
```
cmelchior, line 664: I'm not sure we should use `collectionType.name` directly here as it will result in some weird looking and internal looking names. This will turn into "SETs do not support`.... I would add a `description` or something to this enum and then use that and then set that to the public API name, so e.g `collectionType.description` would turn into `RealmSet`

cmelchior: Same in other error messages

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -676,18 +725,27 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
                     "String" -> PropertyType.RLM_PROPERTY_TYPE_STRING
                     "RealmInstant" -> PropertyType.RLM_PROPERTY_TYPE_TIMESTAMP
                     "ObjectId" -> PropertyType.RLM_PROPERTY_TYPE_OBJECT_ID
-                    else ->
-                        if (inheritsFromRealmObject(type.supertypes())) {
+                    else -> {
```
cmelchior, line 728: Not sure I understand why this has to change? I assume it is because of the change in `inheritsFromRealmObject`, but I don't understand why that was changed?

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
```diff
@@ -82,6 +84,33 @@ class Sample : RealmObject {
     var nullableTimestampListField: RealmList<RealmInstant?> = realmListOf()
     var nullableObjectIdListField: RealmList<ObjectId?> = realmListOf()
 
+    // Set types
```
cmelchior, line 87: Do we support `Set<List<T>>`? I don't remember 🙈 If we don't, maybe add a comment about it as well.

test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/EmbeddedTests.kt
```diff
@@ -62,6 +62,6 @@ class EmbeddedTests {
             )
         )
         assertEquals(KotlinCompilation.ExitCode.COMPILATION_ERROR, result.exitCode)
-        assertTrue(result.messages.contains("RealmLists does not support nullable realm objects element type"))
+        assertTrue(result.messages.contains("LISTs do not support nullable realm objects element types"))
```
cmelchior, line 65: See my comment about error messages further up. This would prevent these changes here.

cmelchior (commented):
Good start 👍 Had a few comments, that are probably easy to fix.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmSet.kt
```diff
@@ -0,0 +1,5 @@
+package io.realm.kotlin.types
+
+public interface RealmSet<E>
```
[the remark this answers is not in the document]
edualonso: ™️ sentence incoming: it is done in the next PR 😅

I won't change it just yet since I'm working against a feature branch so it will be reviewed before it goes into master.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
edualonso: It shouldn't, this function is only called when working with collections. I will change this so that it is more obvious and also clarify that dictionaries are not supported yet.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
edualonso: It's because sets don't support embedded objects. Previously we would consider a property was an object if it was either an ordinary or an embedded object. I needed to make the distinction between the two now so that we could reject sets of embedded objects.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
edualonso: Yes, good point 👍

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
edualonso: Thanks for highlighting this. I just realised I'm missing compiler tests for sets too. I will add them and test it there.

edualonso: Oh, or do you mean whether Core itself supports that... I'll ask it but I doubt it.

edualonso: Apparently not.

test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/set/SetTests.kt
```diff
@@ -0,0 +1,211 @@
...
+    // RealmSet<E>
+    // - supported types
+    // - unsupported type fails
+    // ------------------------------------------------
+
+    @Test
+    fun `non-nullable set`() {
+        allSupportedTypes.forEach { primitiveType ->
+            val result = NON_NULLABLE_SET_CODE.format(primitiveType)
+                .let {
```
rorbech, line 52: And maybe make a wrapper method for this repeated `.let { SourceFile... }.let { compilerFromSource ... }` across the tests.

test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/set/SetTests.kt
```diff
@@ -0,0 +1,211 @@
...
+    // ------------------------------------------------
+    // RealmSet<E>
+    // - supported types
+    // - unsupported type fails
+    // ------------------------------------------------
+
+    @Test
+    fun `non-nullable set`() {
+        allSupportedTypes.forEach { primitiveType ->
+            val result = NON_NULLABLE_SET_CODE.format(primitiveType)
```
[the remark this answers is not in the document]
rorbech: Ahh. I see that some of them are reused ... but ... 🤔

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
rorbech: But this looks like you did exactly the same logic as before, just in a different way!? Maybe the naming of the original method should just have been updated to `inheritsFromBaseRealmObject` 🤔  

But maybe a test for compilation errors on sets of embedded objects would reveal if the logic is correct ;)

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -295,6 +304,16 @@ internal object RealmObjectHelper {
...
     }
 
+    internal inline fun <reified T : Any> setSet(
+        obj: RealmObjectReference<out BaseRealmObject>,
+        col: String,
+        set: RealmSet<Any?>,
+        updatePolicy: UpdatePolicy = UpdatePolicy.ALL,
+        cache: ObjectCache = mutableMapOf()
+    ) {
+        // TODO
```
rorbech, line 314: Maybe better with an actual `TODO("Set is not supported")`, but I guess it is not actually tested and will be included in the next PR

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/schema/RealmPropertyType.kt
```diff
@@ -63,3 +63,9 @@ public data class ListPropertyType(
     override val storageType: RealmStorageType,
     override val isNullable: Boolean = false
 ) : RealmPropertyType
+
+// List and Set could be the same type but made them different for visibility
```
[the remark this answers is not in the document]
rorbech: Maybe worth briefly considering it already because some thoughts/constraints around dictionaries could reveal nice concepts for the set type that could be hard to change without breaking it ... but I guess we already have the issue with the list type.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
rorbech: Maybe add the `inheritsFromRealmObject` as a utility method on top of the other more explicit ones to keep the logic centralized.

rorbech (approved):

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
edualonso: I will unify both helpers in one:
```
    private fun inheritsFromRealmObject(
        supertypes: Collection<KotlinType>,
        considerEmbedded: Boolean = false
    ): Boolean = supertypes.any {
        val objectFqNames = when {
            considerEmbedded -> io.realm.kotlin.compiler.realmEmbeddedObjectInterface
            else -> io.realm.kotlin.compiler.realmObjectInterface
        }
        it.constructor.declarationDescriptor?.fqNameSafe in objectFqNames
    }
```
and revert the original function calls since it seems confusing. Only new logic requiring knowing whether a class is embedded or not will use the new codepath.

PR #905, opened 2022-06-24
Fix Compiler Plugin not setting RealmObjectReference generic argument correctly

### LINKED ISSUE #901: Cannot access RealmObject in a multi-module project
Projects with multi-module cannot access data class extended with RealmObject. It returns

```
  bad class file: /../lib1/build/intermediates/compile_library_classes_jar/debug/classes.jar(/lib1/RealmObjectExampleModel.class)
undeclared type variable: T
 Please remove or make sure it appears in the correct subdirectory of the classpath.
```

With Java SDK works fine

cmelchior:
Closes #901 

We did not set the correct generic argument for 

```
  @Nullable
  private RealmObjectReference<T> io_realm_kotlin_objectReference;
```

Inside model classes. This accidentally worked because the generic parameter is type-erased at runtime, but if other libraries like Android Hilt also attempted to introspect or work with the source code it would report errors like `bad class file: <xxx> undeclared type variable: T`

This PR changes the compiler plugin so we now use the model class as the generic argument:

```
  @Nullable
  private RealmObjectReference<Sample> io_realm_kotlin_objectReference;
```
  
The fix has been verified to fix the the repro case here: https://github.com/realm/realm-kotlin/issues/901#issuecomment-1160393534

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 * Fixed JVM memory leak when passing string to C-API. (Issue [#890](https://github.com/realm/realm-kotlin/issues/890))
 * Fixed crash present on release-mode apps using Sync due to missing Proguard exception for `ResponseCallback`.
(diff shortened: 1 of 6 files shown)


rorbech (approved):
LGTM. Ok to rely on the manual verification of the repro case for this PR, but maybe we should consider adding hilt to our compatibility project to catch future issues like this earlier.

PR #907, opened 2022-06-24
Precommit hook that runs ktlint and detekt

clementetb:
Sets a pre-commit hook that runs `ktlintCheck` and `detekt` tasks before a commit, preventing committing any changes if any of them fail.

It can be disabled with the `--no-verify` option.

The checks are only applied to the files that are going to be committed. Any untracked or unstaged changes would be temporarily stashed during the execution of the hook, and would be restored on complete.

It can be enabled by setting the git hooks path with `git config core.hooksPath .githooks`


cmelchior (approved):
Nice 👍

.githooks/pre-commit
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Fail script if any command fails to execute
+set -e
```
rorbech, line 4: You can usually achieve the same by passing the argument to the interpreter with `#!/bin/bash -e`.

rorbech (commented):
Seems ok, but should we maybe consider to do it in `.githooks/pre-push`? It is a bit time consuming, so could be a bit annoying, but I guess there is also good arguments for fixing these things in every commit 🤔 

Maybe also add the notes on how to enable/disable it in `CONTRIBUTING.md`

PR #909, opened 2022-06-27
SyncError - fix assumption that recovery path should always be present

### LINKED ISSUE #895: Crash: Failure when computing UTF-16 size

Sometimes

### Description

Android App crashed intermittently after upgrade to 1.0.0

### Stacktrace & log output

```shell
1655256275.582 E/libc++abi: terminating with uncaught exception of type realm::util::ExceptionWithBacktrace<std::runtime_error>: Failure when computing UTF-16 size error_code = 1; retcode = 16; StringData.size = 57; StringData.data = 
    Exception backtrace:
    <backtrace not supported on this platform>
```

Yes, sometimes

### Reproduction Steps

Crashes any time while the app is open

### Version

1.0.0

### What SDK flavour are you using?

MongoDB Realm (i.e. Sync, auth, functions)

### Are you using encryption?

No, not using encryption

Android 11

### Build environment

Xcode version: ...
Dependency manager and version: ...

edualonso:
Fixed logic that wrongly assumed a recovery path should always exist when processing sync errors.
See https://github.com/realm/realm-kotlin/pull/898#issuecomment-1166207478

Closes #895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Fix JVM memory leak when passing string to C-API. (Issue [#890](https://github.com/realm/realm-kotlin/issues/890))
+* Sync error events not requiring a Client Reset incorrectly assumed they had to include a path to a recovery Realm file.
 
 ### Compatibility
 * This release is compatible with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 * Fix JVM memory leak when passing string to C-API. (Issue [#890](https://github.com/realm/realm-kotlin/issues/890))
-* Sync error events not requiring a Client Reset incorrectly assumed they had to include a path to a recovery Realm file.
+* [Sync] Sync error events not requiring a Client Reset incorrectly assumed they had to include a path to a recovery Realm file. (Issue [#895](https://github.com/realm/realm-kotlin/issues/895))
 
 ### Compatibility
 * This release is compatible with:
--- a/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
+++ b/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
@@ -563,13 +563,18 @@ jobject convert_to_jvm_sync_error(JNIEnv* jenv, const realm_sync_error_t& error)
         realm_sync_error_user_info_t user_info = error.user_info_map[i];
         user_info_map->insert(std::make_pair(user_info.key, user_info.value));
     }
+
     if (error.user_info_length > 0) {
         auto original_it = user_info_map->find(error.c_original_file_path_key);
-        auto recovery_it = user_info_map->find(error.c_recovery_file_path_key);
         auto original_file_path = original_it->second;
-        auto recovery_file_path = recovery_it->second;
         joriginal_file_path = to_jstring(jenv, original_file_path);
-        jrecovery_file_path = to_jstring(jenv, recovery_file_path);
+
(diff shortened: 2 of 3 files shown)


rorbech (commented):
The fix looks good to me, but I think we should have a test that exercise this path.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Fix JVM memory leak when passing string to C-API. (Issue [#890](https://github.com/realm/realm-kotlin/issues/890))
+* Sync error events not requiring a Client Reset incorrectly assumed they had to include a path to a recovery Realm file.
```
cmelchior, line 11: Can we also link to the issue?

cmelchior, line 11:
```suggestion
* [Sync] Sync error events not requiring a Client Reset incorrectly assumed they had to include a path to a recovery Realm file.
```

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -563,13 +563,18 @@ jobject convert_to_jvm_sync_error(JNIEnv* jenv, const realm_sync_error_t& error)
...
         auto original_it = user_info_map->find(error.c_original_file_path_key);
-        auto recovery_it = user_info_map->find(error.c_recovery_file_path_key);
         auto original_file_path = original_it->second;
-        auto recovery_file_path = recovery_it->second;
         joriginal_file_path = to_jstring(jenv, original_file_path);
-        jrecovery_file_path = to_jstring(jenv, recovery_file_path);
+
+        // Sync errors may not have the path to the recovery file unless a Client Reset is requested
+        if (error.is_client_reset_requested) {
+            auto recovery_it = user_info_map->find(error.c_recovery_file_path_key);
```
cmelchior, line 574: Same

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -563,13 +563,18 @@ jobject convert_to_jvm_sync_error(JNIEnv* jenv, const realm_sync_error_t& error)
         realm_sync_error_user_info_t user_info = error.user_info_map[i];
         user_info_map->insert(std::make_pair(user_info.key, user_info.value));
     }
+
     if (error.user_info_length > 0) {
         auto original_it = user_info_map->find(error.c_original_file_path_key);
```
cmelchior, line 568: In order to prevent this bug from happening again, we should probably check that this isn't user_info_map.end(). If this doesn't find anything, we will end up with similar memory corruption issues as we have now.

cmelchior (changes requested):
I think this would actually fix the reported bug, but it would be nice to be sure we catch it if it happens any time in the future or if the key we are looking for changes.

This should also make it easier to write a test for it as It assume any sync error that isn't a client reset will then throw an exception.

edualonso:
> This should also make it easier to write a test for it as It assume any sync error that isn't a client reset will then throw an exception.

I don't think this is strictly true since changing read and write permissions to false triggered a sync error that wasn't a client reset but that contained data in `user_info` though. It smells like there is something weird going on when Core generates certain errors. I will ask someone about it.

edualonso:
According to James:
> When a PermissionDenied error is received, this is a fatal (non-client-reset) error that marks the file for deletion. Having “original path” in the user_info_map is a side effect of using the same code for client reset. But you should only rely on SyncError.is_client_reset_requested to know if it is a client reset error or not. In a manual client reset, the user_info_map will always have the original and recovery file paths populated here as well.

edualonso:
Closing as this should point to `releases`.

PR #910, opened 2022-06-27
`RealmSet` - basic API

edualonso:
The `RealmSet` interface is implemented by unmanaged and managed sets. `UnmanagedRealmSet`s implement the interface delegating on `mutableSetOf()`. `ManagedRealmSet`s are implementing the following:
```
AbstractMutableSet<E>(), RealmSet<E>
```
`AbstractMutableSet` has defaults for quite a lot of operations from the interface which simplifies the implementation of complex operations needed in the Java SDK, e.g. set intersection with `retainAll` (it is implemented using `iterator`). These default operations might not be the most efficient as opposed to letting Core do the operation itself, but given that the C-API doesn't expose intersect or union, this seems like the only solution for now.

The trip down to Core is implemented piggybacking on the `Converters` written for `RealmList`s.

The test framework is inspired by the one used in `RealmListTests` with minor changes - it might be worth considering aligning both 🤔 

Compiler plugin binding:
- [x] object helpers for regular realms
- [x] object helpers for dynamic realms

`RealmSet` interface - in order to have basic testing roundtrip capabilities we need at least these - observe all the other operations are already implemented by `AbstractMutableSet` which internally uses `iterator()`:
- [x] support for unmanaged sets
- [x] size
- [x] add
- [x] clear
- [x] contains
- [x] iterator
- [x] implement `Deleteable`
- [x] flows and changesets
- [x] sync testing
- [x] documentation
- [x] changelog entry

Changes to existing features:
- abstracted operators for collections to become `CollectionOperator`s and piggyback on the existing solution for lists (moved `mediator`, `realmReference` and `converter` to the new interface)
- abstracted notification changesets to be collection-based and not only list-based 
- reused changesets from core internally without exposing indices as the public `Set` API doesn't support them but are needed internally to determine which type of changeset to emit
- unified subscription to add notification callbacks for collections and objects at interop level (not for results)


test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1137 @@
...
+        actual: Collection<ByteArray>
+    ) {
+        assertEquals(expected.size, actual.size)
+
+        // We can't iterate by index on the set and the positions are not guaranteed to be the same
+        // as in the dataset so to compare the values are the same we need to bend over backwards...
+        var assertionSucceeded = 0
+        actual.forEach { actualByteArray ->
+            expected.forEach { expectedByteArray ->
+                try {
```
rorbech, line 889: A bit convoluted. I guess it is overall sound as you verify that size of the two collections are equal in size up front, but would be more clear to iterate and verify that all elements of `actual` is actually in `expected` and break if it isn't. Then you could also print the element not found.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1137 @@
...
+                    assertContentEquals(expectedByteArray, actualByteArray)
+                    assertionSucceeded += 1
+                } catch (e: AssertionError) {
+                    // Do nothing, the byte arrays might be structurally equal in the next iteration
+                }
+            }
+        }
+        if (assertionSucceeded != expected.size) {
+            fail("None of the ByteArray elements were found in the dataset.")
```
rorbech, line 898: The text doesn't match the condition ... copy paste that hasn't been updated.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1137 @@
...
+        }
+    }
+
+    override fun structuralContains(
+        collection: Collection<ByteArray>,
+        element: ByteArray?
+    ): Boolean {
+        var assertionSucceeded = 0
+        collection.forEach { expectedByteArray ->
```
rorbech, line 907: Again a bit convoluted. At the least add a small comment about why it can only have one match (`assertionSucceded == 1`). That would speed up reviewing and maintenance.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1137 @@
...
+        try {
+            block()
+        } catch (e: AssertionError) {
+            throw AssertionError("'${toString()}' failed - ${e.message}", e)
+        }
+    }
+}
+
+/**
+ * TODO
```
rorbech, line 614: ??

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1137 @@
...
+                // else -> null // TODO this is just to avoid having to deal with all supported types when testing something very specific
+            }
+        }
+    }
+
+    @BeforeTest
+    fun setup() {
+        tmpDir = PlatformUtils.createTempDir()
+        val configuration = RealmConfiguration.Builder(
+            setTestSchema + setOf(
```
rorbech, line 94: Seems a bit funny to have a single element set holding the schema for the test .... and then adding a lot of other classes to it to setup the test 🤪

rorbech (changes requested):
Really nice, left my comments ... most concerning is that it seems like there are public breaking changes in the list change API!?

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1365,6 +1458,46 @@ actual object RealmInterop {
...
+                null, // See https://github.com/realm/realm-kotlin/issues/661
+                // Change callback
+                staticCFunction { userdata, change ->
+                    try {
+                        userdata?.asStableRef<Callback<RealmChangesPointer>>()?.get()?.onChange(
+                            CPointerWrapper(
+                                change,
+                                managed = false
+                            )
+                        ) // FIXME use managed pointer https://github.com/realm/realm-kotlin/issues/147
```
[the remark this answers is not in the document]
clementetb: It would be a better approach than copy the data into memory, list fine grained notifications don't use realm_clone because I didn't know about it at that time..

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -980,6 +1068,11 @@ actual object RealmInterop {
                     }
                 }
             }
+            is Link -> {
```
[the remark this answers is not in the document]
edualonso: I don't quite remember the context. I will debug the native tests and will get back to you on this.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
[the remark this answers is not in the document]
edualonso: Thanks for the heads up, I completely missed this one.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
clementetb: SetTestSchema is only used in `getCloseableRealm` and here, couldn't it be removed? That way you remove the union of this two sets here, and you have an explicit schema set in `getCloseableRealm`

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1156 @@
...
+            Double::class to RealmSetContainer::nullableDoubleSetField as KMutableProperty1<RealmSetContainer, RealmSet<Any?>>,
+            RealmInstant::class to RealmSetContainer::nullableTimestampSetField as KMutableProperty1<RealmSetContainer, RealmSet<Any?>>,
+            ObjectId::class to RealmSetContainer::nullableObjectIdSetField as KMutableProperty1<RealmSetContainer, RealmSet<Any?>>,
+            ByteArray::class to RealmSetContainer::nullableBinarySetField as KMutableProperty1<RealmSetContainer, RealmSet<Any?>>
+        ).toMap()
+    }
+}
+
+// Circular dependencies with sets
+class SetLevel1 : RealmObject {
```
clementetb, line 1126: Is it necessary to have three different classes here? Couldn't we have just one?

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/dynamic/DynamicMutableRealmObjectTests.kt
```diff
@@ -375,6 +380,152 @@ class DynamicMutableRealmObjectTests {
...
                 }
+                is SetPropertyType -> {
+                    if (type.isNullable) {
+                        when (type.storageType) {
+                            RealmStorageType.BOOL -> {
+                                val value = true
+                                dynamicSample.getNullableValueSet<Boolean>(property.name).add(value)
+                                dynamicSample.getNullableValueSet<Boolean>(property.name).add(null)
+                                val setOfNullable = dynamicSample.getNullableValueSet(property.name, Boolean::class)
+                                // assertEquals(value, setOfNullable[0])
```
clementetb, line 391: Why is this commented out?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/CollectionOperator.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2021 Realm Inc.
```
clementetb, line 2:
```suggestion
 * Copyright 2022 Realm Inc.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/notifications/SetChange.kt
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2021 Realm Inc.
```
clementetb, line 2:
```suggestion
 * Copyright 2022 Realm Inc.
```

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmSetNotificationsTests.kt
```diff
@@ -0,0 +1,310 @@
+/*
+ * Copyright 2021 Realm Inc.
```
clementetb, line 2:
```suggestion
 * Copyright 2022 Realm Inc.
```

clementetb (approved):
Great job, looks solid 💎

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
edualonso: It follows the pattern we have in lists so I would keep it for consistency.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
edualonso: I'll get rid of it. Perhaps worth thinking about aligning this too when reworking list and set tests again because this approach is also used for list tests.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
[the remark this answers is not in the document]
rorbech: Ok. I guess we could still improve this if we could keep a reference to the `realm_value_t` and pass it back to the C-API as we received it, but that would again require some efforts. Lets just keep this for now, but have the use case in mind for optimization.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -513,10 +512,75 @@ actual object RealmInterop {
...
+        return from_realm_value(cvalue)
+    }
+
+    actual fun realm_set_find(set: RealmSetPointer, value: RealmValue): Pair<Long, Boolean> {
+        val index = LongArray(1)
+        val found = BooleanArray(1)
+        return memScope {
+            realmc.realm_set_find(set.cptr(), managedRealmValue(value), index, found)
+            try {
+                Pair(index[0], found[0])
```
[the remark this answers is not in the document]
rorbech: Hmm. Collections actually seems to have some indexable API (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/element-at.html). How does this add up with our implementation, should we throw or 🤔

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/notifications/internal/SetChangeImpl.kt
```diff
@@ -0,0 +1,32 @@
+/*
```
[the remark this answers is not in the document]
rorbech: Ok. We can rearrange in #941 if needed.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicRealmObject.kt
```diff
@@ -137,38 +138,111 @@ public interface DynamicRealmObject : BaseRealmObject {
...
+     *
+     * @param propertyName the name of the set property to retrieve the set for.
+     * @param clazz the Kotlin class of the set element type.
+     * @param T the type of the set element type.
+     * @return the referenced [RealmSet]
+     * @throws IllegalArgumentException if the class doesn't contain a field with the specific
+     * name, if trying to retrieve values for non-set properties or if `clazz` doesn't match the
+     * property's [RealmStorageType.kClass].
+     */
+    public fun <T : Any> getNullableValueSet(propertyName: String, clazz: KClass<T>): RealmSet<T?>
```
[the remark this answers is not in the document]
rorbech: Yeah, we should probably just track and postpone it until we are fully migrated to 1.7 and are OK with the breaking change

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1087 @@
...
+            findLatest(parent)!!.apply {
+                objectSetField = realmSetOf(
+                    SampleWithPrimaryKey().apply {
+                        primaryKey = 1
+                        stringField = "UPDATED"
+                    }
+                )
+            }
+        }
+        realm.query<SampleWithPrimaryKey>("primaryKey = 1")
```
cmelchior, line 483: Shouldn't we also check that the `objectSetField` contains a Set with only 1 element with the UPDATED value?

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -0,0 +1,1087 @@
...
+    fun iterator_hasNext()
+    fun iterator_next()
+    fun iterator_remove()
+    fun iteratorFailsIfRealmClosed(realm: Realm)
+
+    /**
+     * Asserts structural equality for two given collections. This is needed to evaluate equality
+     * contents of ByteArrays and RealmObjects.
+     */
+    fun assertStructuralEquality(expected: Collection<T>, actual: Collection<T>)
```
cmelchior, line 575: These methods seem to be very generic in nature, should the live outside this class as free floating functions instead?

cmelchior (commented):
Very long PR, so I probably glazed over some parts. Generally looks fine though. Left some smaller comments here and there

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/IterableExt.kt
```diff
@@ -32,3 +34,17 @@ public fun <T> Iterable<T>.toRealmList(): RealmList<T> {
...
 }
+
+/**
+ * Instantiates an **unmanaged** [RealmSet] containing all the elements of this iterable.
+ */
+public fun <T> Iterable<T>.toRealmSet(): RealmSet<T> {
+    if (this is Collection) {
+        return when (size) {
+            0 -> UnmanagedRealmSet()
+            1 -> realmSetOf(if (this is List) get(0) else iterator().next())
```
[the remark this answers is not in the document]
rorbech: The pattern comes from the collection-implementation of `Iterable.toList()` which I copied to `toRealmList()`. I assume it is just an optimization ... at least I found it quite sane to also just skip the iterator when we can just grab the element immediately.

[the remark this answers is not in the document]
cmelchior: Ah yeah, I guess avoiding allocating the object would be nice 👍

cmelchior (approved):
I'm still a bit concerned about the error handling, but as long as we tackle it in a future PR I guess that is fine 👍

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -104,99 +104,92 @@ register_results_notification_cb(realm_results_t *results, jobject callback) {
...
+                jni_check_exception(jenv);
                 jenv->CallVoidMethod(static_cast<jobject>(userdata),
                                      on_change_method,
                                      reinterpret_cast<jlong>(changes));
             },
             [](void *userdata,
                const realm_async_error_t *async_error) {
                 // TODO Propagate errors to callback
-                //  https://github.com/realm/realm-kotlin/issues/303
+                //  https://github.com/realm/realm-kotlin/issues/889
```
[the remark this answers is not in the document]
edualonso: I suspect the changes I added in https://github.com/realm/realm-kotlin/pull/967/files#diff-c71ff0791782da90014db6f7000bdeea6c883ca9729f17bd1b11041d1e7711faR56 could be used here to log simple messages without having to bend over backwards. What do you think?

PR #916, opened 2022-07-01
SyncError - fix assumption that recovery path should always be present

edualonso:
Continuation from https://github.com/realm/realm-kotlin/pull/909 pointing to `releases`.

Fixed logic that wrongly assumed a recovery path should always exist when processing sync errors.
See https://github.com/realm/realm-kotlin/pull/898#issuecomment-1166207478

Added test that triggers a "permission denied" error by modifying the server's sync configuration using the Admin API.

Modified the mapping of sync errors to include permission denied and mark it as an `UnrecoverableSyncException` for users to gracefully handle app restart.

Closes https://github.com/realm/realm-kotlin/issues/895

### DIFF
--- a/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
+++ b/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
@@ -244,20 +244,21 @@ class SyncedRealmTests {
             schema = setOf(ParentPk::class, ChildPk::class),
             user = user,
             partitionValue = partitionValue
-        ).waitForInitialRemoteData()
-            .errorHandler { _, error ->
-                channel.trySend(error)
-            }.build()
+        ).errorHandler { _, error ->
+            channel.trySend(error)
+        }.build()
 
         runBlocking {
             // Remove permissions to generate a sync error containing ONLY the original path
-            // This way we assert we don't read wrong data from the user_info field in the sync error
+            // This way we assert we don't read wrong data from the user_info field
             app.asTestApp.changeSyncPermissions(SyncPermissions(read = false, write = false))
 
             val deferred = async { Realm.open(config) }
 
             val error = channel.receive()
-            assertNotNull(error)
+            val message = error.message
+            assertNotNull(message)
+            assertTrue(message.toLowerCase().contains("permission denied"))
             deferred.cancel()
 
             // Revert permissions before next tests
--- a/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
+++ b/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
@@ -39,6 +39,7 @@ import io.realm.kotlin.test.mongodb.TestApp
 import io.realm.kotlin.test.mongodb.asTestApp
 import io.realm.kotlin.test.mongodb.createUserAndLogIn
 import io.realm.kotlin.test.mongodb.shared.DEFAULT_NAME
+import io.realm.kotlin.test.mongodb.util.SyncPermissions
 import io.realm.kotlin.test.util.TestHelper
 import io.realm.kotlin.test.util.TestHelper.randomEmail
 import io.realm.kotlin.test.util.use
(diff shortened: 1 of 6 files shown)


test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -231,6 +232,40 @@ class SyncedRealmTests {
...
+            user = user,
+            partitionValue = partitionValue
+        ).errorHandler { _, error ->
+            channel.trySend(error)
+        }.build()
+
+        runBlocking {
+            // Remove permissions to generate a sync error containing ONLY the original path
+            // This way we assert we don't read wrong data from the user_info field
+            app.asTestApp.changeSyncPermissions(SyncPermissions(read = false, write = false))
```
clementetb, line 254: We are changing the server config but there is no guarantee to return it back to the original state if the test fails. To prevent cascading issues to other tests it can be wrapped with a finally block.

Another thing is that we could encapsulate this logic on a `testApp.generateSyncError(block: ()->Unit)` that automatically does the permissions change, exception catching and restore. That way we separate the logic of generating the exception from the actual test case.

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -231,6 +232,40 @@ class SyncedRealmTests {
...
+
+            val deferred = async { Realm.open(config) }
+
+            val error = channel.receive()
+            val message = error.message
+            assertNotNull(message)
+            assertTrue(message.toLowerCase().contains("permission denied"))
```
clementetb, line 261: An assert message would give some context on what is failing.

test/sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/AdminApi.kt
```diff
@@ -87,6 +87,12 @@ interface AdminApi {
      */
     suspend fun triggerClientReset(userId: String)
 
+    /**
+     * Changes the permissions for sync. Receives a lambda block which with your test logic which,
```
nhachicha, line 91:
```suggestion
     * Changes the permissions for sync. Receives a lambda block with your test logic.
```

test/sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/AdminApi.kt
```diff
@@ -87,6 +87,12 @@ interface AdminApi {
...
+     * in case it fails, will failsafe to reverting the original server permissions.
```
nhachicha, line 92:
```suggestion
     * It will safely revert to the original permissions even when an exception was thrown.
```

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -231,6 +232,42 @@ class SyncedRealmTests {
...
+
+        // Remove permissions to generate a sync error containing ONLY the original path
+        // This way we assert we don't read wrong data from the user_info field
+        runBlocking {
+            app.asTestApp.changeSyncPermissions(SyncPermissions(read = false, write = false)) {
+                runBlocking {
+                    val deferred = async { Realm.open(config) }
+
+                    val error = channel.receive()
+                    val message = error.message
```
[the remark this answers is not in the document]
edualonso: According to James:
> This flag is controlled by the server. I can’t recall the reasons for this so if you need further explanation, best to ask the server team. In terms of the client’s behaviour, even though the error is not “fatal” the client will disconnect sync, and transition to the “inactive” state.

cc. @nhachicha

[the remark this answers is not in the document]
nhachicha: I think we still need to assert the type of exception and make sure it belongs to the expected category, despite Core not setting the flag accordingly ...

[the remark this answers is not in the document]
edualonso: Now our sync error mapper returns `UnrecoverableSyncException` if we detect a permission denied error. At least users can now see the error returned in the callback is unrecoverable so that they can act upon it.

nhachicha (approved):

PR #922, opened 2022-07-08
Updated credentials key

edualonso:
Updated `REALM_S3_ACCESS_KEY` since it was already present in master but not in releases. This caused the upload debug symbols and documentation tasks to fail while releasing 1.0.1.


clementetb (approved):
Looks good 💎

nhachicha (approved):

PR #931, opened 2022-07-18
Update Core submodule remote

edualonso:
Changed Core submodule's remote to use ssh instead to allow pushing branches to origin directly from submodule.


edualonso:
Core submodule is fetched using https so let's not change that for now. This problem is fixed by changing the remote locally on the developers computer. Closing PR.

PR #933, opened 2022-07-19
Client Reset - automatic recover

edualonso:
Added `RecoverUnsycedChangesStrategy` and `RecoverOrDiscardUnsyncedChangesStrategy` and lifted restrictions on strategies depending on the Sync mode.

Similarly to the Java implementation (https://github.com/realm/realm-java/pull/7698) the API for `RecoverOrDiscardUnsyncedChangesStrategy` has
```
    public fun onAfterRecovery(before: TypedRealm, after: MutableRealm)
    public fun onAfterDiscard(before: TypedRealm, after: MutableRealm)
```
instead of the standard `onAfter` callback.

TODO:
- [x] decide whether we should change the default strategy to be `RecoverOrDiscardUnsycedChanges` to avoid breaking changes? Ask other SDKs about this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+## 1.1.0 (YYYY-MM-DD)
+
+### Breaking Changes
+* [Sync] Changed default recovery mode from `DiscardUnsyncedChangesStrategy` to `RecoverOrDiscardUnsyncedChangesStrategy`
+
+### Enhancements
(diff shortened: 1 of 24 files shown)


edualonso:
The Core commit must be updated to include a fix the includes https://github.com/realm/realm-core/pull/5726 since one of our new tests client reset integration tests triggers a "231" error which isn't present in the C-API thus crashing the test.

edualonso:
Replaced by https://github.com/realm/realm-kotlin/pull/979

Issue #945, opened 2022-07-29
Flexible sync of RealmObject with EmbeddedRealmObject field


klamborowski:
Hi :wave: 

I'm trying to use kotlin realm sdk for Flexible Sync. 

For standard Realm object like this it's work perfect:

```
open class SomeClass(
    @PrimaryKey var _id: String = "",
    var _partition: String = "",
    var card: String? = null,
) : RealmObject 
```

But for embedded object 
```
open class SomeClass(
    @PrimaryKey var _id: String = "",
    var _partition: String = "",
    var card: Card? = null,
) : RealmObject 

open class Card(
    var cardNumber: String? = null,
    var isPrivateCard: Boolean? = null,
    var uuid: String? = null
) : EmbeddedRealmObject
```

I can't pass Card:class to SyncConfiguration like bellow, cause schema argument only takes set of RealmObject, not BaseRealmObject.
```
SyncConfiguration.Builder(
        user = getCurrentUser(),
        schema = setOf(SomeClass::class, Card::class)
    )
```

But if I add only SomeClass:class
```
SyncConfiguration.Builder(
        user = getCurrentUser(),
        schema = setOf(SomeClass::class)
    )
```
I will receive the following error and app crash:

```
     Caused by: io.realm.kotlin.internal.interop.RealmCoreLogicException: [18]: Schema validation failed due to the following errors:
2022-07-29 17:07:53.490 11177-11177/my.app.dev E/AndroidRuntime: - Property 'SomeClass.card' of type 'object' has unknown object type 'Card'
```

Looking forward for your help :)

cmelchior:
Hi @klamborowski Thank you for reporting this. This is an unfortunate bug in our SyncConfiguration that wasn't caught for some reason. We should have a patch ready in a day or two. Sorry for the inconvenience.

cmelchior:
A fix should be ready in `1.0.2-SNAPSHOT`. We will make a proper release shortly.

klamborowski:
Great news! Thanks for a quick fix 😊

cmelchior:
A proper 1.0.2 release is now out.

PR #950, opened 2022-08-03
Allow Embedded Objects to be added to Sync schemas

cmelchior:
Fixes #https://github.com/realm/realm-kotlin/issues/945

### DIFF
--- a/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
+++ b/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("invisible_member", "invisible_reference")
+
 package io.realm.kotlin.test.mongodb.shared
 
 import io.realm.kotlin.CompactOnLaunchCallback
@@ -21,8 +23,12 @@ import io.realm.kotlin.InitialDataCallback
 import io.realm.kotlin.MutableRealm
 import io.realm.kotlin.Realm
 import io.realm.kotlin.TypedRealm
+import io.realm.kotlin.dynamic.DynamicRealmObject
 import io.realm.kotlin.entities.sync.ChildPk
 import io.realm.kotlin.entities.sync.ParentPk
+import io.realm.kotlin.entities.sync.flx.FlexChildObject
+import io.realm.kotlin.entities.sync.flx.FlexEmbeddedObject
+import io.realm.kotlin.entities.sync.flx.FlexParentObject
 import io.realm.kotlin.ext.query
 import io.realm.kotlin.internal.platform.createDefaultSystemLogger
 import io.realm.kotlin.internal.platform.runBlocking
@@ -34,9 +40,11 @@ import io.realm.kotlin.mongodb.exceptions.ClientResetRequiredException
 import io.realm.kotlin.mongodb.exceptions.SyncException
 import io.realm.kotlin.mongodb.sync.DiscardUnsyncedChangesStrategy
 import io.realm.kotlin.mongodb.sync.ManuallyRecoverUnsyncedChangesStrategy
+import io.realm.kotlin.mongodb.sync.PartitionValue.ValueType
 import io.realm.kotlin.mongodb.sync.SyncConfiguration
 import io.realm.kotlin.mongodb.sync.SyncMode
 import io.realm.kotlin.mongodb.sync.SyncSession
+import io.realm.kotlin.test.assertFailsWithMessage
 import io.realm.kotlin.test.mongodb.TestApp
 import io.realm.kotlin.test.mongodb.asTestApp
 import io.realm.kotlin.test.mongodb.createUserAndLogIn
@@ -44,8 +52,10 @@ import io.realm.kotlin.test.util.TestHelper
 import io.realm.kotlin.test.util.TestHelper.getRandomKey
 import io.realm.kotlin.test.util.TestHelper.randomEmail
 import io.realm.kotlin.test.util.use
+import io.realm.kotlin.types.ObjectId
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
+import java.lang.IllegalStateException
 import kotlin.random.Random
 import kotlin.test.AfterTest
 import kotlin.test.BeforeTest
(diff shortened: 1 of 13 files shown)


test/sync/src/commonTest/kotlin/io/realm/kotlin.test/Utils.kt
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.kotlin.test
+
+import kotlin.reflect.KClass
+import kotlin.test.assertFailsWith
+
+// NOTE: Copy from :base:commonTest. It is unclear if there is an easy way to share test code like
```
rorbech, line 21: The pattern is to put it in `:base:commonMain` as `sync` depends on `base`.

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -498,6 +508,81 @@ class SyncConfigTests {
...
+        val config2 = SyncConfiguration.Builder(user, "", supportedSchemaTypes).build()
+        assertEquals(3, config2.schema.size)
+    }
+
+    @Test
+    fun unsupportedSchemaTypesThrowException() {
+        val user = createTestUser()
+        val unsupportedSchemaType = setOf(DynamicRealmObject::class)
+
+        // There is no way for us to detect if a KClass implements a certain interface, this is
```
rorbech, line 573: We could easily do a public predicate method around https://github.com/realm/realm-kotlin/blob/master/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectUtil.kt#L105 that would tell if a KClass was of `BaseRealmObject`. If not in this PR we could just make an issue to track it.

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -498,6 +508,81 @@ class SyncConfigTests {
...
+        // Only do a smoke test for one constructor type as they all go through the same
+        // backing code and testing all combinations didn't seem worth.
+        val config1 = SyncConfiguration.create(user, "", supportedSchemaTypes)
+        assertEquals(3, config1.schema.size)
+        val config2 = SyncConfiguration.Builder(user, "", supportedSchemaTypes).build()
+        assertEquals(3, config2.schema.size)
+    }
+
+    @Test
+    fun unsupportedSchemaTypesThrowException() {
```
rorbech, line 569: Should this then also test the flexible sync configuration?

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -498,6 +508,81 @@ class SyncConfigTests {
...
+            }
+
+            // Flexible Sync
+            validateConfig(SyncConfiguration.create(user, supportedSchemaTypes))
+            validateConfig(SyncConfiguration.Builder(user, supportedSchemaTypes).build())
+        }
+
+        // Only do a smoke test for one constructor type as they all go through the same
+        // backing code and testing all combinations didn't seem worth.
+        val config1 = SyncConfiguration.create(user, "", supportedSchemaTypes)
```
rorbech, line 562: Any reason for having this as a separate step and not just test it in `validateConfig`?

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -498,6 +508,81 @@ class SyncConfigTests {
...
+                //     validateConfig(SyncConfiguration.Builder(user, null as RealmUUID, supportedSchemaTypes).build())
+                // }
+                ValueType.NULL -> {
+                    validateConfig(SyncConfiguration.create(user, null as String?, supportedSchemaTypes))
+                    validateConfig(SyncConfiguration.Builder(user, null as String?, supportedSchemaTypes).build())
+                }
+                else -> TODO("Test for partition type not defined")
+            }
+
+            // Flexible Sync
```
rorbech, line 555: I think it would be a bit more clear if partition and flexible API was tested in separate methods ... but 🤷

rorbech (commented):
LGTM but just some minor comments about sharing code and minor comments on the tests

rorbech:
I don't think there is an immediate way to add differentiation of `BaseRealmObject`s to the infrastructure in `TypeDescriptor.kt`, so not for this PR, but it would be really nice if we could somehow include these variations to ensure that out tests are exhaustive 🤔

test/sync/src/commonTest/kotlin/io/realm/kotlin.test/Utils.kt
cmelchior: I tried that, but the code I'm trying to share is only in `:base:commonTest` and it doesn't seem like the test configurations can be shared like that :/

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
cmelchior: Ah yeah. I could use `realmObjectCompanionOrNull` as a test for this 👍

test/sync/src/commonTest/kotlin/io/realm/kotlin.test/Utils.kt
rorbech: But stuff like `TestHelper` in `base/src/commonMain` is used in ex. `SyncedRealmTest` 🤔

cmelchior: Yes, `commonMain` works fine...but things in `commonTest` don't and our custom asserts methods require the test dependency.

rorbech: But `commonMain` of the `base` is sort or our _common test_-code ... why can't we just put it in that?

rorbech: Or is it too dangerous to have the test dependencies for `main` 🤔

rorbech: Maybe just leave it and the we can pool it with the "reconsider module structure"-task

rorbech (approved):
When CI is green ...

test/sync/src/commonTest/kotlin/io/realm/kotlin.test/Utils.kt
cmelchior: Yeah, not 100% sure what the best approach is. I'll add it to our DevEx list for now

clementetb (approved):
LGTM

PR #951, opened 2022-08-04
Fix crashes on unmapped error codes

clementetb:
The SDK would crash if core throws an exception that is not mapped by the C-API. 

This PR fixes this situation by handling these unmapped error codes into proper sync or app exceptions.


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/exceptions/RealmException.kt
```diff
@@ -9,7 +9,7 @@ package io.realm.kotlin.exceptions
  */
 public open class RealmException : RuntimeException {
     public constructor() : super()
-    public constructor(message: String) : super(message)
-    public constructor(message: String, cause: Throwable) : super(message, cause)
-    public constructor(cause: Throwable) : super(cause)
+    public constructor(message: String?) : super(message)
```
cmelchior, line 12: It was actually on purpose I made these non-null as we should always strive to have some kind of message. I can see you use `null` in the `safeErrorCodeMapping` method, but I would consider if we could just return `""` there instead since it is kind of an edge case.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -52,208 +52,226 @@ internal fun convertSyncErrorCode(error: SyncErrorCode): SyncException {
...
-                ProtocolConnectionErrorCode.RLM_SYNC_ERR_CONNECTION_BAD_DECOMPRESSION, // Error in decompression (UPLOAD)
-                ProtocolConnectionErrorCode.RLM_SYNC_ERR_CONNECTION_BAD_CHANGESET_HEADER_SYNTAX, // Bad syntax in a changeset header (UPLOAD)
-                ProtocolConnectionErrorCode.RLM_SYNC_ERR_CONNECTION_BAD_CHANGESET_SIZE -> { // Bad size specified in changeset header (UPLOAD)
-                    UnrecoverableSyncException(message)
-                }
-                ProtocolConnectionErrorCode.RLM_SYNC_ERR_CONNECTION_SWITCH_TO_FLX_SYNC, // Connected with wrong wire protocol - should switch to FLX sync
-                ProtocolConnectionErrorCode.RLM_SYNC_ERR_CONNECTION_SWITCH_TO_PBS -> { // Connected with wrong wire protocol - should switch to PBS
-                    WrongSyncTypeException(message)
+    return safeErrorCodeMapping(onException = { SyncException(it) }) {
+        when (error.category) {
```
cmelchior, line 56: Probably not worth fixing in this PR, but we would still crash if we suddenly see a category we haven't seen before. It would crash here though, but inside RealmInterop. Maybe just create an issue for it for now?

rorbech (commented):
Maybe I just don't get the overall layers of mapping, but the tests looks more like a unittest of the `AppErrorCode/SyncErrorCode`-constructor than a test verifying that we actually can handle unmapped error codes. Don't know if it possible for Sync errors, but for App-errors we could probably inject a mock HTTP-client that issued unknown error codes. Don't know if it is worth considering for this PR compared to getting the fix out though 🤔 ... has it been verified manually?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -63,7 +61,7 @@ internal fun convertSyncErrorCode(error: SyncErrorCode): SyncException {
             // See https://github.com/realm/realm-core/blob/master/src/realm/sync/protocol.hpp#L200
             // Use https://docs.google.com/spreadsheets/d/1SmiRxhFpD1XojqCKC-xAjjV-LKa9azeeWHg-zgr07lE/edit
             // as guide for how to categorize Connection type errors.
-            when (ProtocolConnectionErrorCode.fromInt(error.value)) {
+            when (error.error) {
```
edualonso, line 64: Redundant naming?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -222,24 +218,16 @@ internal fun convertAppError(error: AppError): Throwable {
...
-        }
-        SyncErrorCodeCategory.RLM_SYNC_ERROR_CATEGORY_SESSION -> {
-            ProtocolSessionErrorCode.fromInt(error.value).description
-        }
+    val categoryDesc = error.category?.description ?: error.categoryCode.toString()
+    val errorCodeDesc: String? = error.error?.description ?: when (error.category) {
         SyncErrorCodeCategory.RLM_SYNC_ERROR_CATEGORY_SYSTEM,
-        SyncErrorCodeCategory.RLM_SYNC_ERROR_CATEGORY_UNKNOWN, -> {
+        SyncErrorCodeCategory.RLM_SYNC_ERROR_CATEGORY_UNKNOWN,
+        -> {
```
edualonso, line 230: Weird indenting?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -250,8 +238,23 @@ internal fun createMessageFromSyncError(error: SyncErrorCode): String {
 
     // Combine all the parts to form an error format that is human-readable.
     // An example could be this: `[Connection][WrongProtocolVersion(104)] Wrong protocol version was used: 25`
-    val errorDesc: String = if (errorCodeDesc == null) error.value.toString() else "$errorCodeDesc(${error.value})"
-    return "[$categoryDesc][$errorDesc] ${error.message}"
+    val errorDesc: String =
+        if (errorCodeDesc == null)
+            error.errorCode.toString()
+        else
+            "$errorCodeDesc(${error.errorCode})"
```
edualonso, line 245: Leave it as a one liner or write braces for the `if`

edualonso (approved):
Only tiny things, otherwise looks good 💯

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
clementetb: Not really, but we can rename the variable `error: `SyncErrorCode into `syncErrorCode: SyncErrorCode` for readability.

cmelchior (approved):
Only minor comments 🚀

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/sync/AppErrorCategory.kt
```diff
@@ -27,17 +27,9 @@ actual enum class AppErrorCategory(actual val description: String, val nativeVal
 
     actual companion object {
 
-        actual fun fromInt(nativeValue: Int): AppErrorCategory {
-            for (value in values()) {
-                if (value.nativeValue.value.toInt() == nativeValue) {
-                    return value
-                }
+        internal actual fun of(nativeValue: Int): AppErrorCategory? =
+            values().first { value ->
```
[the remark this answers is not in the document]
clementetb: 😅 yes

Issue #952, opened 2022-08-04
Alternative to writeEncryptedCopyTo from realm-java


ganfra:
We are migrating our code from realm-java to realm-kotlin. 
We were using `writeEncryptedCopyTo` method, but it doesn't exist on realm-kotlin.
Is there anyway to do the same thing?

cmelchior:
Hi @ganfra There is currently no method like that available. You would need to copy the file manually currently. We are in the process of adding it though and it should hopefully land within the next few days: #955 

Folding this into #570

PR #955, opened 2022-08-08
Support for Realm.writeCopyTo

### LINKED ISSUE #570: Add support for `Realm.writeCopyTo()/writeEncryptedCopyTo()`
Similar to Realm Java we should expose the option of writing copies of Realm files. This enables multiple features

* Make it easy to  create backups
* Encrypt a non-encrypted Realm
* Create asset files that can be bundled in Android APK's

```java
// No proper filesystem support in Kotlin stdlib, use Strings or expose okio Paths?
Realm.writeCopyTo(destination: String)
Realm.writeEncryptedCopyTo(destination: String, key: ByteArray)
```

Note, that in order to support new Android 11 security mechanisms on external storage, we might want to expose a variant on Android that returns an outputbuffer. See https://github.com/realm/realm-java/issues/6974

cmelchior:
Closes #570 #952 

This PR adds support for `Realm.writeCopyTo()` similar to Java which had `Realm.writeCopyTo()` and `Realm.writeEncryptedCopyTo()`

All the SDK's differ slightly with how this feature is implemented, but JS has deprecated doing a copy to a path only and .NET only support using a configuration as input.

While the primary use case is probably making a simple backup of the same type, we do have some dimensions that are tricky to express with its own API, i.e. encryption and different kinds of sync. So instead of creating a lot of overloads of this method, I have only implemented one.

This does mean that the simple use case does require a little bit of code:

```
// Current API
val copyConfig = RealmConfiguration.Builder()
  .name("backup.realm")
  .encryptionKey(config.encryptionKey)
  .build()
realm.writeCopyTo(copyConfig)

// With path parameter, inheriting the current configuration
realm.writeCopyTo("path/to/backup/backup.realm")
```

If there is much demand we could implement something like `Realm.writeCopyToPath(path: String)`, but for now I would prefer to keep the API small.

The currently implementation doesn't support converting local realms to flexible sync Realms, since no subscriptions are created. So this is disallowed, but it tracked in Core here: https://github.com/realm/realm-core/issues/5711

Even though the API does work inside a write transaction, it is awkward having it there. So for now this API is only available on `Realm` with the intent to also add it to `DynamicRealm` once that becomes available. We could move it to `BaseRealm`, but it would just require more documentation and I don't really see a use case for backing up a Realm inside a write.


test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -494,6 +510,123 @@ class RealmTests {
...
+            Realm.open(configBUnencrypted)
+        }
+
+        Realm.open(configBEncrypted).use { realm ->
+            assertEquals(1000, realm.query<Parent>().count().find())
+        }
+    }
+
+    @Test
+    fun writeCopyTo_localEncryptedToLocal() {
```
rorbech, line 586: If we try to cover all cases then I guess we should theoretically also have local-encrypted to local-encrypted, but I don't know if it is worth it 🤷

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -494,6 +510,123 @@ class RealmTests {
...
+                repeat(1000) { i: Int ->
+                    copyToRealm(
+                        Parent().apply {
+                            name = "Object-$i"
+                        }
+                    )
+                }
+            }
+            realm.writeCopyTo(configB)
+        }
```
rorbech, line 538: Pattern repeats in multiple tests. Maybe throw the logic in a method.

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -71,21 +77,25 @@ import kotlin.test.assertNotNull
...
 
+@Suppress("LargeClass")
 class SyncedRealmTests {
 
     companion object {
         // Initial version of any new typed Realm (due to schema being written)
         private val INITIAL_VERSION = VersionId(2)
     }
 
+    private lateinit var tmpDir: String
```
rorbech, line 90: If not needed throughout the test maybe just have this a local variable in `createWriteCopyLocalConfig`

rorbech (approved):
LGTM - Only very minor comments

test/base/src/androidTest/kotlin/io/realm/kotlin/test/RealmTests.kt
```diff
@@ -26,9 +26,7 @@ import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertFailsWith
 import kotlin.test.assertFalse
-import kotlin.time.ExperimentalTime
 
-@OptIn(ExperimentalTime::class)
```
clementetb, line 31: Is this the only ocurrence of `ExperimentalTime`?

clementetb (approved):
LGTM - Missing changelog!

test/base/src/androidTest/kotlin/io/realm/kotlin/test/RealmTests.kt
cmelchior: I could remove it here, because we only use Duration which is stable. Other parts of the Kotlin Time API still needs this annotation.

Issue #953, opened 2022-08-05
Calling sort().first().find() doesn't work


ganfra:
Want to sort a list and retrieve the first one with `sort().first().find()` but it doesn't work.
Using `sort().find().firstOrNull()` will work instead.

cmelchior:
@ganfra When you say it doesn't work, what do you mean exactly? Can you share an example?

ganfra:
Yes sorry, I was not clear.
So the result is not returned sorted as I'd have expected.
But if I just call `find().firstOrNull()`instead it respects the sort operator.

Exemple: 

```
class Task : RealmObject {
    @PrimaryKey
    var id: Long = 0,
    var name: String = ""
    var lastEditTs: Long = 0
}
```

If I want to get the latest edited task: 
`realm.query(Task::class).sort("lastEditTs", Sort.DESCENDING).first().find() `=> will return the first edited 
`realm.query(Task::class).sort("lastEditTs", Sort.DESCENDING).find().firstOrNull()` => will return the last edited

clementetb:
@ganfra Thanks for the clarification, we were able to reproduce the issue.

PR #956, opened 2022-08-08
Calling sort().first() does not respect query ordering.

clementetb:
Ordering is not respected when calling `first` on a query. It looks like a c-api issue, see https://github.com/realm/realm-core/issues/5720

Closes https://github.com/realm/realm-kotlin/issues/953


cmelchior:
This should have been fixed in Core now

clementetb:
Looking good to me 🚀

cmelchior (approved):

PR #967, opened 2022-08-11
Update to Core 12.5.1

### LINKED ISSUE #867: User exceptions are not signalled during DiscardLocalChangesStrategy callbacks on Kotlin native
We are not able to properly handle user exceptions during the `DiscardLocalChangesStrategy.onBeforeReset()` and `DiscardLocalChangesStrategy.onAfterReset()' on mac os. 

We need a way to properly signal core that an exception has ocurred.

The sync thread would crash in this situation. A work around would be to handle the exception manually on those callbacks.

edualonso:
Fixes #867

Updated to Core 12.5.1 and fixed the following breaking changes

- Queries now receive a `realm_query_arg_t` instead of a `realm_value_t` - see https://github.com/realm/realm-core/issues/4266 for more details.
- Anonymous logins now have the option of not reusing the same anonymous user - I've implemented it the same way .NET did, see https://github.com/realm/realm-dotnet/blob/d5b334a1fe5f5d5be1139e2e7f28558e2ebe417d/Realm/Realm/Sync/Credentials.cs#L99  - thoughts @cmelchior?
- `realm_convert_with_config` now receives a boolean that triggers copying over all objects present in the realm specified by the provided configuration, though we default it to `false` to avoid copying things - thoughs @cmelchior?
- Added new recovery modes for client reset (though they are not used yet, they will be added when adding automatic reset with recovery).
- `onBefore` and `onAfter` client reset callbacks now return a boolean to allow routing errors through `onError` in Kotlin Native. As a consequence of this `SyncClientResetIntegrationJVMTests.kt` was no longer needed and was removed.
- Sync-related download and upload completion blocks have been unified under `realm_sync_wait_for_completion_func_t`.

TODO:
- [x] update changelog
- [x] add missing tests for new anonymous login with `reuseExisting = false`


packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1033,10 +1042,14 @@ actual object RealmInterop {
     ): RealmQueryPointer {
         memScoped {
             val count = args.size
-            val cArgs = allocArray<realm_value_t>(count)
+            val cArgs = allocArray<realm_query_arg_t>(count)
```
rorbech, line 1045: Looks like a code block that are used a number of times. Should probably be abstracted into a method and reused instead of copying it.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1057,12 +1060,23 @@ actual object RealmInterop {
...
+        query: String,
+        args: Array<RealmValue>
+    ): RealmQueryPointer {
         val count = args.size
-        val cArgs = realmc.new_valueArray(count)
+        val cArgs = realmc.new_queryArgArray(count)
         return memScope {
             args.mapIndexed { i, arg ->
-                realmc.valueArray_setitem(cArgs, i, managedRealmValue(arg))
+                realm_query_arg_t().apply {
```
rorbech, line 1073: Same as for native ... we should probably put this in a separate method

rorbech (commented):
LGTM - Minor comments ... but haven't evaluated the questions for @cmelchior, so will leave him to do the final approval.

CHANGELOG.md
```diff
@@ -21,7 +22,7 @@
 * Minimum Android SDK: 16.
 
 ### Internal
-* None.
+* Updated to Realm Core 12.5.0, commit 4b0a203b202b7a91caf2a387e17e5623d7b37c4a.
```
cmelchior, line 25: 🥇

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1745,7 +1766,14 @@ actual object RealmInterop {
...
                 val beforeCallback = safeUserData<SyncBeforeClientResetHandler>(userData)
                 val beforeDb = CPointerWrapper<FrozenRealmT>(beforeRealm, false)
-                beforeCallback.onBeforeReset(beforeDb)
+
+                // Check if exceptions have been thrown, return true if all went as it should
+                try {
+                    beforeCallback.onBeforeReset(beforeDb)
+                    true
+                } catch (e: Throwable) {
+                    false
```
cmelchior, line 1775: Should we log the exception here? It seems to be swallowed otherwise?

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1763,8 +1791,18 @@ actual object RealmInterop {
...
+                // afterRealm is wrapped inside a ThreadSafeReference so the pointer needs to be resolved
+                val afterRealmPtr = realm_wrapper.realm_from_thread_safe_reference(afterRealm, null)
+                val afterDb = CPointerWrapper<LiveRealmT>(afterRealmPtr, false)
+
+                // Check if exceptions have been thrown, return true if all went as it should
+                try {
+                    afterCallback.onAfterReset(beforeDb, afterDb, didRecover)
+                    true
+                } catch (e: Throwable) {
+                    false
```
cmelchior, line 1804: Same

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationProvider.kt
```diff
@@ -27,6 +27,7 @@ import io.realm.kotlin.internal.interop.sync.AuthProvider
  */
 public enum class AuthenticationProvider(id: AuthProvider) {
     ANONYMOUS(AuthProvider.RLM_AUTH_PROVIDER_ANONYMOUS),
+    ANONYMOUS_NO_REUSE(AuthProvider.RLM_AUTH_PROVIDER_ANONYMOUS_NO_REUSE),
```
cmelchior, line 30: It is a bit weird that this is its own auth provider, but I. guess we are just exposing the underlying data

cmelchior (approved):

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationProvider.kt
edualonso: Yeah, this is Core's new enum, even though it adds no value having it 🤷 . In fact, its addition caused .NET to have a crash after they updated their Core to 12.5.0: https://github.com/realm/realm-dotnet/issues/2987 so I'd rather add it to avoid problems.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
edualonso: If this returns `false` then the execution follows through `onError` instead, which the intended behaviour we sought after reporting https://github.com/realm/realm-core/pull/5564 to Yavor.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationProvider.kt
cmelchior: It looks like .NET still only have one auth provider. Is there a way for us to do something similar? 🤔

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
cmelchior: Yeah, we shouldn't throw it here, but this error is hidden so people wouldn't know they have buggy code. Could we at least do `println(e.toStrint)` here??

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
edualonso: I've extracted it to a separate function and added a comment to clarify the usage of `realm_query_arg_t` as it is a bit quirky.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationProvider.kt
edualonso: Hm, it's because they don't have an enum watchdog test but we do 😅

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
edualonso: Ah, sorry, I missed you meant log the error. Hm, we can do it easily for native but JVM is a bit more convoluted...

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -664,19 +664,25 @@ before_client_reset(void* userdata, realm_t* before_realm) {
     auto before_pointer = wrap_pointer(env, reinterpret_cast<jlong>(before_realm), false);
     env->CallVoidMethod(static_cast<jobject>(userdata), java_before_callback_function, before_pointer);
     jni_check_exception(env);
+
+    return true;
```
cmelchior, line 668: I couldn't any changes to code on the JVM side, which means that if the callback throws on the JVM, then `jni_check_exception(env);` will throw an exception instead of returning false.

Do we have any unit tests of us manually throwing an exception in `BeforeClientReset`? I

cmelchior (changes requested):
After another discussion, I think there is an edge case we are not accounting for on the JVM

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationProvider.kt
cmelchior: I meant, is there a way for us to only have one auth provider for this in the public API?

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
cmelchior: I think we need to address JVM regardless, since exceptions are leaking. I think if we can use Kotlin println...it will work correctly on both JVM and Android...It isn't ideal, but should be sufficient until we can come up with something better

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
edualonso: Yes, `onBefore` here https://github.com/realm/realm-kotlin/pull/967/files#diff-4496cd276f4c265c4a34d2d6835c8a1e290376d9d2d30380bb4dc60ac49a94ebR547
and `onAfter` here https://github.com/realm/realm-kotlin/pull/967/files#diff-4496cd276f4c265c4a34d2d6835c8a1e290376d9d2d30380bb4dc60ac49a94ebR599

edualonso: I debugged it when I wrote this for automatic client reset in a separate branch and looks like `jni_check_exception` does the catching and forwards the error appropriately.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
edualonso: Native is not an issue at all, I can call println with the exception message. For JVM I will add some helpers to extract the exception message from `api_helpers` from the JNI and then do the whole JNI thing to call `System.out.println`. I'm not sure we can call Kotlin's `println` since it's a top level function which might be tricky to find by JNI.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
edualonso: After carefully inspecting the execution it seems it does work because the error is intercepted by the error handler instead, then we detect the error itself has `isClientResetRequested` so it's routed through the client reset handler. I will change it so that it doesn't take that path and stays within the client reset flow.

edualonso: Funnily enough the before/after callbacks in the C-API throw an exception too:
```
        if (!callback(userdata.get(), &r1)) {
            throw CallbackFailed();
        }
```
so the execution stays the same regardless. I will `return false` for consistency with native though.

clementetb: The exception catch could be done in the cinterop with a lambda.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1879,8 +1882,19 @@ actual object RealmInterop {
...
+                // afterRealm is wrapped inside a ThreadSafeReference so the pointer needs to be resolved
+                val afterRealmPtr = realm_wrapper.realm_from_thread_safe_reference(afterRealm, null)
+                val afterDb = CPointerWrapper<LiveRealmT>(afterRealmPtr, false)
+
+                // Check if exceptions have been thrown, return true if all went as it should
+                try {
+                    afterCallback.onAfterReset(beforeDb, afterDb, didRecover)
+                    true
+                } catch (e: Throwable) {
+                    println(e.message)
```
clementetb, line 1895: Shouldn't we use our internal logger so this message could be captured by a custom logger?

Maybe the exception catch could happen in a higher layer.

clementetb (approved):
Looking good. I added some comments about exception catching on the client reset callbacks.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
cmelchior: The internal logger isn't available in cinterop unfortunately. I'm not sure there is good way to get this to work right now, but it is something we should track and hopefully fix when we look into combining our modules.

edualonso:
I'm merging this after discussing options for the logger, having reached the conclusion that may be worth waiting until a potential package refactor.

PR #969, opened 2022-08-15
Throw proper exception when returning invalid RealmObject from write

### LINKED ISSUE #965: Realm.write crashes with NullPointer if returning deleted managed object.
Hello guys . I found another NullPointerException  in this [code line L144](https://github.com/realm/realm-kotlin/blob/dee8f6fd32b88d0a270d7db549b6069b586f1c29/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableWriter.kt#L144)  .
The stack is following : 
`java.lang.NullPointerException
at io.realm.kotlin.internal.SuspendableWriter.freezeWriteReturnValue(SuspendableWriter.kt:144)
at io.realm.kotlin.internal.SuspendableWriter.access$freezeWriteReturnValue(SuspendableWriter.kt:47)
at io.realm.kotlin.internal.SuspendableWriter$write$2.invokeSuspend(SuspendableWriter.kt:130)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
at android.os.Handler.handleCallback(Handler.java:938)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:223)
at android.os.HandlerThread.run(HandlerThread.java:67)`

There is my [test repo and last commit](https://github.com/XandrMaster/RealmTest) for reproduce that issue .  
So this [function works fine](https://github.com/XandrMaster/RealmTest/blob/main/shared/src/androidTest/kotlin/com/example/realmtest/RepositoryTestAndr.kt#L19)  , but [this one with runBlocking](https://github.com/XandrMaster/RealmTest/blob/main/shared/src/androidTest/kotlin/com/example/realmtest/RepositoryTestAndr.kt#L12) cause to Exception :( . 
For iOs test the same Exception . 
But call the same function from Activity coroutines also works fine . The problem in the tests.
Maybe am I doing something wrong ?
Thanks for help an advance !

cmelchior:
Closes #965

We didn't correctly check if users tried to return deleted objects from a write, which caused a NullPointer when we tried to freeze an invalid object.

We basically have two choices on what to do here:

1) Allow sending back a deleted object without trying to freeze it.
2) Throw some error if it happens.

I opted for 2) as I cannot see any use case where sending back a deleted object would actually be the behavior wanted, but if someone feel that there is a use case, I'll be happy to change it.

Regarding the exception type, I was torn a bit between `IllegalArgument` and `IllegalState`, but opted for `IllegalState` as argument exceptions are usually only used for input to a public function. Again, this can probably be debated, so if anyone can come up with good arguments for making it an `IllegalArgument`, I'll be happy to change it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,27 @@
+## 1.0.3 (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
+### Enhancements
+* None.
+
+### Fixed
+* Returning deleted objects from `Realm.write` and `Realm.writeBlocking` threw a non-sensical `NullPointerException`. Returning such a value is not allowed and now throws an `IllegalStateException`. (Issue [#965](https://github.com/realm/realm-kotlin/issues/965))
+
+### Compatibility
+* This release is compatible with:
+  * Kotlin 1.6.10 and above.
+  * Coroutines 1.6.0-native-mt. Also compatible with Coroutines 1.6.0 but requires enabling of the new memory model and disabling of freezing, see https://github.com/realm/realm-kotlin#kotlin-memory-model-and-coroutine-compatibility for details on that.
+  * AtomicFu 0.17.0.
+* Minimum Gradle version: 6.1.1.  
+* Minimum Android Gradle Plugin version: 4.0.0.
+* Minimum Android SDK: 16.
+
+### Internal
+* None.
+
+
 ## 1.0.2 (2022-08-05)
 
 ### Breaking Changes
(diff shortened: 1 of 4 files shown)


edualonso (approved):
I think it's a fair assumption to throw an exception 👍

rorbech (approved):

PR #978, opened 2022-08-18
Fix test converting a partition-based Realm into a local Realm.

cmelchior:
Local schema checks are more restrictive than synced Realms, so all server-side fields must also be part of the schema or be removed through a migration.

### DIFF
--- a/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
+++ b/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
@@ -570,7 +570,12 @@ class SyncedRealmTests {
         }
     }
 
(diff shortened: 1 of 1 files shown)


test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -683,11 +691,24 @@ class SyncedRealmTests {
...
+            syncRealm.syncSession.downloadAllServerChanges(30.seconds)
+
             // Copy to partition-based Realm
             syncRealm.writeCopyTo(localConfig)
         }
-        // Open Local Realm and check that data can read.
-        Realm.open(localConfig).use { localRealm: Realm ->
+
+        // Opening the local Realm with the same schema will throw a schema mismatch, because
+        // the server schema contains classes and fields not in the local schema.
```
edualonso, line 705: 👍

edualonso (approved):
🚀

PR #979, opened 2022-08-18
Client Reset - automatic recover

### LINKED ISSUE #925: Implement Client Reset w/ Recovery

edualonso:
Closes #925 
Replaces https://github.com/realm/realm-kotlin/pull/933

Added `RecoverUnsycedChangesStrategy` and `RecoverOrDiscardUnsyncedChangesStrategy` and lifted restrictions on strategies depending on the Sync mode.

Similarly to the Java implementation (https://github.com/realm/realm-java/pull/7698) the API for `RecoverOrDiscardUnsyncedChangesStrategy` has
```
    public fun onAfterRecovery(before: TypedRealm, after: MutableRealm)
    public fun onAfterDiscard(before: TypedRealm, after: MutableRealm)
```
instead of the standard `onAfter` callback.

Changes from previous implementations:
- `DiscardUnsynced.onError` is deprecated in favour of `onManualFallback` as per alignment across SDKs.

### Note:
Removing the Command Server has uncovered some defective configuration in the server apps in some cases. It appears we are configuring our test apps in some way that makes some tests non-deterministic and messes executions around across runs and platforms. We will need to revisit the app configuration framework at some point but for now I introduced the following changes:
- Created a test app for each Client Reset test and delete it after each test
- For some reason some JVM tests timed out so it was necessary to increase the timeout setting in the HttpClient for Client Reset tests, not all others. In fact, the client's timeout can now be configured ad-hoc. This increases the execution time, but it seems to be sufficient to avoid timeouts. Overall the execution time has increased by 2-3 minutes, but that was also expected as a few new client reset tests had been added.
- In addition two new client reset apps with recovery disabled have been added to avoid conflicts with existing apps which also get deleted after the execution of their respective tests has finished.

TODO:
- [x] decide whether we should change the default strategy to be `RecoverOrDiscardUnsycedChanges` to avoid breaking changes? Ask other SDKs about this.


edualonso:
Wait for merging this until https://github.com/realm/realm-core/issues/5705 is dealt with.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncConfiguration.kt
```diff
@@ -355,22 +356,28 @@ public interface SyncConfiguration : Configuration {
...
-         * partition-based applications only accept [DiscardUnsyncedChangesStrategy].
-         *
          * In case no strategy is defined a default one (which does not do any data migration and
-         * its only job is logging information on the client reset process itself) will be used.
+         * its only job is logging information on the client reset process itself) will be used in
+         * case Recovery is not enabled on Atlas.
          *
          * @param resetStrategy custom strategy to handle client reset.
          */
+        // Flexible Sync applications only accept [ManuallyRecoverUnsyncedChangesStrategy] whereas
```
clementetb, line 365: It can be removed. Flexible sync allows all reset strategies.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncConfiguration.kt
```diff
@@ -355,22 +356,28 @@ public interface SyncConfiguration : Configuration {
...
+        // partition-based applications only accept [DiscardUnsyncedChangesStrategy].
         public fun syncClientResetStrategy(resetStrategy: SyncClientResetStrategy): Builder =
             apply {
-                if (partitionValue == null && resetStrategy is DiscardUnsyncedChangesStrategy) {
-                    throw IllegalArgumentException("DiscardUnsyncedChangesStrategy is not supported on Flexible Sync")
-                }
-                if (partitionValue != null && resetStrategy is ManuallyRecoverUnsyncedChangesStrategy) {
-                    throw IllegalArgumentException("ManuallyRecoverUnsyncedChangesStrategy is not supported on Partition-based Sync")
-                }
+                // if (partitionValue == null && resetStrategy is DiscardUnsyncedChangesStrategy) {
```
clementetb, line 369: Can be removed.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/SyncConfigurationImpl.kt
```diff
@@ -85,160 +146,248 @@ internal class SyncConfigurationImpl(
...
+        session: SyncSession,
+        appPointer: RealmAppPointer,
+        error: SyncError
+    ) {
+        // If there is a user exception we appoint it as the cause of the client reset
+        strategy.onError(
+            session,
+            ClientResetRequiredException(appPointer, error)
+        )
+        // TODO call both onError and onManualResetFallback until the deprecated function is removed
```
clementetb, line 364: We could create a ticket to track this deprecation.

clementetb (approved):
Some minor stuff. Heads up, in the latest core update the client reset simulation has slightly changed, see https://github.com/realm/realm-java/blob/d7f3fda52cbde0c4f62815d57bce44e3254797fa/realm/realm-library/src/main/cpp/io_realm_mongodb_sync_Sync.cpp#L70

PR #990, opened 2022-08-26
Support for building and publishing each architecture independently.

cmelchior:
Our current `cinterop` package has a few problems if you want to build and publish packages for each platform independently. This PR fixes those issues. They were discovered while trying to add parallel builds on Github Actions.


cmelchior:
Outdated

PR #991, opened 2022-08-29
Upgrade to latest core

rorbech:


packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1382,10 +1382,6 @@ actual object RealmInterop {
                         e.printStackTrace()
                     }
                 },
-                staticCFunction { userdata, asyncError ->
```
cmelchior, line 1385: How are these being reported now?

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+## 1.1.1 (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
+### Enhancements
+* None.
+
+### Fixed
+* Ordering not respected for `RealmQuery.first()`. (Issue [#953](https://github.com/realm/realm-kotlin/issues/953))
```
cmelchior, line 10: I guess we don't really know this until https://github.com/realm/realm-kotlin/pull/956 has been merged?

cmelchior (approved):
I'm a bit concerned about the removal of error reporting. Will we just crash if those callbacks throw? But if CI is green we can merge it.

CHANGELOG.md
rorbech: No, I guess you are right. I just cherry-picked this core-update from #956, but technically you are right. Will remove the fixed-entry.

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
rorbech: Don't know why they were exposed in the C-API in the first place, because it sounds like it never should have been ... https://github.com/realm/realm-core/commit/2f1167e828e20e94fbff2f3d39dd609339f941e8

rorbech:
> I'm a bit concerned about the removal of error reporting. Will we just crash if those callbacks throw?

See https://github.com/realm/realm-kotlin/pull/991#discussion_r957344112

PR #995, opened 2022-09-07
Add support for `MutableRealmInt`

### LINKED ISSUE #588: Datatype support: MutableRealmInteger
Similar to Realm Java, we should add support for a `MutableRealmInteger` with counter semantics when synchronized: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java

edualonso:
Closes #588

`MutableRealmInt`s are mutable, `Long`-like, numeric quantities. When used inside a managed `RealmObject` these properties behave like a conflict-free replicated `Long` providing `increment` and `decrement` functions that automatically resolve conflicts - see https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type.

The API is similar to the Java SDK's (see https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java) with a few adjustments that make it more idiomatic and adapted to the Kotlin language:
- introduction of operators supported by `Long`: `plus`, `minus`, `times`, `div`, `rem`, `inc`, `dec`
- support for nullable values using Kotlin's type system - see below

The API consists of:
- `get`
- `set`
- `increment`/`decrement`

Nullability is handled using Kotlin's type system, i.e. `MutableRealnt` instances **cannot** store null values (as opposed to the Java SDK), but the class property itself is allowed to be nullable:
```
class Sample : RealmObject {
    var counter: MutableRealmInt = MutableRealmInt.of(42)
    var nullableCounter: MutableRealmInt? = null
}
```

`MutableRealmInt` properties are mutable, i.e. must be `var` - though we don't support immutable properties in `RealmObjects`s for the time being so... 🤷 . This represents a change compared to the Java SDK where these properties must be `final` and could also be nullable, which made things a bit convoluted. Sharing managed values across `RealmObject` instances behaves in the same way as with existing Realm primitive type properties:
```
val counter: MutableRealmInt = MutableRealmInt.of(42)
managedFoo.counterField = counter
managedBar.counterField = managedFoo.counterField
managedBar.counterField.increment(1)
println(managedFoo.counterField.get()) // 42
println(managedBar.counterField.get()) // 43
```

We believe these design decisions feel idiomatic. In addition, the .NET SDK implements these semantics as well so alignment has also played an important role.


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/MutableRealmIntegerInternal.kt
```diff
@@ -0,0 +1,92 @@
...
+    override fun equals(other: Any?): Boolean {
+        if (other === this) return true
+        if (other !is MutableRealmInt) return false
+
+        val thisValue = get()
+        val otherValue = other.get()
+        return thisValue == otherValue
+    }
+
+    override fun hashCode(): Int {
```
cmelchior, line 57: Should we also override `toString` to print something meaningful? At least the value.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/MutableRealmIntegerInternal.kt
```diff
@@ -0,0 +1,92 @@
...
+        if (other === this) return true
+        if (other == null) return false
+        if (other !is MutableRealmInt) return false
+
+        val thisValue = get()
+        val otherValue = other.get()
+        return thisValue == otherValue
+    }
+
+    override fun hashCode(): Int = get().hashCode()
```
cmelchior, line 91: Same with toString here

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmIntTests.kt
```diff
@@ -0,0 +1,290 @@
...
+    @AfterTest
+    fun tearDown() {
+        if (this::realm.isInitialized && !realm.isClosed()) {
+            realm.close()
+        }
+        PlatformUtils.deleteTempDir(tmpDir)
+    }
+
+    @Test
+    fun unmanaged_basic() {
```
cmelchior, line 60: The wording and behaviour of this test isn't great IMO. The name indicates that this tests Sample more than MutableRealmInt and the test itself doesn't test the interesting things around set, increment and decrement. I would probably just delete it

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmIntTests.kt
```diff
@@ -0,0 +1,290 @@
...
+        equalityTest(sample1, sample2)
+    }
+
+    @Test
+    fun unmanaged_nullability() {
+        nullabilityTest(Sample())
+    }
+
+    @Test
+    fun unmanaged_compareTo() {
```
cmelchior, line 79: I'm missing specific tests for `increment`, `decrement`, especially corner cases like overflow/underflow, what happens if increment is given a negative number, and things like that.

Also specific tests for `set`

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmIntTests.kt
```diff
@@ -0,0 +1,290 @@
...
+        bar.mutableRealmIntField = foo.mutableRealmIntField
+        bar.mutableRealmIntField.increment(1)
+        val fooValue = foo.mutableRealmIntField.get()
+        val barValue = bar.mutableRealmIntField.get()
+
+        assertEquals(fooValue, barValue)
+    }
+
+    @Test
+    fun managed_basic() {
```
cmelchior, line 105: Same as unmanaged, these tests do not seem relevant for testing MutableRealmInt?

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmIntTests.kt
```diff
@@ -0,0 +1,290 @@
...
+            val managedFoo = copyToRealm(Sample())
+            val managedBar = copyToRealm(Sample())
+
+            managedFoo.mutableRealmIntField = counter
+            managedBar.mutableRealmIntField = managedFoo.mutableRealmIntField
+            managedBar.mutableRealmIntField.increment(1)
+            val managedFooValue = managedFoo.mutableRealmIntField.get()
+            val managedBarValue = managedBar.mutableRealmIntField.get()
+
+            // Values obviously diverge since we don't copy the reference but the value, just as we
```
cmelchior, line 206: We should probably document this behaviour in the Dokka for the class since this behaviour differs between unmanaged and managed objects.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -58,7 +59,10 @@ class RealmSetTests {
     private lateinit var realm: Realm
 
     private val managedTesters: List<SetApiTester<*, RealmSetContainer>> by lazy {
-        descriptors.map {
+        descriptors.filter {
+            // Filter out MutableRealmInts
+            it.elementType.classifier != MutableRealmInt::class
```
cmelchior, line 64: Why?

test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/list/ListTests.kt
```diff
@@ -32,8 +33,8 @@ import kotlin.test.assertTrue
 class ListTests {
 
     private val supportedPrimitiveTypes = TypeDescriptor.classifiers.keys.filter {
-        // Filter out RealmObject
-        it != RealmObject::class
+        // Filter out RealmObject and MutableRealmInt
```
cmelchior, line 36: Why? It seems like having MutableRealmInts in lists should just work?

test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/set/SetTests.kt
```diff
@@ -32,8 +33,8 @@ import kotlin.test.assertTrue
 class SetTests {
 
     private val supportedPrimitiveTypes = TypeDescriptor.classifiers.keys.filter {
-        // Filter out RealmObject
-        it != RealmObject::class
+        // Filter out RealmObject and MutableRealmInt
```
cmelchior, line 36: I guess supporting MutableRealmInt in Sets would be weird, and could easily break very fast.

test/sync/src/commonMain/kotlin/io/realm/kotlin/entities/sync/SyncObjectWithAllTypes.kt
```diff
@@ -118,12 +121,16 @@ class SyncObjectWithAllTypes : RealmObject {
...
                                     { obj: SyncObjectWithAllTypes ->
+                                        obj.mutableRealmIntField = MutableRealmInt.of(42)
+                                        obj.mutableRealmIntNullableField = null
                                         obj.intField = 42
                                         obj.intNullableField = 42
                                         obj.intRealmList = realmListOf(42)
                                         obj.intRealmSet = realmSetOf(42)
                                     },
                                     { obj: SyncObjectWithAllTypes ->
+                                        assertEquals(MutableRealmInt.of(42), obj.mutableRealmIntField)
```
cmelchior, line 132: Also missing tests of `increment` and `decrement` working across Sync. Maybe not that multiple clients converge, but at least that increment and decrement actually works when hitting the server.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,139 @@
...
+ *      .counter
+ *      .increment(1)
+ * ```
+ * Note that the [set] operator must be used with extreme care. It will quash the effects of any
+ * prior calls to [increment] or [decrement]. Although the value of a `MutableRealmInt` will always
+ * converge across devices, the specific value on which it converges will depend on the actual order
+ * in which operations took place. Mixing [set] with [increment] and/or [decrement] is, therefore,
+ * not advised, unless fuzzy counting is acceptable.
+ *
+ * `MutableRealmInt`s **may not be primary keys**.
```
cmelchior, line 46:
```suggestion
 * `MutableRealmInt`s **cannot be primary keys**.
```

cmelchior: Is this tested anywhere? At least I didn't seem to be able to find it

cmelchior (changes requested):
Nice start 💯 ... I have a few comments that I think we need to look at:

1. Testing looks like it can be improved for some of the methods.
2. Why are we not adding List support?
3. Some minor comments about the public API and Dokkal: constructor naming and operator loading

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Sets the `MutableRealmInt` value.
+     *
+     * Calling [set] forcibly sets the `MutableRealmInt` to the provided value. Doing this
+     * obliterates the effects of any calls to [increment] and [decrement] perceived before the call
+     * to `set`.
+     *
+     * @param value new value.
+     */
+    public fun set(value: Long)
```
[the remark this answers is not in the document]
edualonso: I don't think assignment operators allow overloading: https://kotlinlang.org/docs/operator-overloading.html

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
```diff
@@ -57,6 +58,7 @@ class Sample : RealmObject {
     var objectIdField: ObjectId? = ObjectId.create()
     var uuidField: RealmUUID? = RealmUUID.random()
     var byteArrayField: ByteArray? = null
+    var mutableRealmInt: MutableRealmInt? = MutableRealmInt.of(42)
     var child: Child? = null
 
     // List types
```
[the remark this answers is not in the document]
edualonso: I didn't add it because we don't support it in Java. I assumed we would go for the same restriction.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmIntTests.kt
edualonso: That's fine but it should be the same for all other primitive types. Should we add it elsewhere more visible? I find it a bit counterproductive to adding it only for `MutableRealmInt` and not making it visible for all other primitive types.

test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
edualonso: As I mentioned we don't support this data type for lists, collections or sets in Java so I went for the same constraint.

test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/list/ListTests.kt
edualonso: See other comments.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
edualonso: It is, the compiler `PrimaryKeyTests` do it (https://github.com/realm/realm-kotlin/blob/el/mutable-realm-integer/test/base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/PrimaryKeyTests.kt#L53). Also in the library `PrimaryKeyTests` we check the number of supported types matches the expected types written in the test (https://github.com/realm/realm-kotlin/blob/el/mutable-realm-integer/test/base/src/androidTest/kotlin/io/realm/kotlin/test/shared/PrimaryKeyTests.kt#L156)

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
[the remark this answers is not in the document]
edualonso: .NET does not support this either, mentioning it is due to some implementation challenges in Core. I see to ask the Core elders about this 🧙‍♂️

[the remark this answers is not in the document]
cmelchior: Hmm, I guess if it is missing from Core, we cannot add it :D, but would be nice to understand why at least

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,227 @@
...
+
+    override fun toFloat(): Float = get().toFloat()
+
+    override fun toInt(): Int = get().toInt()
+
+    override fun toLong(): Long = get()
+
+    override fun toShort(): Short = get().toShort()
+
+    override fun toString(): String = get().toString()
```
cmelchior, line 154: Maybe : "RealmMutableInt{${get()}}", just so we know it is not a normal Long

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,227 @@
...
+ * ```
+ * val counter: MutableRealmInt = realm.query<User>("user = $0", "Fred")
+ *      .first()
+ *      .find()!!
+ *      .counter
+ * ```
+ * the `counter` holds a reference to the `User` model object from which it was obtained. Neither
+ * can be GCed until all references to both are unreachable.
+ */
+public abstract class MutableRealmInt : Comparable<MutableRealmInt>, Number() {
```
cmelchior, line 70: Turning this into an abstract class looks pretty good 👍

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -236,26 +258,33 @@ internal object RealmObjectHelper {
...
+                    else -> IllegalStateException(
+                        "Cannot set `${obj.className}.$${obj.metadata[key]!!.name}` to `${value.value}`: $NOT_IN_A_TRANSACTION_MSG",
+                        exception
+                    )
                 }
             }
         }
     }
 
+    internal fun setMutableInt(
```
rorbech, line 278: Do we really need specialized versions of this?! I would imagine that we should just define an appropriate converter. Maybe it isn't obvious and worth it since `MutableRealmInt` isn't supported in the collection types, but I would prefer to follow the generic pattern until we decide to do specialized paths for any of the types.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -442,7 +444,7 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
                                     )
                                 }
                                 val indexableTypes = with(pluginContext.irBuiltIns) {
-                                    setOf(byteType, charType, shortType, intType, longType, stringType, realmInstantType, objectIdType, realmUUIDType).map { it.classifierOrFail }
+                                    setOf(byteType, charType, shortType, intType, longType, stringType, realmInstantType, objectIdType, realmUUIDType, realmMutableIntType).map { it.classifierOrFail }
```
rorbech, line 447: I guess this should ideally be using the same infrastructure as the test's `TypeDescriptor` but maybe we could at least for now just move these kind of meta-information to a central place in the compiler plugin source, so that it is not spread across the generators.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,139 @@
...
+        return thisValue.compareTo(otherValue)
+    }
+
+    public companion object {
+        /**
+         * Creates a new, unmanaged [MutableRealmInt] with the specified initial value.
+         *
+         * @param value initial value.
+         */
+        public fun of(value: Long): MutableRealmInt = UnmanagedMutableRealmInt(value)
```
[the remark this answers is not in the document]
rorbech: Hmmm. For some reason we ended up with `RealmInstant.from` in https://github.com/realm/realm-kotlin/pull/857. Can't recall the argument, but maybe the same applies here? 🤪

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,227 @@
...
+
+    /**
+     * Increments the `MutableRealmInt`, adding the value of the argument. Increment/decrement
+     * operations from all devices are reflected in the new value, which is guaranteed to converge.
+     * This operation should not be confused with the [inc] operator, which returns a new value
+     * whereas this one modifies the value itself.
+     *
+     * @param value quantity to be added to the `MutableRealmInt`.
+     */
+    public abstract fun increment(value: Number)
```
rorbech, line 97: Do we really need these (also `decrement`), when there is other ways to achieve the same with the operators? Just adds more test cases, but really convenience 🤷

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/MutableRealmIntInternal.kt
```diff
@@ -0,0 +1,92 @@
...
+import io.realm.kotlin.types.BaseRealmObject
+import io.realm.kotlin.types.MutableRealmInt
+
+internal class ManagedMutableRealmInt(
+    private val obj: RealmObjectReference<out BaseRealmObject>,
+    private val propertyKey: PropertyKey,
+    private val converter: RealmValueConverter<Long>
+) : MutableRealmInt() {
+
+    private enum class Operation {
```
rorbech, line 31: Seems like a very cumbersome way of aligning two different implementation paths just to split them in the `operationInternal` ... why?

rorbech (commented):
Gave it a spin. I tried to review with a different focus that the other reviews, so haven't verified all individual tests. Feels a bit going against the converter infrastructure, but maybe ok if not required for collection support 🤔

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
edualonso: I've tried and it's possible to return unmanaged instances using the current framework (whose behaviour I wasn't very familiar with), though getting it to work for managed instances requires some non-trivial changes in the compiler plugin. I will try to devise a solution that doesn't turn into a timesink, otherwise I will leave it as is and will add a todo and create a task for it.

test/base/src/commonMain/kotlin/io/realm/kotlin/test/util/TypeDescriptor.kt
```diff
@@ -260,8 +262,12 @@ public object TypeDescriptor {
         val collectionType: CollectionType,
         val elementType: ElementType
     ) {
-        val isPrimaryKeySupported: Boolean =
-            collectionType == CollectionType.RLM_COLLECTION_TYPE_NONE && elementType.realmFieldType.primaryKeySupport
+        // MutableRealmInts CANNOT be primary keys, but the CoreFieldType is INT and INTs CAN be.
```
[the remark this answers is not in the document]
edualonso: I think it's fine having it. This logic is rendered useless with my latest changes.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
rorbech: Whao, that feels really weird to me ... wonder what the reasoning about `.inc` and `.dec` not changing the value is 🤔

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/MutableRealmIntInternal.kt
rorbech: But you are introducing an enum just for the sake of being able to `when`-it. Then maybe do a inline method that takes a lambda with the different inner operations and the message.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/MutableRealmInt.kt
```diff
@@ -0,0 +1,139 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.kotlin.types
+
+import io.realm.kotlin.internal.UnmanagedMutableRealmInt
+
+/**
+ * A `MutableRealmInt` is a mutable, [Long]-like, numeric quantity. It behaves almost exactly as a
```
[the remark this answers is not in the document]
edualonso: This was the headline used in the Java SDK and I don't find it particularly descriptive. I think with the description below we can safely remove it and won't need more clarifications.

test/sync/src/commonMain/kotlin/io/realm/kotlin/entities/sync/SyncObjectWithAllTypes.kt
edualonso: Sure, I've added a new specific test as integrating it in the roundtrip test we have is a bit of a hassle.

test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -577,6 +578,64 @@ class SyncedRealmTests {
...
+        // Upload data with default values
+        Realm.open(config).use { realm ->
+            realm.write { copyToRealm(masterObject) }
+            realm.syncSession.uploadAllLocalChanges()
+        }
+
+        // Increment counter twice asynchronously
+        val updates = async {
+            for (i in 0..1) {
+                Realm.open(config).use { realm ->
```
cmelchior, line 602: I'm not sure this actually tests the interesting case. You open the same Realm twice which means that all changes are done to the same file on disk, I would probably open two different Realms.

cmelchior (approved):
Nice 👍

PR #998, opened 2022-09-09
Test of subqueries on a result

rorbech:
This just adds a test to highlight that the current C-API implementation doesn't take filtering of the result into account when it does a subquery on the result.


clementetb (approved):

rorbech:
Superseded by #1013

PR #1000, opened 2022-09-12
Fix assemble crashing in the test project

cmelchior:
Pulling this change from https://github.com/realm/realm-kotlin/pull/782 into `master` to keep the PR smaller.

This PR just fixes a smaller annoyance that calling `./gradlew assemble` crashes in the test project.


rorbech (approved):
Nice 🎉

PR #1001, opened 2022-09-12
Refactor EmailPassword Auth tests.

cmelchior:
In the past, we have seen quite a lot of flakiness in the EmailPasswordAuth tests. The root cause was never discovered. With the refactor to Ktor 2, we started to see them again. This PR refactors the tests, so instead of modifying the config of a single app, we now use different test apps.

This seems to fix the flakiness + has the advantage of improving the speed of running tests.

### DIFF
--- a/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
+++ b/test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
@@ -41,11 +41,11 @@ import io.realm.kotlin.mongodb.sync.SyncSession
 import io.realm.kotlin.mongodb.sync.SyncSession.ErrorHandler
 import io.realm.kotlin.mongodb.syncSession
 import io.realm.kotlin.notifications.ResultsChange
+import io.realm.kotlin.query.RealmResults
 import io.realm.kotlin.test.mongodb.TestApp
 import io.realm.kotlin.test.mongodb.asTestApp
 import io.realm.kotlin.test.mongodb.createUserAndLogIn
 import io.realm.kotlin.test.mongodb.shared.DEFAULT_NAME
-import io.realm.kotlin.test.mongodb.util.SyncPermissions
 import io.realm.kotlin.test.platform.PlatformUtils
 import io.realm.kotlin.test.util.TestHelper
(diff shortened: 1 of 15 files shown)


test/sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -258,24 +260,24 @@ class SyncedRealmTests {
...
+            val deferred = async {
+                Realm.open(config)
+                // Make sure that the test eventually fail. Coroutines can cancel a delay
+                // so this doesn't always block the test for 10 seconds.
+                delay(10 * 1000)
+                channel.trySend(AssertionError("Realm was successfully opened"))
             }
+
+            val error = channel.receive()
+            assertTrue(error is UnrecoverableSyncException, "Was $error")
```
rorbech, line 273: Isn't this a job for `assertFailsWithMessage`?

test/sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/AdminApi.kt
```diff
@@ -94,14 +107,9 @@ interface AdminApi {
...
-     * Set whether or not automatic confirmation is enabled.
-     */
-    suspend fun setAutomaticConfirmation(enabled: Boolean)
-
-    /**
-     * Set whether or not custom confirmation is enabled.
+     * Set type of authentication method used for the EmailAuth provider.
      */
-    suspend fun setCustomConfirmation(enabled: Boolean)
+    suspend fun setAuth(auth: UserConfirmationMode)
```
rorbech, line 112: Do we even use this anymore?

rorbech (commented):
LGTM - Only very minor comments

rorbech (approved):
LGTM - Only very minor comments (in previous comment ... forgot to approve) ... and of course ... when CI is green

test/sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/AdminApi.kt
cmelchior: No, I was a bit afraid of removing it, but I should just get rid of it.

PR #1004, opened 2022-09-13
New project infrastructure for tests

rorbech:
This PR migrates our SDK tests from its own Gradle project (using Gradle composite build with `includeBuild("../packages")`  to subprojects of the SDK project (`packages`). 

To use the new test infrastructure you should open the `packages` project in the IDE and dependent on the Gradle property `testRepository` it will either use all the local submodules or artifacts from the specified repository. See the various options in https://github.com/realm/realm-kotlin/blob/58a194f6558c2a9c84b2beeb47b20f8d4a0287f6/packages/gradle.properties#L35

Since the new setup only applies the compiler plugin on the `packages/test-X` modules, a new `integration-tests` folder have been added to hold projects that test. 

The description of the new setup has been summarized in the updated [`./CONTRIBUTING.md`](https://github.com/realm/realm-kotlin/pull/1004/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055).

**Pros/Cons**
- 👍 The new setup takes advantage of applying the compiler plugin to our test modules with `kotlinCompilerPluginClasspath`/`kotlinNativeCompilerPluginClasspath` which enables us to have fully incremental compilation rippling to the test project when updating the compiler plugin
- 👍 Works around issue with unresolved symbols in composite builds with Kotlin Multiplatform projects (https://youtrack.jetbrains.com/issue/KTIJ-15775/MPP-IDE-Lots-of-red-code-unresolved-references-with-HMPP-and-composite-build)
- 👎 Only applies the compiler plugin in test, so pattern will not work if we start doing additional logic in the top level Gradle plugin

**TODO**
- [x] Move `test/base` to `packages/test-base`  
- [x] Move `test/sync` to `packages/test-sync`  
- [x] Clean up old `test` project - I just wiped it 🤷 
- [x] Fix CI to execute new tests
- [x] Come up with proper pattern for integration test of our whole `gradle-plugin` as the test now applies the compiler plugin directly. Maybe we should just move `examples` into `integration-test` and ensure that we execute one of them on all targets!?
  - A new `integration-tests` folder have been added to hold projects that test application of the top level `io.realm.kotlin:gradle-plugin`-Gradle plugin  
- [ ] Get rid of the last red lines ... Seems like ` commonTest` symbols are not accessible from at least `androidTest` in the test projects 😞 
  - Will be postponed to another PR
- [x] Finalize Documentation 
- [x] Maybe find a pattern to separate SDK modules from test modules so that it is easy to execute tasks only on one of the categories. Right now setting `testRepository` will skip including the SDK-modules, but maybe this logic should be separated from `testRespository` as you currently wouldn't publish anything if none of the SDK-modules are included.
  - It even tricked myself with the implicit logic around `testDirectory`, so added `includeSdkModules` and `includeTestModules` and separated the logic around which modules to include from the `testRepository`. See [`./CONTIBUTING.md`](https://github.com/realm/realm-kotlin/pull/1004/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055)for a description.


packages/gradle.properties
```diff
@@ -31,3 +31,15 @@ android.useAndroidX=true
 # Allow the default dependency name to match the client debug build type. Otherwise the client project has to
 # explicitly specify the debug variant (ex: implementation("io.realm.kotlin:library-base-debug:0.8.0"))
 kotlin.android.buildTypeAttribute.keep=false
+
+# testRepository - will control how the gradle project is configured. The behavior overall falls in
+# two categories.
+# - If unset or the empty string the project is setup for local development where tests
+#   (test-base/test-sync) is executed against sub-projects
+#testRepository=
```
cmelchior, line 39: Remove?

packages/gradle.properties
```diff
@@ -31,3 +31,15 @@ android.useAndroidX=true
...
+# - If set then tests (test-base/test-sync) are executed against the artifacts in the specified
```
cmelchior, line 40: We should probably also mention that doing this will disable all non-test modules.

cmelchior (commented):
Really great... I just played around with it a little bit and at least and most things seems to work. Although I'm probably missing some corner cases.

But a massive improvement 💯

CONTRIBUTING.md
```diff
@@ -7,44 +7,103 @@ We welcomes all contributions! The only requirement we have is that, like many o
...
 
 Checkout repo:
 ```
 git clone --recursive  https://github.com/realm/realm-kotlin.git 
 ```
 
-Build library:
+### Building and running test
+
+To support simultaneous development of the SDK and test, while still selectively allowing to run
```
cmelchior, line 27: This is also described under "Writing Tests", maybe just keep it there?

integration-tests/README.md
```diff
@@ -0,0 +1,9 @@
+# Integration test projects
+
+This folder holds the various integration test projects.
+
+- `gradle-plugin-test` - Smoke test project that verifies that our top level Gradle plugin can be
+  applied on a both  single and a multi platform modules. It is currently currently testing:
```
cmelchior, line 6:
```suggestion
  applied on a both single and a multi platform modules. It is currently testing:
```

integration-tests/gradle-plugin-test/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
+plugins {
+    id("com.android.library") version "7.2.2" apply false
+    id("org.jetbrains.kotlin.android") version "1.7.10" apply false
```
cmelchior, line 3: Should this version come from `buildSrc/Config.kt`? I think that is available here? 🤔

integration-tests/gradle-plugin-test/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// Explicitly adding the plugin to the classpath as it makes it easier to control the version
+// centrally (don't need version in the 'plugins' block). Further, snapshots are not published with
+// marker interface so would need to be added to the classpath manually anyway.
+buildscript {
+    dependencies {
+        // FIXME How to share version through here ... should we go
```
cmelchior, line 11:
```suggestion
        // FIXME How to only define the current version number in one place?
```

integration-tests/gradle-plugin-test/gradle.properties
```diff
@@ -0,0 +1,8 @@
+kotlin.code.style=official
+
+org.gradle.jvmargs=-Xmx4096M
+
+android.useAndroidX=true
+android.enableJetifier=true
+
+kotlin.mpp.stability.nowarn=true
```
cmelchior, line 8:
```suggestion
kotlin.mpp.stability.nowarn=true

```

integration-tests/gradle-plugin-test/multi-platform/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+    kotlin("multiplatform")
+    id("io.realm.kotlin")
+}
+
+kotlin {
+    jvm()
+
+    val hostOs = System.getProperty("os.name")
+    val isMingwX64 = hostOs.startsWith("Windows")
+    val nativeTarget = when {
```
cmelchior, line 11: I guess this is fine for this project, but this means it is kinda hidden which platform we are testing on. Not sure if it matters though?

integration-tests/gradle-plugin-test/multi-platform/src/commonMain/kotlin/io/realm/test/multiplatform/model/TestClass.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import io.realm.kotlin.types.RealmObject
+import io.realm.kotlin.types.annotations.PrimaryKey
+
+class TestClass : RealmObject {
+    @PrimaryKey
+    var id: Long = 0
+
+    var text: String = "INIT"
+}
```
cmelchior, line 11: Looks like we should also enable ktlint/detekt for these projects?

integration-tests/gradle-plugin-test/settings.gradle.kts
```diff
@@ -0,0 +1,20 @@
+rootProject.name = "gradle-plugin-test"
+
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        google()
+        mavenCentral()
+        maven("file://${rootDir.absolutePath}/../../packages/build/m2-buildrepo")
```
cmelchior, line 8: Shouldn't these use the `testRepository` option as well? At least it feels weird that we can customize the repository path and then hardcode it here?

integration-tests/gradle-plugin-test/single-platform/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
cmelchior, line 1: Lets please only use a single top-level gitignore file for the repo.

integration-tests/gradle-plugin-test/single-platform/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+}
+
+dependencies {
+    implementation("androidx.appcompat:appcompat:1.5.1")
+    implementation("io.realm.kotlin:library-base:1.2.0-SNAPSHOT")
```
cmelchior, line 35: Same as before. Would be nice if we can move this version to a central location.

integration-tests/gradle-plugin-test/single-platform/src/androidTest/kotlin/io/realm/test/singleplatform/CrudTests.kt
```diff
@@ -0,0 +1,72 @@
...
+    fun setup() {
+        tmpDir = Files.createTempDirectory("android_tests").absolutePathString()
+        realm = RealmConfiguration.Builder(setOf(TestClass::class))
+            .directory(tmpDir)
+            .build()
+            .let { Realm.open(it) }
+    }
+
+    @After
+    fun teadDown() {
```
cmelchior, line 37:
```suggestion
    fun tearDown() {
```

packages/gradle.properties
```diff
@@ -31,3 +31,8 @@ android.useAndroidX=true
 # Allow the default dependency name to match the client debug build type. Otherwise the client project has to
 # explicitly specify the debug variant (ex: implementation("io.realm.kotlin:library-base-debug:0.8.0"))
 kotlin.android.buildTypeAttribute.keep=false
+
+# Project setup - See './CONTRIBUTING.md' for description of the project structure and various options.
+# includeSdkModules=true
+# includeTestModules=false
```
cmelchior, line 37:
```suggestion
# includeTestModules=true
```

packages/test-sync/build.gradle.kts
```diff
@@ -21,7 +21,32 @@ plugins {
...
-    id("io.realm.kotlin")
+    // Test relies on the compiler plugin, but we cannot apply our full plugin from within the same
+    // gradle run, so we just apply the compiler plugin directly as a dependency below instead
+    // id("io.realm.kotlin")
+}
+
+
+// Test relies on the compiler plugin, but we cannot apply our full plugin from within the same
+// gradle run, so we just apply the compiler plugin directly
+dependencies {
```
cmelchior, line 32: It is a bit magical and quite a lot of code, but since the end result is soo much better I think we can live with this until hopefully composite builds actually work in the IDE.

cmelchior (approved):
Really big upgrade to the experience of working with the SDK code 🥇 💯 🕺 Only left a bunch of minor comments, otherwise this looks really great.

clementetb (commented):
It does look promising, I wasn't able to assemble it locally.

CONTRIBUTING.md
```diff
@@ -7,44 +7,103 @@ We welcomes all contributions! The only requirement we have is that, like many o
...
-### Commands to build from source
+### Obtaining the source code 
 
 Checkout repo:
 ```
 git clone --recursive  https://github.com/realm/realm-kotlin.git 
 ```
 
-Build library:
+### Building and running test
```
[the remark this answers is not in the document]
rorbech: I tried pulling out a brief _"Building and running tests"_ and _"Running tests against Maven artifacts"_ and then put the details in a _"Advanced project setup"_ section ... hope it clears it up a bit? 😬

CONTRIBUTING.md
rorbech: Yeah. I tried to put the motivation into this section ... but has not reworked it in to the _"Advanced project setup"_ section instead.

integration-tests/gradle-plugin-test/build.gradle.kts
rorbech: I was intentionally trying to avoid the `buildSrc`-symlink hack, but since we also require linting-setup I just fell back on it to easily get access to all the versions and plugins 😞

integration-tests/gradle-plugin-test/multi-platform/build.gradle.kts
rorbech: Yes. I didn't really know what was the best. I actually like that it is just given from the host platform what you are testing on, but maybe it is "too perpendicular" (🤪) to our other testing patterns 🤔 ... I think we could just rework it later if we have a better approach ... Maybe when GHA is in place

integration-tests/gradle-plugin-test/multi-platform/src/commonMain/kotlin/io/realm/test/multiplatform/model/TestClass.kt
rorbech: Yep. Enabled our `realm-lint` plugin for this project too.

integration-tests/gradle-plugin-test/settings.gradle.kts
rorbech: I guess it should, but it is a bit tricky as it is a relative path and this project is at a different level than the other projects using it. I would deem it for another PR to fix this up, but I can add a `TODO`.

integration-tests/gradle-plugin-test/single-platform/build.gradle.kts
rorbech: See https://github.com/realm/realm-kotlin/pull/1004#discussion_r976319361

integration-tests/gradle-plugin-test/settings.gradle.kts
cmelchior: TODO is fine

Issue #1024, opened 2022-09-25
Realm -SNAPSHOT seems incompatible with Kotlin 1.7.2


sebleclerc:
Hi! 

I came across this issue: https://github.com/realm/realm-kotlin/issues/1002
I seems to have a similar error but WITH the fix in place. I did update my project with Kotlin 1.7.20-RC and my realm module fails with

```
e: org.jetbrains.kotlin.backend.common.CompilationException: Back-end: Please report this problem https://kotl.in/issue
[...]/persistence/src/commonMain/kotlin/persistence/models/DBAdresse.kt:-1:-1
Details: Internal error in file lowering: java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.ir.declarations.IrTypeParametersContainer org.jetbrains.kotlin.backend.common.ir.IrUtilsKt.getClassIfConstructor(org.jetbrains.kotlin.ir.declarations.IrTypeParametersContainer)'
        at io.realm.kotlin.compiler.IrUtilsKt.copyTo$default(IrUtils.kt:323)
        at io.realm.kotlin.compiler.RealmModelSyntheticPropertiesGeneration.addVariableProperty(RealmModelSyntheticPropertiesGeneration.kt:623)
        at io.realm.kotlin.compiler.RealmModelSyntheticPropertiesGeneration.addRealmObjectInternalProperties(RealmModelSyntheticPropertiesGeneration.kt:197)
        at io.realm.kotlin.compiler.RealmModelLowering.lower(RealmModelLoweringExtension.kt:89)
        at org.jetbrains.kotlin.backend.common.ClassLoweringVisitor.visitClass(Lower.kt:101)
        at org.jetbrains.kotlin.ir.visitors.IrElementVisitorVoid$DefaultImpls.visitClass(IrElementVisitorVoid.kt:111)
        at org.jetbrains.kotlin.backend.common.ClassLoweringVisitor.visitClass(Lower.kt:92)
        at org.jetbrains.kotlin.backend.common.ClassLoweringVisitor.visitClass(Lower.kt:92)
        at org.jetbrains.kotlin.ir.declarations.IrClass.accept(IrClass.kt:64)
        at org.jetbrains.kotlin.ir.declarations.IrFile.acceptChildren(IrFile.kt:36)
        at org.jetbrains.kotlin.ir.visitors.IrVisitorsKt.acceptChildrenVoid(IrVisitors.kt:15)
[...]

PR #1038, opened 2022-09-29
Compile example with Kotlin 1.7.20

cmelchior:
Kotlin 1.7.20 has been released. So bumping our example to that.

We still need to investigate why https://github.com/realm/realm-kotlin/issues/1024 doesn't seem to be caught on CI.

TODO
- [ ] Our kmm-sample does setup in slightly different way, which doesn't seem to catch https://github.com/realm/realm-kotlin/issues/1024. We need to figure out why and fix it. And also fix 1.7.20 support in the process.


cmelchior:
Superceded by https://github.com/realm/realm-kotlin/pull/1059

PR #1039, opened 2022-09-30
Update compatibility section.

cmelchior:
Be more specific about current restrictions around Kotlin and Ktor versions.


JacobOscarGunnarsson (approved):
Seems like CI has failed but these changes shouldn't be that cause for that

cmelchior:
Yeah, i just cancelled CI. There was no need for it to run.

PR #1044, opened 2022-10-03
Jg/user methods

### LINKED ISSUE #751: Add User remaining methods
[x] getIdentities()
[ ] getProviderType()
[ ] getAccessToken()
[ ] getRefreshToken()
[ ] getDeviceId()
[x] linkCredentials(credentials)

- - -
In Kotlin, this is exposed as properties, so:

```
interface User {
  val providerType: AuthenticationProvider
  val refreshToken: String
  val accessToken: String
  val deviceId: String
}
```

JacobOscarGunnarsson:
Add the remaining methods for a User. Closes #751


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +69,34 @@ public interface User {
      */
     public val identities: List<UserIdentity>
 
+    /**
```
JacobOscarGunnarsson, line 72: Should all of these fields be treated as public? Not sure why but feels concerning to have the tokens accessible in this manner

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/UserTests.kt
```diff
@@ -111,6 +111,34 @@ class UserTests {
         }
     }
 
+    @Test
+    fun getProviderType() = runBlocking {
+        val email = randomEmail()
+        val emailUser = createUserAndLogin(email, "123456")
+        assertEquals(AuthenticationProvider.EMAIL_PASSWORD, emailUser.provider)
+    }
+
```
JacobOscarGunnarsson, line 120: Are these tests sufficient?

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -2123,6 +2127,18 @@ actual object RealmInterop {
         return AuthProvider.of(realm_wrapper.realm_auth_credentials_get_provider(credentials.cptr()))
     }
 
+    actual fun realm_user_get_access_token(user: RealmUserPointer): String {
+        return realm_wrapper.realm_user_get_access_token(user.cptr())
```
cmelchior, line 2131: I think you need to use `safeKString` here and for the other strings, otherwise the underlying data might go out of scope.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
cmelchior: Yes, they are part of the public API. The tokens are something developers should treat carefully, but they are needed if you e.g. want to call REST API's on App Services yourself.

Also, since the app is running in a secure sandbox, it shouldn't be possible to access them there.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/UserTests.kt
```diff
@@ -111,6 +111,34 @@ class UserTests {
...
+    @Test
+    fun getAccessToken() = runBlocking {
```
cmelchior, line 122: I would also test what this returns when the user is logged out...I'm actually not sure. I suspect it still returns the token, which the server will just reject. But capturing the behaviour would be nice

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/UserTests.kt
cmelchior: Generally, we try to test:

1) Success case
2) Boundaries
3) Error cases

2/3 doesn't really apply in this case though, except maybe the cases I mentioned where the user is either logged out or deleted.

So, yes, I would say this is probably sufficient.

edualonso (approved):
Looks good 💯

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,34 @@ public interface User {
...
+     * Returns the provider type used to log the user
```
cmelchior, line 72: This seems to cut off.

Maybe "Returns the provider type used to log the user in."

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,34 @@ public interface User {
...
+     *
+     * @return the provider type of the user
```
cmelchior, line 74: We don't really use `@return` for properties. I would just remove this since the above sentence covers everything.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,34 @@ public interface User {
...
+     */
+    public val provider: AuthenticationProvider
+
+    /**
+     * Returns the current access token for the user.
+     *
+     * @return the current access token.
```
cmelchior, line 81: Remove

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,34 @@ public interface User {
...
+     */
+    public val accessToken: String
+
+    /**
+     * Returns the current refresh token for the user.
+     *
+     * @return the current refresh token.
```
cmelchior, line 88: Remove

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
cmelchior: Also, we should document the behaviour when the user logs out.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,34 @@ public interface User {
...
+
+    /**
+     * Returns the current access token for the user.
+     *
+     * @return the current access token.
+     */
+    public val accessToken: String
+
+    /**
+     * Returns the current refresh token for the user.
```
cmelchior, line 86: We should also document the behaviour when logging out.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/UserTests.kt
```diff
@@ -111,6 +111,43 @@ class UserTests {
         }
     }
 
+    @Test
+    fun getProviderType() = runBlocking {
+        val email = randomEmail()
+        val emailUser = createUserAndLogin(email, "123456")
+        assertEquals(AuthenticationProvider.EMAIL_PASSWORD, emailUser.provider)
```
cmelchior, line 118: What happens if you log out?

cmelchior (changes requested):
Nice 👍 .. Only had a few minor comments.

We also need to update the CHANGELOG though.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,30 @@ public interface User {
      */
     public val identities: List<UserIdentity>
 
+    /**
+     * Returns the provider type used to log the user in
```
cmelchior, line 72:
```suggestion
     * Returns the provider type used to log the user in.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,30 @@ public interface User {
...
+     * If a user logs out the authentication provider persists.
```
cmelchior, line 73:
```suggestion
     * If a user logs out, the authentication provider last used to log the user in will still be returned.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,30 @@ public interface User {
...
+     */
+    public val provider: AuthenticationProvider
+
+    /**
+     * Returns the current access token for the user.
+     * If a user logs out the access token is removed.
```
cmelchior, line 79:
```suggestion
     * If a user logs out, an empty access token is returned.
```

Removed is an ambiguous word IMO, so rephrased a little.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,30 @@ public interface User {
...
+     */
+    public val accessToken: String
+
+    /**
+     * Returns the current refresh token for the user.
+     * If a user logs out the refresh token is removed.
```
cmelchior, line 85:
```suggestion
     * If a user logs out an empty refresh token is returned.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -68,6 +68,30 @@ public interface User {
...
+     */
+    public val refreshToken: String
+
+    /**
+     * Returns a unique identifier for the device the user logged in to.
+     * If a user logs out the device id persists.
```
cmelchior, line 91:
```suggestion
```

I would just remove this sentence here since the device ID doesn't seem connected to the user being logged in or not.

cmelchior (changes requested):
Looks good 👍 . Just missing an update to the enhancement section in the CHANGELOG.md file in the root of the project. Then this looks ready to merge.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 
 ### Enhancements
 * Added support for `MutableRealmInt` in model classes. The new type behaves like a reference to a `Long`, but also supports `increment` and `decrement` methods. These methods implement a conflict-free replicated data type, whose value will converge even when changed across distributed devices with poor connections.
+* [Sync] Support for `User.getProviderType()`.
```
cmelchior, line 8: We just released 1.2.0, so these need to go into a new 1.3.0 section

cmelchior (approved):

PR #1048, opened 2022-10-04
Add support for deleteAll()

### LINKED ISSUE #574: Add support for `MutableRealm.deleteAll()`
Similar to Realm Java, we should expose the option of easily deleting data from all types in the schema.

```kotlin
MutableRealm.deleteAll()
```

JacobOscarGunnarsson:
Closes #574 and #573.
Adds functionality that allows a user to wipe their current realm by invoking deleteAll().
Adds functionality to delete all objects of a specified class with delete(classname) or delete(KClass).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ### Breaking Changes
 
 ### Enhancements
+* Support for `MutableRealm.deleteAll()`.
+* Support for `MutableRealm.delete()`.
+* Support for `DynamicMutableRealm.deleteAll()`.
+* Support for `DynamicMutableRealm.delete()`.
 * [Sync] Support for `User.getProviderType()`.
 * [Sync] Support for `User.getAccessToken()`.
 * [Sync] Support for `User.getRefreshToken()`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 ### Enhancements
 * Support for `MutableRealm.deleteAll()`.
-* Support for `MutableRealm.delete()`.
+* Support for `MutableRealm.delete(KClass)`.
 * Support for `DynamicMutableRealm.deleteAll()`.
-* Support for `DynamicMutableRealm.delete()`.
+* Support for `DynamicMutableRealm.delete(className)`.
 * [Sync] Support for `User.getProviderType()`.
 * [Sync] Support for `User.getAccessToken()`.
 * [Sync] Support for `User.getRefreshToken()`.
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
@@ -126,4 +126,11 @@ public interface MutableRealm : TypedRealm {
      * @throws IllegalArgumentException if the object is invalid, frozen or not managed by Realm.
      */
     public fun delete(deleteable: Deleteable)
+
+    /**
+     * Deletes all objects from this Realm.
+     *
+     * @throws IllegalStateException if the Realm is closed.
+     */
+    public fun deleteAll()
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
@@ -136,10 +136,16 @@ public interface MutableRealm : TypedRealm {
      * Deletes all objects of the specified class from the Realm.
      *
      * @param KClass the class whose objects should be removed.
+     * @throws IllegalStateException if the class does not exist within the schema.
      */
     public fun delete(schemaClass: KClass<out BaseRealmObject>)
 }
 
+/**
+ * Deletes all objects of the specified class from the Realm.
+ *
+ * Reified convenience wrapper of [MutableRealm.delete].
+ */
 public inline fun <reified T : BaseRealmObject> MutableRealm.delete() {
     delete(T::class)
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
@@ -129,8 +129,6 @@ public interface MutableRealm : TypedRealm {
 
     /**
      * Deletes all objects from this Realm.
-     *
-     * @throws IllegalStateException if the Realm is closed.
      */
     public fun deleteAll()
 
@@ -138,7 +136,6 @@ public interface MutableRealm : TypedRealm {
      * Deletes all objects of the specified class from the Realm.
      *
      * @param KClass the class whose objects should be removed.
-     * @throws IllegalStateException if the Realm is closed.
      */
     public fun delete(schemaClass: KClass<out BaseRealmObject>)
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
@@ -135,7 +135,7 @@ public interface MutableRealm : TypedRealm {
     /**
      * Deletes all objects of the specified class from the Realm.
      *
-     * @param KClass the class whose objects should be removed.
+     * @param schemaClass the class whose objects should be removed.
      * @throws IllegalStateException if the class does not exist within the schema.
      */
     public fun delete(schemaClass: KClass<out BaseRealmObject>)
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
@@ -136,7 +136,7 @@ public interface MutableRealm : TypedRealm {
      * Deletes all objects of the specified class from the Realm.
      *
      * @param schemaClass the class whose objects should be removed.
-     * @throws IllegalStateException if the class does not exist within the schema.
+     * @throws IllegalArgumentException if the class does not exist within the schema.
      */
     public fun delete(schemaClass: KClass<out BaseRealmObject>)
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
@@ -133,4 +133,16 @@ public interface MutableRealm : TypedRealm {
      * @throws IllegalStateException if the Realm is closed.
      */
     public fun deleteAll()
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param KClass the class whose objects should be removed.
+     * @throws IllegalStateException if the Realm is closed.
+     */
+    public fun delete(schemaClass: KClass<out BaseRealmObject>)
+}
+
+public inline fun <reified T : BaseRealmObject> MutableRealm.delete() {
+    delete(T::class)
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicMutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicMutableRealm.kt
@@ -102,4 +102,11 @@ public interface DynamicMutableRealm : DynamicRealm {
      * @throws IllegalArgumentException if the object is invalid, frozen or not managed by Realm.
      */
     public fun delete(deleteable: Deleteable)
+
+    /**
+     * Deletes all objects from this Realm.
+     *
+     * @throws IllegalStateException if the Realm is closed.
+     */
+    public fun deleteAll()
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicMutableRealm.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicMutableRealm.kt
@@ -106,12 +106,13 @@ public interface DynamicMutableRealm : DynamicRealm {
     /**
      * Deletes all objects of the specified class from the Realm.
      *
-     * @param schemaClassName the class whose objects should be removed.
+     * @param className the class whose objects should be removed.
+     * @throws IllegalArgumentException if the class does not exist within the schema.
      */
-    public fun delete(schemaClassName: String)
+    public fun delete(className: String)
 
     /**
-     * Deletes all objects from this Realm.
+     * Deletes all objects from the defined schema in the current Realm.
      */
     public fun deleteAll()
 }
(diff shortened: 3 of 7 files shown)


JacobOscarGunnarsson:
Since [this PR](https://github.com/realm/realm-kotlin/pull/1044) has updated the changeling I'll wait for that to be released before merging master into this, in that way I can use the new CHANGELOG.md to update this branch

JacobOscarGunnarsson:
Also need to implement this functionality for dynamic Realms

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -675,4 +675,28 @@ class MutableRealmTests {
             }
         }
     }
+
```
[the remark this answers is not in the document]
cmelchior: You can try to do this instead:

```
fun deleteAll_shouldThrowOnClosedRealm() { 
  realm.write { 
    realm.close()  
    assertFailsWith<IllegalStateException> { deleteAll() } 
  } 
}
```

But I'm not sure that is a use case we care about, e.g. `MutableRealm.delete()` doesn't mention anything about what happens when a Realm is closed.

So I would probably just remove the documentation and test around what happens when a Realm is closed. Our current API means that you _really_ have to go out of your way to make that happen.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -675,4 +675,28 @@ class MutableRealmTests {
...
+    @Test
+    fun deleteAll() {
+        realm.writeBlocking {
+            for (i in 0..9) {
+                copyToRealm(Parent())
```
cmelchior, line 683: I would probably cut down on the number of types in here. Having 5 seems overkill, especially if there are no differences between them

I would try to aim to test different schema types, i.e. `RealmObject`, `RealmObject` with primary key and an `EmbeddedRealmObject`

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/dynamic/DynamicMutableRealmTests.kt
```diff
@@ -642,4 +642,25 @@ class DynamicMutableRealmTests {
             }
         }
     }
+
+    @Test
+    fun deleteAll() {
+        dynamicMutableRealm.run {
+            for (i in 0..9) {
+                copyToRealm(DynamicMutableRealmObject.create("Sample"))
```
cmelchior, line 650: Same as above... Having 4 similar types seem overkill?

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/dynamic/DynamicMutableRealmTests.kt
```diff
@@ -642,4 +642,25 @@ class DynamicMutableRealmTests {
             }
         }
     }
+
+    @Test
+    fun deleteAll() {
```
cmelchior, line 647: For dynamic Realms, I would also test the case where the schema changes, i.e. what happens if you have a realm with data for `schema = setOf(Person, Dog)` but then open the same Realm with `schema = setOf(Person)`..what happens with the dog elements? (I know the answer, but it would be nice with a test for it).

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
```diff
@@ -126,4 +126,11 @@ public interface MutableRealm : TypedRealm {
      * @throws IllegalArgumentException if the object is invalid, frozen or not managed by Realm.
      */
     public fun delete(deleteable: Deleteable)
+
+    /**
+     * Deletes all objects from this Realm.
+     *
+     * @throws IllegalStateException if the Realm is closed.
```
cmelchior, line 133: See below, but I think we can just remove this.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicMutableRealm.kt
```diff
@@ -102,4 +102,11 @@ public interface DynamicMutableRealm : DynamicRealm {
      * @throws IllegalArgumentException if the object is invalid, frozen or not managed by Realm.
      */
     public fun delete(deleteable: Deleteable)
+
+    /**
+     * Deletes all objects from this Realm.
+     *
+     * @throws IllegalStateException if the Realm is closed.
```
cmelchior, line 109: Same

cmelchior (changes requested):
Nice. Great start 👍 .... Only minor comments and a suggestion for an additional test that will capture the current behaviour

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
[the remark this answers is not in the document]
JacobOscarGunnarsson: seems like you're unable to invoke realm.close() within a write transaction as well (it causes an `IllegalStateException`), so in that sense I find it hard to call the delete method on a closed realm since realm.write also causes an exception when invoked on a closed realm

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
```diff
@@ -126,4 +126,23 @@ public interface MutableRealm : TypedRealm {
...
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param KClass the class whose objects should be removed.
+     * @throws IllegalStateException if the Realm is closed.
+     */
+    public fun delete(schemaClass: KClass<out BaseRealmObject>)
+}
+
```
edualonso, line 145:
```suggestion
    /**
     * Deletes all objects of the specified class from the Realm.
     * 
     * Reified convenience wrapper of [MutableRealm.delete].
     */
```
See how other reified extension functions are documented, for example
```
/**
 * Returns a [RealmQuery] matching the predicate represented by [query].
 *
 * Reified convenience wrapper of [TypedRealm.query].
 *
 * @param query the Realm Query Language predicate to append.
 * @param args Realm values for the predicate.
 */
public inline fun <reified T : BaseRealmObject> TypedRealm.query(
    query: String = "TRUEPREDICATE",
    vararg args: Any?
): RealmQuery<T> = query(T::class, query, *args)
```

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 ### Breaking Changes
 
 ### Enhancements
+* Support for `MutableRealm.deleteAll()`.
+* Support for `MutableRealm.delete()`.
```
cmelchior, line 7:
```suggestion
* Support for `MutableRealm.delete(KClass)`.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/dynamic/DynamicMutableRealm.kt
```diff
@@ -102,4 +102,16 @@ public interface DynamicMutableRealm : DynamicRealm {
      * @throws IllegalArgumentException if the object is invalid, frozen or not managed by Realm.
      */
     public fun delete(deleteable: Deleteable)
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param schemaClassName the class whose objects should be removed.
+     */
+    public fun delete(schemaClassName: String)
```
cmelchior, line 111: What happens if you provide a className that doesn't exist? I assume we throw an `IllegalArgumentException`. In that case we should add a `@throws` to the Dokka.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
```diff
@@ -126,4 +126,20 @@ public interface MutableRealm : TypedRealm {
...
+     * Deletes all objects from this Realm.
+     */
+    public fun deleteAll()
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param KClass the class whose objects should be removed.
+     */
+    public fun delete(schemaClass: KClass<out BaseRealmObject>)
```
cmelchior, line 140: What happens if you provide a class that doesn't exist? I assume we throw an IllegalArgumentException. In that case we should add a @throws to the Dokka.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -675,4 +678,31 @@ class MutableRealmTests {
             }
         }
     }
+
+    @Test
+    fun deleteAll() {
```
cmelchior, line 683: We are missing tests for `delete(className)` both in the success case and if you provide a classname that isn't part of the schema both for typed and dynamic realms.

cmelchior (changes requested):
Nice...we are lacking tests for the new `delete(className)` methods though.

CHANGELOG.md
JacobOscarGunnarsson: yeah, the `delete(className)` is implicitly tested through `deleteAll()` but is probably better to have explicit test cases for it as well

cmelchior: Yes, we try to have explicit unit tests of all public API's. This way we make sure we always test it, even if we refactor `deleteAll()` at some point.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
```diff
@@ -126,4 +126,26 @@ public interface MutableRealm : TypedRealm {
...
+
+    /**
+     * Deletes all objects from this Realm.
+     */
+    public fun deleteAll()
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param KClass the class whose objects should be removed.
```
cmelchior, line 138:
```suggestion
     * @param schemaClass the class whose objects should be removed.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
```diff
@@ -126,4 +126,26 @@ public interface MutableRealm : TypedRealm {
...
+     * @throws IllegalStateException if the class does not exist within the schema.
```
cmelchior, line 139: It feels a bit weird that we are throwing `IllegalStateException` here. Especially since the DynamicRealm is thrown `IllegalArgumentException` I don't know if this is something we can change at the lower layers, but otherwise, we can catch the IllegalState, check the message and remap to `IllegalArgumentException` as needed. We are doing that elsewhere in a few places.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -675,4 +680,98 @@ class MutableRealmTests {
...
+            delete(EmbeddedParent::class)
+            assertEquals(0, query<EmbeddedParent>().count().find())
+            assertEquals(0, query<EmbeddedChild>().count().find())
+        }
+    }
+
+    @Test
+    fun delete_nonExistingClassThrows() {
+        realm.writeBlocking {
+            assertFailsWith<IllegalStateException> {
```
cmelchior, line 772: See comment above. It feels wrong that this is `IllegalStateException`. We should find a way to make it a ` IllegalArgumentException`

cmelchior (changes requested):

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/MutableRealm.kt
```diff
@@ -126,4 +126,26 @@ public interface MutableRealm : TypedRealm {
...
+    /**
+     * Deletes all objects from this Realm.
+     */
+    public fun deleteAll()
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param schemaClass the class whose objects should be removed.
+     * @throws IllegalStateException if the class does not exist within the schema.
```
cmelchior, line 139: We need to change this to `IllegalArgumentException`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalMutableRealm.kt
```diff
@@ -66,4 +67,22 @@ internal interface InternalMutableRealm : MutableRealm {
     override fun delete(deleteable: Deleteable) {
         deleteable.asInternalDeleteable().delete()
     }
+
+    override fun delete(schemaClass: KClass<out BaseRealmObject>) {
+        try {
+            delete(query(schemaClass).find())
+        } catch (err: IllegalStateException) {
+            if (err.message?.contains("not part of this configuration schema") == true) {
+                throw IllegalArgumentException()
```
cmelchior, line 76: I would pass on the message here, so `throw IllegalArgumentException(err.message)`

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -675,4 +680,98 @@ class MutableRealmTests {
...
+        realm2.writeBlocking {
+            assertEquals(10, query<Sample>().count().find())
+            deleteAll()
+            assertEquals(0, query<Sample>().count().find())
+        }
+        realm.writeBlocking {
+            assertEquals(0, query<Sample>().count().find())
+            assertEquals(10, query<SampleWithPrimaryKey>().count().find())
+            assertEquals(10, query<EmbeddedParent>().count().find())
+            assertEquals(10, query<EmbeddedChild>().count().find())
```
cmelchior, line 747: These asserts don't need to be inside a `writeBlocking`. They should work fine when moved outside and then using `realm.query()`.

cmelchior: You can also do this `with(realm) { asssertX }` or `realm.run { assertX }`

cmelchior: We don't have any strict guidelines about which to use, but I would probably consider `with(realm) { }` the most readable

cmelchior (commented):
Almost there...just a few minor comments about docs and code style

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/MutableRealmTests.kt
```diff
@@ -675,4 +680,97 @@ class MutableRealmTests {
...
+        }
+        val realm2 = Realm.open(RealmConfiguration.Builder(schema = setOf(Sample::class)).directory(tmpDir).build())
+        realm2.writeBlocking {
+            assertEquals(10, query<Sample>().count().find())
+            deleteAll()
+            assertEquals(0, query<Sample>().count().find())
+        }
+        realm.writeBlocking {
```
cmelchior, line 736: Nitpick: I would probably leave a comment here that you are doing the write to force the Realm to the newest version

cmelchior (approved):
Nice 🚀

Issue #1046, opened 2022-10-03
Slow bootstrap time for synced Realms


cmelchior:
It looks like Realm Kotlin is bootstrapping sync data in a way that might be much slower than Realm Java.

Basically, Realm Java wraps Object Store Async Open https://github.com/realm/realm-java/blob/5be74cfabc6674ef13db92c0180c6eab6eb6c34e/realm/realm-library/src/objectServer/java/io/realm/internal/objectstore/OsAsyncOpenTask.java#L20 here: https://github.com/realm/realm-java/blob/8d54cdba51df3a8cce8abccb58a6c9aac16799d2/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java#L276

Whereas Kotlin creates the Realm locally, then calls `SyncSession::wait_for_download_completion`: https://github.com/realm/realm-kotlin/blob/master/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/SyncConfigurationImpl.kt#L73

For an empty server Realm, this distinction doesn't matter. But for a large Realm, it means the server is integrating all of the server changes against the local state, which can take quite a while. 

When using async open, we are instead downloading the full serve Realm and then validates the local schema against that. 

In my case, the difference was something like 8 minutes on Kotlin vs. 4 seconds on Java.

cmelchior:
Fixed in 1.3.0

PR #1050, opened 2022-10-04
Use async open when opening partition-based Realms.

cmelchior:
Closing #1046 

This has been verified to fix the difference reported in https://jira.mongodb.org/browse/HELP-37904.

There are some open questions about how to clean up threadsafe references. It looks like we might be leaking memory in the client reset tests.

### DIFF
--- a/packages/jni-swig-stub/realm.i
+++ b/packages/jni-swig-stub/realm.i
@@ -164,7 +164,7 @@ std::string rlm_stdstr(realm_string_t val)
     };
 }
 
-// reuse void callback type as template for `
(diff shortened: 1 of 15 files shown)


packages/jni-swig-stub/realm.i
```diff
@@ -164,6 +164,19 @@ std::string rlm_stdstr(realm_string_t val)
     };
 }
 
+// reuse void callback type as template for `
```
clementetb, line 167:
```suggestion
// reuse void callback type as template for `realm_async_open_task_completion_func_t` function
```

clementetb (approved):
Looking good!

PR #1052, opened 2022-10-05
Add RealmInstant.now()

JacobOscarGunnarsson:
Adds ability to fetch a RealmInstant representing how long it was since the Unix epoch

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
@@ -2,6 +2,7 @@ package io.realm.kotlin.internal.platform
 
 import io.realm.kotlin.log.LogLevel
 import io.realm.kotlin.log.RealmLogger
+import io.realm.kotlin.types.RealmInstant
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KType
 
@@ -108,6 +109,11 @@ public expect fun threadId(): ULong
  * Returns UNIX epoch time in seconds.
  */
 public expect fun epochInSeconds(): Long
+
+/**
+ * Returns a RealmInstant representing the time that has passed since the Unix epoch.
+ */
+public expect fun internalNow(): RealmInstant
 
 /**
  * Returns the type of a mutable property.
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmInstant.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmInstant.kt
@@ -16,6 +16,7 @@
 package io.realm.kotlin.types
 
 import io.realm.kotlin.internal.RealmInstantImpl
+import io.realm.kotlin.internal.platform.internalNow
 
 /**
  * A representation of a Realm timestamp. A timestamp represent a single point in time defined as
@@ -72,6 +73,11 @@ public interface RealmInstant : Comparable<RealmInstant> {
                 return if (epochSeconds < 0) MIN else MAX
             }
             return RealmInstantImpl(s, ns)
+        }
+
+        public fun now(): RealmInstant {
+            val internalNow = internalNow()
+            return RealmInstantImpl(internalNow.epochSeconds, internalNow.nanosecondsOfSecond)
         }
     }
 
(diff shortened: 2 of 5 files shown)


packages/library-base/src/darwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -44,6 +46,33 @@ public actual fun threadId(): ULong {
...
+ * and https://github.com/Kotlin/kotlinx-datetime/blob/master/core/native/src/Instant.kt.
+ *
+ * Even though Darwin only uses millisecond precision, it is possible that [date] uses larger resolution, storing
+ * microseconds or even nanoseconds. In this case, the sub-millisecond parts of [date] are rounded to the nearest
+ * millisecond, given that they are likely to be conversion artifacts.
+ *
+ * Since internalNow() should only logically return a value after the Unix epoch, it is safe to create a RealmInstant
+ * without considering having to pass negative nanoseconds.
+ */
+@Suppress("MagicNumber")
```
JacobOscarGunnarsson, line 60: seems like the linter didn't appreciate `secs` even though it's not really constant... is there a way to bypass this?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -109,6 +110,11 @@ public expect fun threadId(): ULong
  */
 public expect fun epochInSeconds(): Long
 
+/**
+ * Returns a RealmInstant representing the time that has passed since the Unix epoch.
+ */
+public expect fun internalNow(): RealmInstant
```
cmelchior, line 116: We only make this `public` if there is a reason to. All access to this seems to happen inside the `library-base` module, so it should be safe to mark this `internal`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmInstant.kt
```diff
@@ -73,6 +74,11 @@ public interface RealmInstant : Comparable<RealmInstant> {
             }
             return RealmInstantImpl(s, ns)
         }
+
+        public fun now(): RealmInstant {
```
cmelchior, line 78: This method seems to be missing documentation. Generally we strive to document all of our public API methods=

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmInstant.kt
```diff
@@ -73,6 +74,11 @@ public interface RealmInstant : Comparable<RealmInstant> {
...
+            val internalNow = internalNow()
+            return RealmInstantImpl(internalNow.epochSeconds, internalNow.nanosecondsOfSecond)
```
cmelchior, line 80: Since `internalNow()` already returns a `RealmInstant` there is no need to create a new instance. We can just return it directly the user

packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -1,9 +1,12 @@
 package io.realm.kotlin.internal.platform
 
+import io.realm.kotlin.internal.RealmInstantImpl
+import io.realm.kotlin.types.RealmInstant
 import java.io.File
 import java.util.concurrent.TimeUnit
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KType
+import java.time.Clock as jtClock
```
cmelchior, line 9: Any reason you are renaming to `jtClock` here?

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInstantTests.kt
```diff
@@ -151,6 +151,17 @@ class RealmInstantTests {
...
 
+    /**
+     * When this method was implemented the unix epoch time was 1664980145.
+     * Nanoseconds are too granular to perform tests on especially since the
+     * darwin implementation rounds the timestamp to a precision of milliseconds.
+     */
+    @Test
+    fun now() {
+        val ts = RealmInstant.now()
+        assertTrue(ts.epochSeconds > 1664980145)
```
cmelchior, line 162: We should also test that `ts.nanoSeconds` are positive.

cmelchior (changes requested):
Wow, that was quick 🚀  

I have a few comments and we are also missing a CHANGELOG update.

packages/library-base/src/darwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
cmelchior: This is the way to bypass it. Alternatively you need to configure an exception in https://github.com/realm/realm-kotlin/blob/master/config/detekt/detekt.yml#L565

Issue #858, opened 2022-06-01
Remove compatibility hacks for the compiler plugin


nhachicha:
When upgrading to Kotlin 1.7 as a minimum supported version we need to revert some workaround done to maintain compatibility from `1.6.10` 

- https://github.com/realm/realm-kotlin/pull/856
- https://github.com/realm/realm-kotlin/pull/854
- https://github.com/realm/realm-kotlin/pull/845 (this could be fixed in K2)
- https://github.com/realm/realm-kotlin/pull/666
- https://github.com/realm/realm-kotlin/pull/1009 (Change between 1.7.10 and 1.7.20)

PR #1059, opened 2022-10-07
Update to Kotlin 1.7.20 as minimum version.

cmelchior:
Update the project to use Kotlin 1.7.20 as the minimum version

- Rewires all our freeze methods, so they can easily be enabled/disabled from one location. We could probably just delete them outright, but maybe it would be worth keeping them around for a wee bit longer while we are still transitioning fully to the new memory model across all branches.

- Makes the new memory model on Kotlin Native required. The old memory model is no longer supported.

- Fixes listed in https://github.com/realm/realm-kotlin/issues/858 will be cleaned up in a separate PR.

- From [Kotlin 1.7.0](https://kotlinlang.org/docs/whatsnew17.html#unified-compiler-plugin-abi-with-jvm-and-js-ir-backends
) having the shaded compiler artifact is no longer needed, but removing it completely is causing problems in our test dependency substitution. So I'm keeping it around for now and will create an issue for fixing that issue. I have rewired the compiler plugin so it always uses the non-shaded version though. This means that the shaded artifact is only being used in our internal setup.

The error being reported was the below.
```
* What went wrong:
Could not determine the dependencies of task ':plugin-compiler:compileJava'.
> Could not resolve all dependencies for configuration ':plugin-compiler:compileClasspath'.
   > Cannot change dependencies of dependency configuration ':plugin-compiler:api' after it has been included in dependency resolution.
```


README.md
```diff
@@ -277,16 +277,18 @@ See [Config.kt](buildSrc/src/main/kotlin/Config.kt#L2txt) for the latest version
 
 # Kotlin Memory Model and Coroutine compatibility
 
-Realm Kotlin is implemented against Kotlin's default memory model (the old one), but still supports running with the new memory model if enabled in the consuming project. See https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md#switch-to-the-new-mm for details on enabled the new memory model.
+Realm Kotlin 1.3.0 and above *only* works with the new Kotlin Native memory model. This is also the default memory model from Kotlin 1.7.20 and onwards. This mean that you need the default Kotlin Coroutine library 1.6.0 and above and not the `-mt` variant, which have also been [deprecated](https://blog.jetbrains.com/kotlin/2021/12/introducing-kotlinx-coroutines-1-6-0/).
```
nhachicha, line 280: Should we hint to the other branch/version one should use if we want to continue working with the old memory model (i.e the Jetpack Compose stable)

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -64,13 +64,13 @@ object Versions {
...
+    const val atomicfu = "0.18.3" // https://github.com/Kotlin/kotlinx.atomicfu
     const val autoService = "1.0" // https://mvnrepository.com/artifact/com.google.auto.service/auto-service
     // Not currently used, so mostly here for documentation. Core requires minimum 3.15, but 3.18.1 is available through the Android SDK.
     // Build also tested successfully with 3.21.4 (latest release).
     const val cmake = "3.22.1"
-    const val coroutines = "1.6.0-native-mt" // https://mvnrepository.com/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-core
-    const val datetime = "0.3.2" // https://github.com/Kotlin/kotlinx-datetime
+    const val coroutines = "1.6.4" // https://mvnrepository.com/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-core
+    const val datetime = "0.4.0" // https://github.com/Kotlin/kotlinx-datetime
     const val detektPlugin = "1.19.0-RC1" // https://github.com/detekt/detekt
```
nhachicha, line 74: 1.21 is released

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -79,14 +79,14 @@ object Versions {
     const val junit = "4.13.2" // https://mvnrepository.com/artifact/junit/junit
     const val jvmTarget = "1.8"
     // When updating the Kotlin version, also remember to update /examples/min-android-sample/build.gradle.kts
-    const val kotlin = "1.6.10" // https://github.com/JetBrains/kotlin and https://kotlinlang.org/docs/releases.html#release-details
-    const val latestKotlin = "1.7.20-RC" // https://kotlinlang.org/docs/eap.html#build-details
+    const val kotlin = "1.7.20" // https://github.com/JetBrains/kotlin and https://kotlinlang.org/docs/releases.html#release-details
+    const val latestKotlin = "1.7.20" // https://kotlinlang.org/docs/eap.html#build-details
     const val kotlinCompileTesting = "1.4.9" // https://github.com/tschuchortdev/kotlin-compile-testing
     const val ktlint = "0.45.2" // https://github.com/pinterest/ktlint
```
nhachicha, line 85: 0.47.1 is released

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -79,14 +79,14 @@ object Versions {
...
     const val ktor = "1.6.8" // https://github.com/ktorio/ktor
     const val nexusPublishPlugin = "1.1.0" // https://github.com/gradle-nexus/publish-plugin
     const val okio = "3.0.0" // https://square.github.io/okio/#releases
```
nhachicha, line 88: 3.2.0 is out

packages/cinterop/src/darwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -114,9 +114,13 @@ import realm_wrapper.realm_user_t
 import realm_wrapper.realm_value_t
 import realm_wrapper.realm_value_type
 import realm_wrapper.realm_version_id_t
-import kotlin.native.concurrent.freeze
 import kotlin.native.internal.createCleaner
 
+private fun <T> T.freeze(): T {
```
nhachicha, line 119: inline ?

packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmCompilerSubplugin.kt
```diff
@@ -50,8 +49,7 @@ class RealmCompilerSubplugin : KotlinCompilerPluginSupportPlugin {
     }
 
     override fun getPluginArtifactForNative(): SubpluginArtifact {
```
nhachicha, line 51: Add todo to remove this in a separate PR

packages/library-sync/build.gradle.kts
```diff
@@ -51,9 +51,7 @@ kotlin {
                 // Cinterop does not hold anything required by users
                 implementation(project(":cinterop"))
 
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.coroutines}") {
-                    version { strictly(Versions.coroutines) }
```
nhachicha, line 55: We can use the [range notation](https://docs.gradle.org/current/userguide/single_versions.html#simple_version_declaration_semantics) to declare the minimum version we support which uses the new memory model ...

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/dynamic/DynamicRealmObjectTests.kt
```diff
@@ -344,6 +347,115 @@ class DynamicRealmObjectTests {
                         }
                     }
                 }
+                is SetPropertyType -> {
+                    if (type.isNullable) {
+                        when (type.storageType) {
+                            RealmStorageType.BOOL -> {
+                                assertEquals(null, dynamicSample.getNullableValueSet<Boolean>(property.name).first())
```
nhachicha, line 354: use `kotlin.test.assertNull` instead?

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/dynamic/DynamicRealmObjectTests.kt
```diff
@@ -344,6 +347,115 @@ class DynamicRealmObjectTests {
...
+                            RealmStorageType.BOOL -> {
+                                val expectedValue = defaultSample.booleanField
+                                assertEquals(expectedValue, dynamicSample.getValueSet<Boolean>(property.name).first())
+                                assertEquals(expectedValue, dynamicSample.getValueSet(property.name, Boolean::class).first())
+                            }
+                            RealmStorageType.INT -> {
+                                val expectedValue: Long? = when (property.name) {
+                                    "byteSetField" -> defaultSample.byteField.toLong()
+                                    "charSetField" -> defaultSample.charField.code.toLong()
+                                    "charSetField" -> defaultSample.charField.code.toLong()
```
nhachicha, line 406: duplicate

nhachicha (approved):
LGTM 👍 minor things

README.md
cmelchior: I think we agreed there was no need to mention other releases until we actually needed them. Since Compose right now supports 1.7.20, they don't actually need to care about the difference.

buildSrc/src/main/kotlin/Config.kt
cmelchior: See Yup, I tried to bump this and ktlint in this PR but especially Ktlint had a large amount of changes. So I opted for doing that in separate PR

packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmCompilerSubplugin.kt
cmelchior: We cannot remove this method as it is part of the interface. It is marked as deprecated though, so I would expect this to be gone in a future Kotlin release.

PR #1061, opened 2022-10-07
first draft of subqueries for list

JacobOscarGunnarsson:


packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,241 @@
...
+                assertIs<ResultsChange<*>>(resultsChange)
+                assertNotNull(resultsChange.list)
+                assertEquals(5, resultsChange.list.size)
+            }
+            observer.cancel()
+            channel.close()
+        }
+    }
+
+    @Test
```
JacobOscarGunnarsson, line 214: Not entirely sure if this corresponds to deleting a list

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,241 @@
...
+    fun realmList_simpleQueryEmbeddedObject() {
+        val parent = realm.query<EmbeddedParent>("id == $0", "1").find().first()
+        val result = parent.children.query("TRUEPREDICATE").find()
+        assertEquals(3, result.size)
+    }
+
+    @Test
+    fun realmList_queryMax() {
```
JacobOscarGunnarsson, line 82: could also squash these aggregate tests into the same test, but that depends on what we prefer

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,250 @@
...
+        assertEquals(0, result)
+    }
+
+    @Test
+    fun realmList_querySumWithNullValues() {
+        val querySample = realm.query<QuerySample>("intField == $0", 1).find().first()
+        val result = querySample.objectListField.query("TRUEPREDICATE").sum<Int>("nullableIntField").find()
+        assertEquals(6, result)
+    }
+
```
cmelchior, line 157: We also miss tests for the `IllegalArgumentException` being thrown if the query string is invalid or the argument input is wrong

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,250 @@
...
+    fun tearDown() {
+        if (!realm.isClosed()) {
+            realm.close()
+        }
+        PlatformUtils.deleteTempDir(tmpDir)
+    }
+
+    @Test
+    fun realmList_simpleQuery() {
+        val querySample = realm.query<QuerySample>("intField == $0", 1).find().first()
```
cmelchior, line 69: You don't need to query for this object in all the tests. You could just get the reference in the `setup` instead

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,250 @@
...
+            val channel = Channel<ResultsChange<*>>(capacity = 1)
+            val observer = async {
+                querySample.objectListField
+                    .query("TRUEPREDICATE").asFlow()
+                    .collect { flowList ->
+                        channel.send(flowList)
+                    }
+            }
+            channel.receive().let { resultsChange ->
+                assertIs<ResultsChange<*>>(resultsChange)
```
cmelchior, line 202: This is always true since `ResultsChange` is the top-level class. Instead this should be checking against `InitialResults`

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,250 @@
...
+                assertNotNull(resultsChange.list)
+                assertEquals(6, resultsChange.list.size)
+            }
+            realm.writeBlocking {
+                val queriedSample = findLatest(querySample)
+                queriedSample!!.objectListField.removeFirst()
+            }
+            channel.receive().let { resultsChange ->
+                assertIs<ResultsChange<*>>(resultsChange)
```
cmelchior, line 211: Same, this should check against UpdatedResults

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    @Test
+    fun realmList_flowDeleteElement() {
+        runBlocking {
+            val querySample = realm.query<QuerySample>("intField == $0", 1).find().first()
+            val channel = Channel<ResultsChange<*>>(capacity = 1)
+            val observer = async {
+                querySample.objectListField
+                    .query("TRUEPREDICATE").asFlow()
```
cmelchior, line 196: I would probably make some kind of predicate here. Just because this might skip creating a query altogether in the underlying implementation. It could be something simple as `intField >= 0`

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmListQueryTests.kt
```diff
@@ -0,0 +1,250 @@
...
+            channel.receive().let { resultsChange ->
+                assertIs<ResultsChange<*>>(resultsChange)
+                assertNotNull(resultsChange.list)
+                assertEquals(6, resultsChange.list.size)
+            }
+            realm.writeBlocking {
+                val queriedSample = findLatest(querySample)
+                queriedSample!!.objectListField = realmListOf()
+            }
+            channel.receive().let { resultsChange ->
```
cmelchior, line 241: Does this test actually parse. I would have expected the flow to complete without sending an event. I.e. you need something like:

```
                    .query("TRUEPREDICATE").asFlow()
                    .onCompletion { channel.send(/*something*/) }
                    .collect { flowList ->
                        channel.send(flowList)
                    }
```
```

cmelchior (changes requested):
Nice 👍. Implementation looks great, I only have a few suggestions for improving the tests. Also missing a CHANGELOG update.

rorbech:
Superseded by #1132

Issue #1051, opened 2022-10-05
Flaky test: fun mutableRealmInt_convergesAcrossClients()


cmelchior:
Seems to lack synchronization between incrementing the two writes: https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-1048/7/tests

PR #1065, opened 2022-10-10
Remove race conditions in tests.

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/1051

Before this PR, it was possible for the writes on Realm 2 and 3 to happen at the same "time". Which would result in Realm one seeing two writes in the same notification. Which would break the test.

I rewrote the logic to test for the value after each step.


JacobOscarGunnarsson (approved):
👍

PR #1066, opened 2022-10-10
Adding device farm test

nhachicha:


.github/actions/run-android-device-farm-test/action.yml
```diff
@@ -0,0 +1,58 @@
...
+          version: 0.1
+          phases:
+            install:
+              commands:
+                - export PYTHON_VERSION=3
+            test:
+              commands:
+                - adb shell am instrument -w -r io.realm.testapp.test/androidx.test.runner.AndroidJUnitRunner | egrep 'OK \([0-9]+ test[s]?\)'
+
+    - run: |
```
cmelchior, line 43: How are failing tests being reported?

.github/actions/run-android-device-farm-test/action.yml
```diff
@@ -0,0 +1,58 @@
...
+    value: ${{ steps.get-test-results.outputs.results-path }}
+runs:
+  using: "composite"
+  steps:
+    - name: Run the tests
+      uses: realm/aws-devicefarm/test-application@d94e739490340474f8464f74289a0f48706dd3a3
+      id: run-tests
+      with:
+        project_arn: ${{ inputs.project-arn }}
+        device_pool_arn: ${{ inputs.device-pool-arn }}
```
cmelchior, line 24: How do we configure the list of devices? E.g. if we want to run on more than one device?

cmelchior (commented):
Awesome getting this in. I'm not sure I fully understand what the action does though, so have asked some clarifying questions. We should probably attempt to document some of this somewhere.

.github/actions/run-android-device-farm-test/action.yml
nhachicha: The instrument run doesn't build the JUnit XML results to pull from the device. 
I modified the behaviour to always pull the Logcat from the device and publish it in case we want to investigate a failure/crash

.github/workflows/pr.yml
```diff
@@ -1058,12 +999,6 @@ jobs:
           name: packages-android-${{ needs.check-cache.outputs.version-label }}
           path: ./packages/build/m2-buildrepo 
 
-      - name: Restore Android artifacts
```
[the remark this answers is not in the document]
nhachicha: Yes it 's a copy of https://github.com/realm/realm-kotlin/pull/1066/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160L1052 :)

.github/actions/run-android-device-farm-test/action.yml
nhachicha: The list of devices is configured in AWS device farm, you can create different device pool that you reference by their `arn`. The issue with running on multiple devices is pulling resources out (logs etc.) and aggregating the results. The current .NET is also running on 1 device from the pool of selected device for simplicity

cmelchior (approved):
Looks good. I think we need to document somewhere how to log into Device Farm and control the DEVICEFARM_POOL_ARN.

It is feels a bit unfortunate if we are sharing a device pool with .NET. We probably want our own, so we can modify what it points to without fearing it impacting other teams.

nhachicha:
@cmelchior we're not sharing a device pool with .NET. Each repo has its own project/device pool(s)

PR #1069, opened 2022-10-13
Add support for ApiKeyAuthProvider

JacobOscarGunnarsson:
Early version of #432, some optimisations still needs to be done, and discussions about error handling needs to be made as well.

This depends on us being able to merge changes from `jm/named-apikey-callbacks` in core.

need to update `CredentialsTests` and `CHANGELOG.md`

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -18,6 +18,7 @@
 
 package io.realm.kotlin.internal.interop
 
+import io.realm.kotlin.internal.interop.sync.ApiKeyWrapper
 import io.realm.kotlin.internal.interop.sync.AuthProvider
 import io.realm.kotlin.internal.interop.sync.CoreSubscriptionSetState
 import io.realm.kotlin.internal.interop.sync.CoreSyncSessionState
@@ -312,6 +313,12 @@ expect object RealmInterop {
         syncConfig: RealmSyncConfigurationPointer,
         overriddenName: String?
     ): String
+    fun realm_app_user_apikey_provider_client_create_apikey(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        name: String,
+        callback: AppCallback<ApiKeyWrapper>
+    )
 
     // User
     fun realm_user_get_all_identities(user: RealmUserPointer): List<SyncUserIdentity>
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -348,6 +348,12 @@ expect object RealmInterop {
         callback: AppCallback<ApiKeyWrapper>,
     )
 
+    fun realm_app_user_apikey_provider_client_fetch_apikeys(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        callback: AppCallback<Array<ApiKeyWrapper>>,
+    )
+
     // User
     fun realm_user_get_all_identities(user: RealmUserPointer): List<SyncUserIdentity>
     fun realm_user_get_identity(user: RealmUserPointer): String
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -334,6 +334,13 @@ expect object RealmInterop {
         callback: AppCallback<Unit>,
     )
 
+    fun realm_app_user_apikey_provider_client_enable_apikey(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        id: ObjectIdWrapper,
+        callback: AppCallback<Unit>,
+    )
+
     // User
     fun realm_user_get_all_identities(user: RealmUserPointer): List<SyncUserIdentity>
     fun realm_user_get_identity(user: RealmUserPointer): String
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -327,6 +327,13 @@ expect object RealmInterop {
         callback: AppCallback<Unit>,
     )
 
+    fun realm_app_user_apikey_provider_client_disable_apikey(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        id: ObjectIdWrapper,
+        callback: AppCallback<Unit>,
+    )
+
     // User
     fun realm_user_get_all_identities(user: RealmUserPointer): List<SyncUserIdentity>
     fun realm_user_get_identity(user: RealmUserPointer): String
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -320,6 +320,13 @@ expect object RealmInterop {
         callback: AppCallback<ApiKeyWrapper>
     )
 
+    fun realm_app_user_apikey_provider_client_delete_apikey(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        id: ObjectIdWrapper,
+        callback: AppCallback<Unit>,
+    )
+
     // User
     fun realm_user_get_all_identities(user: RealmUserPointer): List<SyncUserIdentity>
     fun realm_user_get_identity(user: RealmUserPointer): String
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -341,6 +341,13 @@ expect object RealmInterop {
         callback: AppCallback<Unit>,
     )
 
+    fun realm_app_user_apikey_provider_client_fetch_apikey(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        id: ObjectIdWrapper,
+        callback: AppCallback<ApiKeyWrapper>,
+    )
+
     // User
     fun realm_user_get_all_identities(user: RealmUserPointer): List<SyncUserIdentity>
     fun realm_user_get_identity(user: RealmUserPointer): String
--- a/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
+++ b/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
@@ -338,10 +338,42 @@ void app_complete_result_callback(void* userdata, void* result, const realm_app_
     }
 }
 
-void app_api_key_callback(realm_userdata_t userdata, realm_app_user_apikey_t* api_key, const realm_app_error_t* api_error) {
-    api_key->
+void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* apikey, const realm_app_error_t* error) {
+    auto env = get_env(true);
+    static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
+    static JavaMethod api_key_wrapper_constructor(env, api_key_wrapper_class, "<init>", "([BLjava/lang/String;Ljava/lang/String;Z)V");
+
+    static JavaClass java_callback_class(env, "io/realm/kotlin/internal/interop/AppCallback");
+    static JavaMethod java_notify_onerror(env, java_callback_class, "onError",
+                                          "(Lio/realm/kotlin/internal/interop/sync/AppError;)V");
+    static JavaMethod java_notify_onsuccess(env, java_callback_class, "onSuccess",
+                                            "(Ljava/lang/Object;)V");
+    if (error) {
+        jobject app_exception = convert_to_jvm_app_error(env, error);
+        env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onerror, app_exception);
+        jni_check_exception(env);
+    } else {
+        jbyteArray id = env->NewByteArray(sizeof(apikey->id.bytes));
+        jstring key = to_jstring(env, apikey->key);
+        jstring name = to_jstring(env, apikey->name);
+        jboolean disabled = apikey->disabled;
+        jobject api_key_wrapper_obj = env->NewObject(api_key_wrapper_class,
+                                                     api_key_wrapper_constructor,
+                                                     id,
+                                                     key,
+                                                     name,
+                                                     disabled,
+                                                     false);
+
+        env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onsuccess, api_key_wrapper_obj);
+        jni_check_exception(env);
+        env->DeleteLocalRef(id);
+    }
 }
 
+void app_apikey_list_callback(realm_userdata_t userdata, realm_app_user_apikey_t[], size_t count, realm_app_error_t*) {
+    // Return list of api keys
+}
 
 bool realm_should_compact_callback(void* userdata, uint64_t total_bytes, uint64_t used_bytes) {
     auto env = get_env(true);
--- a/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
+++ b/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
@@ -338,17 +338,28 @@ void app_complete_result_callback(void* userdata, void* result, const realm_app_
     }
 }
 
-jobject create_api_key_wrapper(const realm_app_user_apikey_t* key_data) {
-
+jobject create_api_key_wrapper(JNIEnv* env, const realm_app_user_apikey_t* key_data) {
+        auto id_size = sizeof(key_data->id.bytes);
+        static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
+        static JavaMethod api_key_wrapper_constructor(env, api_key_wrapper_class, "<init>", "([BLjava/lang/String;Ljava/lang/String;Z)V");
+        jbyteArray id = env->NewByteArray(id_size);
+        env->SetByteArrayRegion(id, 0, id_size, reinterpret_cast<const jbyte*>(key_data->id.bytes));
+        jstring key = to_jstring(env, key_data->key);
+        jstring name = to_jstring(env, key_data->name);
+        jboolean disabled = key_data->disabled;
+        return env->NewObject(api_key_wrapper_class,
+                                                     api_key_wrapper_constructor,
+                                                     id,
+                                                     key,
+                                                     name,
+                                                     disabled,
+                                                     false);
 }
 
 
 
 void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* apikey, const realm_app_error_t* error) {
     auto env = get_env(true);
-    static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
-    static JavaMethod api_key_wrapper_constructor(env, api_key_wrapper_class, "<init>", "([BLjava/lang/String;Ljava/lang/String;Z)V");
-
     static JavaClass java_callback_class(env, "io/realm/kotlin/internal/interop/AppCallback");
     static JavaMethod java_notify_onerror(env, java_callback_class, "onError",
                                           "(Lio/realm/kotlin/internal/interop/sync/AppError;)V");
@@ -359,20 +370,7 @@ void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* api
         env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onerror, app_exception);
         jni_check_exception(env);
     } else {
-        jobject api_key_wrapper_obj = create_api_key_wrapper(apikey);
-//        auto id_size = sizeof(apikey->id.bytes);
-//        jbyteArray id = env->NewByteArray(id_size);
-//        env->SetByteArrayRegion(id, 0, id_size, reinterpret_cast<const jbyte*>(apikey->id.bytes));
-//        jstring key = to_jstring(env, apikey->key);
-//        jstring name = to_jstring(env, apikey->name);
-//        jboolean disabled = apikey->disabled;
-//        jobject api_key_wrapper_obj = env->NewObject(api_key_wrapper_class,
-//                                                     api_key_wrapper_constructor,
-//                                                     id,
-//                                                     key,
-//                                                     name,
-//                                                     disabled,
-//                                                     false);
+        jobject api_key_wrapper_obj = create_api_key_wrapper(env, apikey);
 
         env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onsuccess, api_key_wrapper_obj);
         jni_check_exception(env);
@@ -383,7 +381,6 @@ void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* api
 void app_apikey_list_callback(realm_userdata_t userdata, realm_app_user_apikey_t* keys, size_t count, realm_app_error_t* error) {
     auto env = get_env(true);
     static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
-    static JavaMethod api_key_wrapper_constructor(env, api_key_wrapper_class, "<init>", "([BLjava/lang/String;Ljava/lang/String;Z)V");
 
     static JavaClass java_callback_class(env, "io/realm/kotlin/internal/interop/AppCallback");
     static JavaMethod java_notify_onerror(env, java_callback_class, "onError",
@@ -401,22 +398,7 @@ void app_apikey_list_callback(realm_userdata_t userdata, realm_app_user_apikey_t
         // For each ApiKey, create the Kotlin Wrapper and insert into array
         for (int i = 0; i < count; i++) {
             realm_app_user_apikey_t api_key = keys[i];
-            jobject api_key_wrapper_obj = create_api_key_wrapper(&api_key);
-//
-//
-//            auto id_size = sizeof(api_key.id.bytes);
-//            jbyteArray id = env->NewByteArray(id_size);
-//            env->SetByteArrayRegion(id, 0, id_size, reinterpret_cast<const jbyte*>(api_key.id.bytes));
-//            jstring key = to_jstring(env, api_key.key);
-//            jstring name = to_jstring(env, api_key.name);
-//            jboolean disabled = api_key.disabled;
-//            jobject api_key_wrapper_obj = env->NewObject(api_key_wrapper_class,
-//                                                         api_key_wrapper_constructor,
-//                                                         id,
-//                                                         key,
-//                                                         name,
-//                                                         disabled,
-//                                                         false);
+            jobject api_key_wrapper_obj = create_api_key_wrapper(env, &api_key);
             env->SetObjectArrayElement(key_array, i, api_key_wrapper_obj);
         }
 
--- a/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
+++ b/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
@@ -339,21 +339,21 @@ void app_complete_result_callback(void* userdata, void* result, const realm_app_
 }
 
 jobject create_api_key_wrapper(JNIEnv* env, const realm_app_user_apikey_t* key_data) {
-        auto id_size = sizeof(key_data->id.bytes);
         static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
         static JavaMethod api_key_wrapper_constructor(env, api_key_wrapper_class, "<init>", "([BLjava/lang/String;Ljava/lang/String;Z)V");
+        auto id_size = sizeof(key_data->id.bytes);
         jbyteArray id = env->NewByteArray(id_size);
         env->SetByteArrayRegion(id, 0, id_size, reinterpret_cast<const jbyte*>(key_data->id.bytes));
         jstring key = to_jstring(env, key_data->key);
         jstring name = to_jstring(env, key_data->name);
         jboolean disabled = key_data->disabled;
         return env->NewObject(api_key_wrapper_class,
-                                                     api_key_wrapper_constructor,
-                                                     id,
-                                                     key,
-                                                     name,
-                                                     disabled,
-                                                     false);
+                              api_key_wrapper_constructor,
+                              id,
+                              key,
+                              name,
+                              disabled,
+                              false);
 }
 
 
@@ -371,7 +371,6 @@ void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* api
         jni_check_exception(env);
     } else {
         jobject api_key_wrapper_obj = create_api_key_wrapper(env, apikey);
-
         env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onsuccess, api_key_wrapper_obj);
         jni_check_exception(env);
     }
--- a/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
+++ b/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
@@ -369,12 +369,50 @@ void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* api
 
         env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onsuccess, api_key_wrapper_obj);
         jni_check_exception(env);
-        env->DeleteLocalRef(id);
     }
 }
 
-void app_apikey_list_callback(realm_userdata_t userdata, realm_app_user_apikey_t[], size_t count, realm_app_error_t*) {
-    // Return list of api keys
+void app_apikey_list_callback(realm_userdata_t userdata, realm_app_user_apikey_t* keys, size_t count, realm_app_error_t* error) {
+    auto env = get_env(true);
+    static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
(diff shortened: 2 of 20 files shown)


packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    fun create_throwsWithNoName(): Unit = runBlocking {
+        // worth creating a more specific exception?
+        assertFailsWith<ServiceException> {
+            provider.create("")
+        }
+    }
+
```
JacobOscarGunnarsson, line 62: In Realm Java there also exists tests like

```
 @Test
    fun createApiKey_invalidArgumentThrows() {
        testNullArg { provider.create(TestHelper.getNull()) }
        testNullArg { provider.create("") }
        looperThread.runBlocking {
            provider.createAsync(TestHelper.getNull(), checkNullInApiKeyCallback)
        }
        looperThread.runBlocking {
            provider.createAsync("", checkNullInApiKeyCallback)
        }
    }
```

But I'm not entirely sure how this is replicated in kotlin, passing null does not work here due to type safety

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -18,6 +18,7 @@
```
JacobOscarGunnarsson, line 18: need to update `CredentialsTests as well`

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,122 @@
...
+                        !apiKeyData.disabled
+                    )
+                }.freeze()
+            )
+            return channel.receive()
+                .getOrThrow()
+        }
+    }
+
+    override suspend fun delete(id: ObjectId) {
```
nirinchev, line 36: Not sure if that's something you want to do, but in .NET and Dart we ignore the 404 error case for this method as deleting an API key that doesn't exist should be a no-op.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -337,6 +338,76 @@ void app_complete_result_callback(void* userdata, void* result, const realm_app_
     }
 }
 
+jobject create_api_key_wrapper(JNIEnv* env, const realm_app_user_apikey_t* key_data) {
+        auto id_size = sizeof(key_data->id.bytes);
+        static JavaClass api_key_wrapper_class(env, "io/realm/kotlin/internal/interop/sync/ApiKeyWrapper");
```
cmelchior, line 343: Nitpick, I would move these static variables above `id_size`

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -337,6 +338,76 @@ void app_complete_result_callback(void* userdata, void* result, const realm_app_
...
+        static JavaMethod api_key_wrapper_constructor(env, api_key_wrapper_class, "<init>", "([BLjava/lang/String;Ljava/lang/String;Z)V");
+        jbyteArray id = env->NewByteArray(id_size);
+        env->SetByteArrayRegion(id, 0, id_size, reinterpret_cast<const jbyte*>(key_data->id.bytes));
+        jstring key = to_jstring(env, key_data->key);
+        jstring name = to_jstring(env, key_data->name);
+        jboolean disabled = key_data->disabled;
+        return env->NewObject(api_key_wrapper_class,
+                                                     api_key_wrapper_constructor,
```
cmelchior, line 351: Something about indentation is not right here. They should be at the same level as `api_key_wrapper_class`

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -337,6 +338,76 @@ void app_complete_result_callback(void* userdata, void* result, const realm_app_
...
+                                          "(Lio/realm/kotlin/internal/interop/sync/AppError;)V");
+    static JavaMethod java_notify_onsuccess(env, java_callback_class, "onSuccess",
+                                            "(Ljava/lang/Object;)V");
+    if (error) {
+        jobject app_exception = convert_to_jvm_app_error(env, error);
+        env->CallVoidMethod(static_cast<jobject>(userdata), java_notify_onerror, app_exception);
+        jni_check_exception(env);
+    } else {
+        jobject api_key_wrapper_obj = create_api_key_wrapper(env, apikey);
+
```
cmelchior, line 374:
```suggestion
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -36,6 +37,11 @@ public interface User {
      */
     public val app: App
 
+    /**
+     * TODO
```
cmelchior, line 41: Missing documentation

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKey.kt
```diff
@@ -0,0 +1,5 @@
+package io.realm.kotlin.mongodb.auth
+
+import io.realm.kotlin.types.ObjectId
+
+public data class ApiKey public constructor(public val id: ObjectId, public val value: String?, public val name: String, public val enabled: Boolean)
```
cmelchior, line 5: Missing documentation of class and each parameter (since they are also class properties)

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,12 @@
+package io.realm.kotlin.mongodb.auth
+
+import io.realm.kotlin.types.ObjectId
+
+public interface ApiKeyAuth {
```
cmelchior, line 5: Class and all methods are missing documentation, including which errors each can throw.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
cmelchior: Agree that throwing is probably the worst option. We talked about either just ignoring it or returning a boolean indicating whether the deletion was a success. I'm leaning toward returning a boolean as that also mirrors Kotlin/Java collection API's.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,124 @@
...
+internal class ApiKeyAuthImpl(val app: AppImpl, val user: UserImpl) : ApiKeyAuth {
+
+    override suspend fun create(name: String): ApiKey {
+        Channel<Result<ApiKey>>(1).use { channel ->
+            RealmInterop.realm_app_user_apikey_provider_client_create_apikey(
+                app.nativePointer,
+                user.nativePointer,
+                name,
+                channelResultCallback<ApiKeyWrapper, ApiKey>(channel) { apiKeyData ->
+                    ApiKey(
```
cmelchior, line 23: I would consider moving this mapping into a helper method since it is shared between 3 methods and it is easy to get the boolean logic wrong.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        assertEquals("foo", key.name)
+        assertNotNull(key.value)
+        assertNotNull(key.id)
+        assertTrue(key.enabled)
+    }
+
+    @Test
+    fun create_throwsWithInvalidName(): Unit = runBlocking {
+        // worth creating a more specific exception?
+        assertFailsWith<ServiceException> {
```
cmelchior, line 50: In this case I would use `assertFailsWithMessage` as that makes it easier to reason about the test doing the right thing, see e.g https://github.com/realm/realm-kotlin/blob/12ad0117b3fd554d5edd714b921ab8d579785451/packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/dynamic/DynamicRealmObjectTests.kt#L477

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        val key = provider.create("foo")
+        assertEquals("foo", key.name)
+        assertNotNull(key.value)
+        assertNotNull(key.id)
+        assertTrue(key.enabled)
+    }
+
+    @Test
+    fun create_throwsWithInvalidName(): Unit = runBlocking {
+        // worth creating a more specific exception?
```
cmelchior, line 49: I would lean towards *not*, at least for now, but we know the rules, we should document them in the method documentation.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        assertFailsWith<ServiceException> {
+            provider.create("%s")
+        }
+    }
+
+    @Test
+    fun create_throwsWithNoName(): Unit = runBlocking {
+        // worth creating a more specific exception?
```
cmelchior, line 57: The behaviour of this should follow the behaviour above, so I would say no.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        assertFailsWith<ServiceException> {
```
cmelchior, line 58: Same as above. I would use `assertFailsWithMessage`

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
cmelchior: Yes, these kinds of tests are not relevant in Kotlin since the type system prevents these errors.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        assertEquals(key1.name, key2.name)
+        assertNull(key2.value)
+        assertEquals(key1.enabled, key2.enabled)
+    }
+
+    @Test
+    fun fetch_nonExistingKeyThrows(): Unit = runBlocking {
+        // worth creating a more specific exception?
+        assertFailsWith<ServiceException> {
+            provider.fetch(ObjectId.create())
```
cmelchior, line 77: Not finding a key seems pretty normal use case, so I would return `null` in this case instead of throwing an exception and making the return type nullable, i.e. `fetch(id: ObjectId): ApiKey?`

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    @Test
+    fun fetchAll() = runBlocking {
+        val key1 = provider.create("foo")
```
cmelchior, line 83: I would also add a test for the empty case, ie. calling `provider.fetchAll()` and check size 0 before starting to make keys.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        provider.disable(key.id)
+        assertFalse(provider.fetch(key.id).enabled)
+    }
+
+    @Test
+    fun disable_nonExistingKeyThrows(): Unit = runBlocking {
+        assertFailsWith<ServiceException> {
+            provider.disable(ObjectId.create())
+        }
+    }
```
cmelchior, line 159: Missing tests for what happens when use is logged out. This is e.g. the tests from Java:

```
   @Test
    fun callMethodWithLoggedOutUser() {
        user.logOut()
        for (method in Method.values()) {
            try {
                when(method) {
                    Method.CREATE -> provider.create("name")
                    Method.FETCH_SINGLE -> provider.fetch(ObjectId())
                    Method.FETCH_ALL -> provider.fetchAll()
                    Method.DELETE -> provider.delete(ObjectId())
                    Method.ENABLE -> provider.enable(ObjectId())
                    Method.DISABLE -> provider.disable(ObjectId())
                }
                fail("$method should have thrown an exception")
            } catch (error: AppException) {
                assertEquals(ErrorCode.SERVICE_UNKNOWN, error.errorCode)
            }
        }
    }
```

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        val key2 = provider.create("bar")
+        val keys = provider.fetchAll()
+        assertEquals(2, keys.size)
+        assertTrue(keys.any { it.id == key1.id })
+        assertTrue(keys.any { it.id == key2.id })
+    }
+
+    @Test
+    fun delete(): Unit = runBlocking {
+        val key1 = provider.create("foo")
```
cmelchior, line 93: With comments from above, I think a cleaner use of the API could look like this:

```
runBlocking {
  val key1 = provider.create("foo")
  assertNotNull(provider.fetch(key1.id))
  assertTrue(provider.delete(key1.id))
  assertNull(provider.fetch(key1.id))
}
```

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        provider.disable(key.id)
+        assertFalse(provider.fetch(key.id).enabled)
+        provider.enable(key.id)
+        assertTrue(provider.fetch(key.id).enabled)
+        provider.enable(key.id)
+        assertTrue(provider.fetch(key.id).enabled)
+    }
+    @Test
+    fun enable_nonExistingKeyThrows(): Unit = runBlocking {
+        assertFailsWith<ServiceException> {
```
cmelchior, line 133: `assertFailsWithMessage` to make the error scenario more clear

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,160 @@
...
+        val key = provider.create("foo")
+        provider.disable(key.id)
+        assertFalse(provider.fetch(key.id).enabled)
+        provider.disable(key.id)
+        assertFalse(provider.fetch(key.id).enabled)
+    }
+
+    @Test
+    fun disable_nonExistingKeyThrows(): Unit = runBlocking {
+        assertFailsWith<ServiceException> {
```
cmelchior, line 156: `assertFailsWithErrorMesssage`

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
cmelchior: Also missing tests for accessing `user` and `app` properties.

cmelchior (changes requested):
Great start...the implementation looks pretty solid. Mostly documentation is missing and I have some suggestions for improving the API. Also missing a CHANGELOG update.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
JacobOscarGunnarsson: I ended up just making it into a no-op, since I believe it's difficult mapping the boolean value to the different scenarios

removes key successfully -> true
deletes a key that doesn't exist -> true/false?? 
fails performing the delete operation -> false

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKey.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Class representing an API key for a User. An API can be used to represent the
+ * user when logging instead of using email and password.
+ * Note that a keys value is only available when the key is created, after that it is not
+ * visible. So anyone creating an API key is responsible for storing it safely after that.
+ * @param ObjectId, an id uniquely identifying the key.
+ * @param value, the value of this key. Only returned when the key is created.
+ * @param name, the name of the key.
+ * @param enabled, whether the key is enabled or not.
+ */
```
edualonso, line 13:
```suggestion
/**
 * Class representing an API key for a [User]. An API key can be used to represent a
 * user when logging in instead of using email and password.
 * Note that the value of a key will only be available immediately after the key is created, after which point it is not
 * visible anymore. This means that keys returned by [ApiKeyAuth.fetch] and [ApiKeyAuth.fetchAll] will have a `null` [value]. Anyone creating an API key is responsible for storing it safely after that.
 * @param id an [ObjectId] uniquely identifying the key.
 * @param value the value of this key, only returned when the key is created, `null` otherwise.
 * @param name the name of the key.
 * @param enabled whether the key is enabled.
 */
```
Reworded some things plus added kdoc links - observe you also need to add `import`s for those linked classes used in the docs!

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKey.kt
```diff
@@ -0,0 +1,14 @@
...
+public data class ApiKey public constructor(public val id: ObjectId, public val value: String?, public val name: String, public val enabled: Boolean)
```
edualonso, line 14: Also, for the sake of better visibility I would put each parameter in its own separate line.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
+package io.realm.kotlin.mongodb.auth
+
+import io.realm.kotlin.mongodb.App
+import io.realm.kotlin.mongodb.User
+import io.realm.kotlin.types.ObjectId
+
+/**
+ * Exposes functionality for a user to manage API keys under their control.
```
edualonso, line 8:
```suggestion
 * Exposes functionality for users to manage API keys under their control.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+ */
+public interface ApiKeyAuth {
+
+    /**
+     * The User that this instance in associated with.
```
edualonso, line 13:
```suggestion
     * [User] to which this instance is associated.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public val user: User
+
+    /**
+     * The App that this instance in associated with.
```
edualonso, line 18:
```suggestion
     * [App] to which this instance is associated.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Creates a user API key that can be used to authenticate as the user.
+     * The value of the key must be persisted at this time as this is the only time it is visible.
+     * The key is enabled when created. It can be disabled by calling the disable method.
+     *
+     * @param name the name of the key
+     * @throws IllegalArgumentException if an invalid name for the key is sent to the server.
+     * @return the new API key for the user.
+     */
```
edualonso, line 30:
```suggestion
    /**
     * Creates an [ApiKey] that can be used to authenticate as the user.
     * The value of the key must be persisted at this time as this is the only time it is visible.
     * The key is enabled by default when created. It can be disabled by calling [disable].
     *
     * @param name the name of the key.
     * @throws `IllegalArgumentException` if an invalid name for the key is sent to the server.
     * @return the new API key for the user.
     */
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+    public suspend fun create(name: String): ApiKey
+
+    /**
+     * Deletes a specific API key created by the user.
+     * Returns silently if no key is deleted.
+     * @param id the id of the key to delete.
+     */
```
edualonso, line 37:
```suggestion
    /**
     * Deletes a specific [ApiKey] created by the user.
     * Returns silently if no key is deleted.
     *
     * @param id the id of the key to delete.
     */
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+    public suspend fun delete(id: ObjectId)
+
+    /**
+     * Disables a specific API key created by the user.
+     *
+     * @param id the id of the key to disable.
+     * @throws IllegalArgumentException if a non existing API key is disabled.
+     */
```
edualonso, line 45:
```suggestion
    /**
     * Disables a specific [ApiKey] created by the user.
     *
     * @param id the id of the key to disable.
     * @throws `IllegalArgumentException` if a non-existing API key is disabled.
     */
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+    public suspend fun disable(id: ObjectId)
+
+    /**
+     * Enables a specific API key created by the user.
+     *
+     * @param id the id of the key to disable.
+     * @throws IllegalArgumentException if a non existing API key is enabled.
+     */
```
edualonso, line 53:
```suggestion
    /**
     * Enables a specific [ApiKey] created by the user.
     *
     * @param id the id of the key to disable.
     * @throws `IllegalArgumentException` if a non existing API key is enabled.
     */
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+    public suspend fun enable(id: ObjectId)
+
+    /**
+     * Fetches a specific user API key associated with the user.
+     *
+     * @param id the id of the key to fetch.
+     * @throws IllegalArgumentException if a non existing API key is fetched.
+     */
```
edualonso, line 61:
```suggestion
    /**
     * Fetches a specific user [ApiKey] associated with the user.
     *
     * @param id the id of the key to fetch.
     * @throws `IllegalArgumentException` if a non existing API key is fetched.
     */
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,69 @@
...
+    public suspend fun fetch(id: ObjectId): ApiKey?
+
+    /**
+     * Fetches all API keys associated with the user.
+     * Returns an empty list if no key is found.
+     */
```
edualonso, line 67:
```suggestion
    /**
     * Fetches all [ApiKey]s associated with the user.
     * Returns an empty list if no key is found.
     */
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+                    }.freeze()
+                )
+                return channel.receive().getOrThrow()
+            }
+        } catch (ex: AppException) {
+            if (ex.message?.contains("[Service][Unknown(-1)] expected Authorization header with JWT (Bearer schema).") == true) {
+                throw ex
+            } else {
+                // No-op
+            }
```
edualonso, line 70: What is this?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+                )
+                return channel.receive()
+                    .getOrThrow()
+            }
+        } catch (ex: ServiceException) {
+            if (ex.message?.contains("[Service][ApiKeyNotFound(35)] API key not found.") == true) {
+                throw IllegalArgumentException(ex.message!!)
+            } else {
+                throw ex
+            }
```
edualonso, line 93: Same here. If this is written to handle the unexpected behaviour you spotted while disabling the provider in the server you should write a `TODO` comment pointing to the github ticket where you reported that issue.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,162 @@
...
```
edualonso, line 116: Same here

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+                )
+                return channel.receive()
+                    .getOrThrow()
+            }
+        } catch (ex: ServiceException) {
+            if (ex.message?.contains("[Service][ApiKeyNotFound(35)] API key not found.") == true) {
+                return null
+            } else {
+                throw ex
+            }
```
edualonso, line 139: Same here

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/ApiKeyAuthImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+            apiKeyData.value,
+            apiKeyData.name,
+            !apiKeyData.disabled
+        )
+    }
+
+    override suspend fun create(name: String): ApiKey {
+        try {
+            Channel<Result<ApiKey>>(1).use { channel ->
+                RealmInterop.realm_app_user_apikey_provider_client_create_apikey(
```
edualonso, line 30: Similarly to the comment in `RealmInterop`, it looks like you could unify some of these parts under shared logic.

edualonso (changes requested):
Looks nice. Mostly my comments are suggestions about documentation and minor stuff. I would look into refactoring some shared logic across different operations. Also, try to get hold of someone from the server team to ask them about the unexpected behaviour you experienced while disabling the authentication provider, it would be nice to know if it's a bug or an expected behaviour so as to avoid adding custom error handling in our implementation.

Also, I'm not sure why your Core pointer appears as a change, maybe we need to look into that again 🤔

clementetb (changes requested):
Nice job, it looks great overall.

There are missing copyright headers on all new files.
Error handling in the `ApiKeyAuthImpl.kt` might need some work.
The changelog needs an entry for the core update.

⭐️

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,84 @@
...
+    public val app: App
+
+    /**
+     * Creates a user API key that can be used to authenticate as the user.
+     * The value of the key must be persisted at this time as this is the only time it is visible.
+     * The key is enabled when created. It can be disabled by calling the disable method.
+     *
+     * @param name the name of the key
+     * @throws IllegalArgumentException if an invalid name for the key is sent to the server.
```
cmelchior, line 43: All of these methods can also throw a general `ServiceException`. You can probably just copy the message from `EmailPasswordAuth`

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/ApiKeyAuthTests.kt
```diff
@@ -0,0 +1,237 @@
...
+                    fail("$method should have thrown an exception")
+                } catch (error: ServiceException) {
+                    assertEquals("[Service][Unknown(-1)] expected Authorization header with JWT (Bearer schema).", error.message)
+                }
+            }
+        }
+    }
+
+    @Test
+    @Ignore("There seems to be a bug in the endpoint used for this test, need to wait until it is completed")
```
cmelchior, line 222: Do we have an issue tracking this. We should put the link here then.

cmelchior (commented):
Mostly minor comments. CI also isn't passing but looks like it just might be formatting.

cmelchior (approved):
Looks good to me 👍 Only thing is that we should bump to the latest Core commit since the required changes have been merged (+update changelog) so we don't depend on a custom branch in Core.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKey.kt
clementetb: The constructor parameters could be formatted in multiple lines to faciltate reading.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKey.kt
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.kotlin.mongodb.auth
+
+import io.realm.kotlin.types.ObjectId
+/**
+ * Class representing an API key for a [User]. An API key can be used to represent a
+ * user when logging in instead of using email and password.
+ * Note that the value of a key will only be available immediately after the key is created, after
```
clementetb, line 22: Add an empty line to separate paragraphs.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKey.kt
```diff
@@ -0,0 +1,31 @@
...
+ * which point it is not visible anymore. This means that keys returned by [ApiKeyAuth.fetch] and
+ * [ApikeyAuth.fetchall] will have a `null` [value]. Anyone creating an API key is responsible for
+ * storing it safely after that.
+ * @param id an [ObjectId] uniquely identifying the key.
```
clementetb, line 26: Add an empty line to separate from the parameters.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public val user: User
+
+    /**
+     * The App that this instance in associated with.
+     */
+    public val app: App
+
+    /**
+     * Creates a user API key that can be used to authenticate as the user.
```
clementetb, line 39: These documentation blocks can improve formatting take a look into this example:
https://github.com/realm/realm-kotlin/blob/760055a5bfa1df1c1bb98eb24335772fcd198e74/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmResults.kt#L38

Separate paragraphs, @throws to the end

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/ApiKeyAuth.kt
```diff
@@ -0,0 +1,97 @@
...
+     * @throws IllegalArgumentException if an invalid name for the key is sent to the server.
+     * @throws io.realm.kotlin.mongodb.exceptions.ServiceException for other failures that can happen when
+     * communicating with App Services. See [AppException] for details.
+     * @return the new API key for the user.
+     */
+    public suspend fun create(name: String): ApiKey
+
+    /**
+     * Deletes a specific API key created by the user.
+     * Returns silently if no key is deleted.
```
clementetb, line 53:
```suggestion
     * Deletes a specific API key created by the user. The function would complete normally if the provided key does not exist.
```

clementetb (approved):
Added some minor changes, otherwise it looks solid. Happy when CI is green.

🚀

cmelchior (approved):
There still seem to be a few outstanding comments around docs formatting, but otherwise this looks ready to merge 💯

Issue #1068, opened 2022-10-13
SyncSession.resume/pause crash after receiving remote update


cmelchior:
It looks like the `SyncSessionImpl.nativePointer` is somehow being outdated and crashes the app when being used (e.g. when calling pause/resume).

This is reproducible by having two apps communicating. After Device A sends changes to Device B, Device B will crash when interacting with the SyncSession pointer.

cmelchior:
The root problem was due to a pointer going out of date. The sequence was as follows:

1) Device A: Open Flexible Sync with `initialSubscriptions`. This would trigger the creation of the SyncedContext which stored a copy of the current `dbPointer`.
2) Device B writes a change
3) Device A receives this change, integrates it, updates the Realm pointer and closes the old one
4) Device A access the SyncSession. This tried to lazily create the session pointer using the, now closed, dbPointer.
5) Device A would crash with a SEGFAULT.

Creating both Subscriptions and SyncSession immediately when accessed the first time fixes the problem.

Issue #1070, opened 2022-10-13
Realm.asFlow() is not being triggered on Sync changes.


cmelchior:
Looks like our `Realm.asFlow()` isn't being triggered when there is a remote change, whereas queries are being triggered correctly:

```
realm.asFlow().collect {
  // Is not being triggered when changes are coming from another device
  println("$it -> ${it.realm.version()}")
}
realm.query<FlexParentObject>().asFlow().collect {
  // Works if changes are coming from another device
  println("$it -> ${it.list.size}")
}
```

This test reproduces the issue: https://github.com/realm/realm-kotlin/pull/1071/files#diff-65ad5644786458a5a2bc5d99f679e8533faf91be5c5ec9aedf806cba336acaefR214

rorbech:
Seems to be that we are only initializing the `nofitier`'s internal `LiveRealm` when we are actually listening for something other that top level realm update. So the `realm_add_realm_changed_callback` is not registered before listening for something else. An immediate workaround until a fix is ready would be to just trigger some notification, ex.
```
async { realm.query<X>.asFlow().first {} }
```

PR #1071, opened 2022-10-13
Fix SyncSession crashing when accessed after receiving remote changes

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/1068

Note, when creating this patch I discovered another bug: https://github.com/realm/realm-kotlin/issues/1070.

A test has been written that catch this but it is marked as `@Ignore` for now.

### DIFF
--- a/packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
+++ b/packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
@@ -31,15 +31,12 @@ import io.realm.kotlin.ext.query
 import io.realm.kotlin.internal.platform.fileExists
 import io.realm.kotlin.internal.platform.freeze
 import io.realm.kotlin.internal.platform.runBlocking
-import io.realm.kotlin.log.LogLevel
 import io.realm.kotlin.mongodb.App
 import io.realm.kotlin.mongodb.User
 import io.realm.kotlin.mongodb.exceptions.DownloadingRealmTimeOutException
 import io.realm.kotlin.mongodb.exceptions.SyncException
 import io.realm.kotlin.mongodb.exceptions.UnrecoverableSyncException
-import io.realm.kotlin.mongodb.subscriptions
 import io.realm.kotlin.mongodb.sync.InitialSubscriptionsCallback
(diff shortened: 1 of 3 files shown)


packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -768,6 +836,7 @@ class SyncedRealmTests {
                 localRealm.writeCopyTo(flexSyncConfig)
             }
         }
+        flexApp.close()
```
edualonso, line 839: 👍

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
```diff
@@ -969,6 +1044,76 @@ class SyncedRealmTests {
...
+            // 3. On the next update of Realm, we run through the weak list of all previous
+            //    RealmReferences and close all native pointers with their JVM object GC'ed.
+            //    This should now include the object created in step 1.
+            realm1.write { }
+        }
+
+        // 4. With the original native dbPointer now being closed, accessing the syncSession for
+        //    the first time should still work.
+        try {
+            realm1.syncSession.pause()
```
edualonso, line 1110: If an error happens here it will get swallowed by the `finally` and we won't know whether the session is still accessible. I would check the session state has actually been updated after calling `pause`.

edualonso (approved):
Looks great, just one comment about the newly added test.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/shared/SyncedRealmTests.kt
cmelchior: Hmm, `finally` doesn't swallow the exception, it should still propagate and fail the test, but you are right. Checking the state would improve the test.

edualonso: Yes, I didn't mean an exception gets swallowed but rather that the state itself doesn't get updated for some reason. Poor wording 🙊

PR #1074, opened 2022-10-15
Add Gradle Kotlin instructions

astatio:
The readme contained only instructions on how to use SNAPSHOT version if Groovy for Gradle was being used. Since it's a KMM project, it makes sense to provide the Kotlin instructions as well.


astatio:
@cla-bot check

nhachicha (approved):
Thanks @astatio !

PR #1078, opened 2022-10-18
Remove anonymous union native workaround

rorbech:


nhachicha (approved):
LGTM 👍 I didn't know they fixed this since 1.5

PR #1082, opened 2022-10-20
Add kbson library and support for BsonObjectId

clementetb:
This PR adds the new kbson to the SDK. The library has its own implementation of `ObjectId` and this PR adds it as a supported modeling type. 

To avoid breaking the API both the Realm `ObjectId` and the kbson can be used. 

The Realm one has been marked as deprecated and would be removed in the next major in favor of the kbson implementation.

It adds support for using the BsonObjectId as a partition key, and replaces the `PartitionValue` class with a `BsonValue`.


packages/cinterop/build.gradle.kts
```diff
@@ -179,6 +179,8 @@ kotlin {
             dependencies {
                 implementation(kotlin("stdlib-common"))
                 implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.coroutines}")
+
+                api("org.mongodb.kbson:kbson:${Versions.kbson}")
```
cmelchior, line 183: Does the API visibility propagate all the way to end users if defined in `cinterop`? Or do we also need to define it in `library-base`?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -603,8 +603,8 @@ internal object RealmObjectHelper {
         obj.checkValid()
 
         val propertyMetadata = checkPropertyType(obj, propertyName, value)
-        val clazz = RealmStorageTypeImpl.fromCorePropertyType(propertyMetadata.type).kClass.let {
-            if (it == BaseRealmObject::class) DynamicMutableRealmObject::class else it
+        val clazz = RealmStorageTypeImpl.fromCorePropertyType(propertyMetadata.type).kClass.let { it ->
+            if (it == BaseRealmObject::class) DynamicMutableRealmObject::class else value?.let { it::class } ?: it
```
cmelchior, line 607: Not 100% sure why this change was needed?

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/ImportTests.kt
```diff
@@ -83,7 +83,7 @@ class ImportTests {
                 Float::class -> assertEquals(3.14f, managed.floatField)
                 Double::class -> assertEquals(1.19840122, managed.doubleField)
                 RealmInstant::class -> assertEquals(RealmInstant.from(100, 1000), managed.timestampField)
-                ObjectId::class -> assertEquals(ObjectId.from("507f1f77bcf86cd799439011"), managed.objectIdField)
+                ObjectId::class -> assertEquals(ObjectId.from("507f1f77bcf86cd799439011"), managed.realmObjectIdField)
```
cmelchior, line 86: Don't we need to keep tests for both kbson objectID and our own?

cmelchior (commented):
Looking good.... Looks like we need to enable Multidex in our integration tests + obligatory missing CHANGELOG entry ;)

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
clementetb: It is needed to allow both `ObjectId` and `BsonObjectId` to be in the dynamic set, if not it always resolves to the one defined in the core property types (BsonObjectId).

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/ImportTests.kt
clementetb: Yes, I marked it as reviewable too soon there were some missing tests + support on the config missing, and the changelog entry 🫣

packages/cinterop/build.gradle.kts
clementetb: Yes,  it propagates down to the end users.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/QueryTests.kt
```diff
@@ -141,6 +142,10 @@ class QueryTests {
                 }
                 RealmStorageType.OBJECT_ID -> {
                     realm.query<QuerySample>("objectIdField = $0", ObjectId.from("507f191e810c19729de860ea"))
+                    realm.query<QuerySample>("bsonObjectIdField = $0", ObjectId.from("507f191e810c19729de860ea"))
+
+                    realm.query<QuerySample>("objectIdField = $0", BsonObjectId("507f191e810c19729de860ea"))
```
cmelchior, line 147: Copy/paste mistake?

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
```diff
@@ -73,6 +72,10 @@ import kotlin.time.Duration.Companion.seconds
 
 const val DEFAULT_NAME = "test.realm"
 
+internal enum class ValueType {
+    STRING, LONG, INT, NULL, OBJECT_ID, UUID
```
cmelchior, line 76: Are these being used for anything?

cmelchior (approved):
Nice 👍 ... only had minor comments.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/QueryTests.kt
clementetb: We are testing the combination of the two `ObjectId` query arguments on the two fields defined by the different `ObjectId` implementations. They should be interchangeable.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt
clementetb: Yes, they are using in some tests below to do exhaustive testing see: https://github.com/realm/realm-kotlin/blob/090b4bc6bfd6ecec3955c90f6201836f962589bf/packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncConfigTests.kt#L331

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/SyncConfigurationImpl.kt
```diff
@@ -53,11 +52,12 @@ import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
+import org.mongodb.kbson.BsonValue
 
 @Suppress("LongParameterList")
 internal class SyncConfigurationImpl(
     private val configuration: InternalConfiguration,
-    internal val partitionValue: PartitionValue?,
+    internal val partitionValue: BsonValue?,
```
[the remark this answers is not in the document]
clementetb: I wasn't aware, I removed it because the `PartitionValue` had some implementation details on how to serialize `BsonValues`.. so I see an opportunity to remove duplicate code :)

integration-tests/gradle-plugin-test/single-platform/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ android {
     defaultConfig {
         minSdk = Versions.Android.minSdk
         targetSdk = Versions.Android.targetSdk
+        multiDexEnabled = true
```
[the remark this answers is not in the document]
clementetb: It defines 483 methods and does not come with any ProGuard definition.

PR #1084, opened 2022-10-21
Jg/configuration in memory

JacobOscarGunnarsson:
Allows realms to be created in memory rather than directly on disk

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
@@ -361,7 +361,6 @@ public interface Configuration {
         public fun initialData(callback: InitialDataCallback): S =
             apply { initialDataCallback = callback } as S
 
-
         /**
          * Setting this will create an in-memory Realm instead of saving it to disk. In-memory Realms might still use
          * disk space if memory is running low, but all files created by an in-memory Realm will be deleted when the
@@ -371,7 +370,7 @@ public interface Configuration {
          * reference to the in-memory Realm object with the specific name as long as you want the data to last.
          */
         public fun inMemory(): S =
-           apply { this.durability = Durability.MEM_ONLY } as S
+            apply { this.durability = Durability.MEM_ONLY } as S
 
         /**
          * Removes the default system logger from being installed. If no custom loggers have
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
@@ -165,6 +165,12 @@ public interface Configuration {
      */
     public val initialDataCallback: InitialDataCallback?
 
+    public enum class Durability {
+        FULL, MEM_ONLY
+    }
+
+    public val durability: Durability
+
     /**
      * Base class for configuration builders that holds properties available to both
      * [RealmConfiguration] and [SyncConfiguration].
@@ -205,6 +211,7 @@ public interface Configuration {
         protected var encryptionKey: ByteArray? = null
         protected var compactOnLaunchCallback: CompactOnLaunchCallback? = null
         protected var initialDataCallback: InitialDataCallback? = null
+        protected var durability: Durability = Durability.FULL
 
         /**
          * Sets the filename of the realm file.
@@ -353,6 +360,18 @@ public interface Configuration {
          */
         public fun initialData(callback: InitialDataCallback): S =
             apply { initialDataCallback = callback } as S
+
+
+        /**
+         * Setting this will create an in-memory Realm instead of saving it to disk. In-memory Realms might still use
+         * disk space if memory is running low, but all files created by an in-memory Realm will be deleted when the
+         * Realm is closed.
+         *
+         * Note that because in-memory Realms are not persisted, you must be sure to hold on to at least one non-closed
+         * reference to the in-memory Realm object with the specific name as long as you want the data to last.
+         */
+        public fun inMemory(): S =
+           apply { this.durability = Durability.MEM_ONLY } as S
 
         /**
          * Removes the default system logger from being installed. If no custom loggers have
--- a/packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
+++ b/packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
@@ -5,6 +5,8 @@ import io.realm.kotlin.Realm
 import io.realm.kotlin.RealmConfiguration
 import io.realm.kotlin.entities.Sample
 import io.realm.kotlin.ext.query
+import io.realm.kotlin.internal.RealmImpl
+import io.realm.kotlin.internal.platform.runBlocking
 import io.realm.kotlin.test.platform.PlatformUtils
 import io.realm.kotlin.test.util.TestHelper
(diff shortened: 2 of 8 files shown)


packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
```diff
@@ -0,0 +1,243 @@
...
+            .encryptionKey(randomKey)
+            .build()
+            .let { conf ->
+                assertFailsWith(IllegalArgumentException::class, "Encrypted Realm should not be openable with a wrong encryption key") {
+                    Realm.open(conf)
+                }
+            }
+    }
+
+    // Tests writeCopyTo result when called in a transaction.
```
JacobOscarGunnarsson, line 142: This behaves somewhat differently from the Java SDK, my suspicion is that the semantics differ between

`realm.beginTransaction()` and `realm.commitTransaction()`

in comparison to `realm.writeBlocking{}`

JacobOscarGunnarsson: the key difference is that the assertEquals in the test has a value of 1 in realm Java while in kotlin it is 0

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
```diff
@@ -0,0 +1,243 @@
...
+        val realm2 = Realm.open(configuration2)
+        realm2.writeBlocking {
+            copyToRealm(Sample().apply { stringField = "bar" })
+        }
+        assertEquals(1, realm.query(Sample::class).count().find())
+        assertEquals("foo", realm.query<Sample>("stringField == 'foo'").find().first().stringField)
+        assertEquals(1, realm2.query(Sample::class).count().find())
+        assertEquals("bar", realm2.query<Sample>("stringField == 'bar'").find().first().stringField)
+        realm2.close()
+        PlatformUtils.deleteTempDir(tmpDir2)
```
cmelchior, line 82: I would move this into a `try { } finally { }` block with the asserts inside try and cleanup inside `finally`. That way we will always do cleanup even if the test fails.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Test
+    fun inMemoryRealm_delete() {
+        try {
```
cmelchior, line 87: Use `assertFailsWith<IllegalStateException>` instead

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
cmelchior: Yes, If I remember correctly there is a difference between Kotlin and Java here, so this is fine.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
```diff
@@ -0,0 +1,243 @@
...
+    // Test below scenario:
+    // 1. Creates a in-memory Realm instance in the main thread.
+    // 2. Creates a in-memory Realm with same name in another thread.
+    // 3. Closes the in-memory Realm instance in the main thread and the Realm data should not be released since
+    //    another instance is still held by the other thread.
+    // 4. Closes the in-memory Realm instance and the Realm data should be released since no more instance with the
+    //    specific name exists.
+    @Test
+    @Throws(InterruptedException::class, ExecutionException::class)
+    fun multiThread() {
```
cmelchior, line 172: This test looks very Java-like, and we could probably rewrite it so it uses coroutines instead. That said, if it runs fine on both JVM and Native keeping as is, is probably fine for now.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -165,6 +165,20 @@ public interface Configuration {
...
+     * of the realm is not persistent if it is closed while having a durability of [MEM_ONLY].
+     */
+    public enum class Durability {
+        FULL, MEM_ONLY
+    }
+
+    /**
+     * Describes whether the realm should reside in memory or on disk.
+     */
+    public val durability: Durability
```
cmelchior, line 180: We should also add tests to `RealmConfigurationTests` and `SyncConfigurationTests` that verify the value of this.

It also looks like you didn't add support for SyncConfigurations (judging from CI), so having that test there would also catch that.

packages/test-base/src/androidTest/kotlin/io/realm/kotlin/test/shared/RealmInMemoryTests.kt
```diff
@@ -0,0 +1,243 @@
...
+import org.junit.Test
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeUnit
+import kotlin.random.Random
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.fail
+
+class RealmInMemoryTests {
```
cmelchior, line 24: I know we didn't check for this in Java, but having a test checking that files are actually deleted on disk would be helpful.

cmelchior (changes requested):
Looking good 👍 ... I left some minor comments about wording/naming in the public API + some refactorings in the tests.

Also, it looks like SyncConfiguration support is missing (which is crashing CI).

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -165,6 +165,28 @@ public interface Configuration {
      */
     public val initialDataCallback: InitialDataCallback?
 
+    /**
+     * Enum describing how the Realm file is persisted.
+     */
+    public enum class Durability {
```
rorbech, line 171: Why do we have the public concept of `Durability` but set this via `inMemory`  instead of either setting the `durability(IN_MEMORY)` on the configuration or just returning the configuration state as a getter `isInMemory()`?

cmelchior: Good points. Just having a `isInMemory()` is a lot smaller change. Exposing the enum would mostly be valuable if we had more than 2 types, which we don't have any plans for adding.

packages/test-sync/src/androidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FlexibleSyncConfigurationTests.kt
```diff
@@ -155,6 +156,18 @@ class FlexibleSyncConfigurationTests {
         assertTrue(config.initialSubscriptions!!.rerunOnOpen)
     }
 
+    @Test
+    fun durability() {
```
cmelchior, line 160: Also missing a general `SyncConfiguration` test for partition-based sync. Could be placed in `SyncConfigTests`

cmelchior (approved):
🚀

PR #1085, opened 2022-10-21
Enabling hierarchical multiplatform project structure

### LINKED ISSUE #1006: Enable hierarchical project structure
We should move to proper hierarchical project setup as the new default and we could clean up our source set configuration and take advantage of the new cinterop commonizer to share code across MacOS and iOS, etc.

https://kotlinlang.org/docs/multiplatform-hierarchy.html

rorbech:
This PR rearranges the source sets and enables [Hierarchical Multi Platform Project structure ](https://kotlinlang.org/docs/multiplatform-hierarchy.html)

Closes #1006


packages/cinterop/build.gradle.kts
```diff
@@ -206,37 +206,36 @@ kotlin {
...
             }
         }
-        val macosMain by getting {
-            // TODO HIERARCHICAL-BUILD From 1.5.30-M1 we should be able to commonize cinterops using
-            //  kotlin.mpp.enableCInteropCommonization=true (https://youtrack.jetbrains.com/issue/KT-40975)
-            //  This would also require us to enable hierarchical setup, which is currently blocked by
-            //  https://youtrack.jetbrains.com/issue/KT-48153
-            // FIXME HIERARCHICAL-BUILD Rename to nativeDarwin
-            kotlin.srcDir("src/darwin/kotlin")
+        val nativeDarwin by creating {
```
cmelchior, line 209: Aren't this one created automatically? https://kotlinlang.org/docs/multiplatform-share-on-platforms.html#share-code-in-libraries

This would also impact a lot of wiring below in this file.

Update:
After trying to build without creating this, I can see it is failing, which I guess is why you are creating it and after re-reading the documentation it became clear that the above link was just an example, and the only supported shortcuts are these ones 🙈 https://kotlinlang.org/docs/multiplatform-share-on-platforms.html#use-target-shortcuts

packages/gradle.properties
```diff
@@ -21,15 +21,8 @@ kotlin.mpp.stability.nowarn=true
 kotlin.code.style=official
 android.useAndroidX=true
 
-# We cannot enable hierarchical setup as common symbols are currently not visible in Android source
-# sets in consuming modules (https://youtrack.jetbrains.com/issue/KT-48153)
```
cmelchior, line 25: Should we keep a reference to this YouTrack, since it is still relevant?

packages/library-sync/build.gradle.kts
```diff
@@ -38,12 +38,13 @@ kotlin {
     android("android") {
         publishLibraryVariants("release", "debug")
     }
-    ios()
+    iosX64()
+    iosArm64()
     iosSimulatorArm64()
-    macosX64("macos")
+    macosX64 {}
```
cmelchior, line 44:
```suggestion
    macosX64()
```

packages/test-base/build.gradle.kts
```diff
@@ -180,41 +176,28 @@ kotlin {
...
+        val commonMain by getting
+        val commonTest by getting
+        val nativeDarwin by creating {
+            dependsOn(commonMain)
         }
-    }
-}
-
-// Needs running emulator
-tasks.named("iosTest") {
```
cmelchior, line 205: So the emulator is started automatically for us now?

packages/test-sync/build.gradle.kts
```diff
@@ -60,6 +60,10 @@ kotlin {
                 // FIXME AUTO-SETUP Removed automatic dependency injection to ensure observability of
                 //  requirements for now
                 implementation(project(":test-base"))
+                // IDE Doesn't resolve library-base symbols if not adding it as an explicit
+                // dependency. Probably due to our own custom dependency substitution above, but
+                // should be an issue as it is already a transitive dependency of library-sync.
```
cmelchior, line 65:
```suggestion
                // shouldn't be an issue as it is already a transitive dependency of library-sync.
```

cmelchior (commented):
Looking really nice. Still playing around with so just leaving initial feedback. We should also update https://github.com/realm/realm-kotlin/blob/master/CONTRIBUTING.md#multiplatform-source-layout to reflect the new hierachy.

packages/test-base/build.gradle.kts
rorbech: Yep. Handled by the Kotlin gradle plugin ... the output after removing this is 
```
> Task :test-base:iosX64Test
Starting process 'command '/usr/bin/xcrun''. Working directory: /Users/claus.rorbech/proj/realm-kotlin/packages/test-base Command: /usr/bin/xcrun simctl spawn --standalone iPhone 12 /Users/claus.rorbech/proj/realm-kotlin/packages/test-base/build/bin/iosX64/debugTest/test.kexe -- --ktest_no_exit_code --ktest_logger=TEAMCITY
```

packages/gradle.properties
rorbech: The issue is specific to composite gradle builds which we don't use anymore. And since we are in fact enabling hierarchical structure it seems a bit weird to keep. If we should keep it, then it would maybe better to place it at the workaround for composite builds not working at https://github.com/realm/realm-kotlin/blob/master/packages/test-base/build.gradle.kts#L23?

packages/cinterop/build.gradle.kts
rorbech: 🤪  ... 👍

packages/gradle.properties
rorbech: I guess it is actually already referenced (or at least the duplicated issue) in https://github.com/realm/realm-kotlin/blob/master/CONTRIBUTING.md#advanced-project-setup which is probably the right place to keep the reference as it covers the overall project setup.

packages/cinterop/build.gradle.kts
```diff
@@ -614,6 +613,16 @@ tasks.named("jvmProcessResources") {
     dependsOn(buildJVMSharedLibs)
 }
 
+// Add generic macosTest task that execute macos tests according to the current host architecture
+tasks.register("macosTest") {
+    val arch = when (System.getProperty("os.arch")) {
+        "aarch64" -> "Arm64"
+        "x86_64" -> "X64"
+        else -> "Unsupported MacOs architecture"
```
nhachicha, line 621:
```suggestion
        else -> "Unsupported macOS architecture"
```

packages/gradle.properties
```diff
@@ -21,15 +21,8 @@ kotlin.mpp.stability.nowarn=true
...
-#kotlin.mpp.enableGranularSourceSetsMetadata=true
-#kotlin.mpp.enableCompatibilityMetadataVariant=true
-#kotlin.native.enableDependencyPropagation=false
-
-# Hierachal builds are the default in Kotlin 1.6.20. Use old layout until https://github.com/realm/realm-kotlin/issues/1006
-# is fixed.
-kotlin.mpp.hierarchicalStructureSupport=false
+kotlin.mpp.hierarchicalStructureSupport=true
```
nhachicha, line 24: Since we're using Kotlin `1.7.20` on master these flags are not necessary I think (they're  opt-in pre 1.6.20)

packages/library-base/build.gradle.kts
```diff
@@ -94,30 +94,29 @@ kotlin {
...
-//        getByName("macosMain") {
-//            // TODO HMPP Should be shared source set
-//        }
         val macosArm64Main by getting {
-            kotlin.srcDir("src/darwin/kotlin")
-            kotlin.srcDir("src/macosMain/kotlin")
+            dependsOn(nativeMacos)
         }
-
         getByName("iosArm64Main") {
```
nhachicha, line 112: can't we use `val iosArm64Main by getting`?

packages/test-base/build.gradle.kts
nhachicha, line 205: I suppose we can still set the `iosDevice` property in Gradle properties/Env to request a specific simulator to align local/CI tests

nhachicha (commented):
Nice refactor 🙌. minor comments, also I still have some red lines in `nativeDarwinTest`regarding some imports
image
not sure here why `coreErrorAsThrowable` is implemented inside an object for JVM and as global function for darwin?

cmelchior (approved):
Nice...now I can run all tests from the IDE 🚀 

IMO this looks ready to merge once the merge conflicts are fixed. One thought though. Does it make sense that we symlink folders to native and jvm anymore? Don't we only need the symlink in `androidAndroidTest`?  But it isn't critical and can be addressed later.

packages/gradle.properties
rorbech: But isn't it better to apply what we need explicitly? 🤔

packages/test-base/build.gradle.kts
rorbech: Maybe a good option to keep. Hooked the `iosDevice-property into the Gradle plugins infrastructure with https://github.com/realm/realm-kotlin/pull/1085/files#diff-ad7d5c068deb652e9518763651bed6874ac941679d187d260bec712efec3cf27R198

rorbech:
> IMO this looks ready to merge once the merge conflicts are fixed. One thought though. Does it make sense that we symlink folders to native and jvm anymore? Don't we only need the symlink in `androidAndroidTest`? But it isn't critical and can be addressed later.

The reason why we are sym-linking tests are not due to hierarchical builds not working, but because we would otherwise not be able to trigger Android tests from the IDE. Details are in https://github.com/realm/realm-kotlin/blob/master/CONTRIBUTING.md#test-organization, but basically it is because the KMP project for some reason have chosen to [execute Android tests without an Android context](https://youtrack.jetbrains.com/issue/KT-42298/KMM-plugin-Run-common-tests-on-Android-devices#focus=Comments-27-4657970.0-0) instead of just using `jvmTest` for that 🤯 🤷

rorbech:
We could consider to move the actual test files to `commonTest` and change the direction of the symlink so that `androidAndroidTest` links to the `commonTest` ones. This would enable us to have common test where you select platform from the gutter icon dialog, except for `androidTests. This would require explicit removal of `commonTest` source set from the `androidTest` to avoid duplicate symbols a bit like suggested in https://youtrack.jetbrains.com/issue/KT-34712/MPP-Gradle-Plugin-provide-a-way-to-suppress-commonTest-dependsOn#focus=Comments-27-6290261.0-0

PR #1089, opened 2022-10-26
Break reference cycles to realm from realm callbacks

### LINKED ISSUE #671: Unregister realm and schema change callbacks when realm is GC'ed

rorbech:
Closes #671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * [Sync] Added support for API key authentication. (Issue [#432](https://github.com/realm/realm-kotlin/issues/432))
 
 ### Fixed
-* Close underlying realm if it is no longer referenced by any Kotlin object.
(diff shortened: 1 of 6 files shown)


cmelchior (approved):
Nice 🥳

PR #1090, opened 2022-10-26
Update core reference to next-major

### LINKED ISSUE #1007: Realms file grows "infinitely"
I have a realms file that grows "infinitely" (theoretically).
I update this file around 5 times a second and I get the latest version using a write transaction (doing the query outside a write transaction and then updating inside of it will throw an exception so this is the only way possible) and I always put each one of these write transactions inside a try-with-resources. 

The logic is looking up an ID and getting a long field and add a quantity to it everytime the function is triggered. 

The file grows immensity big and within a month the file grew to 23GB! When I open it with Realm Studio and export as a new realm file the resulting size is a mere 705KB - far more accurate when taking into account the small amount of data inside. I was unable to find anything in the documentation that could justify this behaviour but it's really bad to have a file grow this large unnecessarily. Why does this happen?

rorbech:
Preliminary branch to test out https://github.com/realm/realm-core/tree/next-major including [reclaiming of intermediate frozen versions](https://github.com/realm/realm-core/pull/5440) 

Closes #1007

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Enhancements
 * Add support for in-memory Realms.
+* Reclaim intermediate non-referenced version. (Issue [#1007](https://github.com/realm/realm-kotlin/issues/1007))
 * [Sync] Added support for API key authentication. (Issue [#432](https://github.com/realm/realm-kotlin/issues/432))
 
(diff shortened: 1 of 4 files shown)


rorbech:
Superseded by #1137

Issue #1081, opened 2022-10-19
Write transactions increase Realm version number on each call


Nek-12:
Hi! This is more of a question of understanding how to keep realm versions low

I'm using the following code to write to the database:

```kotlin
override suspend fun save(e: Entry): Entry = Entry(db.write { upsert(e) })

// equal to, using inline functions:

suspend fun save(e: Entry): Entry = realm.write {
     val rid: RealmUUID = e.id
     
     val update: EntryEntity.() -> Unit = {
          property = e.property
     }
     val entity = query<EntryEntity>("id == $0", rid).first().find() ?: EntryEntity().apply { id = rid }
     copyToRealm(entity.apply(update), UpdatePolicy.ERROR)
}.mapToEntryDataClass() // maps the object to a domain model immediately outside of the db.write {} block
```
The save() function is invoked on a background thread (Dispatchers.IO or Dispatchers.Default)

I'm using a global singleton instance provided with Koin to my app

```kotlin
single(createdAtStart = true) { provideDatabase() } // <-- provideDatabase() creates and opens a Realm instance
```
And a few of the DAO objects, singletons too, accessed on random dispatchers.

But while queries run on Dispatchers.IO using this code:
```kotlin
internal suspend inline fun <T : BaseRealmObject> RealmQuery<T>.findSuspend(
    context: CoroutineContext = Dispatchers.IO
) = withContext(context) { find() }

// result is immediately mapped to domain object
```
all run fine, each write operation produces the following Logcat output:
```
     D  Updating Realm version: VersionId(version=50) -> VersionId(version=51)
     D  Closing unreferenced version: 49
     D  Updating Realm version: VersionId(version=51) -> VersionId(version=51)
     D  Updating Realm version: VersionId(version=51) -> VersionId(version=52)
     // each write produces 2 version increments
     D  Updating Realm version: VersionId(version=52) -> VersionId(version=52)
     D  Updating Realm version: VersionId(version=52) -> VersionId(version=53)
     
     D  Updating Realm version: VersionId(version=53) -> VersionId(version=53)
     D  Updating Realm version: VersionId(version=53) -> VersionId(version=54)
     
     D  Updating Realm version: VersionId(version=54) -> VersionId(version=54)
     D  Updating Realm version: VersionId(version=54) -> VersionId(version=55)
     
     D  Updating Realm version: VersionId(version=55) -> VersionId(version=55)
     D  Updating Realm version: VersionId(version=55) -> VersionId(version=56)
     
[...]

PR #1092, opened 2022-10-27
Add CopyFromRealm

### LINKED ISSUE #549: Add support for copyFromRealm
It would be convenient to add a method like copyFromRealm in order to be able to get unmanaged objects from the DB.

cmelchior:
Closes #549 

This is still just a POC, since we haven't agreed on an API yet. But it seems very easy to implement with our current infrastructure.

Given the use cases we have seen surface in https://github.com/realm/realm-kotlin/issues/1081, I would propose the following API:

```
// Main API on `Realm`
interface TypedRealm {
  fun copyFromRealm(obj: T, depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = false): T
  fun copyFromRealm(objects: Iterable<T>, depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = false): List<T>
}

// Extension functions on RealmObject, RealmList, RealmList and RealmSet. By using extension 
// functions we can get the return type correct without involving generic parameters

fun RealmObject> RealmObject.copyFromRealm(depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = false): T
fun RealmList.copyFromRealm(depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = false): List<T>
fun RealmSet.copyFromRealm(depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = false): Set<T>
fun RealmResults.copyFromRealm(depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = false): List<T>
```

This PR also reworks how we release pointers. `RealmInterop.realm_release` has been made internal and instead a `NativePointer.release()` method has been added. This was needed so we can track whether or not a pointer has already been released and don't accidentally release it twice. This does add the overhead of 1 extra object allocation pr. native ref (AtomicBoolean), but I couldn't come up with something simpler. Any suggestions are welcome. 

Benchmarks for the modified handling of pointers. It seems to have neglible effect on accessors:

```
// With updated NativePointer which includes check on every access

Benchmark                            Mode  Cnt          Score         Error  Units
AccessorTests.managedReadBoolean    thrpt  200    1572429,318 ±  103100,881  ops/s
AccessorTests.managedReadDouble     thrpt  200    1628898,485 ±  115850,915  ops/s
AccessorTests.managedReadList       thrpt  200     941460,036 ±   82575,846  ops/s
AccessorTests.managedReadLong       thrpt  200    1826582,350 ±  133130,227  ops/s
AccessorTests.managedReadObject     thrpt  200     433014,501 ±   34624,683  ops/s
AccessorTests.managedReadString     thrpt  200     982106,002 ±   64695,591  ops/s

---

// Current Master

AccessorTests.managedReadBoolean    thrpt  200    1545694,681 ±   105862,798  ops/s
AccessorTests.managedReadDouble     thrpt  200    1615993,717 ±   112399,448  ops/s
AccessorTests.managedReadList       thrpt  200     956839,824 ±    84332,838  ops/s
AccessorTests.managedReadLong       thrpt  200    1776812,581 ±   135987,562  ops/s
AccessorTests.managedReadObject     thrpt  200     428942,463 ±    31123,892  ops/s
AccessorTests.managedReadString     thrpt  200     905974,727 ±    59592,816  ops/s
```

**TODO**
- [x] How to track and close intermediate objects created during the copy? 
- [x] Also close managed collections


cmelchior:
This PR is now ready for review. There is still an outstanding issue about closing objects in the internal cache after the copy is done, but that should prevent starting on review since there is quite a few concepts in here that might take a few iterations.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,67 @@ internal interface InternalTypedRealm : TypedRealm {
...
+            throw IllegalArgumentException("This object is unmanaged. Only managed objects can be copied: $obj.")
+        }
+        if (!obj.isValid()) {
+            throw IllegalArgumentException(
+                "Only valid objects can be copied from Realm. " +
+                    "This object was either deleted or the Realm has been closed, making this " +
+                    "object invalid: $obj."
+            )
+        }
+        if (depth < 0) {
```
nhachicha, line 54: Shouldn't depth be `1` at least? unless you consider the root object at level `0`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,67 @@ internal interface InternalTypedRealm : TypedRealm {
...
+            throw IllegalArgumentException("Only a depth of 0 or more is allowed. Depth was: $depth.")
+        }
+        if (obj is RealmObjectInternal) {
+            val objectRef: RealmObjectReference<out BaseRealmObject> = obj.io_realm_kotlin_objectReference!!
+            val realmRef: RealmReference = objectRef.owner
+            val mediator: Mediator = realmRef.owner.configuration.mediator
+            return createDetachedCopy(mediator, obj, 0, depth, closeAfterCopy, cache)
+        } else {
+            throw IllegalStateException()
```
nhachicha, line 63: Can we add a message for context

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,67 @@ internal interface InternalTypedRealm : TypedRealm {
...
+        }
+    }
+
+    override fun <T : TypedRealmObject> copyFromRealm(obj: T, depth: Int, closeAfterCopy: Boolean): T {
+        return copyObjectFromRealm(obj, depth, closeAfterCopy, mutableMapOf())
+    }
+    override fun <T : TypedRealmObject> copyFromRealm(collection: Iterable<T>, depth: Int, closeAfterCopy: Boolean): List<T> {
+        val (valid: Boolean, nativePointer: NativePointer<*>?) = when (collection) {
+            is ManagedRealmList -> Pair(collection.isValid(), collection.nativePointer)
```
nhachicha, line 72: you can fold 
```Kotlin
is ManagedRealmList, ManagedRealmSet
```
Since they return the same `Pair`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,67 @@ internal interface InternalTypedRealm : TypedRealm {
...
+            is ManagedRealmSet -> Pair(collection.isValid(), collection.nativePointer)
+            is RealmResultsImpl -> Pair(!collection.realm.isClosed(), collection.nativePointer)
```
nhachicha, line 74: why `RealmResultsImpl` doesn't have an `isValid()`?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,67 @@ internal interface InternalTypedRealm : TypedRealm {
...
+        } else {
+            // Else we need to just do the naive approach
+            val result = ArrayList<T>()
+            collection.forEach { obj: T ->
+                val copiedObj = copyObjectFromRealm(obj, depth, closeAfterCopy, cache)
+                result.add(copiedObj)
+            }
+            result
+        }.also {
+            // If
```
nhachicha, line 100: Shouldn't we release if the flag is set only (maybe why you left the comment above)?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -759,6 +761,141 @@ internal object RealmObjectHelper {
...
+        currentDepth: Int,
+        maxDepth: Int,
+        closeAfterCopy: Boolean,
+        cache: ManagedToUnmanagedObjectCache
+    ) {
+        val metadata: ClassMetadata = source.realmObjectReference!!.metadata
+        for (property in metadata.properties) {
+            val accessor = property.accessor
+                ?: sdkError("Typed object should always have an accessor")
+            if (property.isComputed) {
```
nhachicha, line 778: We don't populate backlinks? we already have the reference to the targets...

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -759,6 +761,141 @@ internal object RealmObjectHelper {
...
+                CollectionType.RLM_COLLECTION_TYPE_NONE -> when (property.type) {
+                    PropertyType.RLM_PROPERTY_TYPE_OBJECT -> {
+                        if (currentDepth == maxDepth) {
+                            accessor.set(target, null)
+                        } else {
+                            val realmObject: BaseRealmObject? = accessor.get(source) as BaseRealmObject?
+                            if (realmObject != null) {
+                                accessor.set(
+                                    target,
+                                    createDetachedCopy(
```
nhachicha, line 792: Can we leverage an implementation using `tailrec`?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -759,6 +761,141 @@ internal object RealmObjectHelper {
...
+                            }
+                            accessor.set(target, set)
+                        }
+                        else -> {
+                            throw IllegalStateException("Unknown type: ${property.type}")
+                        }
+                    }
+                }
+                else -> {
+                    TODO("Collection type ${property.collectionType} is not supported.")
```
nhachicha, line 893: Still a todo or an `IllegalStateException` ?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUtils.kt
```diff
@@ -17,19 +17,48 @@
...
 import kotlin.reflect.KProperty1
 
-internal typealias ObjectCache = MutableMap<BaseRealmObject, BaseRealmObject>
+// This cache is only valid for unmanaged realm objects as, for them we only consider the users
+// `equals` method, which in general just is the memory address of the object.
+internal typealias UnmanagedToManagedObjectCache = MutableMap<BaseRealmObject, BaseRealmObject> // Map<OriginalUnmanagedObject, CachedManagedObject>
+
+// For managed realm objects we use `<ClassKey, ObjectKey, Version>` as a unique identifier
+// We are using a hash on the Kotlin side so we can use a HashMap for O(1) lookup rather than
+// having to do 0(n) filter with a JNI call for `realm_equals` for each element.
```
nhachicha, line 49:
```suggestion
// having to do O(n) filter with a JNI call for `realm_equals` for each element.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUtils.kt
```diff
@@ -183,3 +212,93 @@ internal fun <T : BaseRealmObject> copyToRealm(
...
+    }
+}
+public fun <T : BaseRealm> RealmSet<*>.getRealm(): T? {
+    return when (this) {
+        is UnmanagedRealmSet -> null
+        is ManagedRealmSet -> {
+            return this.operator.realmReference.owner as T
+        }
+        else -> {
+            TODO("Unsupported set type: ${this::class}")
```
nhachicha, line 293: ditto

nhachicha (commented):
Very promising 🏅 a couple of minor ones.

I didn't go through the test yet :) but I wonder what happens if an object has a non-nullable link to another type, but we reached the depth of the deepcopy ? setting null on this link is not permitted since it's not nullable.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
cmelchior: This should be an IllegalStateException

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
cmelchior: Mostly because we don't support unmanaged versions of `LinkingObjects` it will throw an exception normally. So it would feel weird if it was allowed for objects that had been copied.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
cmelchior: Mostly because it isn't available in the C-API.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
cmelchior: The root is considered level 0, this should also be part of the Dokka (although It was a massive wall of text).

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/LongPointerWrapper.kt
```diff
@@ -26,14 +27,35 @@ import java.lang.Long.toHexString
 //  runtime-api mpp-module, which ruins IDE solving of the while type hierarchy around the
 //  pointers, which makes in annoying to work with.
 //  https://issuetracker.google.com/issues/174162078
-class LongPointerWrapper<T : CapiT>(val ptr: Long, managed: Boolean = true) : NativePointer<T> {
+public class LongPointerWrapper<T : CapiT>(ptr: Long, managed: Boolean = true) : NativePointer<T> {
+    // Tracks whether or not the `ptr` has already been released
+    internal val released = AtomicBoolean(false)
```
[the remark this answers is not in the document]
cmelchior: I was also afraid of the overhead, but the JVM benchmarks I posted above seem to indicate that the difference is neglible, i.e. it doesn't even show up in those benchmarks.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
cmelchior: This does not seem to work since the compiler cannot locate `isValid` in that case. I assume, because both are just normal instance methods on each class

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
cmelchior: I don't think so, judging from the document, the tailrec function must be the last method called inside the function and this function is called from multiple places inside here and from the public `copyFromRealm` method.

But neat idea actually ... I didn't really have `tailrec` on my radar.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,747 @@
...
+        assertFalse(RealmInterop.realm_equals(o1p, o2p))
+
+        val unmanagedObjects = realm.copyFromRealm(listOf(managedSample1, managedSample2))
+        assertEquals(2, unmanagedObjects.size)
+        val unmanagedSample1 = unmanagedObjects.first()
+        val unmanagedSample2 = unmanagedObjects.last()
+
+        assertNotSame(unmanagedSample1, unmanagedSample2)
+        assertNotEquals(unmanagedSample1, unmanagedSample2)
+        assertEquals(3, unmanagedSample1.intField)
```
rorbech, line 514: Maybe also assert that `stringField == "v1"` to highlight that the two copies originate from the same object.

rorbech (commented):
Nice work. 

I still haven't reviewed all the tests, but really looks like a lot of common infrastructure/functionality that should be abstracted on a `Referenced` and `NativeReferenceHolder` interfaces (or alike) that could give access to the `RealmReference` and the `NativePointer`. Then all methods like `getRealm`, `isValid`, etc. could be on those and not require five-fold implementations 🤯

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/TypedRealm.kt
```diff
@@ -25,4 +26,47 @@ public interface TypedRealm : BaseRealm {
...
+     * an object is closed, `RealmObject.isValid()` will return `false` and further access to it
+     * will throw an [IllegalStateException]. This can be beneficial as managed RealmObjects contain
+     * a reference to a chunck of native memory. This memory is normally freed when the object is
+     * garbage collected by Kotlin. However, manually closing the object allow Realm to free that
+     * memory immediately, allowing for better native memory management and control over the size
+     * of the Realm file.
+     * @returns a in-memory copy of the input object.
+     * @throws IllegalArgumentException if depth < 0 or [obj] is not a valid object to copy.
+     */
+    public fun <T : TypedRealmObject> copyFromRealm(obj: T, depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = true): T
```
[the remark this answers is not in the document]
cmelchior: After an internal discussion we ended up on keeping `copyFromRealm` but changed the default for closing from `true` to `false`.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -157,14 +158,33 @@ private fun <T : CPointed> checkedPointerResult(pointer: CPointer<T>?): CPointer
 // FIXME API-INTERNAL Consider making NativePointer/CPointerWrapper generic to enforce typing
 
 class CPointerWrapper<T : CapiT>(ptr: CPointer<out CPointed>?, managed: Boolean = true) : NativePointer<T> {
-    val ptr: CPointer<out CPointed>? = checkedPointerResult(ptr)
+    private val released: AtomicBoolean = atomic(false)
```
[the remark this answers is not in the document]
cmelchior: My understanding is no. Once you delete a pointer, access to it afterward is undefined behaviour, so Core has no way of tracking this.

packages/plugin-compiler/src/test/resources/sample/expected/00_ValidateIrBeforeLowering.ir
```diff
@@ -949,7 +949,7 @@ MODULE_FRAGMENT name:<main>
...
-                  then: CALL 'internal final fun setObject (obj: io.realm.kotlin.internal.RealmObjectReference<out io.realm.kotlin.types.BaseRealmObject>, propertyName: kotlin.String, value: io.realm.kotlin.types.BaseRealmObject?, updatePolicy: io.realm.kotlin.UpdatePolicy, cache: kotlin.collections.MutableMap<io.realm.kotlin.types.BaseRealmObject, io.realm.kotlin.types.BaseRealmObject>{ io.realm.kotlin.internal.RealmUtilsKt.ObjectCache }): kotlin.Unit [inline] declared in io.realm.kotlin.internal.RealmObjectHelper' type=kotlin.Unit origin=GET_PROPERTY
+                  then: CALL 'internal final fun setObject (obj: io.realm.kotlin.internal.RealmObjectReference<out io.realm.kotlin.types.BaseRealmObject>, propertyName: kotlin.String, value: io.realm.kotlin.types.BaseRealmObject?, updatePolicy: io.realm.kotlin.UpdatePolicy, cache: kotlin.collections.MutableMap<io.realm.kotlin.types.BaseRealmObject, io.realm.kotlin.types.BaseRealmObject>{ io.realm.kotlin.internal.RealmUtilsKt.UnmanagedToManagedObjectCache }): kotlin.Unit [inline] declared in io.realm.kotlin.internal.RealmObjectHelper' type=kotlin.Unit origin=GET_PROPERTY
```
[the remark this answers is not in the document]
cmelchior: Yes, that was also something that concerned me. However, I don't think we ever promised binary compatibility, so I would not consider it a breaking change.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/TypedRealm.kt
```diff
@@ -25,4 +26,47 @@ public interface TypedRealm : BaseRealm {
...
+     * is `true`). If an object is closed, `RealmObject.isValid()` will return `false` and further
+     * access to it will throw an [IllegalStateException]. This can be beneficial as managed
+     * RealmObjects contain a reference to a chunck of native memory. This memory is normally freed
+     * when the object is garbage collected by Kotlin. However, manually closing the object allow
+     * Realm to free that memory immediately, allowing for better native memory management and
+     * control over the size of the Realm file.
+     * @returns an in-memory copy of the input object.
+     * @throws IllegalArgumentException if depth < 0 or [obj] is not a valid object to copy.
+     */
+    public fun <T : TypedRealmObject> copyFromRealm(obj: T, depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = true): T
```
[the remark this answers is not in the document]
cmelchior: Forgot it was moved from Experimental to Stable in 1.5, yes that is much better 👍

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/TypedRealmObjectExt.kt
```diff
@@ -13,16 +13,41 @@
...
+ * an object is closed, `RealmObject.isValid()` will return `false` and further access to it
+ * will throw an [IllegalStateException]. This can be beneficial as managed RealmObjects contain
+ * a reference to a chunck of native memory. This memory is normally freed when the object is
+ * garbage collected by Kotlin. However, manually closing the object allow Realm to free that
+ * memory immediately, allowing for better native memory management and control over the size
+ * of the Realm file.
+ * @returns an in-memory copy of the input object.
+ * @throws IllegalArgumentException if depth < 0 or the object  is not a valid object to copy.
+ */
+public inline fun <reified T : RealmObject> T.copyFromRealm(depth: Int = Int.MAX_VALUE, closeAfterCopy: Boolean = true): T {
```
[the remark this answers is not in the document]
cmelchior: Forgot to change the type here. We don't support `copyFromRealm` on dynamic objects just yet, so it should be in this file.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUtils.kt
```diff
@@ -183,3 +212,93 @@ internal fun <T : BaseRealmObject> copyToRealm(
...
+ * advanced or been closed.
+ *
+ * Given that this method can be called from multiple places, e.g. inside and outside write
+ * transactions, the given Realm type must be provided by the caller as a generic argument.
+ *
+ * If a wrong type is provided a `ClassCastException` is thrown.
+ *
+ * If the object is unmanaged, `null` is returned. Error handling is left up to the caller.
+ */
+public fun <T : BaseRealm> RealmObject.getRealm(): T? {
```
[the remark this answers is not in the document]
cmelchior: I agree there is room for improvement here, but I'm not 100% sure if there is an easy fix that doesn't spiral this PR out of control. I linked to https://github.com/realm/realm-kotlin/issues/582 in the method description which I think will also cover refactoring this into something cleaner?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -759,6 +761,141 @@ internal object RealmObjectHelper {
...
+                                }
+                            }
+                            accessor.set(target, list)
+                        }
+                        else -> {
+                            throw IllegalStateException("Unknown type: ${property.type}")
+                        }
+                    }
+                }
+                CollectionType.RLM_COLLECTION_TYPE_SET -> {
```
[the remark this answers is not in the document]
cmelchior: I thought about this as well, but there are enough differences that I think it is going to just make the code pretty unreadable, and there are no current plans to add more collection types.

Currently, the code differs in 3 locations... 1) converting primitive elements to either list/set 2) creating unmanaged list/set and 3) casting the type. This could be extracted into an interface and delegated to an implementation, but IMO the current approach is more readable, even if it results in a bit of duplicated code.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,747 @@
...
+        assertFailsWith<IllegalStateException> {
+            unmanagedCopy.linkingObject // Empty RealmResults
+        }
+        assertFailsWith<IllegalStateException> {
+            unmanagedCopy.nullableObject!!.linkingObject // Have 1 backlink
+        }
+    }
+
+    @Test
+    fun managedObjectClosedAfterCopy() {
```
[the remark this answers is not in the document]
cmelchior: Good point. A test around this shows that we might be holding on to references in places we shouldn't. I'm still investigating.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,65 @@ internal interface InternalTypedRealm : TypedRealm {
...
+            return createDetachedCopy(mediator, obj, 0.toUInt(), depth, closeAfterCopy, cache)
+        } else {
+            throw MISSING_PLUGIN
+        }
+    }
+
+    override fun <T : TypedRealmObject> copyFromRealm(obj: T, depth: UInt, closeAfterCopy: Boolean): T {
+        return copyObjectFromRealm(obj, depth, closeAfterCopy, mutableMapOf())
+    }
+    override fun <T : TypedRealmObject> copyFromRealm(collection: Iterable<T>, depth: UInt, closeAfterCopy: Boolean): List<T> {
```
clementetb, line 67:
```suggestion

    override fun <T : TypedRealmObject> copyFromRealm(collection: Iterable<T>, depth: UInt, closeAfterCopy: Boolean): List<T> {
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealm.kt
```diff
@@ -61,7 +61,8 @@ internal abstract class LiveRealm(val owner: RealmImpl, configuration: InternalC
         get() {
             // Initialize a new snapshot that can be reused until cleared again from onRealmChanged
             if (_snapshot.value == null) {
-                val snapshot = realmReference.snapshot(owner)
+                println("Create snapshot from owner")
```
clementetb, line 64: We could log this messages using `realm.log`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableWriter.kt
```diff
@@ -92,6 +92,7 @@ internal class SuspendableWriter(private val owner: RealmImpl, val dispatcher: C
                 // - onRealmChanged - updating the realm.snapshot to also point to the latest key cache
                 // Seems like order is not guaranteed, but it is synchroneous, so updating snapshot
                 // in both callbacks should ensure that we have the right snapshot here
+                println("Returning snapshot")
```
clementetb, line 95: same

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,65 @@ internal interface InternalTypedRealm : TypedRealm {
...
+
+    private fun <T : BaseRealmObject> copyObjectFromRealm(obj: T, depth: UInt, closeAfterCopy: Boolean, cache: ManagedToUnmanagedObjectCache): T {
+        // Be able to inject a cache here as well, so the Iterable case can share the cache
+        if (!obj.isManaged()) {
+            throw IllegalArgumentException("This object is unmanaged. Only managed objects can be copied: $obj.")
+        }
+        if (!obj.isValid()) {
+            throw IllegalArgumentException(
+                "Only valid objects can be copied from Realm. " +
+                    "This object was either deleted or the Realm has been closed, making this " +
```
clementetb, line 50:
```suggestion
                    "This object was either deleted or closed, or its Realm has been closed, making this " +
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -34,4 +38,65 @@ internal interface InternalTypedRealm : TypedRealm {
...
+        val (valid: Boolean, nativePointer: NativePointer<*>?) = when (collection) {
+            is ManagedRealmList -> Pair(collection.isValid(), collection.nativePointer)
+            is ManagedRealmSet -> Pair(collection.isValid(), collection.nativePointer)
+            is RealmResultsImpl -> Pair(!collection.realm.isClosed(), collection.nativePointer)
+            else -> Pair(true, null)
+        }
+        if (!valid) {
+            throw IllegalArgumentException(
+                "Only valid collections can be copied from Realm. " +
+                    "This collection was either deleted or the Realm has been closed, making this " +
```
clementetb, line 77:
```suggestion
                    "This collection was either deleted or closed, or its Realm has been closed, making this " +
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableWriter.kt
```diff
@@ -125,6 +126,7 @@ internal class SuspendableWriter(private val owner: RealmImpl, val dispatcher: C
             // the transaction is committed and we freeze it.
             // TODO Can we guarantee the Dispatcher is single-threaded? Or otherwise
             //  lock this code?
+            println("Creating snapshot")
```
clementetb, line 129: same

clementetb (commented):
Looking good 💎

My only concern is that with this PR we are introducing the manual `close` object feature, is this the solution that we like to avoid version pinning? There are other objects (queries, or collections) with a `RealmReference` that prevents releasing the instance. Having a half-backed solution might confuse our users that will see that pinned versions even if they choose to close the objects.

By the way, I see that `linkingObjects` has not been renamed to `backlinks`, I guess that we still need to merge `master`?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+
+    @Test
+    fun realmObjectReferences() {
+        val innerSample = Sample().apply { stringField = "inner" }
+
+        val insertedObj = realm.writeBlocking {
+            copyToRealm(Sample().apply { nullableObject = innerSample })
+        }
+        val unmanagedObj: Sample = insertedObj.copyFromRealm()
+        assertNotSame(insertedObj, unmanagedObj)
```
rorbech, line 126: Should we close the realm as in the other tests?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+            }
+        }
+
+        // Round-trip object through `copyToRealm` and `copyFromRealm`.
+        val unmanagedCopy = realm.writeBlocking {
+            copyToRealm(originalObject).copyFromRealm()
+        }
+
+        // Close Realm to ensure data is decoupled from Realm
+        realm.close()
```
rorbech, line 99: Maybe just assert that touching the `originalObject` is in fact throwing to highlight that we are not just linking back to the original 🤪

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+        schemaProperties.forEach { prop: RealmProperty ->
+            if (prop.type is SetPropertyType) {
+                val accessor: KMutableProperty1<BaseRealmObject, Any?> = fields[prop.name] as KMutableProperty1<BaseRealmObject, Any?>
+                val set: Set<Any?> = createPrimitiveSetData(prop, accessor)
+
+                if (prop.type.storageType == RealmStorageType.BINARY) {
+                    val copy = accessor.get(unmanagedCopy) as Set<ByteArray?>
+                    assertEquals(set.size, copy.size)
+                    copy.forEach { copiedValue: ByteArray? ->
+                        // Order is not guaranteed in the set when round-tripped through Core.
```
rorbech, line 285: You can just assert that you can remove each element and assert that the set is empty afterwards

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+        val copiedSet: RealmSet<Sample> = unmanagedObj.objectSetField
+        assertEquals(5, copiedSet.size)
+        copiedSet.forEach { copiedObject ->
+            assertEquals(
+                1,
+                sample.objectSetField.filter {
+                    copiedObject.stringField == it.stringField
+                }.size
+            )
+        }
```
rorbech, line 340: Again just remove with a predicate.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+
+        // Close Realm to ensure data is decoupled from Realm
+        realm.close()
+
+        assertEquals(1, unmanagedCopy.size)
+        assertEquals("sample", unmanagedCopy.first().stringField)
+    }
+
+    @Test
+    fun invalidObject_throws() {
```
rorbech, line 362: I would separate the test for delete vs. close objects to make it simpler to reason about

rorbech: Or maybe group all the _deleted_ tests together and all the _closed_ tests together ... but I guess it is just a matter of which axises that feels more natural to you 🤪

rorbech: Also lacking tests for `copyFromRealm` on closed results.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+            copyToRealm(sample).copyFromRealm()
+        }
+
+        assertSame(unmanagedCopy, unmanagedCopy.nullableObject)
+        assertSame(unmanagedCopy, unmanagedCopy.objectListField.first())
+        assertSame(unmanagedCopy, unmanagedCopy.objectSetField.first())
+    }
+
+    @Test
+    fun objectGraphShareMemRefForSharedObjects() {
```
rorbech, line 449: How is this different from `circularObjectGraph()`?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+        }
+        assertEquals("obj-depth-2", managedObj.nullableObject!!.nullableObject!!.stringField)
+        assertEquals("list-depth-2", managedObj.objectListField.first().objectListField.first().stringField)
+        assertEquals("set-depth-2", managedObj.objectSetField.first().objectSetField.first().stringField)
+
+        val unmanagedCopy = managedObj.copyFromRealm(depth = 1u)
+        assertEquals("obj-depth-1", unmanagedCopy.nullableObject!!.stringField)
+        assertEquals("list-depth-1", unmanagedCopy.objectListField.first().stringField)
+        assertEquals("set-depth-1", unmanagedCopy.objectSetField.first().stringField)
+        assertNull(unmanagedCopy.nullableObject!!.nullableObject)
```
rorbech, line 550: I guess this would also be my standard expectation, but seeing it against the copied string fields makes me wonder ... 🤔 ... but until we could express an _unset_ property or something like that it is maybe better just to `null` it 😅

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+        assertEquals(0, unmanagedCopy.objectListField.first().objectListField.size)
+        assertEquals(0, unmanagedCopy.objectSetField.first().objectSetField.size)
+    }
+
+    @Test
+    fun depth_primitiveListsAndSetsWhenDepthIsReached() {
```
rorbech, line 556: Seems a bit weird to have this in a separate test instead of just having it the above `depth_nullAfterDepthIsReached`

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+        sample.copyFromRealm(closeAfterCopy = true)
+        assertFalse(sample.isValid())
+        assertFailsWithMessage<IllegalStateException>("Cannot perform this operation on an invalid/deleted object") {
+            sample.stringField
+        }
+    }
+
+    // Verify that closing an object also release the underlying version
+    @Test
+    fun closedObjectsFreeVersion() = runBlocking {
```
rorbech, line 616: Playing a bit around with this I guess we are not really freeing the version of the source to `copyFromRealm` until they managed objects are GC'ed. They store a reference to the `RealmReference` and as this is tracked in the master `Realm` then it is not cleaned up before all objects are GC'ed too. Thus, this connection renders the `closeAfterCopy` useless, so I think we should remove that until we can guarantee immediate effect.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
```diff
@@ -0,0 +1,755 @@
...
+        val unmanagedObj: Sample = insertedObj.copyFromRealm()
+        assertNotSame(insertedObj, unmanagedObj)
+        assertNotNull(unmanagedObj.nullableObject)
+        val innerCopy = unmanagedObj.nullableObject!!
+        assertFalse(innerCopy.isManaged())
+        assertEquals("inner", innerCopy.stringField)
+    }
+
+    @Test
+    fun embeddedObjectReferences() {
```
rorbech, line 134: Maybe also a test of `embeddedObject.copyFromRealm()`, since you can actually copy it out without a parent. I can't see why it should work, but would be nice to have the test to catch regressions if changing the implementation.

rorbech (changes requested):
Completed the review of the tests after checking it out and playing with especially the memory management. 

The current implementation will not close the underlying versions in core just by using `closeAfterCopy`. There are still valid references to the `RealmReference` and as long as these are alive the underlying version is not released.

My immediate thought is that still could be useful if you want to hold on to a detached version of an object, but have a hard time to evaluate if this use case is actually already sufficiently addressed by the "intermediate version reclaimer" mentioned in #1007. If we want to keep the `copyFromRealm` then we should at least eliminate the `closeAfterCopy` as it gives false impressions. 

Besides that, I only have some minor comments on the tests and I am fine with pushing the alignment and abstractions of the of internal methods to a later PR (maybe until we rework how `RealmReference`s are tracked so we can close the versions immediately 😉)

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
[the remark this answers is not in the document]
cmelchior: After testing this, we came to the conclusion that closing the object pointers doesn't really help releasing versions faster, so this option has been removed from the public API (but retained in the internal API while we iterate on solutions).

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealm.kt
cmelchior: This was debug output. Removing it.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
cmelchior: Not sure it would make things easier? The problem with ByteArray also manifests on `remove` as `HashSet` would use the equals implementation which doesn't work correctly.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
cmelchior: Not sure what you mean, since this is copied objects, using `RealmObject.equals` is not going to work?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
rorbech: Maybe I am just a bit confused about the logic and or comment. I think it would be more clear with (if I understand it correctly):
```
copy.forEach { copiedValue ->
    assertTrue {
        set.removeIf {
            copiedValue == null || (it as ByteArray).contentEquals(copiedValue)
        }
    }
}
assertTrue { set.isEmpty() }
```
Maybe it is just the huge comment that confuses me 🤪

rorbech: I think I would just have spun an implementation around `removeIf` and verify that we removed all elements that matched a predicate from the `copiedValue`, but I guess it maybe becomes more tricky to do as `removeIf` could remove multiple elements in one go. Maybe I was also just confused about the bloated comment 🤣

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CopyFromRealmTests.kt
rorbech: Same as https://github.com/realm/realm-kotlin/pull/1092#discussion_r1022559921, but probably just safer just to leave as is.

cmelchior:
@rorbech I think I addressed all your comments

rorbech (approved):
LGTM except for this itchy thing about polluting the public API with these replicated extension methods (similar to the concerns I raised about the internal infrastructure). I guess it will blow up this PR and could probably spiral into too many discussions/decisions, so maybe not now ... but I think we should consider approaching it sooner than later as it will impact the API and maintenance burden.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/RealmListExt.kt
```diff
@@ -16,12 +16,29 @@
...
+/**
+ * Makes an unmanaged in-memory copy of the elements in a managed [RealmList]. This is a deep copy
+ * that will copy all referenced objects.
+ *
+ * @param depth limit of the deep copy. All object references after this depth will be `null`.
+ * [RealmList]s and [RealmSet]s containing objects will be empty. Starting depth is 0.
+ * @returns an in-memory copy of all input objects.
+ * @throws IllegalArgumentException if depth < 0 or, or the list is not valid to copy.
+ */
+public inline fun <reified T : TypedRealmObject> RealmList<T>.copyFromRealm(depth: UInt = UInt.MAX_VALUE): List<T> {
```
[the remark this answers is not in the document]
cmelchior: I added a comment to #582 about this

PR #1095, opened 2022-10-28
Fix using object references as query argument

### LINKED ISSUE #1098: Query by object reference is throwing
Performing a query with an object reference like:
```
val child = realm.query<QuerySample>().find().first()
realm.query<QuerySample>("nullableRealmObject = $0", child)
```
will throw an 
```
kotlin.NotImplementedError: An operation is not implemented: Unsupported type for capiRealmValue `QuerySample`
```

rorbech:
Closes #1098

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,29 @@
+## 1.4.1 (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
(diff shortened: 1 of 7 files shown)


edualonso (approved):
Looks great 👍  I don't foresee big challenges adding this fix in the `realm_value_t` branch 😄

PR #1099, opened 2022-10-31
Expose configuration and user on SyncSession.

### LINKED ISSUE #431: SyncSession - Utility methods
Expose the various utility methods we also have in Realm Java:

```java
interface SyncSession {
  val configuration: SyncConfiguration
  val user: User
  val serverUrl: String
}
}
```

elle-j:
## Description

Closes #431.

Makes the `configuration` and `user` available on `SyncSession`.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
```diff
@@ -73,6 +74,21 @@ public interface SyncSession {
...
+
+    /**
+     * The [User] defined by the [SyncConfiguration] to authenticate the session on Atlas App Services.
+     */
+    public val user: User
+
+    /**
+     * The URL describing the remote Realm that this session connects to and synchronizes changes with.
+     */
+    public val serverUrl: String
```
nirinchev, line 90: Not sure we should be exposing this - for pretty much all users, this will be `realm.mongodb.com` and the internal users that absolutely need to look it up can go through the user.

cmelchior: Good point. This is probably a leftover from the pre-v10 API's. But you are correct, there is no reason we need to expose this in Kotlin. I just forgot to think more closely about it when creating this issue.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Added support for reverse relationships through the `linkingObjects` delegate. See the function documentation for more details. (Issue [#1021](https://github.com/realm/realm-kotlin/pull/1021))
 * Added support for `BsonObjectId` and its typealias `org.mongodb.kbson.ObjectId` as a replacement for `ObjectId`. `io.realm.kotlin.types.ObjectId` is still functional but has been marked as deprecated.
 * [Sync] Added support for `BsonObjectId` as partition value.
+* [Sync] Exposed `configuration`, `user`, and `serverUrl` on `SyncSession`. (Issue [#431](https://github.com/realm/realm-kotlin/issues/431))
```
cmelchior, line 12: 🥇

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -496,6 +496,30 @@ class SyncSessionTests {
         }
     }
 
+    @Test
+    fun session_getConfiguration() {
```
cmelchior, line 500: We should also test this inside the error handler so we capture the exception

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
```diff
@@ -73,6 +74,21 @@ public interface SyncSession {
      */
     public val state: State
 
+    /**
+     * The [SyncConfiguration] responsible for controlling the session.
+     */
+    public val configuration: SyncConfiguration
+
+    /**
+     * The [User] defined by the [SyncConfiguration] to authenticate the session on Atlas App Services.
```
cmelchior, line 83:
```suggestion
     * The [User] used to authenticate the session on Atlas App Services.
```

I would reword this slightly to just include the most important bits.

cmelchior (changes requested):
Great start 💯... There is a few minor comments + we should also find a way to test this inside the SyncConfiguration error handler.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
elle-j: Ah okay! I'll revert that one 👍

nhachicha: The only use case I can think of is testing, If we spin up a local BAAS?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
```diff
@@ -73,6 +74,16 @@ public interface SyncSession {
      */
     public val state: State
 
+    /**
+     * The [SyncConfiguration] responsible for controlling the session.
```
clementetb, line 78: We could use a comment explaining in what conditions will it throw.

clementetb (approved):
Looking good to me, added a minor comment 🚀.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -496,6 +499,69 @@ class SyncSessionTests {
...
+            schema = setOf(ChildPk::class),
+            user = user,
+            partitionValue = partitionValue
+        )
+            .name("test1.realm")
+            .build()
+        val realm1 = Realm.open(config1)
+        assertNotNull(realm1)
+
+        // Open another realm with the same entity but change the type of a field
```
cmelchior, line 528: Good way of testing this 👍

cmelchior (approved):
💯 👍

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
elle-j: Suggestion:
```kt
/**
 * The [SyncConfiguration] responsible for controlling the session.
 * 
 * @throws IllegalStateException if accessed from inside a [SyncSession.ErrorHandler] due to session errors.
 */
```

PR #1100, opened 2022-11-01
Add support for encrypting user metadata used by Sync.

### LINKED ISSUE #413: AppConfiguration - encryptionKey
See [https://github.com/realm/realm-kotlin/issues/406](https://github.com/realm/realm-kotlin/issues/406)

Similar to Realm Java we need to expose `AppConfiguration.Builder.encryptionKey()` which makes it possible to encrypt the metadata Realm used by Sync.

elle-j:
## Description

Closes #413.

Users can encrypt user metadata via `AppConfiguration.Builder.encryptionKey()` and access their encryption key via `AppConfiguration.encryptionKey`.

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
@@ -52,8 +52,8 @@ public interface AppConfiguration {
     //  requires ktor as api dependency
     public val baseUrl: String
     public val encryptionKey: ByteArray?
-    public val networkTransport: NetworkTransport
     public val metadataMode: MetadataMode
+    public val networkTransport: NetworkTransport
     public val syncRootDirectory: String
 
     public companion object {
@@ -103,8 +103,8 @@ public interface AppConfiguration {
         private var userLoggers: List<RealmLogger> = listOf()
 
         /**
-         * Sets the encryption key used to encrypt user metadata only. Individual Realms need to
-         * use [SyncConfiguration.Builder.encryptionKey] to encrypt them.
+         * Sets the encryption key used to encrypt the user metadata Realm only. Individual
+         * Realms need to use [SyncConfiguration.Builder.encryptionKey] to encrypt them.
          *
          * @param key a 64 byte encryption key.
          * @return the Builder instance used.
@@ -228,6 +228,9 @@ public interface AppConfiguration {
                 appId = appId,
                 baseUrl = baseUrl,
                 encryptionKey = encryptionKey,
+                metadataMode = if (encryptionKey == null)
+                    MetadataMode.RLM_SYNC_CLIENT_METADATA_MODE_PLAINTEXT
+                    else MetadataMode.RLM_SYNC_CLIENT_METADATA_MODE_ENCRYPTED,
                 networkTransport = networkTransport,
                 syncRootDirectory = syncRootDirectory,
                 log = appLogger
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
@@ -36,6 +36,7 @@ import io.realm.kotlin.log.LogLevel
 import io.realm.kotlin.log.RealmLogger
 import io.realm.kotlin.mongodb.internal.AppConfigurationImpl
 import io.realm.kotlin.mongodb.internal.KtorNetworkTransport
+import io.realm.kotlin.mongodb.sync.SyncConfiguration
 import kotlinx.coroutines.CoroutineDispatcher
 
 /**
@@ -50,6 +51,7 @@ public interface AppConfiguration {
     // TODO Consider replacing with URL type, but didn't want to include io.ktor.http.Url as it
     //  requires ktor as api dependency
     public val baseUrl: String
+    public val encryptionKey: ByteArray?
     public val networkTransport: NetworkTransport
     public val metadataMode: MetadataMode
     public val syncRootDirectory: String
@@ -94,11 +96,27 @@ public interface AppConfiguration {
         // TODO We should use a multi threaded dispatcher
         //  https://github.com/realm/realm-kotlin/issues/501
         private var dispatcher: CoroutineDispatcher = singleThreadDispatcher("dispatcher-$appId")
-
+        private var encryptionKey: ByteArray? = null
         private var logLevel: LogLevel = LogLevel.WARN
         private var removeSystemLogger: Boolean = false
         private var syncRootDirectory: String = appFilesDirectory()
         private var userLoggers: List<RealmLogger> = listOf()
+
+        /**
+         * Sets the encryption key used to encrypt user metadata only. Individual Realms need to
+         * use [SyncConfiguration.Builder.encryptionKey] to encrypt them.
+         *
+         * @param key a 64 byte encryption key.
+         * @return the Builder instance used.
+         * @throws IllegalArgumentException if the key is not 64 bytes long.
+         */
+        public fun encryptionKey(key: ByteArray): Builder = apply {
+            if (key.size != Realm.ENCRYPTION_KEY_LENGTH) {
+                throw IllegalArgumentException("The provided key must be ${Realm.ENCRYPTION_KEY_LENGTH} bytes. Yours was: ${key.size}.")
+            }
+
+            this.encryptionKey = key.copyOf()
+        }
 
         /**
          * Sets the base url for the App Services Application. The default value is
@@ -209,6 +227,7 @@ public interface AppConfiguration {
             return AppConfigurationImpl(
                 appId = appId,
                 baseUrl = baseUrl,
+                encryptionKey = encryptionKey,
                 networkTransport = networkTransport,
                 syncRootDirectory = syncRootDirectory,
                 log = appLogger
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
@@ -230,7 +230,7 @@ public interface AppConfiguration {
                 encryptionKey = encryptionKey,
                 metadataMode = if (encryptionKey == null)
                     MetadataMode.RLM_SYNC_CLIENT_METADATA_MODE_PLAINTEXT
-                    else MetadataMode.RLM_SYNC_CLIENT_METADATA_MODE_ENCRYPTED,
+                else MetadataMode.RLM_SYNC_CLIENT_METADATA_MODE_ENCRYPTED,
                 networkTransport = networkTransport,
                 syncRootDirectory = syncRootDirectory,
                 log = appLogger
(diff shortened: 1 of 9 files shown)


packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppConfigurationTests.kt
```diff
@@ -394,6 +397,37 @@ class AppConfigurationTests {
...
+        val key = TestHelper.getRandomKey()
+        val config = AppConfiguration.Builder(APP_ID)
+            .encryptionKey(key)
+            .build()
+
+        assertNotSame(key, config.encryptionKey)
+    }
+
+    @Test
+    fun encryptionKey_wrongLength() {
```
cmelchior, line 421: We are missing tests that actually verify that the user data Realm has been encrypted. That can be achieved by creating a local `RealmConfiguration` that points to it and then verify that you need the encryption key to open it.

See e.g. the similar tests in Realm Java here https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/AppTests.kt#L273

That test uses a DynamicRealm, in Kotlin I think you can get away with just using an empty schema..

cmelchior (changes requested):
Really nice. Only thing missing seem to be a test verifying that we actually encrypt the metadata realm.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppConfigurationTests.kt
cmelchior: The metadata realm should be located in `<syncRoot>/<appId>/metadata/sync_metadata.realm`

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppConfigurationImpl.kt
```diff
@@ -37,6 +37,7 @@ import io.realm.kotlin.mongodb.AppConfiguration.Companion.DEFAULT_BASE_URL
 public class AppConfigurationImpl constructor(
```
cmelchior, line 37: This class now fails static analysis, but you can just `@Suppress("LongParameterList")` the warning. It is generally not a warning we are that concerned with fixing. At least not in the internal API.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppConfigurationTests.kt
elle-j: Thanks Christian! 👍

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
```diff
@@ -94,12 +96,28 @@ public interface AppConfiguration {
...
         private var removeSystemLogger: Boolean = false
         private var syncRootDirectory: String = appFilesDirectory()
         private var userLoggers: List<RealmLogger> = listOf()
 
+        /**
+         * Sets the encryption key used to encrypt the user metadata Realm only. Individual
+         * Realms need to use [SyncConfiguration.Builder.encryptionKey] to encrypt them.
+         *
+         * @param key a 64 byte encryption key.
+         * @return the Builder instance used.
```
elle-j, line 110: I added `@return` to the docs; however, the other methods using `apply` do not have this explicitly in the docs. What are your thoughts on this?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
```diff
@@ -292,42 +302,41 @@ class AppTests {
...
+            val syncConfig = SyncConfiguration
+                .Builder(user, setOf(ParentPk::class, ChildPk::class))
+                .build()
+            Realm.open(syncConfig).close()
+
+            // Create a configuration pointing to the metadata Realm for that app
+            val config = RealmConfiguration
+                .Builder(setOf())
+                .name(SYNC_METADATA_REALM_NAME)
+                .directory(getMetadataDirectory(app))
```
elle-j, line 329: For now, `SYNC_METADATA_REALM_NAME` and `getMetadataDirectory()` are only used by this test. I've seen many hardcoded and repeated values elsewhere, but what are your thoughts on extracting these two?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
cmelchior: Yeah, I noticed that we missed adding these before. We should probably add it to all the other ones as well. But this is fine 👍

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
cmelchior: When they are just used in one place, I would keep them inlined here. If we ever need to use them from multiple places, we can look into extracting it then. But it is not a big deal.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
```diff
@@ -292,42 +302,42 @@ class AppTests {
...
+                .build()
+            assertTrue(File(config.path).exists())
+
+            // Open the metadata realm file without a valid encryption key
```
cmelchior, line 334: You should also test that you _can_ open it with the encryption key.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
```diff
@@ -292,42 +302,42 @@ class AppTests {
...
-//            assertFailsWith<RealmFileException> {
-//                DynamicRealm.getInstance(config)
-//            }
-//        } finally {
-//            testApp.close()
-//        }
-//    }
+
+    @Test
+    fun encryption() {
```
cmelchior, line 307: 🥳

cmelchior (approved):
Looks good. I only have one smaller comment around the test and it needs to parse CI, but otherwise this looks ready to merge 👍

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
```diff
@@ -292,42 +301,122 @@ class AppTests {
...
+            Realm.open(syncConfig).close()
+
+            // Create a configuration pointing to the metadata Realm for that app
+            val metadataDir = "${app.configuration.syncRootDirectory}/mongodb-realm/${app.configuration.appId}/server-utility/metadata/"
+            val config = RealmConfiguration
+                .Builder(setOf())
+                .name("sync_metadata.realm")
+                .directory(metadataDir)
+                .encryptionKey(key)
+                .deleteRealmIfMigrationNeeded()
```
elle-j, line 333: Without this there will be an error `Caused by: io.realm.kotlin.internal.interop.RealmCoreLogicException: [18]: Provided schema version 0 is less than last set version 6.`

cmelchior: Hmm, that line actually invalidates the test, because the Realm gets deleted now.

Does it work if you set `RealmConfiguration.schemaVersion(6)` instead?

elle-j: But doesn't it recreate the realm with the new schema? I'm also checking `assertTrue(fileExists(config.path))` which is the path to the `sync_metadata.realm` file.

Yes schemaVersion(6) works, but would be nice to not hard code 6 if that was to change?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
```diff
@@ -292,42 +301,122 @@ class AppTests {
...
+            val metadataDir = "${app.configuration.syncRootDirectory}/mongodb-realm/${app.configuration.appId}/server-utility/metadata/"
+            val config = RealmConfiguration
+                .Builder(setOf())
+                .name("sync_metadata.realm")
+                .directory(metadataDir)
+                .build()
+            assertTrue(fileExists(config.path))
+
+            // Open the metadata realm file without a valid encryption key
+            assertFailsWith<IllegalArgumentException> { Realm.open(config) }
```
nhachicha, line 414: Maybe assert the exception message to be sure it's the expected encryption exception and not another unrelated throw

elle-j: Agreed, thanks :)

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppConfigurationImpl.kt
```diff
@@ -34,11 +34,13 @@ import io.realm.kotlin.mongodb.AppConfiguration
 import io.realm.kotlin.mongodb.AppConfiguration.Companion.DEFAULT_BASE_URL
 
 // TODO Public due to being a transitive dependency to AppImpl
+@Suppress("LongParameterList")
 public class AppConfigurationImpl constructor(
     override val appId: String,
     override val baseUrl: String = DEFAULT_BASE_URL,
+    override val encryptionKey: ByteArray?,
```
clementetb, line 41: Should we keep a copy of the encryption key at SDK level? I think that we should not keep any copy, as they should be safely stored by the user or core.

`RealmConfiguration` also stores a copy. For that case we have a core function to retrieve the key, that would help testing, but we lack of such function for the metadata encryption key.

We could address this in another PR.

clementetb (approved):
Looks great 🚀

cmelchior (approved):
Nice 🚀

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppConfigurationImpl.kt
cmelchior: I agree, this is a bit of a smell, but I wouldn't consider it a big deal, since if you have access to read the memory where this key resides, you also have access to the open Realm and can just ask Core for the data.

elle-j: Interesting points. Currently, however, the exposed core function `realm_config_get_encryption_key` (not for metadata) is not used in any configuration class. The `actual` functions are implemented but yes, a copy is still being stored.

PR #1101, opened 2022-11-02
Update `@Index` annotation docs to reflect new supported types

edualonso:


cmelchior (approved):
Looks good. 👍 once the build is green (I'm not sure I understand why it would fail here)

edualonso:
Oh, I know why it's failing. I think the bson library is only available in master, not in releases yet 🤦 I will create a PR against master since we will make a minor release soon anyway.

PR #1103, opened 2022-11-03
Update ktlint

cmelchior:
Update to latest version of ktlint.

I copied in the new default rules file .There has been some changes to some rules, which prompted a few changes here and there.


nhachicha (approved):
Looks good! wonder why they're still on an RC version 🤷‍♂️

PR #1108, opened 2022-11-04
Clarify and add instructions to set-up files.

elle-j:
Makes set-up files a bit more detailed in order to improve and accelerate the onboarding for new members.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,9 +92,8 @@ the following [Advanced Project Setup](#advanced-project-setup)-section
 
 The overall setup of the project is done to support simultaneous development of the SDK and test,
 while still selectively allowing to run tests against Maven artifacts. This is why the tests are
-separated into separate Gradle modules. Due to issues with IntelliJ/Android Studio not being able to
-resolve symbols in Kotlin Multiplatform projects in Composite Gradle
-projects (https://youtrack.jetbrains.com/issue/KTIJ-15775/MPP-IDE-Lots-of-red-code-unresolved-references-with-HMPP-and-composite-build)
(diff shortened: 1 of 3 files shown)


tools/sync_test_server/start_local_server.sh
```diff
@@ -25,6 +36,12 @@ function echo_step () {
 }
 
 function check_dependencies () {
+  if  ! which -s node; then
+    echo "Error: NodeJS not found. Install using 'brew install node'" && exit 1
+  fi
+  if ! which -s yarn; then
+    echo "Error: Yarn not found. Install using 'brew install yarn'" && exit 1
+  fi
```
elle-j, line 44: Moved this before the check for `realm-cli --version` since that requires npm if that check fails.

CONTRIBUTING.md
```diff
@@ -152,21 +177,22 @@ We have three branches for shared development: `master`, `releases` and `next-ma
 We use the offical [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html) from Kotlin which is enforced using [ktlint](https://github.com/pinterest/ktlint) and [detekt](https://github.com/detekt/detekt).
 
 ```sh
-# Call from root folder to check if code is compliant.
+# Call from packages folder to check if code is compliant.
+cd packages
```
cmelchior, line 181: This is actually not correct. We do mean the root folder, i.e. `<path>/realm-kotlin`. This is because these commands also check the rest of the project, not just the library code inside `packages`

cmelchior (approved):
Really nice improvements. I only noticed one mistake around `ktlintFormat`, but otherwise this is some great edits 💯

CONTRIBUTING.md
elle-j: Okay, seemed to work only from `packages` for me but will try to resolve that. Thanks!

cmelchior: Calling it from the root requires that you have called `./gradlew publishAllPublicationsToTestRepository` first.

clementetb (approved):
Looks great 🚀

cmelchior (approved):

Issue #1107, opened 2022-11-03
Proguard rules update


tarasmorskyi-mediamonks:
add missing rules

-keep class io.realm.kotlin.types.EmbeddedRealmObject
-keep class * implements io.realm.kotlin.types.EmbeddedRealmObject { *; }

nhachicha:
Thanks @tarasmorskyi-mediamonks 
@rorbech I think we missed to add proguard rules in https://github.com/realm/realm-kotlin/pull/797/files

nhachicha:
fixed via https://github.com/realm/realm-kotlin/pull/1109

PR #1109, opened 2022-11-04
Add missing proguard rules

clementetb:
Adds an entry for `NativePointer` into proguard. This a crash when performing a migration.


packages/library-base/proguard-rules-consumer-common.pro
```diff
@@ -90,6 +90,9 @@
 -keep class io.realm.kotlin.internal.interop.AsyncOpenCallback {
     *;
 }
+-keep class io.realm.kotlin.internal.interop.NativePointer {
```
cmelchior, line 93: We should also add these I think? 🤔 

https://github.com/realm/realm-kotlin/issues/1107

nhachicha (approved):

PR #1110, opened 2022-11-04
Functions support

clementetb:
### App function support

Add support for authenticated users to invoke App Services Application functions via the SDK.

Due to the serialization engine does not support third-party libraries yet, there are some limitations in what types can be used as arguments and return types:
- Primitives, `Bson`, lists, and maps are valid argument types.
- Results can only be deserialized to primitives or `Bson` types.

These limitations would be lifted once the `Kserializer` decoder for third-party libraries becomes stable.

The current implementation of the `Ejson` decoder does a case-by-case decoding based on the `deserializationStrategy` used. This would allow us to avoid future breaking changes, as it uses the same function signatures that we would be using in the future.

### Pending

- [x] Use latest kserializer version.
- [x] Add tests for RealmInstant new extension functions.
- [x] Use `kbson` stable version.
- [x] Update the Changelog with the Kbson library version.

### Depends on
- https://github.com/mongodb/kbson/pull/7
- https://github.com/realm/realm-core/pull/5996
- https://github.com/mongodb/kbson/pull/8


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/MutableRealmIntInternal.kt
```diff
@@ -83,3 +88,15 @@ internal class UnmanagedMutableRealmInt(
 
     override fun decrement(value: Number) = increment(-value.toLong())
 }
+
+internal object ManagedMutableRealmIntSerializer : KSerializer<MutableRealmInt> {
```
cmelchior, line 92: Just food for though, but would it make more sense to gather all the serializers in a single file? Since, if we ever have to change them, we probably need to change all of them?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
```diff
@@ -85,6 +85,15 @@ public interface App {
...
     public fun allUsers(): Map<String, User>
 
+    /**
+     * Returns a wrapper for invoking App Services Application functions as the specified [user].
+     *
+     * [Atlas Functions documentation](https://www.mongodb.com/docs/atlas/app-services/functions/)
+     *
+     * @param user user with to execute the function.
+     */
+    public fun functions(user: User): Functions
```
cmelchior, line 95: I know we had these on `App` in Java, but we started to move things that required a user to `User` instead so you didn't had to pass in the user all the time, then you would access this through `app.currentUser.functions` instead.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
```diff
@@ -85,6 +85,15 @@ public interface App {
      */
     public fun allUsers(): Map<String, User>
 
+    /**
+     * Returns a wrapper for invoking App Services Application functions as the specified [user].
```
cmelchior, line 89:
```suggestion
     * Returns a wrapper for invoking App Services Application functions as the specified [user].
```

I think this is just called Atlas Functions? https://www.mongodb.com/docs/atlas/app-services/functions/

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+import io.realm.kotlin.mongodb.internal.BsonEncoder
+import kotlinx.serialization.DeserializationStrategy
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.serializer
+import org.mongodb.kbson.BsonArray
+import org.mongodb.kbson.BsonDocument
+import org.mongodb.kbson.BsonValue
+
+public interface Functions {
+    public val app: App
```
cmelchior, line 28: Missing Dokka for these two

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+    public val user: User
+
+    /**
+     * Invokes an App Services Application function.
```
cmelchior, line 32:
```suggestion
     * Invokes an Atlas Function.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * Due to the serialization engine does not support third-party libraries yet, there are some
```
cmelchior, line 34: Maybe put a link their their documentation here?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+     * The Bson implementations for arrays or maps are [BsonArray] and [BsonDocument], and they can be
+     * used as valid return types.
+     *
+     * @param name Name of the function to call.
+     * @param args Arguments to the function.
+     * @param deserializationStrategy Deserialization strategy for decoding the results.
+     * @param T The type for the functions response.
+     * @return Result of the function.
+     *
+     * @throws AppException if the request failed in some way.
```
cmelchior, line 49: I think there are a number of error codes associated with functions:

```
    RLM_APP_ERR_SERVICE_FUNCTION_NOT_FOUND,
    RLM_APP_ERR_SERVICE_FUNCTION_ALREADY_EXISTS,
    RLM_APP_ERR_SERVICE_FUNCTION_DUPLICATE_NAME,
    RLM_APP_ERR_SERVICE_FUNCTION_SYNTAX_ERROR,
    RLM_APP_ERR_SERVICE_FUNCTION_INVALID,
```

We should consider how many of these that should be turned into public actionable exception types

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public suspend fun <T : Any?> invoke(
+        name: String,
+        args: List<Any?>,
+        deserializationStrategy: DeserializationStrategy<T>
```
cmelchior, line 54: Can't we provide a default here? At least it is very unclear to me what a user is supposed to set here.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+public suspend inline fun <reified T : Any?> Functions.invoke(
+    name: String,
+    args: List<Any?>
+): T = invoke<T>(
+    name = name,
+    args = args,
+    deserializationStrategy = serializerOrDefault()
+)
+
+public inline fun <reified T : Any?> serializerOrDefault(): KSerializer<T> =
```
cmelchior, line 86: As this is public, it needs documentation

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * @param name Name of the function to call.
+     * @param args Arguments to the function.
+     * @param deserializationStrategy Deserialization strategy for decoding the results.
+     * @param T The type for the functions response.
+     * @return Result of the function.
+     *
+     * @throws AppException if the request failed in some way.
+     */
+    public suspend fun <T : Any?> invoke(
```
cmelchior, line 51: Do we want to expose calling functions like this? 
At least I find the syntax `app.currentUser.functions.call(...)` more readable than `app.currentUser.functions(...)`

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -97,6 +97,13 @@ public interface User {
      */
     public val deviceId: String
 
+    /**
+     * Returns a wrapper for invoking App Services Functions.
```
cmelchior, line 101: Name again

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -97,6 +97,13 @@ public interface User {
...
+     *
+     * [Atlas Functions documentation](https://www.mongodb.com/docs/atlas/app-services/functions/)
```
cmelchior, line 103: 👍

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,292 @@
...
+import org.mongodb.kbson.BsonRegularExpression
+import org.mongodb.kbson.BsonString
+import org.mongodb.kbson.BsonSymbol
+import org.mongodb.kbson.BsonTimestamp
+import org.mongodb.kbson.BsonType
+import org.mongodb.kbson.BsonUndefined
+import org.mongodb.kbson.BsonValue
+
+/**
+ * Bson encoder based on the `Json` encoder from the `KSerializer`. To avoid using any
```
cmelchior, line 64: I think this is a good first step, but would it be possible to support custom types if we just documented that the support is experimental?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,292 @@
...
+ * experimental `KSerializer` APIs and to maximize compatibility it only supports a predefined set
+ * of types:
+ * - Primitives, Realm, Bson, Lists and Map types for encoding.
+ * - Primitives, Realm and Bson types for decoding.
+ */
+public object BsonEncoder {
+    // We need to define these serializers as contextual because they are not available in the
+    // serializer module.
+    private val realmSerializersModule = SerializersModule {
```
cmelchior, line 73: What about list/sets? Or do they work out of the box?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,292 @@
...
+        contextual(ObjectId::class, ObjectIdSerializer)
+        contextual(RealmUUID::class, RealmUUIDSerializer)
+        contextual(RealmInstant::class, RealmInstantSerializer)
+    }
+
+    public val serializersModule: SerializersModule =
+        Json.serializersModule.plus(realmSerializersModule)
+
+    /**
+     * Encodes a given value into a [BsonValue]. Only primitives, Realm, Bson, lists and maps types
```
cmelchior, line 83: What about RealmSet, it could be turned into a list for transport?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,292 @@
...
+     * are supported.
+     *
+     * @param value value to encode.
+     * @return Ejson encoded String.
```
cmelchior, line 87: Isn't it EJSON or `ejson`? Don't think I have seen it written like this before?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,292 @@
...
+import org.mongodb.kbson.BsonValue
+
+/**
+ * Bson encoder based on the `Json` encoder from the `KSerializer`. To avoid using any
+ * experimental `KSerializer` APIs and to maximize compatibility it only supports a predefined set
+ * of types:
+ * - Primitives, Realm, Bson, Lists and Map types for encoding.
+ * - Primitives, Realm and Bson types for decoding.
+ */
+public object BsonEncoder {
```
cmelchior, line 70: Any reason this is public?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,630 @@
...
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+import kotlin.test.fail
+
+class FunctionsTests {
+    // Pojo class for testing custom encoder/decoder
```
cmelchior, line 71: Missing cleanup?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,630 @@
...
+//    private val faultyIntegerRegistry = CodecRegistries.fromRegistries(
+//        CodecRegistries.fromProviders(IterableCodecProvider()),
+//        CodecRegistries.fromCodecs(StringCodec(), faultyIntegerCodec)
+//    )
+
+    @BeforeTest
+    fun setup() {
+        app = TestApp("functions") { app: BaasApp, service: Service ->
+            initializeDefault(app, service)
+            app.addFunction(FIRST_ARG_FUNCTION)
```
cmelchior, line 128: Nice 👍

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,630 @@
...
+        assertFailsWithMessage<ServiceException>("[Service][Unknown(-1)] expected Authorization header with JWT") {
+            runBlocking {
+                functions.call(FIRST_ARG_FUNCTION.name, 1, 2, 3)
+            }
+        }
+    }
+
+    // Tests that functions that should not execute based on "canevalute"-expression fails.
+    @Test
+    fun callFunction_authorizedOnly() {
```
cmelchior, line 535: Given the error codes I posted earlier, I suspect we are missing some edge cases here like calling a non-existent function, wrong parameters and things like that. 

Also what happens here if the function throws an exception on the server

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,630 @@
...
+
+    @AfterTest
+    fun tearDown() {
+        if (this::app.isInitialized) {
+            app.close()
+        }
+    }
+
+    @Test
+    fun roundtripWithSupportedTypes() {
```
cmelchior, line 149: Missing tests for `RealmList` and `RealmSet`

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/serializer/BsonEncoderTests.kt
```diff
@@ -0,0 +1,311 @@
...
+            assertEquals(
+                bsonValue,
+                BsonEncoder.encodeToBsonValue(value)
+            )
+        }
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    internal fun serializer(value: KClass<*>): KSerializer<*> = when (value) {
+        ObjectIdImpl::class -> BsonEncoder.serializersModule.serializer<ObjectId>()
```
cmelchior, line 115: It looks easy to let this function get out of sync with changes to the codebase, but not sure what we can do about it ? 🤔

cmelchior (commented):
Just initial review, biggest touchpoints are:

- Looks like we are missing support for RealmList and RealmSet
- I have some smaller concerns about the public API
- Error reporting seems like it could be improved

But otherwise it looks like a great start 🥳

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
clementetb: Much better 👍

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
clementetb: And for the interface 🫣

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -89,7 +89,9 @@ object Versions {
     const val serialization = "1.4.0" // https://kotlinlang.org/docs/releases.html#release-details
     const val shadowJar =  "6.1.0" // https://mvnrepository.com/artifact/com.github.johnrengelman.shadow/com.github.johnrengelman.shadow.gradle.plugin?repo=gradle-plugins
     const val multidex = "2.0.1" // https://developer.android.com/jetpack/androidx/releases/multidex
-    const val kbson = "0.1.0" // https://github.com/mongodb/kbson
+    const val kbson = "0.2.0-SNAPSHOT" // https://github.com/mongodb/kbson
+    const val kserializerPlugin = "1.6.10" // https://github.com/Kotlin/kotlinx.serialization
```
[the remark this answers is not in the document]
clementetb: Updated to match

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
clementetb: I have added `FunctionExecutionException` that maps
RLM_APP_ERR_SERVICE_FUNCTION_NOT_FOUND
RLM_APP_ERR_SERVICE_FUNCTION_EXECUTION_ERROR

The other cases are not applicable as they triggered when creating/modifying functions.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
clementetb: We cannot provide a default as it is tied to the [T] type parameter.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
clementetb: Added support for sets.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
clementetb: Added support for collections

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
clementetb: It is accessed for a bunch of inlined extension functions.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/serializer/BsonEncoderTests.kt
clementetb: Yeah, I am not sure how we could get it tracked.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/serializer/BsonEncoderTests.kt
```diff
@@ -0,0 +1,321 @@
...
+            deserializationStrategy = serializer(RealmInstant::class),
+            requiredBsonType = BsonDateTime::class,
+            invalidBsonValue = BsonString("")
+        ),
+    )
+
+    @Test
+    fun decodeFromBsonElement_throwsWrongType() {
+        (primitiveAsserters + realmAsserters).forEach {
+            it.assert()
```
rorbech, line 273: I think we should have the `assertFailsWithMessage` here and not burried inside the `WrongTypeAsserter`

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,623 @@
...
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+import kotlin.test.fail
+
+class FunctionsTests {
+//    @Serializable
+//    private data class Dog(var name: String? = null)
+
+//    private val looperThread = BlockingLooperThread()
```
rorbech, line 72: ??

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,623 @@
...
+                        BsonDocument("KEY", BsonString("VALUE"))
+                    )
+                    assertEquals(
+                        documents[0],
+                        functions.callBlocking<BsonDocument>(FIRST_ARG_FUNCTION.name, *documents)
+                    )
+                }
+                BsonType.DATE_TIME -> {
+                    // JVM and Darwing platform's RealmInstant have better precission than BsonDateTime
+                    // we create a RealmInstant with loose of precision
```
rorbech, line 284: Not sure I get the comment??

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,623 @@
...
+                BsonType.REGULAR_EXPRESSION,
+                BsonType.SYMBOL,
+                BsonType.DB_POINTER,
+                BsonType.JAVASCRIPT,
+                BsonType.JAVASCRIPT_WITH_SCOPE,
+                BsonType.TIMESTAMP,
+                BsonType.END_OF_DOCUMENT,
+                BsonType.MIN_KEY,
+                BsonType.MAX_KEY -> {
+                    // Relying on org.bson codec providers for conversion, so skipping explicit
```
rorbech, line 306: Is there an issue for this that we can reference?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,623 @@
...
+        }
+    }
+
+    private inline fun <reified T : Any> assertTypeOfFirstArgFunction(
+        value: T
+    ): T = functions.callBlocking<T>(FIRST_ARG_FUNCTION.name, value).also {
+        assertEquals(value, it)
+    }
+
+//    @Test
```
rorbech, line 322: Irrelevant so could just remove it

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,623 @@
...
+//                assertEquals(output, result.orThrow)
+//            } finally {
+//                looperThread.testComplete()
+//            }
+//        })
+//    }
+
+    @Test
+    fun unknownFunction() {
+        assertFailsWithMessage<FunctionExecutionException>("[Service][FunctionNotFound(26)] function not found: 'unknown'") {
```
rorbech, line 456: Maybe just assert on the relevant explanatory part of the message "function not found"

rorbech (commented):
Really nice. Added my comments here and there. Maybe most notably ... Why do we support deserializers already but not serializers. 

Further
- We shouldn't add the snapshot repository to all the non-SDK projects as this might cause us to not catch missing dependencies
- Clean up the tests. I don't know if we normally preserve all the _not yet supported_ tests. Maybe we could just reference the Java file that you used for inspiration. But at least prune the tests that would never make sense, i.e. all the async ones, etc.

benchmarks/build.gradle.kts
```diff
@@ -29,6 +29,7 @@ allprojects {
         }
         google()
         mavenCentral()
+        maven(uri("https://oss.sonatype.org/content/repositories/snapshots"))
```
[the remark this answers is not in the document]
clementetb: We were using a kbson snapshot version, I forgot to remove the repository after we moved to a stable version.

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -565,6 +565,14 @@ expect object RealmInterop {
         callback: AppCallback<Unit>
     )
 
+    fun realm_app_call_function(
+        app: RealmAppPointer,
+        user: RealmUserPointer,
+        name: String,
+        serializedArgs: String,
```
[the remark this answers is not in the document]
clementetb: I am renaming the parameter to serializedEjsonArgs

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -404,6 +404,36 @@ void app_apikey_callback(realm_userdata_t userdata, realm_app_user_apikey_t* api
     }
 }
 
+void app_string_callback(realm_userdata_t userdata, const char *serialized_ejson_response,
+                         const realm_app_error_t *error) {
+    auto env = get_env(true);
+    static JavaClass java_callback_class(
```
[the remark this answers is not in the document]
clementetb: I was just following through other implementations that use `AppCallback` but it is true, we could catch it.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmInstantImpl.kt
```diff
@@ -25,3 +26,16 @@ public data class RealmInstantImpl(override val seconds: Long, override val nano
         return "RealmInstant(epochSeconds=$epochSeconds, nanosecondsOfSecond=$nanosecondsOfSecond)"
     }
 }
+
+private const val MILLI_AS_NANOSECOND: Int = 1_000_000
+private const val SEC_AS_MILLISECOND: Int = 1_000
+
+internal fun RealmInstant.toMillis(): Long {
```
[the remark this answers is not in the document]
clementetb: Oh much better, I didn't know about Duration

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,291 @@
...
+import org.mongodb.kbson.BsonTimestamp
+import org.mongodb.kbson.BsonType
+import org.mongodb.kbson.BsonUndefined
+import org.mongodb.kbson.BsonValue
+
+/**
+ * Bson encoder based on the `Json` encoder from the `KSerializer`. To avoid using any
+ * experimental `KSerializer` APIs and to maximize compatibility it only supports a predefined set
+ * of types:
+ * - Primitives, Realm, Bson, Lists and Map types for encoding.
```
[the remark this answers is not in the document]
clementetb: Realm types, the types is at the end of the sentence. RealmList, RealmSet, RealmMap, MutableInt..

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,291 @@
...
+        Json.serializersModule.plus(realmSerializersModule)
+
+    /**
+     * Encodes a given value into a [BsonValue]. Only primitives, Realm, Bson, lists and maps types
+     * are supported.
+     *
+     * @param value value to encode.
+     * @return EJSON encoded String.
+     */
+    public fun encodeToBsonValue(value: Any?): BsonValue = toBsonValue(value)
```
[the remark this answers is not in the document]
clementetb: Initially I tried to encode directly from/to strings, forgot to update the documentation.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,291 @@
...
+    @Suppress("UNCHECKED_CAST", "IMPLICIT_CAST_TO_ANY", "ComplexMethod", "LongMethod")
+    public fun <T : Any?> decodeFromBsonValue(
+        deserializationStrategy: DeserializationStrategy<T>,
+        bsonValue: BsonValue,
+    ): T {
+        return if (bsonValue == BsonNull && deserializationStrategy != serializer<BsonNull>()) {
+            null
+        } else {
+            with(serializersModule) {
+                when (deserializationStrategy) {
```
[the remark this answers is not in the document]
clementetb: It is a temporary solution that uses no experimental serializer APIs. I haven't been able to come up with something more elegant.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
clementetb: The file is a copy-paste from Java, haven't trimmed any of the unused code. I was following what we do in other files, I prefer not to have these comments hanging out, we already have it in the Java project.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
clementetb: `RealmInstant` has nanoseconds precision while `BsonDateTime` has milliseconds. I just wrote it in a funny way 🫣

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
clementetb: I think you cannot call to internal methods from the inlinings, even if suppressing warnings.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/BsonEncoder.kt
```diff
@@ -0,0 +1,291 @@
...
+ * - Primitives, Realm, Bson, Lists and Map types for encoding.
+ * - Primitives, Realm and Bson types for decoding.
+ */
+public object BsonEncoder {
+    // We need to define these serializers as contextual because they are not available in the
+    // serializer module.
+    private val realmSerializersModule = SerializersModule {
+        contextual(ObjectId::class, ObjectIdSerializer)
+        contextual(RealmUUID::class, RealmUUIDSerializer)
+        contextual(RealmInstant::class, RealmInstantSerializer)
```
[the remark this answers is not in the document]
clementetb: The @Serializable tag in `MutableRealmInt` automatically adds it to the serializer module, but it is because it is an abstract class. The other types are interfaces (polymorphic) and we have to state what serializer would be used on the interface type.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,103 @@
...
+     * @param args Arguments to the function.
+     * @param deserializationStrategy Deserialization strategy for decoding the results.
+     * @param T The type for the functions response.
+     * @return Result of the function.
+     *
+     * @throws FunctionExecutionException if the request failed in some way.
+     */
+    public suspend fun <T : Any?> call(
+        name: String,
+        deserializationStrategy: DeserializationStrategy<T>,
```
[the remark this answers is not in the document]
clementetb: The serializers are picked automatically from the decoder being used, by resolving it with the type of the argument. 

To be able to deserialize the return value, we need to resolve the serializer from the type, thus the reified parameter. We could pass the return value class and then later resolve it, but it is easier to do in the inline function.

The Json decoder API works like that, you can decode using an [inline-reified](https://github.com/Kotlin/kotlinx.serialization/blob/16a85df254f4f1e317554eb61ee1fbe914800aa4/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt#L209) function or[ manually pass the serializer.](https://github.com/Kotlin/kotlinx.serialization/blob/16a85df254f4f1e317554eb61ee1fbe914800aa4/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt#L114)

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
clementetb: I copied it from Java but you are right that its worth having them roundtripped.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * Add better error messages when inheriting `RealmObject` with unsupported class types. (Issue [#1086](https://github.com/realm/realm-kotlin/issues/1086))
 * [Sync] `App.close()` have been added so it is possible to close underlying ressources used by the app instance.
 * [Sync] Add support for progress listeners with `SyncSession.progress`. (Issue [#428](https://github.com/realm/realm-kotlin/issues/428))
+* Add better error messages when inheriting `RealmObject` with unsupported class types. (Issue [#1086](https://github.com/realm/realm-kotlin/issues/1086))
```
cmelchior, line 16: This looks like it is in the wrong place and shouldn't even be part of this PR?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmInstantImpl.kt
```diff
@@ -25,3 +29,13 @@ public data class RealmInstantImpl(override val seconds: Long, override val nano
...
     }
 }
+
+internal fun RealmInstant.toDuration(): Duration {
+    return epochSeconds.seconds + nanosecondsOfSecond.nanoseconds
+}
+
+internal fun Duration.toRealmInstant(): RealmInstant {
+    val seconds = this.inWholeSeconds.seconds
+    val nanos = (this - seconds)
```
cmelchior, line 39: This looks a bit suspicious to me? Why is the end result of this nanoseconds?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,103 @@
...
+     * The Bson implementations for arrays or maps are [BsonArray] and [BsonDocument], and they can be
+     * used as valid return types.
+     *
+     * @param name Name of the function to call.
+     * @param args Arguments to the function.
+     * @param deserializationStrategy Deserialization strategy for decoding the results.
+     * @param T The type for the functions response.
+     * @return Result of the function.
+     *
+     * @throws FunctionExecutionException if the request failed in some way.
```
cmelchior, line 69:
```suggestion
     * @throws FunctionExecutionException if the function failed in some way.
```

cmelchior: We should probably also add 

```
     * @throws io.realm.kotlin.mongodb.exceptions.ServiceException for other failures that can happen when
     * communicating with App Services. See [AppException] for details.
```

Similar to what we do in ApiKeyAuth and EmailPasswordAuth

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -247,6 +248,10 @@ internal fun convertAppError(appError: AppError): Throwable {
                 ServiceErrorCode.RLM_APP_ERR_SERVICE_BAD_REQUEST -> {
                     BadRequestException(msg)
                 }
+                ServiceErrorCode.RLM_APP_ERR_SERVICE_FUNCTION_NOT_FOUND,
+                ServiceErrorCode.RLM_APP_ERR_SERVICE_FUNCTION_EXECUTION_ERROR -> {
+                    FunctionExecutionException(msg)
```
cmelchior, line 253: 👍

cmelchior (approved):
Nice, only nitpick comments

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmInstantImpl.kt
cmelchior: Ah yeah, seconds is also a `Duration` 🙈 ...it would help if the types where visible here.

CHANGELOG.md
clementetb: Not part from this PR, it comes from merging main.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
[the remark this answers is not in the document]
rorbech: I had a more detailed look at this and I don't feel this is the way to go. We are currently setting up the `DeserializationStrategy`-based API just to be able to identify the type but are not using the actual deserializers at all. I also don't think that the users should be bothered with doing conversion to EJSON but only be bothered with converting objects to BSON-values. This functionality of converting objects should probably go into the BSON library just as the codec support of the Java BSON library. 

Given how easy it is to do a `@RequiresOptIn`-annotation (see https://github.com/realm/realm-kotlin/commit/dc83ae7c1e689cfdd0ce9dfc2630473637814d07) I think we should just keep the most basic functions API with something like
```
public suspend fun call(name: String, arg: BsonValue): BsonValue
```
Then we can do a very basic temporary `BsonConverter` with the default primitive type conversions until we have a more solid idea of the overall API should look like. This would also be symmetric around yielding Basically what I outlined in https://github.com/realm/realm-kotlin/commit/41a9744538b642c6b871b9167b51494ff5536cbf ... Didn't rip out all the `DeserializationStrategy` so also didn't rip out the `BsonEncoder` just yet. Anyway ... up for discussing the details.

rorbech (changes requested):
I think we should discuss the need for the `DeserializerStrategy` in this PR before merging. See details in https://github.com/realm/realm-kotlin/pull/1110#discussion_r1056414200

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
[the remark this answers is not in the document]
clementetb: The kserializer API for writing third-party encoders is still experimental. 

We designed the API to support in the short future a non-experimental serializer with reduced type support (current implementation) and a full blown experimental one.

To do so I took it from the starting point of what the final API should look like, I took the Json encoder as a reference, as the Bson encoder is going to have the same api except for the method names.

I see the point of your concern on the API wise. I think that we might strip the BsonEncoder thing from this PR, and leave it in the bare minimum, and address it in a future PR.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/serializer/BsonEncoderTests.kt
```diff
@@ -0,0 +1,371 @@
...
+        DecoderAsserter(
+            deserializedType = RealmInstant::class,
+            bsonType = BsonDateTime::class,
+            bsonValue = BsonString("")
+        ),
+    )
+
+    @Test
+    fun decodeFromBsonElement_throwsWrongType() {
+        (wrongPrimitiveAsserters + wrongRealmValueAsserters).forEach {
```
rorbech, line 315: Nothing serious, but since we don't really uses these variables elsewhere we could maybe just define them locally in the method. Also a bit hard to track because you need to look up the `DecoderAsserter` class to understand it. Why not just use a `Triple` and have all the logic here and then destruct them into named variables to that it was easier to follow. Anyway, nothing critical 🤷

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param name Name of the function to call.
+     * @param args Arguments to the function.
+     * @param resultType The KType of the return value.
+     * @param T The type of the return value.
+     * @return Result of the function.
+     *
+     * @throws FunctionExecutionException if the function failed in some way.
+     */
+    public suspend fun <T> call(
```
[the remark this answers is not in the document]
rorbech: Also a bit funny to have the generic parameter throughout the `BsonConverter.decodeFromBsonValue` but maybe it is ok 🤔 ... But if we keep the `T` along with a `KType` that is not bound to be the same then we should definitely be very explicit about that it will throw a `ClassCastException` if they don't match.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
rorbech: It's still here 🤪 ... we wouldn't ever do _sync/async_ differentiated tests because we have a suspending API

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,85 @@
...
+package io.realm.kotlin.mongodb
+
+import io.realm.kotlin.mongodb.exceptions.FunctionExecutionException
+import org.mongodb.kbson.BsonArray
+import org.mongodb.kbson.BsonDocument
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+/**
+ * A Functions manager to call remote Atlas Functions for the associated Atlas App services Application.
```
cmelchior, line 25:
```suggestion
 * A Functions manager to call remote Atlas Functions for the associated Atlas App Services Application.
```

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,648 @@
...
+//                assertEquals(32, result.orThrow.toInt())
+//            } finally {
+//                looperThread.testComplete()
+//            }
+//        }
+//    }
+//
+//
+//    @Test
+//    fun codecArgumentFailure() {
```
rorbech, line 360: I think it still makes sense to have some tests that verifies the exception that we throw when we cannot encode arguments and decode return values. Both when just passing some unsupported types, but also if just asking for the wrong return type, etc.

rorbech: And also that how we handle `T`/`KType` mismatches if we decide to preserve the ability to pass both in the public API

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
```diff
@@ -0,0 +1,85 @@
...
+     * @param args Arguments to the function.
+     * @param resultType The KType of the return value.
+     * @param T The type of the return value.
+     * @return Result of the function.
+     *
+     * @throws FunctionExecutionException if the function failed in some way.
+     */
+    public suspend fun <T> call(
+        name: String,
+        resultType: KType,
```
cmelchior, line 68: It seems like it is possible to misuse this API as you could put in another KType than T here? I guess the intention is that everyone is going through the reified version, so I don't know how big of a problem this is.

cmelchior: Looking at how this is used, it seems that that we call it here `resultType.classifier`, but a KClass is already a `KClassifier`. Is there something I'm missing?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Functions.kt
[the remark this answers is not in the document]
clementetb: Oh, I didn't think it thoughtfully. Yes, we might like to internalize the method. And yes, the T value doesn't make sense in the BsonEncoder.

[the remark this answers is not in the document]
clementetb: I made it internal and accessible to the extension function via de `@PublishedApi` annotation.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/serializer/BsonEncoderTests.kt
clementetb: The asserters data structure has 4 parameters and is being reused in several tests. 

Yes, we could use a `Triple` because not all parameters are used in every test, but it would mean that we would have to define multiple similar datasets in several tests..

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
clementetb: These cases are being validated in `BsonEncoderTests`. I am going to add some tests to validate that we capture these cases here too.

cmelchior (approved):

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,663 @@
...
+        functions.call(name, *args)
+    }
+
+    @Test
+    fun roundtripWithSupportedTypes() {
+        val i32 = 42
+        val i64 = 42L
+
+        for (type in BsonType.values()) {
+            println(type)
```
rorbech, line 167: Debug 👀

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FunctionsTests.kt
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    private inline fun <reified T : Any> assertTypeOfFirstArgFunction(
+        value: T
+    ): T = functions.callBlocking<T>(FIRST_ARG_FUNCTION.name, value).also {
+        assertEquals(value, it)
+    }
+
+    @Test
+    fun unsupportedReturnTypeThrows() {
```
rorbech, line 349: Seems that you swapped the names of these two methods 😅

rorbech (approved):
Awesome. Only some minor comments ... and then of course CI 😬  Seems to fail due to asserting on class names that are obfuscated. Either remove the actual class name in the expected methods or maybe add `keep` the specific classes for testing purposes, just as the other rules in https://github.com/realm/realm-kotlin/blob/main/packages/test-base/proguard-rules-test.pro

PR #1111, opened 2022-11-04
Rename linking objects to backlinks

clementetb:
Renames `LinkingObjects` to `Backlinks`. 

Core still references linking objects in some error messages.


nhachicha (approved):
Back to `Backlinks' 😭

Issue #1105, opened 2022-11-03
Try System.loadLibrary before trying to extract native libraries from a JAR


manosbatsis:
We use Realm in a multi-platform compose app, one that includes a compose desktop build. Some JVM installers move native shared/JNI libraries are out of the JAR, for a [number of reasons](https://conveyor.hydraulic.dev/1.0/configs/jvm/#native-libraries),  including Windows and macOS security systems' requirement to verify all native code is signed. Conveyor specifically recommends:

> Therefore your software should always attempt to load shared libraries by using System.loadLibrary first, before trying to extract native libraries from a JAR. Alternatively you can use System.load in combination with the java.home system property but remember to add either lib on UNIX or bin on Windows.

From [what i can see here](https://github.com/realm/realm-kotlin/blob/6187b9e2934841796d9f465be876f074707d6048/packages/jni-swig-stub/realm.i),  it seems Realm will only use `System.loadLibrary` on Android, which results in exception when trying to initialize our app on desktop JVM apps:

```
Caused by: java.lang.RuntimeException: Couldn't load Realm native libraries
        at io.realm.kotlin.internal.interop.realmc.<clinit>(realmc.java:28)
        ... 85 more
Caused by: java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at io.realm.kotlin.internal.interop.realmc.<clinit>(realmc.java:26)
        ... 85 more
Caused by: java.lang.NullPointerException: lib must not be null
        at io.realm.kotlin.jvm.SoLoader.unpackAndInstall(SoLoader.kt:98)
        at io.realm.kotlin.jvm.SoLoader.load(SoLoader.kt:62)
        at io.realm.kotlin.jvm.SoLoader.load(SoLoader.kt:47)
        ... 90 more
```

cmelchior:
Hi @manosbatsis Thank you for the report.

[...]

PR #1116, opened 2022-11-07
Add support for loading native library code from a custom location on JVM

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/1105

This PR adds support for bundling Realm inside apps distributed using 3rd party JVM installers like [Hydraulic Conveyor](https://conveyor.hydraulic.dev/4.0/).

It does this by changing how we load the native library. After this change we first attempt to load the library using `System.loadLibrary(name)` which will search all paths defined by `java.libraries.path`. If this fails we will fall back to finding the native library inside the JAR file as before this change.

We make no attempt at detecting if the correct native code is linked as users might have modified the binary, so any checks we employ could fail. But that also means that it is up to developers to make sure that the correct native file is loaded. I will create a ticket in Docs to make sure this is documented.

This change was tested manually on Mac. 

According to https://stackoverflow.com/questions/37203247/while-loading-jni-library-how-the-mapping-happens-with-the-actual-library-name our current naming strategy should be good enough for people to just copy/paste the file from our JAR. If not, they would be responsible for naming it to something that `System.loadLibrary` would find using `System.mapLibraryName(name)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 * None.
 
 ### Enhancements
-* Add support for in-memory Realms.
+* Realm will now use `System.loadLibrary()` first when loading native code on JVM, adding support for 3rd party JVM installers. If this fails, it will fallback to the current method of loading the native library from the JAR file. (Issue [#1105](https://github.com/realm/realm-kotlin/issues/1105)). 
+* Added support for in-memory Realms.
 * [Sync] Added support for API key authentication. (Issue [#432](https://github.com/realm/realm-kotlin/issues/432))
 * Added support for reverse relationships through the `linkingObjects` delegate. See the function documentation for more details. (Issue [#1021](https://github.com/realm/realm-kotlin/pull/1021))
 * Added support for `BsonObjectId` and its typealias `org.mongodb.kbson.ObjectId` as a replacement for `ObjectId`. `io.realm.kotlin.types.ObjectId` is still functional but has been marked as deprecated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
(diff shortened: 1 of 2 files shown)


packages/cinterop/src/jvmMain/kotlin/io/realm/kotlin/jvm/SoLoader.kt
```diff
@@ -41,10 +41,24 @@ class SoLoader {
...
+        // default way of shipping dependencies to JVM Desktop apps, but it does require that the
+        // author of these apps have extracted the library from our JAR file and manually placed it
+        // in the location defined by "java.libraries.path".
+        //
+        // If this fails, we will fallback to finding the native code inside the JAR file and
+        // store it in the users cache directory.
+        //
+        // See https://github.com/realm/realm-kotlin/issues/1105 for more information.
+        try {
+            System.loadLibrary("realmc")
```
clementetb, line 56: How come we don't iterate over the libs as we do in with `load()`?

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 * None.
 
 ### Enhancements
-* Add support for in-memory Realms.
+* Realm will now use `System.loadLibrary()` first when loading native code on JVM, adding support for 3rd party JVM installers. If this fails, it will fallback to the current method of loading the native library from the JAR file. (Issue [#1105](https://github.com/realm/realm-kotlin/issues/1105)).
```
nhachicha, line 7:
```suggestion
* Realm will now use `System.loadLibrary()` first when loading native code on JVM, adding support for 3rd party JVM installers. If this fails, it will fallback to the current method of extracting and loading the native library from the JAR file. (Issue [#1105](https://github.com/realm/realm-kotlin/issues/1105)). 
```

nhachicha (approved):
We can use snapshot to validate this works for https://github.com/realm/realm-kotlin/issues/1105

PR #1119, opened 2022-11-08
Update to latest Core commit

cmelchior:
Pulling in https://github.com/realm/realm-core/pull/5999/files in order to fix some performance issues downloading very large server side Realms


cmelchior:
Fast-merging to unblock some testing

PR #1124, opened 2022-11-10
Build release builds for all JVM targets

cmelchior:
We released debug builds for JVM on Mac and Linux, while Windows did correctly use release builds. This PR fixes this, so all platforms now use release builds.

Regarding symbols on JVM platforms. After talking to both the .NET and JS teams, none of them are doing anything special with symbols and are happy wuth using the release builds as is. They also don't seems to have problems reproducing whatever problems are being reported on those platforms. 

So I would recommend we follow suit on that for the time being and re-evaluate later if we experience this is not enough.

Some comments from Slack:

> Binaries on Windows don’t contain debug symbols in any iteration by default to begin with. They’re stored in a separate .pdb file. You have to explicitly opt into those being merged into the binary itself. And even if you do have symbols there isn’t a built-in facility that gives you a stack trace to symbolicate anyway, and Core’s backtrace wasn’t implemented on Win32 either. In any case, our experience is that for the most part the C# stacktrace alone is enough to point us in the right direction, or we’ve just been lucky that our community is forthcoming with repro instructions and sample code.

> FWIW in my experience only the NDK toolchain makes explicitly stripping binaries off debug symbols necessary. I think you’ll find other platforms do that by default.

> Symbol names are always included by default in ELF and Mach-O binaries unless you exclude them. But if you need line numbers too you can build with -gline-tables-only which includes only line number information without impacting the size or performance negatively. It’s ideal for long-term symbol storage if you need to symbolicate a stacktrace address into a file name and line number format from a crash report containing stack frame addresses.

TODO:
- [x] Figure out how to handle stripping of symbols on all platforms
- [x] Find a better way to align shared CMAKE properties.


Jenkinsfile
```diff
@@ -19,7 +19,7 @@ import groovy.json.JsonOutput
 @Library('realm-ci') _
 
 // Branches from which we release SNAPSHOT's. Only release branches need to run on actual hardware.
-releaseBranches = [ 'master', 'releases', 'next-major', 'release/ktor_2.0.0' ]
+releaseBranches = [ 'master', 'releases', 'next-major', 'cm/release-builds' ]
```
cmelchior, line 22: Revert once this PR is finalized

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -17,7 +17,7 @@
 
 object Realm {
     val ciBuild = (System.getenv("JENKINS_HOME") != null)
-    const val version = "1.5.0-SNAPSHOT"
+    const val version = "1.5.0-PERFORMANCETEST-SNAPSHOT"
```
cmelchior, line 20: Revert once this PR is finalized

edualonso (approved):

packages/cinterop/build.gradle.kts
```diff
@@ -318,22 +342,47 @@ android {
 
 // Building Mach-O universal binary with 2 architectures: [x86_64] [arm64] (Apple M1) for macOS
 val capiMacosUniversal by tasks.registering {
-    build_C_API_Macos_Universal(releaseBuild = isReleaseBuild)
+    build_C_API_Macos_Universal(buildType = buildType)
```
nhachicha, line 345: `buildType` name shadowing with https://github.com/realm/realm-kotlin/blob/09c10bad10060635674abb40b697e0e58e631c12/packages/cinterop/build.gradle.kts#L53 :) maybe rename the global one

nhachicha (approved):
🥇

PR #1131, opened 2022-11-16
Implement custom App Services Function authentication

### LINKED ISSUE #741: Credentials: Add support for Custom Function
Add support for `Credentials.customFunction(args: Document)`.

This requires BSON support first: https://github.com/realm/realm-kotlin/issues/648

clementetb:
Implements custom App Services function authentication. It uses the same serializer implemented during Function invocation https://github.com/realm/realm-kotlin/pull/1110 to serialize the authentication payload.

Currently, the payload is limited to `maps` or `BsonDocuments` but with the experimental serializer, we would be able to serialize credentials object modeled by kotlin classes.

Closes #741


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/Credentials.kt
```diff
@@ -131,5 +132,18 @@ public interface Credentials {
...
+        /**
+         * Creates credentials representing a login using an App Services Function. The payload would
+         * be serialized and parsed as an argument to the remote function. It is required to be
+         * a [Map] or [BsonDocument]. The payload keys must match the format and names the function expects.
+         *
+         * @param payload The payload that will be passed as an argument to the server function.
+         * @return a set of credentials that can be used to log into an App Services Application
+         * using [App.login].
+         */
+        public fun customFunction(payload: Any): Credentials {
```
cmelchior, line 145: Given the current restrictions, should we just create two functions `customFunction(payload: Map<String, Any?>)` and `customFunction(payload: BsonDocument)`....then when/if we add support for full serialization, we could combine them into `customFunction(payload: Any)`?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
```diff
@@ -124,6 +124,10 @@ class AppTests {
                     // There doesn't seem to be a reliable way to throw "InvalidCredentials" for these.
                     null
                 }
+                AuthenticationProvider.CUSTOM_FUNCTION -> {
+                    // There is no way to capture custom authentication credentials exceptions.
```
cmelchior, line 128: I'm not sure this is true? What happens if the remote auth function fails or throws an exception?

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/TestAppInitializer.kt
```diff
@@ -470,10 +470,10 @@ object TestAppInitializer {
         source =
         """
         exports = ({mail, id}) => {
-            // Auth function will fail for emails with a domain different to @androidtest.realm.io
+            // Auth function will fail for emails with a domain different to @10gen.com
             // or with id lower than 666
-            if (!new RegExp("@androidtest.realm.io${'$'}").test(mail) || id < 666) {
-                return 0;
+            if (!new RegExp("@10gen.com${'$'}").test(mail) || id < 666) {
+                throw new Error(`Authentication failed`);
```
cmelchior, line 476: I suspect we need to test both `return 0` and `throw` here.

We should verify in tests that proper AuthException errors are being thrown in both cases. I would expect  `InvalidCredentialsException` if the credentials fail or `AuthException` if an exception is thrown.

You can see in our error mapping logic that we are doing some string comparisons to be able to throw `InvalidCredentialsException` for some of the other credential types.

cmelchior (changes requested):
Nice ... I have some questions about the public API though and it looks like we are not testing all error scenarios.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
clementetb: There might be other reasons to throw than `InvalidCredentials`, the function might fail while communicating with an external service for example.

Functions are user-defined and there is no standard way to communicate that there was an exception in the credentials. All exceptions in the remote function are communicated as AuthException with some standard message appending the user error.

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/TestAppInitializer.kt
clementetb: The custom function should return the user id as a `String` or `Object` or throw an exception in case of failure. If we return 0 core would fail to parse the user id. 

It generates an AuthFailure too, but with an error message stating the parsing failure.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppTests.kt
cmelchior: Good point. I guess that is also fair. I can see that we also document in `App.login` that it will throw an AuthException. I don't see any tests verifying that we are getting an AuthException though?

clementetb: Oh I was fiddling with the behaviors and forgot to add the tests.

rorbech (approved):

Issue #1037, opened 2022-09-29
RealmList should support subquries


cmelchior:
Currently, `RealmList` does not support queries. This should be added similar to Realm Java.

Note, that queries will only work on lists of objects. So either we add this as a type-safe extension method or directly in the API, but then throw if you call it on list of primitives.

```
public fun <E: RealmObject> RealmList<E>.query(): RealmQuery<E> = TODO()
```

PR #1132, opened 2022-11-16
Add support for querying on RealmLists

rorbech:
Add support for querying `RealmList`s. 

```
/**
 * Query the objects of a list by the `filter` and `arguments`.
 *
 * @param filter the Realm Query Language predicate to append.
 * @param arguments Realm values for the predicate.
 */
public fun <T : BaseRealmObject> RealmList<T>.query(
    filter: String = TRUE_PREDICATE,
    vararg arguments: Any?
): RealmQuery<T>
```
This is added as an extension method as we cannot add the method `fun query(...): RealmQuery<T>` on the `RealmList` interface as `RealmQuery` has an `BaseRealmObject` upper bound which `RealmList` do not. 

TODO:
- [x] Changelog entry
- [x] DynamicRealmObject support
- [x] Exception categorization 

Closes: #1037 
Closes: #1147


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 ### Enhancements
 * Add support for `Realm.copyFromRealm()`. All RealmObjects, RealmResults, RealmList and RealmSets now also have a `copyFromRealm()` extension method.
+* Add support for querying on RealmList with `RealmList.query(...)`.  (Issue [#1037](https://github.com/realm/realm-kotlin/issues/1037))
```
```suggestion
* Add support for querying on RealmLists containing objects with `RealmList.query(...)`.  (Issue [#1037](https://github.com/realm/realm-kotlin/issues/1037))
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/RealmListExt.kt
```diff
@@ -42,3 +47,22 @@ public inline fun <reified T : TypedRealmObject> RealmList<T>.copyFromRealm(dept
...
+ * Query the objects of a list by the `filter` and `arguments`.
+ *
+ * @param filter the Realm Query Language predicate to append.
+ * @param arguments Realm values for the predicate.
+ */
+// Added as an extension method as we cannot add the method `fun query(...): RealmQuery<T>` to the
+// `RealmList` interface as `RealmQuery` has an `BaseRealmObject` upper bound which `RealmList` do
+// not.
+public fun <T : BaseRealmObject> RealmList<T>.query(
+    filter: String = TRUE_PREDICATE,
```
cmelchior, line 61: It is called `query` in our other API's

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/RealmListExt.kt
```diff
@@ -42,3 +47,22 @@ public inline fun <reified T : TypedRealmObject> RealmList<T>.copyFromRealm(dept
...
         ?: throw IllegalArgumentException("This RealmList is unmanaged. Only managed lists can be copied.")
 }
+
+/**
+ * Query the objects of a list by the `filter` and `arguments`.
+ *
+ * @param filter the Realm Query Language predicate to append.
+ * @param arguments Realm values for the predicate.
+ */
+// Added as an extension method as we cannot add the method `fun query(...): RealmQuery<T>` to the
```
cmelchior, line 57: Nitpick... I would move this inside the method...only because it is a little annoying this comment separate the public Dokka from the public interface definition

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/Converters.kt
```diff
@@ -292,7 +292,7 @@ internal inline fun realmObjectToRealmValueWithImport(
 
 // Returns a converter fixed to convert objects of the given type in the context of the given mediator/realm
 internal fun <T> converter(
-    clazz: KClass<*>,
+    clazz: KClass<T & Any>,
```
cmelchior, line 295: Hmm, not sure I understand this change?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/InternalTypedRealm.kt
```diff
@@ -36,7 +36,7 @@ internal interface InternalTypedRealm : TypedRealm {
 
     override fun <T : BaseRealmObject> query(clazz: KClass<T>, query: String, vararg args: Any?): RealmQuery<T> {
         val className = configuration.mediator.companionOf(clazz).`io_realm_kotlin_className`
-        return ObjectQuery(realmReference, realmReference.schemaMetadata.getOrThrow(className).classKey, clazz, configuration.mediator, null, query, *args)
+        return ObjectQuery(realmReference, realmReference.schemaMetadata.getOrThrow(className).classKey, clazz, configuration.mediator, query, args)
```
cmelchior, line 39: Nice change...having to remember the vararg across the entire internal API was pretty errorprone

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmQuery.kt
```diff
@@ -23,6 +23,8 @@ import io.realm.kotlin.types.RealmList
 import kotlinx.coroutines.flow.Flow
 import kotlin.reflect.KClass
 
+public const val TRUE_PREDICATE: String = "TRUEPREDICATE"
```
cmelchior, line 26: Nitpick, should we rename this to `ALWAYS_TRUE` to better signal what it does? I always found `true_predicate` pretty hard to reason about...but not a big deal.

cmelchior: As a public property this should also have documentation

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmListTests.kt
```diff
@@ -411,6 +414,97 @@ class RealmListTests {
...
+            delete(instance)
+            assertFailsWithMessage<IllegalStateException>("Access to invalidated Collection") {
+                objectListField.query()
+            }
+        }
+        Unit
+    }
+
+    @Test
+    fun query_throwsOnClosedList() = runBlocking {
```
cmelchior, line 497: There seem to be a test missing showing that the flow complete if you delete the object containing the list while the flow is running.

cmelchior (commented):
Nice, only had minor comments. I am a bit concerned that CI has a StackOverflow when building, but nothing in the PR caused any alarm to me..maybe something about the query changes 🤔

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/Converters.kt
rorbech: The converter's `T` can be a nullable type e.g. `converter<String?>` but `KClass` has an upper bound of `Any` (which is not nullable), so this is a way of getting the always non-null part of `T`. With this change we are able to connect the `clazz` with other type references which wasn't possible before https://kotlinlang.org/docs/whatsnew17.html#stable-definitely-non-nullable-types

rorbech:
> I am a bit concerned that CI has a StackOverflow when building, but nothing in the PR caused any alarm to me..maybe something about the query changes 🤔

The `StackOverflow` is just gradle that doesn't report the errors correctly. The actual build error was just some build error from making the `queryPointer` private. Thought the `@Supress("invisible...")` didn't care. Making it `internal` again fixed it ... at least locally

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/RealmListExt.kt
rorbech: Seem like we use them both somehow spread across the SDK 🤪 ... maybe `predicate` that would also align well with 
```
const val ALWAYS_TRUE: String = "TRUE_PREDICATE"
```
?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
```diff
@@ -153,45 +153,59 @@ internal object RealmObjectHelper {
...
+                ) as ListOperator<R>
+            }
         }
     }
 
     internal inline fun <reified R : Any> getSet(
         obj: RealmObjectReference<out BaseRealmObject>,
         propertyName: String
-    ): ManagedRealmSet<Any?> {
+    ): ManagedRealmSet<R?> {
```
edualonso, line 208: Does this comment https://github.com/realm/realm-kotlin/blob/324a4f69ea272b96d6337c888e079e57a0ee9c54/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt#L68 not apply anymore? I haven't been able to use generics on native in the `realm_value_t` pull request.

packages/plugin-compiler/src/test/kotlin/io/realm/kotlin/compiler/GenerationExtensionTest.kt
```diff
@@ -89,7 +89,7 @@ class GenerationExtensionTest {
         fun assertGeneratedIR() {
             val outputFile = File("${outputDir()}/main/01_AFTER.ValidateIrBeforeLowering.ir")
             stripInputPath(outputFile, fileMap)
-            val expected = File("${expectedDir()}/00_ValidateIrBeforeLowering.ir").readText()
+            val expected = File("${expectedDir()}/01_AFTER.ValidateIrBeforeLowering.ir").readText()
```
edualonso, line 92: What has changed here?

edualonso (commented):
Looks nifty with the new nullable types. I guess it will take a few iterations before we internalise the concept.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmObjectHelper.kt
rorbech: The comment is really old, so I just thought I would try to use generics properly the places that I had to touch. It seems to work for most of it, but haven't gone through all the cases relevant to this comment, so didn't remove the comment just yet.

packages/plugin-compiler/src/test/kotlin/io/realm/kotlin/compiler/GenerationExtensionTest.kt
rorbech: The output file generated from the compiler was changed a while ago. This just renames the `expected` files to the same name that the compiler generates

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/RealmListExt.kt
rorbech: Agreed that we would keep `filter` as we already use that in various places and `query` is annoying as it is shadowing the method name.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/ext/RealmListExt.kt
rorbech: It absolutely doesn't fit inside the method, so moved it above the docs ... 🤷 ... is that better?

rorbech:
@cmelchior The missing tests showed quite some issues with our flow, so probably sane with another round. Also discovered this dangling [PR](https://github.com/realm/realm-kotlin/pull/1061/files) with part of the same functionality but didn't deem the tests of all the aggregates essential for this PR. Is this really needed?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmNotificationsTests.kt
```diff
@@ -211,4 +218,56 @@ class RealmNotificationsTests : NotificationTests {
             c.close()
         }
     }
+
+    // Will crash if we the collecting context doesn't follow up on processing notifications
+    // https://github.com/realm/realm-kotlin/issues/1147
+    @Ignore
```
cmelchior, line 224: Could we enable this test and just capture the crash instead?

cmelchior (commented):
Still trying to wrap my head around the buffer tests, but otherwise this looks really great 👍

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmListTests.kt
```diff
@@ -411,6 +423,202 @@ class RealmListTests {
...
+                .onEach { (old, current) ->
+                    assertTrue { old == -1 || current - old == 1 }
+                }
+                // Close gracefully when reaching 100
+                .takeWhile { (_, current) -> current < 100 }
+            withTimeout(10.seconds) {
+                flow.collect { (_, _) ->
+                    delay(30.milliseconds)
+                }
+            }
```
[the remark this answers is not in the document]
rorbech: The `runBlocking` won't return before all coroutines exits, so will only if `collect` or `withTimeout` returns, so either we are done reaching 100 elements or we fail due to the timeout. Shouldn't that be sufficient?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmListTests.kt
```diff
@@ -411,6 +423,202 @@ class RealmListTests {
...
+                    current to new
+                }
+                // Keep on going until we skip ids
+                .takeWhile { (old, current) ->
+                    old == -1 || current - old == 1
+                }
+            withTimeout(10.seconds) {
+                flow.collect { (_, current) ->
+                    // Assert that we haven't reached the last event
+                    assertTrue { current != 100 }
```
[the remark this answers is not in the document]
rorbech: Yes, but I don't really have any good alternatives. I am all open for input 😅

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmNotificationsTests.kt
rorbech: No, because the `throw` isn't propagated to the recipient side so it is actually crashing the app. It should probably be propagated through `channel.close(exception)` instead to be signalled correctly, but need to experiment a bit with buffer control before I have a good feeling on the best way to handle backpressure without ruin our internal logic on events.

rorbech: I have tried to rework the way we report if we cannot deliver events on the notification flow and reworked the `ObjectBoundRealmResults` so that it doesn't fuse any external backpressure strategy and drops the internal events that we are heavily dependent on. Now failure to send events will close the flow with the exception issued in https://github.com/realm/realm-kotlin/pull/1132/files#diff-f05a939e8f402867f4467d8cabb245e48c8ebd28f3c5f92b752dab8122aa866dR142

clementetb (commented):
Looking good! Great that you spotted these issues with Bound queries 🫣

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmListTests.kt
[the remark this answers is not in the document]
rorbech: The `withTimeout` only throws if the `async` is actually `await`ed 🤔  ... so added an `await()` to guard finalizing the test. Maybe that also addresses your concern.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmQuery.kt
```diff
@@ -23,6 +23,11 @@ import io.realm.kotlin.types.RealmList
 import kotlinx.coroutines.flow.Flow
 import kotlin.reflect.KClass
 
+/**
+ * A query filter string that will select all objects.
+ */
+public const val ALWAYS_TRUE: String = "TRUEPREDICATE"
```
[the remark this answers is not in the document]
cmelchior: After discussing it we ended up on 

```
TRUE_PREDICATE
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
```diff
@@ -179,6 +185,35 @@ internal class ManagedRealmList<E>(
...
+            parent,
+            ObjectQuery(
+                operator.realmReference,
+                operator.classKey,
+                operator.clazz,
+                operator.mediator,
+                queryPointer,
+            )
+        )
+//    } catch (exception: Throwable) {
```
clementetb, line 209: Whats the use of this comment?

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
```diff
@@ -19,6 +19,9 @@ package sample.input
 import io.realm.kotlin.ext.backlinks
 import io.realm.kotlin.ext.realmListOf
 import io.realm.kotlin.ext.realmSetOf
+import io.realm.kotlin.internal.RealmObjectCompanion
```
clementetb, line 22: Unused imports

clementetb (approved):
Added minor comments, happy once green

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/ObjectBoundRealmResults.kt
```diff
@@ -75,14 +76,27 @@ internal class ObjectBoundRealmResults<E : BaseRealmObject>(
...
+internal fun <T> Flow<T>.bind(reference: RealmObjectReference<out BaseRealmObject>): Flow<T> =
+    channelFlow {
+        // Listen for object deletions
+        reference.asFlow()
+            .map { it is DeletedObject }
+            .onStart { false }
+            .distinctUntilChanged()
+            // Combine object deletion events with the actual result changes
+            .combine(this@bind) { deleted: Boolean, resultsChange: T -> deleted to resultsChange }
+            // We shouldn't use produceIn as that will fuse backpressure strategy to our
```
cmelchior, line 88:
```suggestion
            // We should use produceIn as that will fuse backpressure strategy to our
```

cmelchior (approved):

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
rorbech: I discovered that we had the same logic in the `tryCatchCoreException` so used that and forgot to remove it after verifying all the tests 😅

JorisBodin:
Hello,

How come it's still not available in the snapshot?

PR #1133, opened 2022-11-17
Add support for @PersistedName annotation for fields

### LINKED ISSUE #590: Add support for `@PersistedName` annotation for fields to override underlying name
Similar to Realm Java, we should provide the option of mapping Kotlin field names to their underlying (persisted) representation:

```
class Example : RealmObject {
  @PersistedName(name = "myPersistedName")
  var myKotlinName: String = ""
}
```

In Realm Java, we allow for different naming strategies, overriding all names at once. This idea was pulled from GSON, but I'm not sure it is worth porting to Java as it didn't seemed to be used really.

Other API's will also be affected by this:

* RealmClass should return the persisted name, not the public (Kotlin) name.
* RQL queries should support both the persisted and public names.

TODO:

// Step 1

* [ ] Create PersistedName annotation in `io.realm.kotlin.types.annotations`
* [ ] Modify `AccessorModifierIrGeneration.modifyPropertiesAndCollectSchema` to account for internal values
* [ ] Modify `RealmModelSyntheticPropertiesGeneration` (see line 501, maybe other places)
* [ ] Validate that persisted and public names are a unique set. They are not allowed to overlap. This check should happen at compile time.
* [ ] Verify in Realm Studio that schema is modified correctly

// Step 2

* [ ] Make sure that `RealmClass` returns the mapped values, not public names in the model class.

// Step 3

* [ ] Modify `PropertyInfo` to correctly set publicName
* [ ] Verify that queries can use both persisted and public names in queries.

elle-j:
## Description

Closes #590.

Adds support for `@PersistedName` annotation to map a public (Kotlin) field name to a persisted (Realm) field name.

Queries can be made using either of the names:

```kt
// Class with field using `@PersistedName`
class Example : RealmObject {
    @PersistedName("myPersistedName")     // or: @PersistedName(name = "myPersistedName")
    var myKotlinName = "My value"
}

// Query by Kotlin name
realm.query<Example>("myKotlinName = $0", "My value")
    .first()
    .find()?
    .myKotlinName

// Query by persisted name
realm.query<Example>("myPersistedName = $0", "My value")
    .first()
    .find()?
    .myKotlinName
```

##  TODO

* [x] Changelog entry
* [x] Tests (if applicable)


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/schema/RealmClass.kt
```diff
@@ -42,7 +42,7 @@ public interface RealmClass {
     public val isEmbedded: Boolean
 
     /**
-     * Index operator to lookup a specific [RealmProperty] from it's property name.
+     * Index operator to lookup a specific [RealmProperty] from it's internal property name.
```
cmelchior, line 45:
```suggestion
     * Index operator to lookup a specific [RealmProperty] from it's persisted property name.
```

`internal` feels a bit too, well internal, given our discussion, maybe persisted is as better way to phrase it

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/schema/RealmProperty.kt
```diff
@@ -21,7 +21,10 @@ package io.realm.kotlin.schema
  */
 public interface RealmProperty {
     /**
-     * Returns the name of the property in the object model.
+     * Returns the internal name of the property in the object model.
```
cmelchior, line 24:
```suggestion
     * Returns the persisted name of the property in the object model.
```

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -367,6 +383,9 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
         val fields: MutableMap<String, SchemaProperty> =
             SchemaCollector.properties.getOrDefault(irClass, mutableMapOf())
 
+        // A map for tracking the names (and their locations) used to ensure uniqueness
+        val internalAndPublicNames = mutableMapOf<String, CompilerMessageSourceLocation>()
```
cmelchior, line 387: #naming The name of this property is a bit vague.... but not sure I have an obvious better name... but this feels more like a `internalNameToSourceLocation` kind of thing...that one is also a mindful though

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmFieldTests.kt
```diff
@@ -0,0 +1,206 @@
...
+    // --------------------------------------------------
+    // Schema
+    // --------------------------------------------------
+
+    @Test
+    fun schema_realmProperty_usesInternalName() {
+        val classFromSchema = realm.schema()[RealmFieldSample::class.simpleName!!]!!
+
+        var realmFieldAnnotatedProperty = classFromSchema["internalNameStringField"]
+        assertNotNull(realmFieldAnnotatedProperty)
```
cmelchior, line 165: I would also assert that the storage type is correct

cmelchior (commented):
Nice 🚀 ...just left some minor comments for now

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
elle-j: Agreed, suggestions:
* `persistedAndPublicNamesToLocations`
* `persistedAndPublicNameToLocation`
* `persistedAndPublicNameLocations`
(or replace `Location` with `SourceLocation`)

packages/plugin-compiler/src/test/resources/invalid-sample/expected/00_ValidateIrBeforeLowering.ir
```diff
@@ -0,0 +1,698 @@
+// --- IR for <main> after Validate IR before lowering
```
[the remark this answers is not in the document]
elle-j: Don't think we actually need the "expected" file in this case. I had a small comment for that test:
```kt
// TODO Unnecessary check?
inputs.assertGeneratedIR()
```

But I assume I can remove that check as well as the expected file.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmFieldTests.kt
```diff
@@ -0,0 +1,206 @@
...
+
+        realmFieldAnnotatedProperty = classFromSchema["publicNameStringField"]
+        assertNull(realmFieldAnnotatedProperty)
+    }
+}
+
+private class RealmFieldSample : RealmObject {
```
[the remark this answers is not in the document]
elle-j: Depending on the upcoming tests to write I may add special classes using `@PersistedName`, but can see once those are done if it makes more sense to populate the existing `Sample` instead :)

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
cmelchior: IMO `persistedAndPublicNameToLocation` would be more descriptive than the current one.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
```diff
@@ -0,0 +1,312 @@
...
+            .find()
+            .single()
+            .run {
+                assertNotNull(this)
+                assertEquals(value, property.getValue(this, property))
+            }
+    }
+}
+
+class PersistedNameSample : RealmObject {
```
cmelchior, line 280: I would try to have 1 or 2 fields stress the unicode support, ie. what happens if we use an emojii as a persisted name? Core should support this.

cmelchior (changes requested):
Really nice. I mostly have some comments around testing

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -733,4 +766,39 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
...
+                "Names must contain at least 1 character.",
+                location
+            )
+        }
+        if (existingNames.containsKey(name)) {
+            val duplicationLocation = existingNames[name]!!
+            if (location.line == duplicationLocation.line) {
+                // TODO Use `logWarn` instead, but it seems that it is not part of the compiler messages
+                //      needed for the test `generate compilation error for invalid PersistedName annotations`
+                logError(
```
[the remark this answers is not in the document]
elle-j: The `messages` from `KotlinCompilation.Result` still does not seem to contain the message when using `logInfo` or `logWarn`, only `logError` and `logDebug`.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -560,6 +591,8 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
                                     putValueArgument(
                                         arg++,
                                         if (type == linkingObjectType) {
+                                            // FIXME Link property name must be the persisted name of the backlink (linking object)
```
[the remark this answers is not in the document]
elle-j: Added it more as a reminder for this PR (along with the TODO item in the PR description) since it needs to be fixed together with these changes, otherwise it will break backlinks if adding `@PersistedName` annotation to a backlink. So in this case, a separate issue shouldn't be created right?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
elle-j: Good idea, will add this.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
```diff
@@ -0,0 +1,312 @@
...
+
+    @PersistedName("persistedNameStringListField")
+    var publicNameStringListField: RealmList<String> = realmListOf()
+
+    @PersistedName("persistedNameStringSetField")
+    var publicNameStringSetField: RealmSet<String> = realmSetOf()
+
+    @PersistedName("persistedNameChildField")
+    var publicNameChild: PersistedNameSample? = null
+}
```
elle-j, line 312: Which of these remaining fields do you think we should keep for testing purposes? (i.e. we could perhaps remove some)

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/annotations/PersistedName.kt
```diff
@@ -17,7 +17,7 @@
 package io.realm.kotlin.types.annotations
 
 @Retention(AnnotationRetention.SOURCE)
-@Target(AnnotationTarget.FIELD)
+@Target(AnnotationTarget.FIELD, AnnotationTarget.PROPERTY)
```
gagik, line 20: not sure if this is needed actually, I was originally considering a different scenario where the backlinks child has a persisted name rather than parent

packages/plugin-compiler/src/test/resources/sample/input/Sample.kt
```diff
@@ -144,6 +144,9 @@ class Sample : RealmObject {
     val linkingObjectsByList by backlinks(Sample::objectListField)
     val linkingObjectsBySet by backlinks(Sample::objectSetField)
 
+    @PersistedName(name ="persistedLinkingObjects")
+    val linkingObjectsWithPersistedName by backlinks(Sample::objectListField)
```
gagik, line 148: again, not sure if one would want a test for this, this is different from the test that we had so far but is probably still a scenario worth investigating

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/IrUtils.kt
```diff
@@ -531,10 +531,15 @@ fun getBacklinksTargetPropertyType(declaration: IrProperty): IrType? {
     }
 }
 
-fun getLinkingObjectPropertyName(backingField: IrField): Name {
+fun getLinkingObjectPropertyName(property: IrProperty): String {
```
gagik, line 534: I decided to change this to take `IrProperty` and return `String` to make this fit more general scenarios but I could revert that and try to do this a different way if this is not best behavior

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/IrUtils.kt
```diff
@@ -531,10 +531,15 @@ fun getBacklinksTargetPropertyType(declaration: IrProperty): IrType? {
...
+    if (property.hasAnnotation(PERSISTED_NAME_ANNOTATION)) {
+        val annotation: IrConstImpl<String> = property.getAnnotation(PERSISTED_NAME_ANNOTATION).getValueArgument(0) as IrConstImpl<String>
```
gagik, line 536: I am not sure if this is the nicest way to do this in Kotlin, is there a better syntax which would do a check whether annotation exists and then do a cast and return value?

gagik: outdated: will most likely change this back

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/IrUtils.kt
gagik: will need to change this back since it targets backingField again now

CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * [Sync] `App.close()` have been added so it is possible to close underlying ressources used by the app instance.
 * [Sync] Add support for progress listeners with `SyncSession.progress`. (Issue [#428](https://github.com/realm/realm-kotlin/issues/428))
 * Add better error messages when inheriting `RealmObject` with unsupported class types. (Issue [#1086](https://github.com/realm/realm-kotlin/issues/1086))
+* Added support for `@PersistedName` annotations for mapping a Kotlin field name to the underlying field name persisted in the Realm. (Issue [#590](https://github.com/realm/realm-kotlin/issues/590))
+  ```kt
```
cmelchior, line 16: we don't have code snippets in the Changelog but leave that for the source code documentation. I would remove this

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/annotations/PersistedName.kt
cmelchior: I was looking a bit for a description between what is a field and what is a property. I found this discussion here: https://discuss.kotlinlang.org/t/property-vs-field/8179/12

And after some experiments, I think we only want to apply this annotation to `AnnotationTarget.FIELD` which is the field backing a property and the one we store in Realm.

E.g. you can have properties without backing fields (which Realm cannot persist), e.g.:

```
    // AnnotationTarget.FIELD cannot be applied to this
    var name
        get() = "foo"
        set(value) {
            _other_field = value
        }
```

packages/plugin-compiler/src/test/kotlin/io/realm/kotlin/compiler/GenerationExtensionTest.kt
```diff
@@ -163,6 +163,10 @@ class GenerationExtensionTest {
             // Primary key
             "id" to PropertyType.RLM_PROPERTY_TYPE_INT,
 
+            // @PersistedName annotated fields
+            "persistedNameStringField1" to PropertyType.RLM_PROPERTY_TYPE_STRING,
```
clementetb, line 167: Would be great to have a persisted name on a linking objects too

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/PersistedNameTests.kt
```diff
@@ -0,0 +1,111 @@
...
+                        @PersistedName("sameName")
+                        var sameName: String? = ""
+                    }
+                """.trimIndent()
+            )
+        )
+
+        assertEquals(KotlinCompilation.ExitCode.OK, result.exitCode)
+        val compilerLog = result.messages
+        assertContains(compilerLog, "The Kotlin name and the persisted name are the same value: 'sameName'")
```
cmelchior, line 71: Nitpick, throwing a build error on this is a little pedantic since it would be valid, just redundant. Can we turn this into a warning instead?

cmelchior (approved):
Great 🚀 . Only minor comments left then this can be merged.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
```diff
@@ -0,0 +1,370 @@
...
+            }
+        }
+
+        val queriedChild = realm.query<PersistedNameChildSample>()
+            .find()
+            .single()
+
+        assertEquals(5, queriedChild.parents.size)
+        assertEquals(1, queriedChild.parents.query("id = 3").find().size)
+    }
```
clementetb, line 219: I miss a test to validate that persisted names are accessible via named backlink queries: https://www.mongodb.com/docs/realm/realm-query-language/#backlink-queries

examples/kmm-sample/androidApp/src/main/java/io/realm/example/kmmsample/androidApp/PersistedNameTest.kt
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2022 JetBrains s.r.o.
```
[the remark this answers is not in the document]
gagik: @elle-j was there a reason for that?

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/PersistedNameTests.kt
gagik: Would this be a build error? Wouldn't it be an assertion error at runtime?

cmelchior: Hmm, maybe I misread this. I can see that the return value is `KotlinCompilation.ExitCode.OK` which seems to indicate that it works correctly.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/annotations/PersistedName.kt
gagik: I am not sure if I understood this correctly but would it not actually imply we would still be targeting properties like @clementetb suggests? I think the backing field is not important here because the annotation is applied by the user onto the original property in the Realm object rather than its backing field. Also for cases like backlinks like
```
    @PersistedName(name = "persistedNameParent")
    val publicNameParent by backlinks(Sample::publicNameChild)
```
would error if we only target fields

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/PersistedNameTests.kt
elle-j: Yes this is only a warning (not an error) via `logDebug()`.

examples/kmm-sample/androidApp/src/main/java/io/realm/example/kmmsample/androidApp/PersistedNameTest.kt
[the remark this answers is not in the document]
elle-j: No that was probably added when copied. It can be removed.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/annotations/PersistedName.kt
cmelchior: Hmm, yeah, I forgot about backlinks. True. In that case, supporting just properties are fine.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -495,10 +526,10 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
                                 ).apply {
                                     dispatchReceiver = irGetObject(propertyClass.companionObject()!!.symbol)
                                     var arg = 0
-                                    // Name
-                                    putValueArgument(arg++, irString(entry.key))
+                                    // Persisted name
+                                    putValueArgument(arg++, irString(persistedName))
                                     // Public name
-                                    putValueArgument(arg++, irString(""))
+                                    putValueArgument(arg++, irString(publicName))
```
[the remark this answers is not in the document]
elle-j: Wouldn't [this test](https://github.com/realm/realm-kotlin/blob/27308439fd02e57d2e08a976b7c401d19a5d816b/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt#L319) cover that scenario?

Or do you mean specifically test for when (a) the public name is `""` then (b) is set to a non-empty string? (Rather than (a) non-empty string then (b) a different non-empty string as in the existing test). In this case, I think we can just add a similar test as the one I linked to but have the `oldSchema` be a class with a property without the `@PersistedName` (since all public names will be `""` if not using the annotation), and have the `newSchema` be a class with a property with a `@PersistedName`.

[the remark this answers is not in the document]
rorbech: Ahhh. I thought that `publicName` was set to the same as `persistedName` when there was no `@PersistedName` on it. I guess we don't change anything in what is fed to the C-API then. 

I needed at some point we could verify that we don't trigger a migration by adding a pre-change realm-file and verify that we don't trigger migrations, but I guess that would require #577

examples/kmm-sample/androidApp/src/main/java/io/realm/example/kmmsample/androidApp/PersistedNameTest.kt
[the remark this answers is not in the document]
rorbech: The whole `kmmsample` was originally cloned from the official Jetbrains KMM Sample so we kept the reference.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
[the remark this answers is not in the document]
elle-j: Okay, yeah no the public name is always `""` unless `@PersistedName` is used. Yeah when #577 is done we could add another test using a realm file.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -532,13 +550,14 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
-                        fields[name] = SchemaProperty(
+                        val schemaProperty = SchemaProperty(
                             propertyType = PropertyType.RLM_PROPERTY_TYPE_OBJECT,
                             declaration = declaration,
                             collectionType = CollectionType.NONE
                         )
+                        fields[name] = schemaProperty
                         modifyAccessor(
-                            declaration,
+                            schemaProperty,
```
elle-j, line 560:
```suggestion
                            property = schemaProperty,
```

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/AccessorModifierIrGeneration.kt
```diff
@@ -549,13 +568,16 @@ class AccessorModifierIrGeneration(private val pluginContext: IrPluginContext) {
...
                             propertyType = PropertyType.RLM_PROPERTY_TYPE_OBJECT,
                             declaration = declaration,
                             collectionType = CollectionType.NONE
                         )
+                        fields[name] = schemaProperty
+                        // Current getObject/setObject has it's own public->storagetype->realmvalue
+                        // conversion so bypass any converters in accessors
                         modifyAccessor(
-                            declaration,
+                            schemaProperty,
```
elle-j, line 580:
```suggestion
                            property = schemaProperty,
```

PR #1139, opened 2022-11-18
Improve error message when forgetting plugin.

### LINKED ISSUE #1060: Improve error message when forgetting to apply plugin
When you forget to apply the Realm Kotlin plugin, the build will fail with a rather unhelpful error message as in https://github.com/realm/realm-kotlin/issues/1023

We should improve the error message to indicate that this might be the problem

elle-j:
## Description

Closes #1060.

If KMM users forget to add the Kotlin plugin, the error message is now more appropriate.

##  TODO

* [ ] <s>Changelog entry</s>
* [x] Tests (if applicable)

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
@@ -16,6 +16,7 @@
 
 package io.realm.kotlin
 
+import io.realm.kotlin.internal.MISSING_PLUGIN_MESSAGE
 import io.realm.kotlin.internal.REALM_FILE_EXTENSION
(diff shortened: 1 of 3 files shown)


rorbech (commented):
Nice, but a bit dangerous to include a link as theres pages could be moving around. Maybe reuse/align it with https://github.com/realm/realm-kotlin/blob/6cc4f1c48bdf64abc4d842e7c5b9f358ecb4ce1e/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmUtils.kt#L55 so we only have one phrasing of this and avoid the reference.

rorbech (approved):
You can signal that you don't think this requires a changelog entry by ~~striking it through~~ in the TODOs

Issue #1123, opened 2022-11-10
Duplicate UUID generated with RealmUUID.random()


mohit-sharma-87:
Hello, 

I have noticed that UUID random function is not returning unique IDs and the issue gets fixed only when the app is reinstalled. 

My sample implementation : 

![Code snippets](https://user-images.githubusercontent.com/1336626/201086410-792a0bde-1cad-41cc-b982-51662ba821bb.png)
![Code snippets](https://user-images.githubusercontent.com/1336626/201086416-c8b73407-5842-4c04-b089-0433a3dd420d.png)

cmelchior:
I have seen the same issue, but only when rerunning apps on the Android Emulator, so I suspect it is a problem with the random generator reusing the same seed in those cases. It should not impact release apps, but we should probably still switch to SecureRandom (which isn't available in Kotlin Multiplatform, but we should be able to get platform implementations for it)

PR #1146, opened 2022-11-22
Add secure random UUID generation

gagik:
Fixes https://github.com/realm/realm-kotlin/issues/1123


packages/library-base/src/androidMain/kotlin/io/realm/kotlin/internal/platform/PRNGFixes.kt
```diff
@@ -0,0 +1,358 @@
...
+import java.security.Security
+
+/**
+ * Fixes for the output of the default PRNG having low entropy.
+ *
+ * The fixes need to be applied via [.apply] before any use of Java
+ * Cryptography Architecture primitives. A good place to invoke them is in the
+ * application's `onCreate`.
+ */
+@Suppress("TooGenericExceptionThrown", "MagicNumber")
```
gagik, line 60: I am generally leaving the code as unchanged as possible, so I am suppressing the generic exception error here as well but if better exceptions are preferred I could change that, same for others.

gagik:
Seems like the motivation behind PRNGFixes (and the actual code snippet) is taken directly from [Android Dev Blogs](https://android-developers.googleblog.com/2013/08/some-securerandom-thoughts.html). I am not exactly well-versed in cryptographic things and I am unsure whether the UUIDs would need to be cryptographically random for security purposes (I'd assume not considering they have not been), so perhaps including something like that is an overkill and maybe even a bad idea? I also [found an issue where it might lead to some errors with some devices](https://stackoverflow.com/questions/19767349/android-securerandom-fix-prngfixes-and-prevention-infomation-message) so I removed it for now to see how it is like with CI.

packages/library-base/src/androidMain/kotlin/io/realm/kotlin/internal/platform/PRNGFixes.kt
```diff
@@ -0,0 +1,358 @@
...
+object PRNGFixes {
```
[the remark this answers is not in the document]
gagik: Seems like the motivation behind PRNGFixes (and the actual code snippet) is taken directly from [Android Dev Blogs](https://android-developers.googleblog.com/2013/08/some-securerandom-thoughts.html). I am not exactly well-versed in cryptographic things and I am unsure whether the UUIDs would need to be cryptographically random for security purposes (I'd assume not considering they have not been), so perhaps including something like that is an overkill and maybe even a bad idea? I also [found an issue where it might lead to some errors with some devices](https://stackoverflow.com/questions/19767349/android-securerandom-fix-prngfixes-and-prevention-infomation-message) so I removed it for now to see how it is like with CI.

packages/library-base/src/androidMain/kotlin/io/realm/kotlin/internal/platform/SecureRandomAndroid.kt
```diff
@@ -0,0 +1,45 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+
+ * See: https://github.com/korlibs/korge/tree/main/krypto
+*/
+
+package io.realm.kotlin.internal.platform
```
cmelchior, line 31: We don't need to duplicate this aacross both `androidMain` and `jvmMain`. It can be placed in `jvm` which is shared between the two.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+package io.realm.kotlin.internal.platform
+
+import kotlin.random.Random
+
+internal expect fun fillRandomBytes(array: ByteArray)
+internal expect fun seedExtraRandomBytes(array: ByteArray)
+
+internal fun arraycopy(src: ByteArray, srcPos: Int, dst: ByteArray, dstPos: Int, count: Int) = src.copyInto(dst, dstPos, srcPos, srcPos + count)
```
cmelchior, line 38: It looks like Kotlin has this built in: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/copy-into.html so I would use that instead

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
```diff
@@ -0,0 +1,66 @@
...
+
+@Suppress("MagicNumber")
+internal object SecureRandom : Random() {
```
cmelchior, line 41: The methods implemented here seem to be enough to cover all the methods available on `Random`, but that is a bit unclear right now. I would probably add some class documentation about it, just so it is clear.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
```diff
@@ -0,0 +1,66 @@
...
+        val temp = ByteArray(4)
+        fillRandomBytes(temp)
+        val a = temp[0].toInt() and 0xFF
+        val b = temp[1].toInt() and 0xFF
+        val c = temp[2].toInt() and 0xFF
+        val d = temp[3].toInt() and 0xFF
+        return (a shl 24) or (b shl 16) or (c shl 8) or (d shl 0)
+    }
+
+    override fun nextBytes(array: ByteArray, fromIndex: Int, toIndex: Int): ByteArray {
```
cmelchior, line 56: Any reason we are overriding this? `Random` has an implementation that fallback to `nextBits`...but maybe this is just more efficient?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
cmelchior: We are also missing tests of this class. Looking at the implementation of `Random`, just testing `nextBites` and `addSeed` seems to be enough, but if we override `nextBytes` we should also test that.

cmelchior (changes requested):
Great start 👍 . Some minor comments and we are missing a changelog update (i.e. fixing https://github.com/realm/realm-kotlin/issues/1123) and tests.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
gagik: I largely just went with Korge's implementation so did not make this decision myself, but yeah I imagine it might be slightly more efficient as I suppose they might be making this nextBytes-centric(?) since this is what gets used more for their (and our?) cases?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/SecureRandomTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+package io.realm.kotlin.test
+
+import io.realm.kotlin.internal.platform.SecureRandom
+import kotlin.test.Test
+import kotlin.test.assertNotEquals
+
+internal class SecureRandomTest {
+    @Test
+    fun nextBytes_ReturnsDifferentBytes() {
```
gagik, line 29: Not sure what other tests would be meaningful, this is also basically as far as testing goes in Korge

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
gagik: Hm, not sure why but removing `nextBytes` override does lead to errors, I am not sure why the fallback to `nextBits` is not working because it does run those lines but ends up with just empty lists

cmelchior: You cannot have a `fun nextBytes` method in SecureRandom if `Random` contains the same method. Was that what you where seeing?

cmelchior: In any case, I think having this is fine, as long as we test it.

gagik: Nope, I was trying to remove the function all together so it would inherit the `Random.nextBytes` but I think something would go wrong with that. It'd run the `nextBytes` code from `Random` but just give out 0s. I will look into this for a little more just to make sure this is not a problem with `nextBits` it uses.

cmelchior: Yeah, that definitely sounds like something that needs to be looked into 🤔

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
```diff
@@ -0,0 +1,63 @@
...
+        val random = ByteArray(toIndex - fromIndex)
+        fillRandomBytes(random)
+        random.copyInto(array, fromIndex, 0, random.size)
+        return array
+    }
+
+    override fun nextBits(bitCount: Int): Int {
+        /** Takes upper [bitCount] bits (0..32) from random integer.
+         *  See: https://github.com/JetBrains/kotlin/blob/fe81ad0bbe413af2b0485afec4ff0317b7b27888/
```
gagik, line 58: So interestingly (and perhaps worryingly) enough I am pretty sure the `nextBits` implementation was not working correctly, i.e. if one ran `SecureRandom.nextInt()` you would always get 0, which is why when I removed the `nextBytes` implementation, it would not work correctly since the fallback uses `nextInt`. I changed this part of the logic to replicate what Kotlin is doing now (`getInt` does seem to work correctly, so it's was an issue with regards to doing the right bitwise shifts). I also added some `nextInt` tests since those would have failed before with the incorrect `nextBits`. This does not matter for UUID purposes since that used `nextBytes` anyways, but now this should work properly as a drop-in `Random` replacement. Curious if Korge uses this anywhere and if it is actually wrong.

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * Internal dispatcher threads would leak when closing Realms. (Issue [#818](https://github.com/realm/realm-kotlin/issues/818))
 * Realm finalizer thread would prevent JVM main thread from exiting. (Issue [#818](https://github.com/realm/realm-kotlin/issues/818))
 * `RealmUUID` did not calculate the correct `hashCode`, so putting it in a `HashSet` resulted in duplicates.
-* JVM apps on Mac and Linux would use a native file built in debug mode, making it slower than needed. The correct native binary built in release mode is now used. Windows was not affected. (Isse [#1124](https://github.com/realm/realm-kotlin/pull/1124))
+* JVM apps on Mac and Linux would use a native file built in debug mode, making it slower than needed. The correct native binary built in release mode is now used. Windows was not affected. (Issue [#1124](https://github.com/realm/realm-kotlin/pull/1124))
```
gagik, line 18: Figured it was worthwhile to fix a minor typo here

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
gagik: Yeah, I think something was wrong: https://github.com/realm/realm-kotlin/pull/1146#discussion_r1036924404

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
...
+* `RealmUUID.random()` did not return unique IDs unless the app is reinstalled. (Issue [#1123](https://github.com/realm/realm-kotlin/pull/1123))
```
cmelchior, line 19: This is probably a bit more alarmist than it needed to be, so I reworded it a bit.

```suggestion
* RealmUUID.random()` would generate the same values when an app was re-launched from Android Studio during development. (Issue [#1123](https://github.com/realm/realm-kotlin/pull/1123)) 
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/platform/SecureRandom.kt
cmelchior: Great catch 👍

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/SecureRandomTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        assertNotEquals(SecureRandom.nextBytes(16).toList(), SecureRandom.nextBytes(16).toList())
```
cmelchior, line 30: These can be written as:

```
repeat(3) {
    assertNotEquals(SecureRandom.nextBytes(16).toList(), SecureRandom.nextBytes(16).toList())
}
```

Which also makes it easier to modify if needed to.

cmelchior (approved):
Looks good. I have a few suggestions for improvements, but otherwise this is good to merge when CI is green (I fixed the issue, so it can run now).

PR #1156, opened 2022-12-01
GHA: Deploy snapshots

cmelchior:
Enable releasing snapshots from CI. 

This is done through a custom Kotlin Script that uses mvn under the hood to upload SNAPSHOTs to Maven Central.

While kts is experimental from JetBrains, it is much more powerful than simple shell script (+uses the same language as the rest of the repo) and even in the worst case where kts support might be deprecated, we can move the code to a cmdline jar file.

The script is rather verbose, which is by design as I wanted to make it easy to debug any failures it might run into.

This PR also contains some fixes to Windows caching that I discovered while working on this.

---
Sources uses for inspiration:
Looks like we might be able to use Maven to do this: https://theoverengineered.blog/posts/publishing-my-first-artifact-to-maven-central-using-github-actions

https://jeroenmols.com/blog/2021/03/24/migrate-artifacts-mavencentral/


cmelchior:
While this PR is being merged into a feature branch that is WIP. It would be nice with some sanity checks around the kotlin release script 🙏

tools/publish_snapshots.main.kts
```diff
@@ -0,0 +1,277 @@
...
+// # Import base64 key with a passphrase
+// echo $BASE64_SIGNING_KEY | base64 -d | gpg --batch --import
+val gpgBase64SigningKey = args[2]
+val gpgPassPhrase = args[3]
+// Maven Central username and password token.
+// Can be found by logging in to https://oss.sonatype.org/#welcome and go to Profile > User Token.
+val sonatypeUsername = args[4]
+val sonatypePassword = args[5]
+// Path to the root of the local m2 repo containing the package to release.
+val localMavenRepo = File("$repoPath/packages/build/m2-buildrepo")
```
rorbech, line 63: Should we consider to somehow hook this up with https://github.com/realm/realm-kotlin/blob/main/packages/gradle.properties#L35

tools/publish_snapshots.main.kts
```diff
@@ -0,0 +1,277 @@
...
+ * Iterate a folder and return all files that needs to be considered as part of a Maven Artifact.
+ */
+fun iteratePackageFiles(directory: File): Sequence<File> {
+    return directory.walkTopDown()
+        .filter { it.isFile }
+        .filterNot { ignoreFiles.contains(it.name) }
+        .filter { file ->
+            ignoredFileTypes.firstOrNull {fileType ->
+                file.name.endsWith(fileType)
+            } == null
```
rorbech, line 145:
```suggestion
            ignoredFileTypes.none { fileType -> file.name.endWith(fileType) }
```

rorbech (approved):
Awesome. Quite some infrastructure. Not completely into GHA and guess it is hard to track all the details about file-filtering, but if it works 🙈 

I guess we lack a selector for which branch to execute this step for or will that be in the feature branch? 

... and maybe we need a specific verifaction step that tests the actual uploaded repository for succeeding. Seems like there will be a big chance that we will fail catching all the necessary files if there are future updates to the Kotlin Multiplatform metadata infrasture 🙈 

Could we consider to use a real sonatype-staging repository for the final release, so that we could publish the artifacts for the various platforms directly from each machine, test it and the just publish the staged repository on succesful completion of tests, etc.

cmelchior:
Right now this runs every time on the feature branch. In later PR's this will be changed so snapshots are only deployed from release branches.

Issue #744, opened 2022-03-24
EmailPasswordAuth: Add support for callResetPasswordFunction


cmelchior:
Add support for `EmailPasswordAuth.callResetPasswordFunction()`. This cannot be implemented until #648 is complete.

PR #1160, opened 2022-12-02
EmailPasswordAuth: Add support for callResetPasswordFunction

gagik:
Closes https://github.com/realm/realm-kotlin/issues/744

This is branched off of [ct/app-functions](https://github.com/realm/realm-kotlin/tree/ct/app-functions) so to make this less confusing for review purposes I am making the PR with that target, but will need to be change once that is in master. This is also dependent on updating to the latest version of Core which includes https://github.com/realm/realm-core/pull/6081.


packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -152,45 +153,50 @@ class EmailPasswordAuthWithAutoConfirmTests {
...
+                adminApi.setResetFunction(enabled = false)
+            }
+        }
+    }
+
+    @Test
+    fun callResetPasswordFunction_invalidArgumentsThrows() {
+        val provider = app.emailPasswordAuth
+        runBlocking {
+            assertFailsWith<IllegalArgumentException> { provider.callResetPasswordFunction("", "password") }
```
gagik, line 196: I assume you cannot use nulls here at runtime so this is the closest invalid argument I came up with based on old tests, not sure if there is more I should check here.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -152,45 +153,50 @@ class EmailPasswordAuthWithAutoConfirmTests {
...
+                adminApi.setResetFunction(enabled = false)
+            }
+        }
+    }
+
+    @Test
+    fun callResetPasswordFunction_invalidServerArgsThrows() {
+        val provider = app.emailPasswordAuth
+        val adminApi = app.asTestApp
+        runBlocking {
```
gagik, line 178: I just modified the old tests slightly, mainly also wrapping them in `runBlocking` to satisfy coroutine scope requirements, but not sure if more test cases are necessary or if there is anything I should change.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -2464,6 +2464,30 @@ actual object RealmInterop {
...
+    ) {
+        memScoped {
+            checkedBooleanResult(
+                realm_wrapper.realm_app_email_password_provider_client_call_reset_password_function(
+                    app.cptr(),
+                    email,
+                    newPassword.toRString(this),
+                    serializedEjsonPayload,
+                    staticCFunction { userData, error ->
+                        handleAppCallback(userData, error) { /* No-op, returns Unit */ }
```
gagik, line 2482: not completely sure if this is correct, just going off of similar functions

gagik (commented):
Just some things I am unsure about

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
cmelchior: This looks fine 👍

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/EmailPasswordAuth.kt
```diff
@@ -87,18 +87,17 @@ public interface EmailPasswordAuth {
...
-    //  *
-    //  * @param email the email of the user.
-    //  * @param newPassword the new password of the user.
-    //  * @param args any additional arguments provided to the reset function. All arguments must
-    //  * be able to be converted to JSON compatible values using `toString()`.
-    //  * @throws AppException if the server failed to confirm the user.
-    //  */
-    // public suspend fun callResetPasswordFunction(email: String, newPassword: String, vararg args: Any?)
+    /**
+     * Call the reset password function configured to the
```
cmelchior, line 91:
```suggestion
     * Call the reset password function configured for the
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/EmailPasswordAuth.kt
```diff
@@ -87,18 +87,17 @@ public interface EmailPasswordAuth {
...
+     * [Credentials.Provider.EMAIL_PASSWORD] provider.
+     *
+     * @param email the email of the user.
+     * @param newPassword the new password of the user.
+     * @param args any additional arguments provided to the reset function. All arguments must
```
cmelchior, line 96: I know this was in our old docs, but I don't think this is the case. The restriction is that the values can be converted to a BsonValue.

So maybe rewrite to something like 

```
@param args any additional arguments provided to the reset function. All arguments be able to be converted to a [BsonValue].
``

File: packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/EmailPasswordAuth.kt [hunk 6]
```diff
@@ -87,18 +87,17 @@ public interface EmailPasswordAuth {
...
+     * be able to be converted to JSON compatible values using `toString()`.
+     * @throws AppException if the server failed to confirm the user.
```
cmelchior, line 98: Is this true? Similar to the discussion here https://github.com/realm/realm-kotlin/pull/1131/files#r1034502924 we should probably return an `AuthException` if there is an error in the function?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/EmailPasswordAuthTests.kt
```diff
@@ -152,45 +153,50 @@ class EmailPasswordAuthWithAutoConfirmTests {
...
+            adminApi.setResetFunction(enabled = true)
+            val email = TestHelper.randomEmail()
+            provider.registerUser(email, "123456")
+            try {
+                provider.callResetPasswordFunction(email, "new-password", "wrong-magic-word")
+            } catch (error: AppException) {
```
cmelchior, line 184: What error code is being returned here? If it is something specific we should consider if it fits in a more specific exception type.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/EmailPasswordAuthTests.kt
cmelchior: That is correct, null pointer errors are not allowed here anymore. But instead we have a new class of errors. If we accept `Any?` as the argument type, we can test here what happens if you use a type that cannot be turned into a `BsonValue`

cmelchior (commented):
Nice 👍 . I only have minor comments:

- Missing changelog entry
- Some questions about throwing `AppException`. It is a very generic error, so if it is possible to do something more fine-grained that would be great.
- And yeah, then we should obviously not merge this until the Core PR has been merged so we can depend on a commit on master.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/EmailPasswordAuth.kt
gagik: it is actually `ServiceException`, which I tried to change but now that I think about it makes sense perhaps since this is (in theory) an issue with the reset password service and not the authentication (at least that's why I assume this is the error? I know `AuthException` is a subset of that but it does fail if I try to assert specifically that)

cmelchior (commented):
Looks like something is off with the Core commit since RealmInterop fails to compile

gagik:
> Looks like something is off with the Core commit since RealmInterop fails to compile

I think the issue is that this is still branched off of the app functions branch so it uses old core and not the latest main branch one. This will have to wait until https://github.com/realm/realm-kotlin/pull/1110 is merged after which it can be rebased and pass CI without a divergent core.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/auth/EmailPasswordAuth.kt
```diff
@@ -87,18 +87,17 @@ public interface EmailPasswordAuth {
...
-    //  */
-    // public suspend fun callResetPasswordFunction(email: String, newPassword: String, vararg args: Any?)
+    /**
+     * Call the reset password function configured for the
+     * [Credentials.Provider.EMAIL_PASSWORD] provider.
+     *
+     * @param email the email of the user.
+     * @param newPassword the new password of the user.
+     * @param args any additional arguments provided to the reset function.
+     * All arguments be able to be converted to a [BsonValue].
```
cmelchior, line 97:
```suggestion
     * All arguments must be able to be converted to a [BsonValue].
```

cmelchior (approved):

Issue #1086, opened 2022-10-21
Error message around Emum support


mohit-sharma-87:
Requesting to add error message whenever Enum class is added in schema set.

```
enum class MyObject : RealmObject {

}
```

PR #1161, opened 2022-12-02
Add error message for unsupported class types and objects

gagik:
Similar to https://github.com/realm/realm-kotlin/pull/698 but for enum classes, object declarations, and anonymous objects.

Closes https://github.com/realm/realm-kotlin/issues/1086

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * Add support for `Realm.copyFromRealm()`. All RealmObjects, RealmResults, RealmList and RealmSets now also have a `copyFromRealm()` extension method.
 * [Sync] `App.close()` have been added so it is possible to close underlying ressources used by the app instance.
 * [Sync] Add support for progress listeners with `SyncSession.progress`. (Issue [#428](https://github.com/realm/realm-kotlin/issues/428))
+* Add better error messages when inheriting `RealmObject` with unsupported class types. (Issue [#1086](https://github.com/realm/realm-kotlin/issues/1086))
 
 ### Fixed
 * Internal dispatcher threads would leak when closing Realms. (Issue [#818](https://github.com/realm/realm-kotlin/issues/818))
--- a/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
+++ b/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
@@ -35,6 +35,7 @@ import org.jetbrains.kotlin.ir.util.companionObject
 import org.jetbrains.kotlin.ir.util.defaultType
 import org.jetbrains.kotlin.ir.util.isAnonymousObject
 import org.jetbrains.kotlin.ir.util.isEnumClass
+import org.jetbrains.kotlin.ir.util.isObject
 import org.jetbrains.kotlin.ir.util.kotlinFqName
 import org.jetbrains.kotlin.ir.util.parentAsClass
 import org.jetbrains.kotlin.ir.util.primaryConstructor
@@ -55,17 +56,18 @@ private class RealmModelLowering(private val pluginContext: IrPluginContext) : C
 
     override fun lower(irClass: IrClass) {
         if (irClass.isBaseRealmObject) {
-            // We don't support data class
+            // Throw error with classes that we do not support
             if (irClass.isData) {
                 error("Data class '${irClass.kotlinFqName}' is not currently supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
             }
-            // We don't support object
-            if (irClass.isAnonymousObject) {
-                error("Object '${irClass.parent.kotlinFqName}' is not supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
-            }
-            // We don't support enum class
             if (irClass.isEnumClass) {
                 error("Enum class '${irClass.kotlinFqName}' is not supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
+            }
+            if (irClass.isObject) {
+                error("Object declarations are not supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
+            }
+            if (irClass.isAnonymousObject) {
+                error("Anonymous objects are not supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
             }
             // For native we add @ModelObject(irClass.Companion::class) as associated object to be
             // able to resolve the companion object during runtime due to absence of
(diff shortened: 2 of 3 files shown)


packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/ModelDefinitionTests.kt
```diff
@@ -81,4 +81,20 @@ class ModelDefinitionTests {
         assertEquals(KotlinCompilation.ExitCode.INTERNAL_ERROR, result.exitCode, "Compilation should fail when using data class")
         assertTrue(result.messages.contains("Data class 'Foo' is not currently supported"))
     }
+
+    @Test
+    fun `enum_class`() {
```
gagik, line 86: Not sure if I should name this differently

cmelchior: Looks fine to me 👍

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
```diff
@@ -57,6 +57,10 @@ private class RealmModelLowering(private val pluginContext: IrPluginContext) : C
             if (irClass.isData) {
                 error("Data class '${irClass.kotlinFqName}' is not currently supported.")
             }
+            // We don't support enum class
+            if (irClass.isEnumClass) {
+                error("Enum class '${irClass.kotlinFqName}' is not currently supported.")
```
cmelchior, line 62:
```suggestion
                error("Enum class '${irClass.kotlinFqName}' is not supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
```
I know this sentence was just copied from the data class above, but IMO we should rewrite it to something like what I proposed. It doesn't promise anything about future compatibility + point to what is the correct solution.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
```diff
@@ -57,6 +57,10 @@ private class RealmModelLowering(private val pluginContext: IrPluginContext) : C
             if (irClass.isData) {
                 error("Data class '${irClass.kotlinFqName}' is not currently supported.")
             }
+            // We don't support enum class
+            if (irClass.isEnumClass) {
```
cmelchior, line 61: I know this is outside the scope of this PR, but since it was this easy to add. Could we also add a check for `object Foo : RealmObject` here? (Singleton objects)

cmelchior (commented):
Really nice :+1: 

I have a small suggestion for the error message + extra tests.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
gagik: That makes sense, I wanted to suggest that when I noticed the word "currently"

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
```diff
@@ -57,6 +57,14 @@ private class RealmModelLowering(private val pluginContext: IrPluginContext) : C
             if (irClass.isData) {
                 error("Data class '${irClass.kotlinFqName}' is not currently supported.")
             }
+            // We don't support object
+            if (irClass.isAnonymousObject) {
+                error("Object '${irClass.parent.kotlinFqName}' is not supported. Only normal classes can inherit from 'RealmObject' or 'EmbeddedRealmObject'.")
```
gagik, line 62: I am calling them just objects but could also call it anonymous objects, I feel like the former is clearer considering the Kotlin keyword. also, because of the `irClass` seems to work with objects, I am referring to its parent's name here, which does yield the object variable name. I also assume we won't be supporting this in the future?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * Add support for `Realm.copyFromRealm()`. All RealmObjects, RealmResults, RealmList and RealmSets now also have a `copyFromRealm()` extension method.
 * [Sync] `App.close()` have been added so it is possible to close underlying ressources used by the app instance.
 * [Sync] Add support for progress listeners with `SyncSession.progress`. (Issue [#428](https://github.com/realm/realm-kotlin/issues/428))
+* Add better error messages when inheriting `RealmObject` with unsupported class types. (Issues [#1086](https://github.com/realm/realm-kotlin/issues/1086))
```
gagik: this issue only mentions enum, so perhaps need to link to somewhere else? i.e. this PR?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * Add support for `Realm.copyFromRealm()`. All RealmObjects, RealmResults, RealmList and RealmSets now also have a `copyFromRealm()` extension method.
 * [Sync] `App.close()` have been added so it is possible to close underlying ressources used by the app instance.
 * [Sync] Add support for progress listeners with `SyncSession.progress`. (Issue [#428](https://github.com/realm/realm-kotlin/issues/428))
+* Add better error messages when inheriting `RealmObject` with unsupported class types. (Issue [#1086](https://github.com/realm/realm-kotlin/issues/1086))
```
gagik, line 14: this issue only mentions enum, so perhaps need to link to somewhere else? i.e. this PR?

cmelchior: I would just update the PR description.

gagik: did you mean the issue description?

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelLoweringExtension.kt
cmelchior: It looks like you added support for detecting anonymous objects instead 😄 (which is great, we should probably have that anyway), like here: https://kotlinlang.org/docs/object-declarations.html#creating-anonymous-objects-from-scratch

So I would rename this to `Anonymous objects classes cannot inherit from RealmObject. Only normal ...`. I would not link to the parent name, that is just confusing IMO, and the compiler should have a link to the line as part of the error.

I was actually thinking about `object declarations` like here: https://kotlinlang.org/docs/object-declarations.html#object-declarations-overview

Not 100% sure how to phrase that well though, maybe something like `Object declarations cannot inherit from 'RealmObject' or 'EmbeddedRealmObject' cannot ....` 

That means the name of the object class will not be part of the name, but maybe that is acceptable.

CHANGELOG.md
cmelchior: Yes

cmelchior: It doesn't matter that much.. The title would probably be more accurate.

cmelchior (approved):
Awesome 🚀

PR #1163, opened 2022-12-07
Kotlin 1.8.0-Beta compatibility

rorbech:
This fixes some Kotlin 1.8.0-Beta compatibility issues due to removed default implementations on some Kotlin Compiler interfaces.

### DIFF
--- a/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticCompanionExtension.kt
+++ b/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticCompanionExtension.kt
@@ -47,6 +47,7 @@ import java.util.ArrayList
  *  interface.
  * Also adds the [RealmObjectCompanion] methods as synthetic methods on the companion.
  */
+@Suppress("EmptyFunctionBlock")
 class RealmModelSyntheticCompanionExtension : SyntheticResolveExtension {
 
     override fun getSyntheticCompanionObjectNameIfNeeded(thisDescriptor: ClassDescriptor): Name? {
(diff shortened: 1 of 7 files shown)


packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticCompanionExtension.kt
```diff
@@ -49,6 +57,48 @@ class RealmModelSyntheticCompanionExtension : SyntheticResolveExtension {
         }
     }
 
+    override fun addSyntheticSupertypes(
```
nhachicha, line 60: The interface [SyntheticResolveExtension](https://github.com/JetBrains/kotlin/blob/v1.8.0-Beta/compiler/frontend/src/org/jetbrains/kotlin/resolve/extensions/SyntheticResolveExtension.kt) didn't change [recently](https://github.com/JetBrains/kotlin/commit/c168a19a588fe78111f43cfa7b34de5d12195bb0#diff-ada58f4f3c851706692d7c4442730e6cae0971d5abc3a83fa85b6f67c9ddca44) I'm not sure why we have to provide these empty overrides below?

rorbech: Because 1.8.0 does not include the default implementations in the interface, so we cannot look up the implementations without it. 
```
Caused by: java.lang.ClassNotFoundException: org.jetbrains.kotlin.resolve.extensions.SyntheticResolveExtension$DefaultImpls
``` 
Probably due to a change in `-Xjvm-default` used for Kotlin

cmelchior (approved):
Nice 👍. Should we add a changelog entry as well? And we probably need to move this to `releases` since we are still debugging issues with Core 13.

PR #1164, opened 2022-12-08
Update to Core 13.1.2

cmelchior:


packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -73,7 +73,11 @@ actual object RealmInterop {
     }
 
     actual fun realm_refresh(realm: RealmPointer) {
-        realmc.realm_refresh(realm.cptr())
+        // Only returns `true` if the version changed, `false` if the version
+        // was already at the latest. Errors will be represented by the actual
+        // return value, so just ignore this out parameter.
+        val didRefresh = booleanArrayOf(false)
```
nhachicha, line 79: Isn't there a value returning the boolean to indicate whether we advanced to a new version or not?

cmelchior: That was the previous behavior, yes, but that conflicted with all other API's where the bool return value indicated if an error was thrown or not.

clementetb (approved):
Looking good! Happy once green on CI

PR #1165, opened 2022-12-09
Pretend to build Core as a submodule for MacOS

cmelchior:
This prevents generating parser files which made it harder to switch Git Submodule versions. This fix was verified locally to work.


clementetb (approved):
Thanks for solving this! It works, tested locally.

PR #1166, opened 2022-12-09
Bump to KBSON 0.2.0

### LINKED ISSUE #1153: Expected member name or ';' after declaration specifiers on iOS
When I upgrade **realm-kotlin** to 1.5.0, iOS app will report the following error, the error message is like this:

```
Expected member name or ';' after declaration specifiers
```

There are two lines on `shared.h` caused the error:

image

I think the reason is because **TRUE** and **FALSE** are keywords on Swift, I am not sure whether I didn't configure the Realm correctly or there already has the solution to solve the problem.

cmelchior:
Closes #1153 

I tried to find a way to test this, but apparently, this only surfaces when actually building the iPhone app and not when only building the shared code. So I just tested this manually and will defer testing this to another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,10 @@
 * None.
 
 ### Fixed
+* Fixed problem with KBSON using reservered keywords in Swift. (Issue [#1153](https://github.com/realm/realm-kotlin/issues/))
 * Fixed database corruption and encryption issues on apple platforms. (Issue [#5076](https://github.com/realm/realm-js/issues/5076))
 * [Sync] Bootstraps will not be applied in a single write transaction - they will be applied 1MB of changesets at a time. (Issue [#5999](https://github.com/realm/realm-core/pull/5999)).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Fixed database corruption and encryption issues on apple platforms. (Issue [#5076](https://github.com/realm/realm-js/issues/5076))
 * [Sync] Bootstraps will not be applied in a single write transaction - they will be applied 1MB of changesets at a time. (Issue [#5999](https://github.com/realm/realm-core/pull/5999)).
 * [Sync] Fixed a race condition which could result in operation cancelled errors being delivered to `Realm.open` rather than the actual sync error which caused things to fail. (Issue [#5968](https://github.com/realm/realm-core/pull/5968)).
+* [Sync] Fixed problem with KBSON using reservered keywords in Swift. (Issue [#1153](https://github.com/realm/realm-kotlin/issues/))
```
rorbech, line 13: It is not only related sync as the bson-dependency is in `cinterop`.

rorbech (approved):
Only a minor CHANGELOG detail. Otherwise LGTM

PR #1184, opened 2022-12-16
Fix `Long` overflow when comparing current time to `RealmInstant.MAX` and `MIN` values

### LINKED ISSUE #1182: Strange behavior with RealmInstant.now() on iOS
I am seeing some strange behavior with RealmInstant.now() on iOS. It works correctly on Android, but on iOS RealmInstant.now().epochSeconds is always 9223372036854775807. I am new to both KMM and Realm and probably doing something wrong, but cannot figure out what.

I created a new KMM App using Android Studio (2021.3.1 Patch 1), added realm dependencies to the build files, then changed the code in function greet() to the following:
```
fun greet(): String {
        val now = RealmInstant.now().epochSeconds
        println(now)
        return "Hello, ${platform.name}!\nnow: $now"
 }
```

Here are some other details:
Realm version: 1.5.1
Kotlin version: 1.7.20
iOS Simulator: iPhone 14 - iOS 16.2

edualonso:
Closes #1182 

We were hoping that converting the `RealmInstant.MAX` and `MIN` values to milliseconds would allow us to compare them to the current milliseconds value returned by `NSDate().timeIntervalSince1970`. However, the `MAX` and `MIN` values use `Long.MAX` and `Long.MIN` respectively and by multiplying them by `1000` we were overflowing causing the value to be `0` and always return `RealmInstant.MAX`.

This has been solved by comparing the value returned by `NSDate().timeIntervalSince1970` to the `MAX` and `MIN` values in seconds.

### DIFF
--- a/packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
+++ b/packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
@@ -20,7 +20,6 @@ import platform.Foundation.timeIntervalSince1970
 import platform.posix.pthread_threadid_np
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KType
-import kotlin.time.Duration.Companion.nanoseconds
 
 @Suppress("MayBeConst") // Cannot make expect/actual const
 public actual val RUNTIME: String = "Native"
--- a/packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
+++ b/packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
@@ -56,11 +56,12 @@ public actual fun epochInSeconds(): Long =
  */
 @Suppress("MagicNumber")
 internal actual fun currentTime(): RealmInstant {
-    val secs = NSDate().timeIntervalSince1970
+    val secs: Double = NSDate().timeIntervalSince1970
     return when {
-        // We can't convert the MIN value to ms as it will cause Long overflow so we have to compare directly against seconds
+        // We can't convert the MIN value to ms - it is initialized with Long.MIN_VALUE and
+        // multiplying it by 1000 will cause overflow. We have to compare directly against seconds
         secs < RealmInstant.MIN.epochSeconds -> RealmInstant.MIN
-        // Similarly here, compare to seconds instead
+        // Similarly here, compare to seconds instead to avoid overflow with Long.MAX_VALUE
         secs > RealmInstant.MAX.epochSeconds -> RealmInstant.MAX
         else -> {
             val millis = (secs * 1000 + if (secs > 0) 0.5 else -0.5).toLong()
--- a/packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
+++ b/packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
@@ -20,6 +20,7 @@ import platform.Foundation.timeIntervalSince1970
 import platform.posix.pthread_threadid_np
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KType
+import kotlin.time.Duration.Companion.nanoseconds
 
 @Suppress("MayBeConst") // Cannot make expect/actual const
 public actual val RUNTIME: String = "Native"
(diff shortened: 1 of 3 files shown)


packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -57,16 +57,16 @@ public actual fun epochInSeconds(): Long =
...
-    return if (millis < RealmInstant.MIN.epochSeconds * 1000) {
-        RealmInstant.MIN
-    } else if (millis > RealmInstant.MAX.epochSeconds * 1000) {
-        RealmInstant.MAX
-    } else {
-        RealmInstantImpl(
-            millis.floorDiv(1000.toLong()),
-            (millis.mod(1000.toLong()) * 1000).toInt()
-        )
+    return when {
```
cmelchior, line 60: Not 100% sure exactly what cases we intend to fix with this change? The Dokka written assumes that we are only calling this method with "reasonable" values which should work with the given code?

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -57,16 +57,16 @@ public actual fun epochInSeconds(): Long =
 @Suppress("MagicNumber")
 internal actual fun currentTime(): RealmInstant {
     val secs = NSDate().timeIntervalSince1970
```
cmelchior, line 59: Please add the type here, it looks to be Double. It was a bit hard to reason about the below stuff without that information.

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -57,16 +57,16 @@ public actual fun epochInSeconds(): Long =
...
-        RealmInstant.MIN
-    } else if (millis > RealmInstant.MAX.epochSeconds * 1000) {
-        RealmInstant.MAX
-    } else {
-        RealmInstantImpl(
-            millis.floorDiv(1000.toLong()),
-            (millis.mod(1000.toLong()) * 1000).toInt()
-        )
+    return when {
+        // We can't convert the MIN value to ms as it will cause Long overflow so we have to compare directly against seconds
```
cmelchior, line 61: IMO, I would be a bit uncomfortable merging this without unit tests, which means we need to refactor this method so we can inject test values.

Looking at the issue, I'm not sure I see how this was causing iOS to overflow. That doesn't feel like that should be possible for reasonable values of NSDate?

cmelchior (changes requested):

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
edualonso: The problem is that the code as it was written would always land here:
```
    } else if (millis > RealmInstant.MAX.epochSeconds * 1000) {
        RealmInstant.MAX // <-------
    } else {
```
because `RealmInstant.MAX` is initialised with `Long.MAX_VALUE`, which when multiplied by `1000` overflows and the result is `0`. Then all calls to `RealmInstant.now()` generate a value for `millis` that is greater than `0` thus the output always is `RealmInstant.MAX` on Darwin.

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
edualonso: That's fair. I will add a test that checks the value is not repeated, which was the case before. See explanation above for overflow.

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -56,17 +56,18 @@ public actual fun epochInSeconds(): Long =
...
-        )
+    val secs: Double = NSDate().timeIntervalSince1970
+    return when {
+        // We can't convert the MIN value to ms - it is initialized with Long.MIN_VALUE and
+        // multiplying it by 1000 will cause overflow. We have to compare directly against seconds
+        secs < RealmInstant.MIN.epochSeconds -> RealmInstant.MIN
+        // Similarly here, compare to seconds instead to avoid overflow with Long.MAX_VALUE
+        secs > RealmInstant.MAX.epochSeconds -> RealmInstant.MAX
+        else -> {
+            val millis = (secs * 1000 + if (secs > 0) 0.5 else -0.5).toLong()
```
clementetb, line 67: I think that we could use the Duration framework to simplify these calculations.

clementetb (approved):
Looking good to me

PR #1190, opened 2022-12-21
Add test stage with minified Android builds

### LINKED ISSUE #1114: Run Android tests with minify enabled
We need to exercise the code path of the API while Proguard is enabled to detect issues similar to https://github.com/realm/realm-kotlin/issues/1106 
We can do this in GHA using an emulator in debug mode but with `minfyEnabled= true` to trigger Proguard obfuscation

rorbech:
This adds a `test-sync` test stage running all sync tests with minification on Android. It was not possible to add minification for `test-base` due to our current project setup, but as `test-sync` holds quite many models and cover reasonably _base_ functionality it will in fact act as a sanity validation of our `library-base/proguard-rules-consumer-common.pro` configuration, especially initialization of the native class and method references. 

If needed it is relatively easy to manually run the full `test-base` test suite with minification following the instructions in `test-base/build.gradle.kts`

Closes #1114


clementetb:
We could enable running `test-base` with minification in CI if we remove its dependency from `test-sync`. We could manually link the shared sources and add the `kotlinx-datetime` dependency to `test-sync`.

packages/test-base/build.gradle.kts
```diff
@@ -131,9 +143,7 @@ android {
 }
 
 kotlin {
-    android("android") {
-        publishLibraryVariants("release")
-    }
+    android("android")
```
cmelchior, line 146: Don't we need to enumerate the library variants we want to release?

cmelchior: Ah, yeah, that makes sense 👍

cmelchior (approved):
If CI is green

PR #1199, opened 2023-01-04
Fix compilation error in main after merging `RealmAny`

edualonso:
Fixed handling of `RealmAny` properties in accessor modifier after `@PersistedName` task was merged.


elle-j (approved):

edualonso:
Will be fixed in https://github.com/realm/realm-kotlin/pull/1196

PR #1201, opened 2023-01-05
`RealmDictionary` - compiler plugin support

edualonso:
- Added draft interface for `RealmDictionary` and support for it in the compiler plugin.
- Added compiler plugin tests for dictionaries - I will align tests for lists, sets and dictionaries in an upcoming PR since there is a lot of similarities among them

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
@@ -30,4 +30,3 @@ public class ManagedRealmDictionary<E> : AbstractMutableMap<String, E>(), RealmD
         TODO("Not yet implemented")
     }
 }
-
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
@@ -20,9 +20,9 @@ public interface RealmMap<K, V> : MutableMap<K, V>
 
 public interface RealmDictionary<E> : RealmMap<String, E>
 
-public class UnmanagedRealmDictionary<E> : RealmDictionary<E>, MutableMap<String, E> by mutableMapOf()
+internal class UnmanagedRealmDictionary<E> : RealmDictionary<E>, MutableMap<String, E> by mutableMapOf()
 
-public class ManagedRealmDictionary<E> : AbstractMutableMap<String, E>(), RealmDictionary<E> {
+internal class ManagedRealmDictionary<E> : AbstractMutableMap<String, E>(), RealmDictionary<E> {
     override val entries: MutableSet<MutableMap.MutableEntry<String, E>>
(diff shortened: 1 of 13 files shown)


packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/Utils.kt
```diff
@@ -2,7 +2,64 @@ package io.realm.kotlin.test.compiler
...
+
+/**
+ * Generates a formatted string containing code that can be built by the compiler.
+ * Use it to generate different types of scenarios for nullability of contained type and the field
+ * itself.
+ */
+fun getCode(
+    collectionType: CollectionType,
+    contentType: String = "",
+    nullableContent: Boolean = false,
```
cmelchior, line 19: Maybe `elementType`?.... IMO `content` is as bit weird name to use for this.

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/Utils.kt
```diff
@@ -2,7 +2,64 @@ package io.realm.kotlin.test.compiler
...
+    nullableField: Boolean = false,
+    userStarProjection: Boolean = false
```
cmelchior, line 21: Nitpick: Since this a type of "contentType", it would be nice with an API that capture that constraint.

Maybe have two variants of `getCode`:

- getCode(collectionType, contentType, nullableContent)
- getCode(collectionType, nullableContent) // or maybe getCodeWithStarProjection

or make star the default value?

```
fun getCode(
    collectionType: CollectionType,
    contentType: String = "*",
    nullableContent: Boolean = false,
    nullableField: Boolean = false
)
```

edualonso: Good point. I think I can keep everything within the same function if I make `*` the default value for `contentType`.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.kotlin.types
+
+public interface RealmMap<K, V> : MutableMap<K, V>
+
+public interface RealmDictionary<E> : RealmMap<String, E>
+
+internal class UnmanagedRealmDictionary<E> : RealmDictionary<E>, MutableMap<String, E> by mutableMapOf()
```
cmelchior, line 23: These should be moved to `io.realm.kotlin.internal` but can be done in a later PR.

cmelchior (approved):

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
edualonso: Yeah, I've moved them in the upcoming PR, I'll leave it as is for now.

PR #1206, opened 2023-01-11
Fix missing realm flow-events from remote updates on synced realms

### LINKED ISSUE #1070: Realm.asFlow() is not being triggered on Sync changes.

rorbech:
Closes #1070

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableNotifier.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableNotifier.kt
@@ -71,16 +71,15 @@ internal class SuspendableNotifier(
     private val realm: NotifierRealm by realmInitializer
 
     /**
-     * FIXME Currently this is a hacked implementation that only does the correct thing if
-     *  other RealmResults or RealmObjects are being observed. But all writes should also flow
-     *  from [SuspendableWriter], so no Realm updates will be lost to end users.
-     *
      * Listen to changes to a Realm.
      *
      * This flow is guaranteed to emit before any other streams listening to individual objects or
      * query results.
      */
-    internal fun realmChanged(): Flow<FrozenRealmReference> {
+    internal suspend fun realmChanged(): Flow<FrozenRealmReference> {
+        // Touching realm will open the underlying realm and register change listeners, but must
+        // happen on the dispatcher as the realm can only be touched on the dispatcher's thread.
+        withContext(dispatcher) { realm }
         return _realmChanged.asSharedFlow()
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * None.
 
 ### Fixed
-* None.
+* Fix missing `Realm.asFlow()`-events from remote updates on synced realms. (Issue [#1070](https://github.com/realm/realm-kotlin/issues/1070))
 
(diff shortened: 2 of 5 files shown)


clementetb (approved):
Looks solid!

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableNotifier.kt
```diff
@@ -71,16 +71,15 @@ internal class SuspendableNotifier(
...
      * Listen to changes to a Realm.
      *
      * This flow is guaranteed to emit before any other streams listening to individual objects or
      * query results.
      */
-    internal fun realmChanged(): Flow<FrozenRealmReference> {
+    internal suspend fun realmChanged(): Flow<FrozenRealmReference> {
+        // Touching realm will open the underlying realm and register change listeners, but must
+        // happen on the dispatcher as the realm can only be touched on the dispatcher's thread.
+        withContext(dispatcher) { realm }
```
cmelchior, line 82: This is very hidden and a pretty important side-effect. Would it be better to have an explicit initializer method we can then call from 

```
            realmScope.launch {
                notifier.initializeRealm()  // Call here
                notifier.realmChanged().collect { realmReference ->
                    updateRealmPointer(realmReference)
                }
            }
```

And then instead of making `realm` in `SuspendableNotifier` lazy, we could make it lateinit?

cmelchior (approved):
Minor suggestion for a refactor, otherwise looking good 👍

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableNotifier.kt
rorbech: I actually had something like that as an initial implementation, but felt that it was really awkward to expose internal requirements like that in the API. Then things had to be called in a fixed order instead of just setting things up as we need them. 

Doesn't it make sense to keep any conventions internally? The current way exactly ensures that things are always set up correctly. If we use a `lateinit var` then we would have to roll our own synchronization around it. `notifier.realmChanged()` is called in a coroutine, so there is no guarantee that it is executed before returning from `Realm.open` and then we could end up registering other notifiers before this is initialized.

You could argue that we could put a comment on `notifer.realmChanged()` saying that the `realm` is opened as a side effect. But that feels really pointless as nobody would expect that you can ask for a flow of changes without opening the realm (except that that was actually what we did ... so some might expect it  🤦🐒 🤣). 

So I would still argue to keep it internal, but I am up for overruling if you feel that it is better?

cmelchior: Sounds fine to me. Lets just keep it the way it is.

PR #1207, opened 2023-01-12
Refactor compiler tests for lists, sets and dictionaries

edualonso:
Compiler tests for lists, sets and dictionaries are very similar so some alignment could be done by moving the shared cases to `CollectionTests` and leave the specific tests of each type in their corresponding classes.
In addition the code generation has been simplified and the tests for embedded objects in lists has been moved to `ListTests`.


clementetb (approved):
Looking good once CI is green

PR #1211, opened 2023-01-13
Statically link Windows C++ runtime

cmelchior:
`realmc.dll` for Windows did not statically link the C++ runtime, which could lead to crashes like `java.lang.UnsatisfiedLinkError: C:\Users\userdavid\AppData\Local\io-realm-kotlin\59a8f7af0c81a9d97dd70c86f4cf5c1618a81fb7\realmc.dll: Can't find dependent libraries` if it wasn't already pre-installed.

A work-around is to https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170 but we should provide it to avoid problems with people using a different version that we might support.


packages/cinterop/src/jvm/CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ set(CMAKE_CXX_STANDARD 17)
 if (CMAKE_SYSTEM_NAME MATCHES "^Windows")
     MESSAGE("Building JNI for Windows")
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    if (NOT WINDOWS_STORE)
+        set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    endif()
```
fealebenpae, line 15:
```suggestion
    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
```
`WINDOWS_STORE` is enabled for UWP builds, which you don't do, so you don't need to worry about that condition - it will always be false in your builds.

fealebenpae (approved):

packages/cinterop/src/jvm/CMakeLists.txt
cmelchior: Thanks 👍

nhachicha (approved):

PR #1213, opened 2023-01-13
Add support for SyncSession ConnectionState

### LINKED ISSUE #429: SyncSession - ConnectionListener
Add support for ConnectionListeners similar to Realm Java:

```java
enum class ConnectionState {
  DISCONNECTED,
  CONNECTING,
  CONNECTED
}

typealias ConnectionListener = (oldState: ConnectionState, newState: ConnectionState) -> Unit

class SyncSession {
  fun addConnectionChangeListener(listener: ConnectionListener)
  var connectionState: ConnectionState 
  fun isConnected(): Boolean
} 


```

rorbech:
This adds support for retrieving the `SyncSession`'s `ConnectionState` and get a `Flow` that emits events on changes in the sessions connection state.

```
/**
 * A **connection state** indicates the state of the underlying connection of the [SyncSession].
 */
public enum class ConnectionState {
    /**
     * Indicates that there is no connection to the server. No data is being transferred even if
     * the session is [SyncSession.State.ACTIVE]. If the connection entered this state due to an
     * error, the error is reported in the [SyncConfiguration.errorHandler].
     */
    DISCONNECTED,

    /**
     * Indicates that a connection is currently in progress of being established. If successful the
     * next state is [CONNECTED], otherwise it will be [DISCONNECTED].
     */
    CONNECTING,

    /**
     * Indicates that a connection is successfully established to the server. If the SyncSession is
     * [SyncSession.State.ACTIVE] then data will now be transferred between the device and the
     * server.
     */
    CONNECTED,
}

public interface SyncSession {
    /**
     * The current [ConnectionState].
     */
    public val connectionState: ConnectionState

    /**
     * Create a [Flow] of _connection state changes_-events.
     *
     * Each change of state in the underlying connection will be emitted as a pair of the old and
     * new [ConnectionState].
     *
     * The flow has an internal buffer of [Channel.BUFFERED] but if the consumer fails to consume
     * the elements in a timely manner the flow will be completed with an [IllegalStateException].
     */
    public fun connectionStateAsFlow(): Flow<Pair<ConnectionState, ConnectionState>>
}
```

Up for discussion:
- [x] Settle on `ConnectionState` being a top level class as `io.realm.kotlin.mongodb.sync.Direction` or inner class of `SyncSession` as `SyncSession.State`
- [x] Settle on doing a specific class for connection change events
- [x] #naming

Closes #429


packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -859,6 +859,17 @@ realm_sync_session_progress_notifier_callback(void *userdata, uint64_t transferr
     jni_check_exception(env);
 }
 
+void
+realm_sync_session_connection_state_change_callback(void *userdata, realm_sync_connection_state_e old_state, realm_sync_connection_state_e new_state) {
+    auto env = get_env(true);
+
+    static JavaMethod java_callback_method(env, JavaClassGlobalDef::connection_state_change_callback(), "onChange", "(II)V");
```
cmelchior, line 866: This needs to be added to the proguard file?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/ConnectionState.kt
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.kotlin.mongodb.sync
+
+/**
+ * A **connection state** indicates the state of the underlying connection of the [SyncSession].
+ */
+public enum class ConnectionState {
```
cmelchior, line 22: Having this as a top-level class makes sense IMO.  Maybe a stretch, but my line of thinking is that that `SyncSession.State` is directly related to the `SyncSession`, similar to `User.State` whereas, ConnectionState is a little more outside the responsibility of SyncSession.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -570,6 +573,34 @@ class SyncSessionTests {
...
+    }
+
+    @Test
+    fun connectionState_asFlow() = runBlocking {
+        Realm.open(createSyncConfig(user)).use { realm: Realm ->
+            val flow = realm.syncSession.connectionState()
+            val initialState = realm.syncSession.connectionState
+            // Adopted from realm-java tests ...
+            // Sometimes the connection is already established and then we cannot expect any
+            // updates, but as this is highly just safely ignore this to avoid flaky tests on CI
```
cmelchior, line 592:
```suggestion
            // updates, but as this is highly likely just safely ignore this to avoid flaky tests on CI
```

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -570,6 +573,34 @@ class SyncSessionTests {
...
+    fun connectionState() = runBlocking {
+        Realm.open(createSyncConfig(user)).use { realm: Realm ->
+            // We don't know what the state will be, but just verify that we can retrieve it
+            // without issues
+            assertNotNull(realm.syncSession.connectionState)
+        }
+    }
+
+    @Test
+    fun connectionState_asFlow() = runBlocking {
```
cmelchior, line 586: There seems to be some tests in Java that indicate that removing the connection listener will return DISCONNECTED. Should we add a `cancel` test for the Flow in Kotlin?

```
    @Test
    fun removeConnectionListener() = looperThread.runBlocking {
        Realm.getInstance(syncConfiguration).use { realm ->
            val session: SyncSession = realm.syncSession
            val listener1 = ConnectionListener { oldState: ConnectionState?, newState: ConnectionState ->
                if (newState == ConnectionState.DISCONNECTED) {
                    fail("Listener should have been removed")
                }
            }
            val listener2 = ConnectionListener { oldState, newState ->
                if (newState == ConnectionState.DISCONNECTED) {
                    looperThread.testComplete()
                }
            }
            session.addConnectionChangeListener(listener1)
            session.addConnectionChangeListener(listener2)
            session.removeConnectionChangeListener(listener1)
        }
    }
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
```diff
@@ -137,6 +142,17 @@ public interface SyncSession {
...
+    /**
+     * Create a [Flow] of _connection state changes_-events.
+     *
+     * Each change of state in the underlying connection will be emitted as a pair of the old and
+     * new [ConnectionState].
+     *
+     * The flow has an internal buffer of [Channel.BUFFERED] but if the consumer fails to consume
+     * the elements in a timely manner the flow will be completed with an [IllegalStateException].
+     */
+    public fun connectionState(): Flow<Pair<ConnectionState, ConnectionState>>
```
cmelchior, line 154: I would turn this into something like `data class ConnectionStateChange(val oldState: ConnectionState, val newState: ConnectionState)` for better readability and consumption. Pairs are a bit annoying to reason when exposed in 3rd party API's.

cmelchior (commented):
Looking good. Have a few minor suggestions for changes.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
rorbech: Yep. I actually left it out intentionally to see if CI was catching it now ... and it did. Have added it now.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
rorbech: The above test verifies that removing the listener is in fact removing it, as it is not failing on a disconnect on close. So there are two things in play. One is canceling the flow, and the other is that realm-java actually receives a disconnect when the realm is closed. I don't see such an event in my current implementation 👀  but should we keep the flow open and await a disconnected event or would we actually rather complete the flow on `close` ... as we have done a number of other places 🤔 

I guess this is a general behavioral pattern that we should then also consider for progress listeners.

cmelchior: Hmm, I think we do need some well-defined semantics for the `close` case. I think we should complete the flow, but whether or not we should emit a DISCONNECT before doing it...it would probably make the most sense, even though it probably ends up making the implementation a bit more complex.

But yeah, we should probably also close the progress listeners.

rorbech: I have added infrastructure to close flows when the realm is closed and have added test for it on both connection state and progress flows.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -291,6 +298,14 @@ public class RealmImpl private constructor(
             }
         }
     }
+    internal enum class State { OPEN, CLOSED, }
```
cmelchior, line 301: Would be nice with some docs on this

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
```diff
@@ -137,6 +142,15 @@ public interface SyncSession {
...
     ): Flow<Progress>
 
+    /**
+     * Create a [Flow] of [ConnectionStateChange]-events to receive notifications of updates to the
+     * session's connection state.
+     *
+     * The flow has an internal buffer of [Channel.BUFFERED] but if the consumer fails to consume
+     * the elements in a timely manner the flow will be completed with an [IllegalStateException].
+     */
+    public fun connectionState(): Flow<ConnectionStateChange>
```
cmelchior, line 152: If feels a bit akward to have both a `connectionState` property and a `connectionState()` method, but I'm not sure I have a better idea.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -570,6 +574,54 @@ class SyncSessionTests {
...
+            } catch (e: TimeoutCancellationException) {
+                assertTrue { realm.syncSession.connectionState == ConnectionState.CONNECTED }
+                // Make some visible sign that we have skipped waiting for events
+                println("Skipping flow tests as connection is already established")
+            }
+        }
+    }
+
+    @Test
+    fun connectionState_completeOnClose() = runBlocking {
```
cmelchior, line 607: We should create an issue for adding the same behaviour to progress listeners

cmelchior (approved):

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
rorbech: It is already fixed by just wrapping the existing `progress` flow in the new [`realm.scopedFlow { ... }`](https://github.com/realm/realm-kotlin/pull/1213/files#diff-b1b23368e96f440a9e9cb5263c2b02ed80409784749dd6a354c799b5b682cac9R116). There is even [a test for it](https://github.com/realm/realm-kotlin/pull/1213/files#diff-980709ee9063ac2e05f8126cb8a69faa20f97521bbbf52808abf2e8060bfcd97R254)

rorbech: But maybe needs a note in the `progress` doc.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
rorbech: Yeah #naming 😞 ... Maybe we should follow the pattern that I suggested in https://github.com/realm/realm-kotlin/pull/1118#discussion_r1017519584. Progress flows are not released yet, so we could switch to `progressAsFlow()` and `connectionStateAsFlow` if we want?

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
cmelchior: Ah yeah 🙈

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
rorbech: Updated it to `xAsFlow` after internal discussions.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -131,6 +139,7 @@ public class RealmImpl private constructor(
                     updateRealmPointer(realmReference)
                 }
             }
+            realmStateFlow.tryEmit(State.OPEN)
```
cmelchior, line 142: Don't we need to check that this succeeds or at least log a warning?

cmelchior (approved):

rorbech:
@cmelchior I included the notifier refresh fix and it is now green on CI, so maybe worth a final peek.

PR #1218, opened 2023-01-17
Update to Core 13.2.0

cmelchior:


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * `RealmUUID.random()` would generate the same values when an app was re-launched from Android Studio during development. (Issue [#1123](https://github.com/realm/realm-kotlin/pull/1123)) 
 * Complete flows with an IllegalStateException instead of crashing when notifications cannot be delivered due to insufficient channel capacity (Issue [#1147](https://github.com/realm/realm-kotlin/issues/1147))
 * [Sync] Custom loggers now correctly see both normal and sync events. Before, sync events were just logged directly to LogCat/StdOut.
+* [Sync] When a `SyncSession` was paused using `SyncSession.pause()`, it would sometimes automatically resume the session. (Core Issue [#6085](https://github.com/realm/realm-core/issues/6085))
```
rorbech, line 35: Shouldn't we somehow mention the new `State.PAUSED. I would deem it ok to add as a side effect of the fix when  it also goes into a minor release, but would expect it to be mentioned here for visibility.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/SyncSession.kt
```diff
@@ -172,11 +172,18 @@ public interface SyncSession {
          * The Realm is open and data will be synchronized between the device and the server
          * if the underlying connection is connected.
          *
-         * The session will remain in this state until the Realm
-         * is closed. In which case it will become [DYING].
+         * The session will remain in this state until the session is either paused or the
+         * Realm is closed. In which case it will become either [PAUSED] or [DYING].
```
nirinchev, line 176: [nit] `In which case it will become either [PAUSED] or [DYING].` is not a proper sentence - you probably want to merge the two sentences together or reword the second one.

PR #1229, opened 2023-01-23
`RealmDictionary` - 4: add support for `values`

edualonso:
Add support for `RealmDictionary.values`

The Core `realm_dictionary_to_results` returns a dictionary's values as a `realm_results_t` object. However, `Map.values` must return a `MutableCollection<V>`, which makes achieving alignment with the existing `RealmResults` API difficult.

As such, a completely detached implementation of results is used for this API since `values` has to support a subset of update operations (deletions derived from `remove` and `retainAll` are allowed, though no additions are supported due to the concept of "key" not existing in this API) and primitive values are allowed as well which is currently unsupported by `RealmResults`.

Retrieving elements from the collection can be done using its `iterator` which in turn allows removal of the last returned `next` element. The iterator is fairly similar to that of `entries` so the for both logic has been aligned for simplicity.

Minimal changes have been done to `RealmResultsImpl` to align with the `realm_results_get` that can return either object or primitive values.

Pull request plan:
- [ ] `containsKey`, `containsValue` - currently missing support in the C-API (https://github.com/realm/realm-core/issues/6181)
- [x] **THIS PR:** `values` - this change entails refactoring `RealmResults` to contain primitive values
- [ ] support for `RealmAny`
- [ ] support for ~~`Deleteable` and~~ (dictionaries don't own objects like sets and lists do so removing them from the dictionary won't remove them from the realm) `asFlow` - currently blocked by https://github.com/realm/realm-core/issues/6228
- [ ] allow `ByteArray`s to be compared structurally to allow `remove` operations on entry set and values
- [ ] support for `copyFromRealm`
- [ ] support for dynamic realms
- [ ] sync testing
- [ ] documentation
- [ ] changelog entry


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -573,6 +628,14 @@ internal interface DictionaryApiTester<T, Container> : ErrorCatcher {
...
     fun entries_removeAll()
     fun entry_equals()
+    fun values_addTrows()
+    fun values_clear()
+    fun values_iteratorNext() // This tests also hasNext
+    fun values_iteratorRemove() // This tests also hasNext
+    fun values_iteratorConcurrentModification()
+    fun values_remove()
+    fun values_removeAll()
+    fun values_retainAll()
```
cmelchior, line 638: Should we also add an `equals` test for the values set?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -1152,6 +1534,10 @@ internal class GenericDictionaryTester<T>(
     override fun assertStructuralEquality(expectedValue: T?, actualValue: T?) {
         assertEquals(expectedValue, actualValue)
```
cmelchior, line 1535: Doesn't this break for byte arrays?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -588,6 +651,8 @@ internal interface DictionaryApiTester<T, Container> : ErrorCatcher {
      */
     fun assertStructuralEquality(expectedValue: T?, actualValue: T?)
 
+    fun structuralContains(receiver: Collection<T?>, element: T?): Boolean
```
cmelchior, line 654: A small doc comment for this function would be nice

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -1113,6 +1190,311 @@ internal abstract class ManagedDictionaryTester<T>(
...
+            realm.writeBlocking {
+                val dictionary = typeSafetyManager.createContainerAndGetCollection(this)
+                val values = dictionary.values
+                assertFailsWithMessage<UnsupportedOperationException>("Adding values to a dictionary through 'dictionary.values' is not allowed") {
+                    values.add(dataSet[0].second)
+                }
+                assertFailsWithMessage<UnsupportedOperationException>("Adding values to a dictionary through 'dictionary.values' is not allowed") {
+                    values.addAll(listOf(dataSet[0].second))
+                }
+                assertTrue(dictionary.isEmpty())
```
cmelchior, line 1206: Aren't this tested in the cleanup method just below? The same seems to happen in all the other tests.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
edualonso: I will also add one for the entry set as well, though observe the restrictions on byte arrays will apply until dealt with in the upcoming PR.

cmelchior (approved):

PR #1231, opened 2023-01-23
Automatic Client Reset

edualonso:
Added Automatic Client Reset and support for `RecoverUnsyncedChangesStrategy` and `RecoverOrDiscardUnsyncedChangesStrategy` after all Sync bugs have been fixed.

I split the code for each test in PBS and FLX versions because after lots of debugging I found it was easier to navigate the code and do individual testing/debugging if they are kept separate rather than under the same test, even though the code is 99% the same.


packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncSessionTests.kt
```diff
@@ -315,70 +311,44 @@ class SyncSessionTests {
...
+        }.build()
 
-                    override fun onAfterReset(before: TypedRealm, after: MutableRealm) {
-                        // Writing "fail("whatever") makes no sense here because Core will redirect
-                        // the execution to onError below
+        runBlocking {
+            val deferred = async { Realm.open(config) }
+            val session = channel.receive()
+            try {
+                assertFailsWithMessage<IllegalStateException>("Operation is not allowed inside a `SyncSession.ErrorHandler`.") {
```
clementetb, line 338: This error message does not tell that the `Uploading and downloading changes is not allowed` are we sure that we are testing it right?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/SyncConfigurationImpl.kt
```diff
@@ -303,3 +208,247 @@ internal class SyncConfigurationImpl(
...
+        SyncSessionResyncMode.RLM_SYNC_SESSION_RESYNC_MODE_RECOVER_OR_DISCARD
+
+    override fun getBefore(): SyncBeforeClientResetHandler =
+        object : SyncBeforeClientResetHandler {
+            override fun onBeforeReset(realmBefore: FrozenRealmPointer) {
+                strategy.onBeforeReset(TypedFrozenRealmImpl(realmBefore, configuration))
+            }
+        }
+
+    override fun getAfter(): SyncAfterClientResetHandler =
```
clementetb, line 254: There are multiple implementations for `getAfter` and `getBefore` that really similar between them, is there any way to reuse the code?

clementetb (commented):
Looking good overall, added some minor comments.

We are not testing capturing client errors or executing the manual reset on all strategies.

edualonso:
I think I have now covered the missing cases for catching errors in all strategies

clementetb (approved):
Looks good 🚀

.github/workflows/issue-needs-attention.yml
```diff
@@ -0,0 +1,12 @@
+# NOTE: This is a common file that is overwritten by realm/ci-actions sync service
```
edualonso, line 1: Fix this file before merging!

Issue #1097, opened 2022-10-30
RealmObject `equals` are probably not correct


cmelchior:
When reading objects from Realm, we use their default `equals` implementation, which compares by memory reference. This means that the following behavior can be observed:

```
realm.writeBlocking {
            val result = copyToRealm(sample)
            val obj1 = result.nullableObject
            val obj2 = result.nullableObject
            assertNotEquals(obj1, obj2) // Managed objects use default `equals`, i.e. compare mem ref. 
}
```

This is probably not what you would expect, and we should consider if managed objects should have their own equals implementation that compare <ClassKey,ObjectKey,Version> instead. It would probably have to be implemented in the Compiler Plugin so we can avoid generating it if users have their own equals method.

We should also take into account `RealmAny` instances containing objects.

tarasmorskyi-mediamonks:
Any update on this? Is currently the only option is to compare value by value aka deepEquals?

tarasmorskyi-mediamonks:
@cmelchior tagging you as reporter

cmelchior:
It appears this issue fell through the cracks and should be fixed soon.

PR #1239, opened 2023-01-25
`RealmDictionary` feature

### LINKED ISSUE #537: Add support for RealmDictionary
Follow the pattern of managed/unmanaged used for lists by using delegates.

An oversimplified API would look like this:
```
// Internal until we support multi-typed keys
internal interface RealmMap<K, V> : MutableMap<K, V>

// Public interface
interface RealmDictionary<V> : RealmMap<String, V>

// Internal implementation for unmanaged dicts
internal class UnmanagedRealmDictionary<V>: RealmDictionary<V>,
    MutableMap<String, V> by mutableMapOf()

// Internal implementation for managed dicts
internal class ManagedRealmDictionary<V>(
    val nativePointer: NativePointer
): AbstractMutableMap<String, V>(), RealmDictionary<V> {
    // TODO
}

/**
 * Instantiates an **unmanaged** [RealmDictionary].
 */
fun <V> realmDictionaryOf(vararg entries: Pair<String, V>): RealmDictionary<V> =
    UnmanagedRealmDictionary()

/**
 * Instantiates an **unmanaged** [RealmDictionary] containing all the entries of this dictionary.
 */
fun <V> Map<String, V>.toRealmDictionary(): RealmDictionary<V> = TODO()


```

Observations:
- **the C-API is still missing a large number of methods and implementations, e.g. `keys`, `values`, `find`, `remove_all`, `contains` and perhaps more.**
- the majority of methods from the `MutableMap` interface are already provided by `AbstractMutableMap` though they might not be optimised, e.g. `putAll` iterates over all values in the provided collection and then call `put`.
- the `entries` property returns a `MutableSet` which might be a slight annoyance since we don't need to enable mutability for said set - it can probably be overridden to throw an exception if users decide to modify it.
- It doesn't seem necessary to rely on `Mixed` as the transport solution since the C-API accepts `realm_value_t` structs, just as `RealmList` does.

edualonso:
Closes #537

### Dictionary API
A `RealmDictionary` is a `Map<K, V>`, where keys are always strings and values are Realm primitives or objects, be it regular or embedded. Dictionaries can also be expressed as:
- `MutableSet<MutableMap.MutableEntry<K, V>>` by calling `RealmDictionary.entries`. The entry set is a live collection connected to its underlying dictionary. Deletions and insertions are allowed and will result in an updated dictionary. Individual entries can also be modified and will result in an update key-value dictionary pair. Unmanaged entries can also be instantiated via factory - see "Factories" section below. Managed entries are also live key-value pairs and mutating them will update the dictionary accordingly.
- `MutableSet<K>` by calling `RealmDictionary.keys`. The key set is a live collection connected to its underlying dictionary. Deletions are allowed though insertions aren't - it isn't possible to insert a key without a value. Core delivers the keys as `realm_results_t` but subscribing for key changes is not exposed for now.
- `MutableCollection<V>` by calling `RealmDictionary.values`. The value collection is a live collection connected to its underlying dictionary. Deletions are allowed though insertions aren't - it isn't possible to insert a value without a key. Core delivers the values as `realm_results_t` so queries on dictionary objects are funnelled through these.

### `RealmDictionary` and `RealmMap` abstractions
Thanks to how the compiler plugin extracts information from the type system it is (almost) possible to work at a `RealmMap` level, though we aren't exposing a public implementation of it and `RealmMap` fields aren't allowed either.

To make the implementation more readable and less painful, a number of type aliases have been added:
- `internal typealias RealmMapEntrySet<K, V> = MutableSet<MutableMap.MutableEntry<K, V>>`
- `internal typealias RealmMapMutableEntry<K, V> = MutableMap.MutableEntry<K, V>`
- `public typealias RealmDictionaryEntrySet<V> = RealmMapEntrySet<String, V>`
- `public typealias RealmDictionaryMutableEntry<V> = RealmMapMutableEntry<String, V>`

### Object dictionaries
Object dictionaries may contain either regular or embedded objects and `null` values are allowed. `RealmDictionary` of objects must be nullable - not the field itself, which is not allowed for any type of Realm collection, but the collection type instead.

### Factories
Unmanaged dictionaries can be instantiated by calling the `realmDictionaryOf` functions that allow the following input values:
- `realmDictionaryOf(vararg elements: Pair<String, T>)`
- `realmDictionaryOf(elements: Collection<Pair<String, T>>)`

Unmanaged `RealmDictionaryMutableEntry` instances can be used to add key-value pairs to the dictionary through the dictionary's entry set and can be created by calling the `realmDictionaryEntryOf` functions which accept the following input values:
- `realmDictionaryEntryOf(pair: Pair<String, V>)`
- `realmDictionaryEntryOf(key: String, value: V)`
- `realmDictionaryEntryOf(entry: Map.Entry<String, V>)`

### Queries on dictionaries
There is no direct way to parse a query on a dictionary object through the C-API. However, we can use the `realm_results_t` object yielded by the C-API when calling `RealmDictionary.values` to parse a query on the results.

### Notifications and change sets
A dictionary change set contains:
- the deleted keys
- the inserted keys
- the changed keys

Pull request plan:
- [x] `containsKey`, `containsValue`~~- currently missing support in the C-API (https://github.com/realm/realm-core/issues/6181)~~
- [x] `values` - this change entails refactoring `RealmResults` to contain primitive values
- [x] support for `RealmAny`
- [ ] support for ~~`Deleteable` and (dictionaries don't own objects like sets and lists do so removing them from the dictionary won't remove them from the realm)~~ `asFlow` - ~~currently blocked by https://github.com/realm/realm-core/issues/6228~~
- [x] allow `ByteArray`s to be compared structurally to allow `remove` operations on entry set and values
- [x] support for `copyFromRealm`
- [x] support for dynamic realms
- [x] support for backlinks
- [x] sync testing
- [x] aggregation queries on dictionaries and add missing test cases for sets and lists
- [x] embedded objects - https://github.com/realm/realm-kotlin/pull/1269
- [x] queries on dictionaries - https://github.com/realm/realm-kotlin/pull/1269
- [x] notifications - https://github.com/realm/realm-kotlin/pull/1274
- [x] documentation
- [x] changelog entry


packages/test-base/src/commonMain/kotlin/io/realm/kotlin/test/util/TypeDescriptor.kt
```diff
@@ -335,7 +345,8 @@ object TypeDescriptor {
         .map { RealmFieldType(CollectionType.RLM_COLLECTION_TYPE_LIST, it) }
     val allSetFieldTypes = elementTypesForSet.filter { it.realmFieldType.setSupport }
         .map { RealmFieldType(CollectionType.RLM_COLLECTION_TYPE_SET, it) }
-    // TODO Dict
+    val allDictionaryFieldTypes = elementTypesForDictionary.filter { it.realmFieldType.dictionarySupport }
+        .map { RealmFieldType(CollectionType.RLM_COLLECTION_TYPE_DICTIONARY, it) }
     val allFieldTypes: List<RealmFieldType> = allSingularFieldTypes + allListFieldTypes
```
clementetb, line 350: allFieldTypes is missing allSetFieldTypes and allDictionaryFieldTypes

edualonso:
Still missing the changelog entry and some documentation but otherwise ready for final review.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmElementQuery.kt
```diff
@@ -17,8 +17,13 @@
 package io.realm.kotlin.query
 
 import io.realm.kotlin.Deleteable
+import io.realm.kotlin.MutableRealm
```
cmelchior, line 20: All these new imports don't seem to be used?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
```diff
@@ -0,0 +1,67 @@
...
+import io.realm.kotlin.internal.RealmMapMutableEntry
+import io.realm.kotlin.notifications.DictionaryChange
+import io.realm.kotlin.notifications.InitialDictionary
+import io.realm.kotlin.notifications.UpdatedDictionary
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * TODO
```
cmelchior, line 30: Missing documentation. We can probably copy most of it from Java

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+public interface RealmMap<K, V> : MutableMap<K, V>
+
+/**
+ * TODO
```
cmelchior, line 35: Missing documentation

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+public interface RealmDictionary<E> : RealmMap<String, E> {
```
cmelchior, line 37: Nitpick, but should we use `<String, V>` instead of `<String, E>`? So it matches the RealmMap names.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
```diff
@@ -0,0 +1,67 @@
...
+     * The change calculations will run on the thread represented by
+     * [RealmConfiguration.Builder.notificationDispatcher].
+     *
+     * The flow has an internal buffer of [Channel.BUFFERED] but if the consumer fails to consume
+     * the elements in a timely manner the coroutine scope will be cancelled with a
+     * [CancellationException].
+     *
+     * @return a flow representing changes to the dictionary.
+     */
+    public fun asFlow(): Flow<DictionaryChange<E>>
```
cmelchior, line 52: Would it make sense to add ` public fun asFlow(): Flow<MapChange<K, V>>` to RealmMap already? And then this needs to be `override public fun asFlow(): Flow<DictionaryChange<E>>`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
```diff
@@ -0,0 +1,67 @@
...
+     * continue running indefinitely until canceled or until the parent object is deleted.
+     *
+     * The change calculations will run on the thread represented by
+     * [RealmConfiguration.Builder.notificationDispatcher].
+     *
+     * The flow has an internal buffer of [Channel.BUFFERED] but if the consumer fails to consume
+     * the elements in a timely manner the coroutine scope will be cancelled with a
+     * [CancellationException].
+     *
+     * @return a flow representing changes to the dictionary.
```
cmelchior, line 50: Missing throws statements.

cmelchior (commented):
Halfway through review, minor comments only so far.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmMap.kt
edualonso: After some more reflecting I think it's better to just have this function in `RealmMap` and not even override it for `RealmDictionary`. The reason for this is that the aliases I added for the notification changes are pretty useless 😞 The IDE doesn't take them into account when telling it to explicitly state the type and uses the parent type instead. So it might seem like a good idea to use aliases for when the types are a mouthful but it seems they useless when trying to get the IDE to use them for convenience. So I'm more inclined to delete them now.

cmelchior (commented):
Only missing 3000 lines of tests now. Looking good so far 😄

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -0,0 +1,2966 @@
...
+    fun tearDown() {
+        if (this::realm.isInitialized && !realm.isClosed()) {
+            realm.close()
+        }
+        PlatformUtils.deleteTempDir(tmpDir)
+    }
+
+    @Test
+    fun unmanaged() {
+        // No need to be exhaustive here, just checking delegation works
```
cmelchior, line 142: Why don't we need to be exhaustive? We should be sure we know how unmanaged objects work as well. That said, It would not block merging this PR for it. Maybe just create an issue and we can add testing for it later.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -0,0 +1,2966 @@
...
+
+/**
+ * Tester interface defining the operations that have to be tested exhaustively.
+ */
+internal interface DictionaryApiTester<T, Container> : ErrorCatcher {
+
+    val realm: Realm
+
+    override fun toString(): String
+    fun copyToRealm()
```
cmelchior, line 1343: Looks like we are missing testing for `equals/hashCode` for the `RealmDictionary` itself?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/QueryTests.kt
```diff
@@ -2019,9 +2023,206 @@ class QueryTests {
...
+                    )
+                }
+            )
+        }
+
+        // Supported fields for aggregations
+        listOf("@sum", "@min", "@max", "@avg")
+            .forEach { aggregator ->
+                listOf(
+                    QuerySample::intListField.name, // Just test one Long value
```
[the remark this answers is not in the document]
edualonso: I have created https://github.com/realm/realm-kotlin/issues/1295

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/BacklinksTests.kt
```diff
@@ -278,24 +336,29 @@ class BacklinksTests {
...
+            }
     }
 
     @Test
     fun deleteSourceObject() {
         val child = realm.writeBlocking {
             val child = this.copyToRealm(Child())
 
-            val parents = Array(5) {
+            val parents = (1..5).map {
```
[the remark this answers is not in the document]
edualonso: The problem with `repeat` is that it doesn't do any mapping and we would have to add the output manually for each iteration. Otherwise I agree, it would look better.

cmelchior (commented):
Done...I think we are missing some corner cases around equality testing for Dictionary but otherwise it looks good 🚀

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -0,0 +1,2966 @@
...
+                val actualAny = assertNotNull(it.nullableRealmAnyDictionaryField["X"])
+                assertEquals(RealmAny.Type.OBJECT, actualAny.type)
+                val actual = actualAny.asRealmObject<RealmDictionaryContainer>()
+                assertEquals(expected.stringField, actual.stringField)
+            }
+        }
+    }
+
+    @Test
+    fun closedRealm_readFails() {
```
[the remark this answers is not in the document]
edualonso: I think I simply forgot about them 🙈 

These will also most likely start to fail once we remove the call to `RealmReference.checkClosed()` so I'll add a TODO.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -0,0 +1,2966 @@
...
+
+    @Test
+    fun entries_removeAll() {
+        for (tester in managedTesters) {
+            tester.entries_removeAll()
+        }
+    }
+
+    @Test
+    fun entries_equals() {
```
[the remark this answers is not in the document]
edualonso: Do you mean comparing two `RealmDictionary.entries.toString()` or two individual `entry.toString()`?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -0,0 +1,2966 @@
...
+
+    @Test
+    fun values_retainAll() {
+        for (tester in managedTesters) {
+            tester.values_retainAll()
+        }
+    }
+
+    @Test
+    fun values_equals() {
```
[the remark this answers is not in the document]
edualonso: Same as with the comment about the entry set. Do you mean the value collection as a whole or individual values?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
edualonso: True, I will add them 👍

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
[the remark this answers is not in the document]
cmelchior: It looks like we are missing all of them.

edualonso:
Added missing `toString` tests and added scaffolding for future `equals` and `hashCode` tests for when https://github.com/realm/realm-kotlin/issues/1097 is fixed

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmMapInternal.kt
```diff
@@ -657,6 +677,11 @@ internal class ManagedRealmDictionary<V> constructor(
                 }
         }
     }
+
+    override fun toString(): String = entries.joinToString { (key, value) -> "[$key,$value]" }
```
cmelchior, line 681: Hmm, I can see where this is coming from, but it looks pretty dangerous, i.e. lets say you have dictionaries with 100's or 1000's of elements. I would probably avoid doing this, but there doesn't seem to be a great alternative either.

Since `toString` is generally used for debugging I would probably rather turn this into something like `RealmDictionary{owner=<classType>,objKey=<objectId>,version=<transactionVersion>}`. It would be easy to compute and you would still get some context for debugging.

cmelchior (changes requested):
Deferring tests to we fix the `equals` issue is fine, but I think the `toString()` implementation is making too many assumptions about dictionaries not having many elements. So we should change it to something else IMO. I proposed a solution in the comment, but happy to discuss alternatives.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmMapInternal.kt
edualonso: Since there was no `toString` implementation for the other collections I took the approach we use for Java lists but I agree it can easily backfire with large datasets. I think your proposal is fair, I would only add `size` for clarity too.

rorbech (approved):
Finally had a look on this. Haven't gone through all the details but only the overall structures and essential mechanisms ... And looks good to me.

PR #1242, opened 2023-01-26
`RealmDictionary` - 5: add general support for `RealmAny`

edualonso:
Added general support for `RealmAny`
Improved testing for accessors to cover all types of dictionaries

Pull request plan:
- [ ] `containsKey`, `containsValue` - currently missing support in the C-API (https://github.com/realm/realm-core/issues/6181)
- [x] `values` - this change entails refactoring `RealmResults` to contain primitive values
- [x] **THIS PR:** support for `RealmAny`
- [ ] support for ~~`Deleteable` and~~ (dictionaries don't own objects like sets and lists do so removing them from the dictionary won't remove them from the realm) `asFlow` - currently blocked by https://github.com/realm/realm-core/issues/6228
- [ ] allow `ByteArray`s to be compared structurally to allow `remove` operations on entry set and values
- [ ] support for `copyFromRealm`
- [ ] support for dynamic realms
- [ ] sync testing
- [ ] documentation
- [ ] changelog entry


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
```diff
@@ -215,49 +222,159 @@ class RealmDictionaryTests {
...
+                }
+        }
+
+        // Cleanup between assertions for convenience
+        realm.writeBlocking {
+            delete(query<RealmDictionaryContainer>())
+        }
+
+        // Test with a dictionary of RealmAny containing an object
+        realmDictionaryOf<RealmAny?>(
```
cmelchior, line 320: I would move this into its own test. This test is bit too big and tests too many things IMO. Same in the other tests

cmelchior (approved):

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmDictionaryTests.kt
edualonso: I'll split both getter and setter sets into the different cases.

PR #1246, opened 2023-01-26
Allow using the same field name as the persisted name.

clementetb:
With this PR we allow to bypass the core restriction on using the same persisted and public names. 

Anytime that the same name is used we ignore that it has been defined and process it normally.


packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/PersistedNameTests.kt
```diff
@@ -67,8 +67,6 @@ class PersistedNameTests {
         )
 
         assertEquals(KotlinCompilation.ExitCode.OK, result.exitCode)
-        val compilerLog = result.messages
```
rorbech, line 70: I guess the whole test should be removed. The intention is now caught by https://github.com/realm/realm-kotlin/pull/1246/files#diff-15e1aebddb1a90c5b093c9072231a81488ac5bcb71f580a96fb476951aed6d55R391 and this test name is completely off compared to what it actually does.

rorbech (commented):
LGTM, but we should probably target `releases` instead.

clementetb:
Changed base from main to releases

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ This release will bump the Realm file format from version 22 to 23. Opening a fi
 * Prevent "Cannot listen for changes on a deleted Realm reference"-exceptions when notifier is not up-to-date with newest updates from write transaction.
 * [Sync] Custom loggers now correctly see both normal and sync events. Before, sync events were just logged directly to LogCat/StdOut.
 * [Sync] When a `SyncSession` was paused using `SyncSession.pause()`, it would sometimes automatically resume the session. `SyncSession.State.PAUSED` has been added, making it explicit when a session is paused. (Core Issue [#6085](https://github.com/realm/realm-core/issues/6085))
+* Allow defining properties with the field name as the persisted name. ([#1240](https://github.com/realm/realm-kotlin/issues/1240))
```
cmelchior, line 41: This should be in a 1.6.1 changelog

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/IrUtils.kt
```diff
@@ -285,14 +286,26 @@ data class SchemaProperty(
     val publicName: String
 
     init {
-        if (hasPersistedNameAnnotation) {
+        val declarationName = declaration.name.identifier
+        val persistedAnnotationName: String? = if (hasPersistedNameAnnotation) getPersistedName(declaration) else null
+
+        // We only set it if the persisted and public names are different because core does
+        // would detect it as a duplicated name.
```
cmelchior, line 293:
```suggestion
        // We only set the public name if the persisted and public names are different 
        // because core would otherwise detect it as a duplicated name and fail.
```

cmelchior (approved):

CHANGELOG.md
clementetb: 🤦

PR #1259, opened 2023-02-01
Add missing generic type parameter for generated property

### LINKED ISSUE #1252: Realm 1.6.0 conflict with dagger
I have a module(only Realm 1.6.0 implemented) with my Database and RealmObjects,my main program has dagger and implementation of module,with this I no longer can create global variables like:
```
    var dog: Dog= Dog()  

    override fun onCreate(savedInstanceState: Bundle?) {
        ....
    }
}
```
it will raise error:
```
error: cannot access Dog
    private gr.mastersdk.database.model.Dog dog;

undeclared type variable: T
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```
But the same code will work if I create this entity inside function like:
```
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)

        var dog = Dog()
    }
}
```
Not really sure why this conflict exists,but having find any way out of it
dagger that I'm using: 'com.google.dagger:hilt-android:2.41'

### LINKED ISSUE #1256: Cannot use Realm Kotlin classes in Java code
If referencing a Realm Kotlin class from a Java class like below

`Sample.kt`
```
class Sample: RealmObject {
    var id: String = "Sample"
}
```
And `Test.java`
```
public class Test {
    Sample  sample = new Sample();
}
```
Then compilation will fail with 
```
Test.java:4: error: cannot access Sample
    Sample  sample = new Sample();
    ^
  bad class file: Sample.class
    undeclared type variable: T
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

rorbech:
Closes #1252 
Closes #1256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Allow defining properties with the field name as the persisted name. ([#1240](https://github.com/realm/realm-kotlin/issues/1240))
+* Fix compilation error when accessing Realm Kotlin model classed from Java code. ([#1256](https://github.com/realm/realm-kotlin/issues/1256))
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
--- a/integration-tests/gradle-plugin-test/single-platform/src/main/java/io/realm/test/singleplatform/model/Pojo.java
+++ b/integration-tests/gradle-plugin-test/single-platform/src/main/java/io/realm/test/singleplatform/model/Pojo.java
@@ -1,0 +1,5 @@
+package io.realm.test.singleplatform.model;
+
+public class Pojo {
+    TestClass o = new TestClass();
+}
--- a/integration-tests/gradle-plugin-test/single-platform/src/main/java/io/realm/test/singleplatform/model/Pojo.java
+++ b/integration-tests/gradle-plugin-test/single-platform/src/main/java/io/realm/test/singleplatform/model/Pojo.java
@@ -1,5 +1,7 @@
 package io.realm.test.singleplatform.model;
 
+// This class is only added to verify that we can access the Realm Model classes from plain old
+// Java source. See issue for further details https://github.com/realm/realm-kotlin/issues/1256
 public class Pojo {
     TestClass o = new TestClass();
 }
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixed
 * Allow defining properties with the field name as the persisted name. ([#1240](https://github.com/realm/realm-kotlin/issues/1240))
+* Fix compilation error when accessing Realm Kotlin model classed from Java code. ([#1256](https://github.com/realm/realm-kotlin/issues/1256))
```
cmelchior, line 11: Can we add an integration test for this? I assume adding a Java class to one of our JVM integration samples should work?

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -258,7 +258,7 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
             pluginContext,
             realmObjectCompanionInterface,
             REALM_OBJECT_COMPANION_CLASS_MEMBER,
-            pluginContext.irBuiltIns.kClassClass.defaultType
+            pluginContext.irBuiltIns.kClassClass.typeWith(clazz.defaultType)
```
cmelchior, line 261: Good find 💯

CHANGELOG.md
rorbech: I tried adding a plain Java class as part of the test suite and it didn't trigger the issue as it did in my reproduction project 🤔

rorbech: Ahh. I did it in the `jvmTest`. Will try in the Android tests instead.

edualonso, line 11:
```suggestion
* Fix compilation error when accessing Realm Kotlin model classes from Java code. ([#1256](https://github.com/realm/realm-kotlin/issues/1256))
```

rorbech: I managed to trigger it in the `integration-tests/gradle-plugin-test/single-platform` with https://github.com/realm/realm-kotlin/pull/1259/files#diff-343ba0d05e9a620f83dfed66a489bb4016ccade98fe3984c6fb5a7eacf66bdf6R1. Apparently it is only an issue if the accessing Java class is in the `java` source directory 🤯. Stopped investigating options to put it inside `test-base` as I didn't want to pollute our tests by adding the additional source setup, but could try to investigate if we rather would like it to be here?

cmelchior: I think having it in an integration test is just fine 👍

integration-tests/gradle-plugin-test/single-platform/src/main/java/io/realm/test/singleplatform/model/Pojo.java
```diff
@@ -0,0 +1,5 @@
+package io.realm.test.singleplatform.model;
+
+public class Pojo {
```
cmelchior, line 3: I would put a reference to this issue here to make the context clear

cmelchior (approved):

PR #1263, opened 2023-02-02
`RealmDictionary` - 9a: add test cases for aggregations on queries

edualonso:
Added aggregation queries for dictionaries.
Added missing aggregation queries for lists and sets.

**NOTE:** this is **NOT** query parsing for dictionaries. That functionality is missing from the C-API but will be added soon.


clementetb (approved):
Looks good, happy once CI is green

PR #1274, opened 2023-02-07
`RealmDictionary` - 9e: notifications

edualonso:
Added infrastructure for notifications on maps/dictionaries.

A dictionary changeset will contain the keys of those elements that have been either inserted, deleted or modified.

Following the pattern used previously in the feature, all "dictionary" classes are type aliases that point to their corresponding `Map<String, V>` class.

### OBSERVE:
The proposed C-API functions in https://github.com/realm/realm-core/pull/6268 need a bit of tweaking so until the corresponding PR is done I am using my own Core branch - keeping as a draft until then.

### TODO:
- [x] point to master in Core once https://github.com/realm/realm-core/pull/6268 adds the missing `get_size` function to the C-API
- [x] update core pointer and version in changelog


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/notifications/MapChange.kt
```diff
@@ -0,0 +1,82 @@
...
+ * ```
+ */
+public sealed interface MapChange<K, V> {
+    public val map: RealmMap<K, V>
+}
+
+/**
+ * Convenience alias for [RealmDictionary] changes. It represents a [MapChange] of `<String, V>`.
+ */
+public typealias DictionaryChange<V> = MapChange<String, V>
```
cmelchior, line 44: Do we need these? Since we are setting these up anyway in our `asFlow` method. I'm not sure what value they bring?

cmelchior: Same for the other ones below.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -0,0 +1,390 @@
...
+                queriedDictionary[dataSet[0].first] = dataSet[0].second
+            }
+
+            channel.receive().let { dictionaryChange ->
+                assertIs<UpdatedDictionary<*>>(dictionaryChange)
+
+                assertNotNull(dictionaryChange.map)
+                assertEquals(1, dictionaryChange.map.size)
+                dictionaryChange.insertions.let { insertions ->
+                    assertEquals(1, insertions.size)
```
cmelchior, line 141: We should also check that changes and deleted are 0

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -0,0 +1,390 @@
...
+                queriedDictionary[dataSet[0].first] = dataSet[1].second
+            }
+
+            channel.receive().let { dictionaryChange ->
+                assertIs<UpdatedDictionary<*>>(dictionaryChange)
+
+                assertNotNull(dictionaryChange.map)
+                assertEquals(1, dictionaryChange.map.size)
+                dictionaryChange.changes.let { changes ->
+                    assertEquals(1, changes.size)
```
cmelchior, line 160: We should also check inserted and deleted. Also in the other asserts below

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -0,0 +1,390 @@
...
+            channel.receive().let { dictionaryChange ->
+                assertIs<UpdatedDictionary<*>>(dictionaryChange)
+
+                assertNotNull(dictionaryChange.map)
+                assertEquals(dataSet.size - 2, dictionaryChange.map.size)
+                assertEquals(1, dictionaryChange.deletions.size)
+                assertEquals(0, dictionaryChange.insertions.size)
+            }
+
+            // Assert notification on removal of elements via entry set iterator
```
cmelchior, line 222: This seems to be more about testing the value iterator. So I would pull this into its own test case.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -0,0 +1,390 @@
...
+
+            observer.cancel()
+            channel.close()
+        }
+    }
+
+    @Test
+    override fun cancelAsFlow() {
+        runBlocking {
+            // Freeze values since native complains if we reference a package-level defined variable
```
cmelchior, line 248: I don't think this is still the case with the new memory model?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -0,0 +1,390 @@
...
+            // Wait for flow completion
+            assertTrue(channel2.receive())
+
+            observer.cancel()
+            channel1.close()
+        }
+    }
+
+    @Test
+    @Ignore // FIXME Wait for https://github.com/realm/realm-kotlin/pull/300 to be merged before fleshing this out
```
cmelchior, line 362: This has been closed

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -0,0 +1,390 @@
...
+    override fun closeRealmInsideFlowThrows() {
+        TODO("Waiting for RealmDictionary support")
+    }
+
+    @Test
+    override fun closingRealmDoesNotCancelFlows() {
```
cmelchior, line 368: Talking to @rorbech about this and we both agreed that this behaviour is wrong and we should close all channels when a Realm is closed, but it is fine to keep a test of the current behaviour for now.

cmelchior (commented):
I have a few reservations about adding a complete hierarchy for `DictionaryChange`, it doesn't seem to bring much of a benefit as users are never asked to write the type for this, instead it add quite a lot of code.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/notifications/MapChange.kt
edualonso: I don't think we need them per se. I thought they might be a more convenient way to establish boundaries to all things dictionary just as I did for `RealmDictionaryEntrySet` and `RealmDictionaryMutableEntry`.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
edualonso: `changes` is missing and will add it, `deletions` are just two lines down.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
edualonso: I think it makes sense to leave it here. Dictionary notifications can be triggered from multiple places: by modifying the entry set, the value collection or the dictionary itself. That is different from other collections. I think having everything under the `asFlow` call proves that the same stream can handle being bombarded from all different change sources just fine.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
edualonso: That's probably true. A clear case of rogue copy-paste 😓

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
edualonso: I've checked this one out and the only way I can think of implementing this is keeping an external reference to the realm we are working with and closing it inside the callback, which actually doesn't cause the flow to throw. Or am I misunderstanding the comment?

I think in order for us to implement this test properly we should wait until we expose the realm in our collections and objects: https://github.com/realm/realm-kotlin/issues/582

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
edualonso: I guess whenever we change the behaviour this test will fail and we will have to fix it so no worries.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/notifications/RealmDictionaryNotificationsTests.kt
cmelchior: Yeah, thinking about this. We probably shouldn't throw when closing the Realm inside the flow, but just gracefully close the flow. This is also related to the test just below. But all of that is for another PR, so fine with resolving this as is 👍

cmelchior (approved):

PR #1281, opened 2023-02-13
Obfuscate sensitive information shown by Ktor

### LINKED ISSUE #410: AppConfiguration - HttpLogObfuscator
Similar to Realm Java we need support for a HttpLogObfuscator so we don't accidentally leak credentials or keys into LogCat.

This issues includes:

* Move `HttpLogObfuscator` and `RegexPatternObfuscator` from Java to Kotlin.
* Add support for these in the NetworkTransport layer.
* Expose `AppConfiguration.Builder.httpLogObfuscator()` method.

edualonso:
Closes #410

Tokens, passwords and custom function arguments are obfuscated when logged by Ktor.
Exposed `AppConfiguration.Builder.logObfuscator()`. The API allows `null` values for the obfuscator, in which case the log entry will not be obfuscated.
We found an issue with regex replacement with rather long strings on K/N (https://github.com/realm/realm-kotlin/issues/1284) so until we figure out what is causing this we will do string replacement for access and refresh tokens instead.


packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/HttpLogObfuscatorTests.kt
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+
+    @Test
+    fun nullObfuscator() = runBlocking {
+        app = TestApp(
+            appName = "null-obfuscator",
+            logLevel = LogLevel.DEBUG,
+            builder = { it.httpLogObfuscator(null) },
+            customLogger = NullObfuscatorLoggerInspector,
```
rorbech, line 179: Seems a bit backwards. I would suggest doing a logger that just collects the messages and verifies the output afterwards like done with the https://github.com/realm/realm-kotlin/blob/main/packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/SyncedRealmTests.kt#L1262.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/HttpLogObfuscatorTests.kt
```diff
@@ -0,0 +1,354 @@
...
+            initialSetup = { app, service ->
+                initializeDefault(app, service)
+                app.addFunction(TestAppInitializer.FIRST_ARG_FUNCTION)
+                app.addFunction(TestAppInitializer.SUM_FUNCTION)
+                app.addFunction(TestAppInitializer.NULL_FUNCTION)
+            }
+        )
+
+        // Create user and log in
+        user = app.createUserAndLogin()
```
rorbech, line 189: The way the test is constructed it is a bit hidden what we actually assert. Why not just collect the messages and assert that the framework actually logged the original request/response messages. You would fail this test if you somehow filtered out all messages due to erroneous log level, etc.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/HttpLogObfuscatorTests.kt
```diff
@@ -0,0 +1,354 @@
...
+            call<BsonNull>(TestAppInitializer.NULL_FUNCTION.name)
+        }
+
+        Unit
+    }
+
+    @Test
+    fun emailPassword_registerAndLogin() = runBlocking {
+        app = initApp()
+        createUserAndLoginAssertions()
```
rorbech, line 231: Again a bit backwards. What about just using the above mentioned `CustomLogCollector` and asserting that it contains the obfuscated entry and doesn't container a non-obfuscated entry. This would make the test way more explicit instead of having to compare the actual tests with the implementation of the log collector.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/LogObfuscator.kt
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+// Replacer for any given feature but custom functions
+// FIXME Access and refresh tokens cannot be replaced by regex due to
+//  https://github.com/realm/realm-kotlin/issues/1284
+//  so until we figure out what is wrong, do manual string replacement and hopefully nothing will
+//  change - if the message string changes our tests will hopefully catch it
+internal class GenericRegexPatternReplacer(
+    private val patternReplacementMap: Map<Regex, String>
```
rorbech, line 165: What about taking `vararg pairs: Pair<Regex, String>` so that you could construct it directly with 
```
GenericRegexPatternReplace(
  x1 to y1,
  x2 to y2,
)
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/LogObfuscator.kt
```diff
@@ -0,0 +1,248 @@
...
+        return pattern.replace(input, replacement)
+    }
+}
+
+internal object LogObfuscatorImpl : HttpLogObfuscator {
+
+    private val urlRegex =
+        Regex("https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b([-a-zA-Z0-9()!@:%_\\+.~#?&\\/\\/=]*)")
+
+    private val features: Collection<ObfuscatorFeature> = setOf(
```
rorbech, line 208: Isn't this just the keys of `regexReplacerMap`? Why don't you just use something like
```
urlRegex.find(input)?.let { matchResult ->
   regexReplacerMap.filterKeys { url.container(it) }.values.map { it.findAndReplace (...) }
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/LogObfuscator.kt
```diff
@@ -0,0 +1,248 @@
...
+                }
+            }
+        }
+        return input
+    }
+}
+
+// Features that will be obfuscated by default. It uses the `urlPath` to tell the replacer when to
+// find the patterns in the logger
+internal enum class ObfuscatorFeature(internal val urlPath: String) {
```
rorbech, line 238: Again seems redundant compared to just use the `urlPath` of these as keys in `defaultFeatureToReplacerMap`

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/LogObfuscator.kt
```diff
@@ -0,0 +1,248 @@
...
+
+        // Defines the first part of the URL path to a feature
+        private const val AUTH = "auth"
+        internal const val FUNCTIONS = "functions/call"
+
+        // Credentials feature
+        private const val CREDENTIALS_PROVIDER = "$AUTH/providers"
+
+        // Credentials type
+        private const val PROVIDER_EMAIL_PASSWORD = "$CREDENTIALS_PROVIDER/local-userpass"
```
rorbech, line 40: Nit-pick (or maybe just a way off comment 🤪), but there are awfully many constants and most of them are just used once. I think I would have strived to somehow build these from some inline `setOf(...).map(...)`, but as said, might be that it would just cause more confusion. At least this is very clean ... though a bit overkill.

rorbech (commented):
Seems OK except for the missing `AppConfiguration.Builder().httpLogObfuscator` test and some minor doc comments ... but I guess I have quite a number of code/test style-ish comments. Maybe browse through it and we could have a chat about if you could be convinced about some of my arguments.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/HttpLogObfuscatorTests.kt
edualonso: It sounds better than sending events when the expected message has arrived. I will change it.

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/AppConfigurationTests.kt
```diff
@@ -349,21 +349,6 @@ class AppConfigurationTests {
 //        assertEquals(configCodecRegistry, config.defaultCodecRegistry)
 //    }
 //
-//    @Test
```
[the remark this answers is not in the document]
edualonso: Hm, I don't remember why I deleted this to be honest. It seems though I need to expose the obfuscator in the app config object itself which doesn't make a lot of sense though 🤔

clementetb (approved):
Looking good, added some minor comments

CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
+## 1.8.0 (YYYY-MM-DD)
```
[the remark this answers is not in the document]
cmelchior: No, this is adding new API's. This isn't just bug fixes, those normally go to `main`.

Issue #1288, opened 2023-02-19
Symbol cannot RealmUUID when emitting null value in MutableStateFlow more than once


pushpulldecoder:
MutableStateFlow throws an exception when null value is emitted more than once in MutableStateFlow<RealmUUID?> but works perfectly fine if Wrapped inside a class or converted to `realmUUId.bytes` or `realmUUId.toString()` and back.

### Code

```
class ExampleUnitTest {

	data class WrappedRealmUUId(val realmUUID : RealmUUID?)

	@Test
	fun realmUUId_null_flow_test() {
		val flow = MutableStateFlow<RealmUUID?>(null)
		runBlocking {
			this.launch {
				flow.collect { println(it) }
			}
			for (i in 0 .. 10) {
				flow.tryEmit(RealmUUID.random())
				delay(100)
				flow.tryEmit(null as RealmUUID?)
				delay(100)
			}
		}
	}

	@Test
	fun wrapped_realmUUId_null_flow_test() {
		val flow = MutableStateFlow<WrappedRealmUUId?>(null)
		runBlocking {
			this.launch {
				flow.collect { println(it) }
			}
			for (i in 0 .. 10) {
				flow.tryEmit(WrappedRealmUUId(RealmUUID.random()))
				delay(100)
				flow.tryEmit(null as WrappedRealmUUId?)
				delay(100)
			}
		}
	}
}
```

### Exception

```
class kotlinx.coroutines.internal.Symbol cannot be cast to class io.realm.kotlin.types.RealmUUID (kotlinx.coroutines.internal.Symbol and io.realm.kotlin.types.RealmUUID are in unnamed module of loader 'app')
java.lang.ClassCastException: class kotlinx.coroutines.internal.Symbol cannot be cast to class io.realm.kotlin.types.RealmUUID (kotlinx.coroutines.internal.Symbol and io.realm.kotlin.types.RealmUUID are in unnamed module of loader 'app')
	at io.realm.kotlin.internal.RealmUUIDImpl.equals(RealmUUIDImpl.kt:58)
	at kotlin.jvm.internal.Intrinsics.areEqual(Intrinsics.java:169)
	at kotlinx.coroutines.flow.StateFlowImpl.updateState(StateFlow.kt:329)
	at kotlinx.coroutines.flow.StateFlowImpl.setValue(StateFlow.kt:318)
	at kotlinx.coroutines.flow.StateFlowImpl.tryEmit(StateFlow.kt:370)
[...]

PR #1289, opened 2023-02-20
Fix RealmUUID equal comparison

clementetb:
Fixes https://github.com/realm/realm-kotlin/issues/1288

Comparing a `RealmUUID` instance with an object instance of a different type resulted on a `ClassCastException`


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmUUIDTests.kt
```diff
@@ -155,5 +155,7 @@ class RealmUUIDTests {
 
         @Suppress("EqualsNullCall")
         assertFalse(uuid1.equals(null)) // Do not assume type coercion in 'equals' implementation
+
+        assertFalse(uuid1.equals("hello world")) // Do not assume type coercion in 'equals' implementation
```
edualonso, line 159: 😰

edualonso (approved):

PR #1291, opened 2023-02-20
Update Core to 13.4.2.

### LINKED ISSUE #1285: Track latest changes to the Admin API in our tests
Some of our flexible sync tests might break after we update to a version of the BaaS server that includes these changes:

- https://github.com/10gen/baas/pull/8719
- https://github.com/10gen/baas/pull/8580

### Permissions: Current Implementation
These PRs change the way in which we define permissions for flexible sync apps. The current payload looks like this:
```
{
    "_id": {
        "$oid": "62f1a1637e9aa286af7198d5"
    },
    "name": "mongodb-svc",
    "type": "mongodb",
    "config": {
        "flexible_sync": {
            "state": "enabled",
            ...
            "permissions": { // flexible sync permissions read from here
                "rules": {
                    "table1": [ ... ],
                    "table2": [ ... ]
                    ...  
                },
                "defaultRoles": [ ... ]
            }
        }
    }
}
```
### Permissions: New Implementation
After these changes are introduced, the `permissions` block will be ignored and permissions will have to be defined outside the config object:
```
{
    "_id": {
        "$oid": "62f1a1637e9aa286af7198d5"
    },
    "name": "mongodb-svc",
    "type": "mongodb",
    "config": {
        "flexible_sync": {
            "state": "enabled",
            ...
            // "permissions" field here is being deprecated - if set
            // it will be ignored by the server
        }
    },
    // flexible sync permissions now being read from the below fields 
    "rules": [ ... ],
    "default_rule": { ... }
}
```
### Roles: Current Implementation
In addition to the changes mentioned above, setting service rules for roles is also changing. Currently the `roles` object looks like this:
```
{
    "roles": [{
        "name": "defaultRole",
        "apply_when": {},
        "insert": true,
        "delete": true,
    }]
}
```
### Roles: New Implementation
But a new `document_filters` filed will be necessary now. A `roles` object that makes all objects and fields readable/writeable will look like this:
```
{
    "roles": [{
        "name": "defaultRole",
        "apply_when": {},
        "document_filters": {
            "read": true,
            "write": true
        },
        "write": true,
        "read": true,
        "insert": true,
        "delete": true,
    }]
}
```

cmelchior:
Closes #1285 

Including updating to the latest BAAS test server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * None.
 
 ### Enhancements
+* Upgrade OpenSSL from 3.0.7 to 3.0.8.
 * Add support for querying on RealmSets containing objects with `RealmSet.query(...)`.  (Issue [#1037](https://github.com/realm/realm-kotlin/issues/1258))
 * [Sync] Add support for setting App Services connection identifiers through `AppConfiguration.appName` and `AppConfiguration.appVersion`, making it easier to identify connections in the server logs.
 . (Issue (#407)[https://github.com/realm/realm-kotlin/issues/407])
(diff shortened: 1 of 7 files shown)


edualonso (approved):
Great 🚀  I somehow expected the BaaS changes would trigger some obscure errors. I'm happy I was wrong 😄

Issue #1268, opened 2023-02-03
Align error handling for operations on collections whose Realm is no longer open


edualonso:
Should we rely on Core throwing an `IllegalStateException` with `Access to invalidated Collection` or rather catch it ourselves earlier by calling `realmReference.checkClosed()`?

Right now `ManagedRealmList<E>.query()` relies on the former, whereas (most of) the overriden `List` interface functions use the latter. Same applies for sets and dictionaries.

rorbech:
We cannot rely on our own check as the realm could be closed between us checking and the actual core operation. So there is not really a choice. I would vote for just getting rid of our own check and rely exclusively on Core, but maybe better to clean up after #889.

Issue #941, opened 2022-07-26
Alignment needed before adding dictionaries


edualonso:
Before starting working on adding dictionaries we should:
- Move all collection internal classes (e.g.`UnmanagedRealmSet`, `ManagedRealmList`, etc.) to separate files
- Align to a reasonable extent the test framework of both lists and sets and consider dictionaries while at it, though it might not be fully possible due to maps not being `Collection`s.
- Align to a reasonable extent the test framework of collection notifications.
- Optimise the notification callback code in `realm_api_helpers` - see comments in `register_notification_cb`.
- More?

PR #1292, opened 2023-02-20
Fix semantics for `RealmSet.iterator` to handle concurrent modifications

### LINKED ISSUE #1220: Fix `RealmSet.iterator()` wrong semantics
The semantics of the rather naive implementation of `RealmSet.iterator()` are wrong in case of deleting elements since it does not keep track of the index for the last returned element, causing it to return or remove the wrong element after the first removal, failing in subsequent removals depending on the size of the set.

edualonso:
Closes #1220 

### WARNING
**DO NOT MERGE UNTIL #1239 IS MERGED SINCE THE CHANGES ARE BUILT ON TOP OF THE DICTIONARY FEATURE!!!**

Fixed implementation for `RealmSet.iterator()` to add concurrent modification handling similarly to what we do for dictionaries. To add full support I had to add custom implementations for `remove` and `removeAll` and refactor some of the calls to the operator functions to keep track of the modifications. Also fixed a wrong string in an exception being thrown in the dictionary iterator.


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -740,6 +764,68 @@ internal abstract class ManagedSetTester<T>(
...
+
+            errorCatcher {
+                realm.writeBlocking {
+                    val set = typeSafetyManager.createContainerAndGetCollection(this)
+                    set.addAll(dataSet)
+                    assertTrue(set.removeAll(dataSet))
+
+                    // TODO https://github.com/realm/realm-kotlin/issues/1097
+                    //  If the RealmAny instance contains an object it will NOT be removed until
+                    //  the issue above is fixed
```
rorbech, line 805: Shouldn't this be tracked as separate issue? One thing is that our `RealmObject.equals` has the wrong semantics, but it is a more critical issue that we are not removing objects just because the current implementation relies on that `equals`. We should probably prioritize the actual issue of removal of `RealmAny` containing a `RealmObject` higher ... and I guess on fix could just be to fix the `equals`.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -818,13 +913,21 @@ internal abstract class ManagedSetTester<T>(
...
 
-                assertFailsWith<NoSuchElementException> { (iterator.next()) }
-                set.addAll(dataSet)
-                while (iterator.hasNext()) {
-                    val element = iterator.next()
-                    assertTrue(structuralContains(dataSet, element))
+                    // Concurrent modification on the same iterator
+                    assertFailsWith<ConcurrentModificationException> {
+                        iterator.next()
+                    }
```
rorbech, line 923: Shouldn't this just go as it is tested in `iterator_concurrentModification`?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -798,11 +884,20 @@ internal abstract class ManagedSetTester<T>(
...
-                set.addAll(dataSet)
-                assertTrue(iterator.hasNext())
+                set.iterator().also { iterator ->
+                    assertFalse(iterator.hasNext())
+                    set.addAll(dataSet)
+
+                    // Concurrent modification on the same iterator
+                    assertFailsWith<ConcurrentModificationException> {
+                        iterator.hasNext()
+                    }
```
rorbech, line 895: Shouldn't this just go as it is tested in iterator_concurrentModification?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
```diff
@@ -870,6 +973,56 @@ internal abstract class ManagedSetTester<T>(
...
+                clearIterator.next()
+                set.clear()
+                assertFailsWith<ConcurrentModificationException> {
+                    clearIterator.remove()
+                }
+            }
+        }
+
+        // Makes no sense to test concurrent modifications outside the transaction, so clean up only
+        assertContainerAndCleanup()
```
rorbech, line 1023: Funny name if we sometimes don't assert. Should the actual clean up just be somewhere else in the test infrastructure?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmSetInternal.kt
```diff
@@ -214,30 +247,73 @@ internal fun <E : BaseRealmObject> ManagedRealmSet<E>.query(
  */
 internal interface SetOperator<E> : CollectionOperator<E, RealmSetPointer> {
 
+    // Modification counter used to detect concurrent writes from the iterator
```
rorbech, line 250: Maybe leave a note and reference if this is _"inspired"_ from Java's AbstractList implementation.

rorbech (approved):
Only minor comments about more explicit comments and stuff like that.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmSetInternal.kt
```diff
@@ -84,47 +83,81 @@ internal class ManagedRealmSet<E> constructor(
...
         return object : MutableIterator<E> {
 
-            private var pos = -1
+            private var expectedModCount = operator.modCount // Current modifications in the map
+            private var cursor = 0 // The position returned by next()
+            private var lastReturned = -1 // The last known returned position
 
-            override fun hasNext(): Boolean = pos + 1 < size
+            override fun hasNext(): Boolean {
+                operator.realmReference.checkClosed()
```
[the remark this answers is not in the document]
edualonso: Shameful copy paste line. I will remove the ones in this iterator, but the others will have to wait until we address https://github.com/realm/realm-kotlin/issues/1268 and https://github.com/realm/realm-kotlin/issues/889.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
edualonso: It can't really be moved to `@AfterTest` because it has to be done after every type is tested inside the same test to avoid overcomplicating the setup process for each type - but of course it could be done differently, perhaps should have been done [here](https://github.com/realm/realm-kotlin/issues/941) but we never prioritised it high enough so 🤷 . In any case, I think it's fine having the assertions as an optional parameter but I grant you the name could be better, but you know, `#naming`...

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmSetTests.kt
edualonso: That's true, the comment is misleading considering the issue doesn't mention `RealmAny`. I will modify the issue to also consider `RealmAny` containing objects as well. I agree we should do this rather soon as there seem to be people interested in it.

Issue #1290, opened 2023-02-20
Native Sync Thread is leaking


astatio:
The number of realm threads has increased exponentially after 1.6.0. In one of my threads dumps that I made to check why my bot was becoming unresponsive i saw 8435 threads in a single realm db file! The thread dump was also over 23MB. This did not happen before. It seems there's a serious memory leak problem. 

I would like to highlight that this comes after the following issue https://github.com/realm/realm-kotlin/issues/1007 was solved and happens under the same conditions: a realm with lots of writes under short amounts of time (3 per 5s, for example)

cmelchior:
Hi @astatio. That doesn't right. It is true that we changed how our threads are managed and currently there is a known leak for Synced Realms, but it should only leak a single thread.

Are you by any chance opening and closing Realms during the writes?

astatio:
It appears that is correct i do indeed call a method that opens and closes the same Realm file to perform a query while still having the same Realm file open. Is that what causes the problem? If so, shouldn't it throw an Exception?

cmelchior:
Yes, that would cause this problem. We will be working on a fix, but note, that in Realm Kotlin (unlike Realm Java), you don't really need to open and close the Realm. You can just open it once and keep it open for the lifetime of the app.

Issue #792, opened 2022-04-19
Internal refactor: Split RealmInterop and common native pointer owner interface


rorbech:
From #713 

(Christian - 2) RealmInterop is too huge and annoying to reason about. I would like to split it into more manageable chunks that mirror the ObjectStore classes: Realm, Results, List, Object, or something along those lines.

(Claus - 2) Interface for objects holding a native pointer. For close() on all managed resources, and in general common operations like realm_release, realm_equals, etc.

PR #1299, opened 2023-03-01
Correctly cleanup Sync Client thread when it is destroyed on the C++ side

cmelchior:
Close https://github.com/realm/realm-kotlin/issues/1290

This PR makes sure we correctly clean up the Sync thread. This is currently leaking, which on the JVM prevents it from shutting down gracefully, and depending on your usage (like in #1290), it might also be a big resource drain.

There is an open question about how to handle exceptions on the Sync thread. Ideally, they should never happen and they are reported back to us as just a string. For now, I just print them (waiting for proper logging), but we could also consider throwing an exception, but it will 100% tear the app down with it. 

In Java, we throw an Exception, and I don't think we have seen

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -539,7 +539,13 @@ expect object RealmInterop {
     fun realm_user_refresh_custom_data(app: RealmAppPointer, user: RealmUserPointer, callback: AppCallback<Unit>)
 
     // Sync client config
-    fun realm_sync_client_config_new(appId: String): RealmSyncClientConfigurationPointer
+    fun realm_sync_client_config_new(): RealmSyncClientConfigurationPointer
+
+    fun realm_sync_client_config_set_default_binding_thread_observer(
+        syncClientConfig: RealmSyncClientConfigurationPointer,
+        appId: String
+    )
+
     fun realm_sync_client_config_set_base_file_path(
         syncClientConfig: RealmSyncClientConfigurationPointer,
         basePath: String
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -539,7 +539,7 @@ expect object RealmInterop {
     fun realm_user_refresh_custom_data(app: RealmAppPointer, user: RealmUserPointer, callback: AppCallback<Unit>)
 
     // Sync client config
-    fun realm_sync_client_config_new(): RealmSyncClientConfigurationPointer
+    fun realm_sync_client_config_new(appId: String): RealmSyncClientConfigurationPointer
     fun realm_sync_client_config_set_base_file_path(
         syncClientConfig: RealmSyncClientConfigurationPointer,
         basePath: String
--- a/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -954,8 +954,33 @@ actual object RealmInterop {
         )
     }
 
-    actual fun realm_sync_client_config_new(): RealmSyncClientConfigurationPointer {
-        return LongPointerWrapper(realmc.realm_sync_client_config_new())
+    actual fun realm_sync_client_config_new(appId: String): RealmSyncClientConfigurationPointer {
+        val ptr: RealmSyncClientConfigurationPointer = LongPointerWrapper(realmc.realm_sync_client_config_new())
+        realmc.realm_sync_client_config_set_default_binding_thread_observer(ptr.cptr(), object: SyncThreadObserver {
+            override fun threadName(): String {
+                return "SyncThread-$appId"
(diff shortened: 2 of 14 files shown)


packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -954,8 +954,41 @@ actual object RealmInterop {
         )
     }
 
-    actual fun realm_sync_client_config_new(): RealmSyncClientConfigurationPointer {
-        return LongPointerWrapper(realmc.realm_sync_client_config_new())
+    actual fun realm_sync_client_config_new(appId: String): RealmSyncClientConfigurationPointer {
+        val ptr: RealmSyncClientConfigurationPointer = LongPointerWrapper(realmc.realm_sync_client_config_new())
+        realmc.realm_sync_client_config_set_default_binding_thread_observer(
```
rorbech, line 959: Hmm. I am a little against having this kind of logic inside the `realm_sync_client_config_new` as `cinterop` usually is isolated to dispatch individual calls to the C-API. But I guess we are also trying to make RealmInterop platform agnostic 🤔 There are other cases where we have pushed common code into RealmInterop which sort of goes the other direction requiring us to keep the platform implementations in sync manually, i.e.
- https://github.com/realm/realm-kotlin/blob/main/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt#L1330
- https://github.com/realm/realm-kotlin/blob/main/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt#L1152

I guess the ideal solution would be to have a platform agnostic API at the `cinterop` layer with ability to include implementation of common things. I guess it is a bit overkill to do as part of this PR and fits better into a task like #792, but I think it is still valid to have in mind going forward and should maybe push for prioritization of #792.

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -539,7 +539,7 @@ expect object RealmInterop {
     fun realm_user_refresh_custom_data(app: RealmAppPointer, user: RealmUserPointer, callback: AppCallback<Unit>)
 
     // Sync client config
-    fun realm_sync_client_config_new(): RealmSyncClientConfigurationPointer
+    fun realm_sync_client_config_new(appId: String): RealmSyncClientConfigurationPointer
```
rorbech, line 542: Maybe add a comment about why we need the `appId` here ... but I guess the need for that ties into my comment against actually registering the thread-notifier-callback in here.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -954,8 +954,41 @@ actual object RealmInterop {
...
+                @Suppress("TooGenericExceptionThrown")
+                override fun onError(error: String) {
+                    // TODO Wait for https://github.com/realm/realm-core/issues/4194 to correctly
+                    //  log errors. For now, just throw an Error. Exceptions from the Sync Client
+                    //  indicate something is fundamentally is wrong on the Sync Thread.
+                    //
+                    //  As any exception here will be thrown out of context with no way to catch it, we
+                    //  turn it into an Error. Also, in Realm Java, these exceptions have only been
+                    //  reported when we integrated features ourselves, which indicates we want to
+                    //  make it very visible when they happen.
```
rorbech, line 986: Hmm. Seems like the PR description and the implementation is a bit schizophrenic about this ... which one do you prefer. If it is right that we shouldn't ever see this I guess I would go with throwing and updating the comment as suggested, but ... 🙈 
```suggestion
                    // TODO Wait for https://github.com/realm/realm-core/issues/4194 to correctly
                    //  log errors. For now, just throw an Error as exceptions from the Sync Client
                    //  indicate that something is fundamentally wrong on the Sync Thread. 
                    //  In Realm Java this has only been reported during development of new
                    //  features, so throwing an Error seems appropriate to increase visibility.
```

rorbech (commented):
LGTM except that it seems like description and implementation is diverging a bit. Are we clear on what we want for errors? 

There are some comments on where and how to share code vs. implementing platform specific things in `RealmInterop`. This was more to raise the concern on the topic than a request for change as I think it is of more fundamental layering concern and probably fits better into #792.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
rorbech: I think I would feel it would be safer, as these kind of things would be less hidden. Imagining introducing a new platform in `cinterop` - then it is just a matter of binding the C-API and now replicating other logic into various methods. But it would then be an issue knowing that `realm_sync_client_config_set_default_binding_thread_observer` is a no-op on Darwin and not trust it to take care of something that it doesn't do. But we could also just do the full implementation on Darwin if it made sense for logging thread creation/teardown 🤔 I guess a single method registration call and a callback wouldn't matter that much on this code path anyway.

cmelchior: I reworked it so there is also now a thread observer on Darwin. We don't need it for cleaning up threads, but at least handling errors on the Sync thread will be done in the same now.

rorbech (approved):

PR #1304, opened 2023-03-08
Allow returning invalid objects from writes

### LINKED ISSUE #1300: Deleting Non-Optional Frozen RealmObject Crashes
Hi!

I was looking at the docs for [deleting a frozen object](https://www.mongodb.com/docs/realm/sdk/kotlin/realm-database/frozen-arch/), and the example shows how to delete an optional RealmObject, which works fine. If, however, I modify the query to return a non-optional like below:

```
val sample: Sample = realm.query<Sample>().find().first()

realm.writeBlocking {
    findLatest(sample)?.also { delete(it) }
}
```

The code crashes with the following error:
```
java.lang.IllegalStateException: A deleted Realm object cannot be returned from a write transaction.
```

But if I add any other line (i.e. `println` or `return`) in the `writeBlocking` clause, it deletes the object without erroring out:
```
realm.writeBlocking {
    findLatest(sample)?.also { delete(it) }
    return@writeBlocking // Replace with anything else
}
```
Is this the intended behavior when deleting frozen, non-optional RealmObjects?

cmelchior:
Closes #1300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * None.
 
 ### Fixed
+* Returning invalid objects from `Realm.write` would throw an `IllegalStateException`. (Issue [#1300](https://github.com/realm/realm-kotlin/issues/1300))
 * Compatibility with Realm Java when using the `io.realm.RealmObject` abstract class. (Issue [#1278](https://github.com/realm/realm-kotlin/issues/1278))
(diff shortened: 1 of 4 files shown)


rorbech (commented):
LGTM. I guess we could consider targeting `release` ... or is it safer to do in a minor release than in a patch release. And when CI is green of course.

rorbech (approved):

Issue #1226, opened 2023-01-20
Throw compile time error when illegal schema types are used


cmelchior:
This compiles fine:

```
class Person: RealmObject {
  var tasks: Array<Task>? = null
}
```

Even though Array is not a supported type. It should result in a compile time warning instead.

cmelchior:
It looks like error handling in general is missing. This model class also compiles fine even though BsonDateTime isn't a supported field:

```
class Talk: RealmObject {
    @PrimaryKey
    var id: ObjectId = ObjectId()
    var title: String = ""
    var description: String = ""
    var durationInMin: Int = 0
    var scheduleOn: BsonDateTime = BsonDateTime()
    var location: String = ""
}
```

cmelchior:
Another example. The property just gets ignored. We should really throw a build error here:

```
class Something: RealmObject {
    var foo = "bar"
    var bar = realmListOf("foo", "bar", "baz")
}

class GamePlayer: RealmObject {
    @PrimaryKey
    var name: String = ""
    var type: String = ""
    var child: Something? = Something()
}
```

PR #1307, opened 2023-03-08
Fail build if model class contain types not supported.

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/1226

This mirrors the behavior in Realm Java, where we also fail at compile time for unknown types.

### DIFF
--- a/examples/realm-java-compatibility/app/src/main/java/io/realm/kotlin/demo/javacompatibility/data/java/JavaRepository.kt
+++ b/examples/realm-java-compatibility/app/src/main/java/io/realm/kotlin/demo/javacompatibility/data/java/JavaRepository.kt
@@ -45,12 +45,12 @@ class JavaRepository(appContext: Context) : Repository {
 
     init {
         Realm.init(appContext)
-        realm = Realm.getInstance(
-            RealmConfiguration.Builder()
-                .name("java.realm")
-                .allowWritesOnUiThread(true)
-                .build()
-        )
+        val config = RealmConfiguration.Builder()
+            .name("java.realm")
+            .allowWritesOnUiThread(true)
+            .build()
+        Realm.deleteRealm(config)
+        realm = Realm.getInstance(config)
         realm.executeTransaction {
             realm.createObject(JavaEntity::class.java)
             val entities = realm.where(JavaEntity::class.java).findAll()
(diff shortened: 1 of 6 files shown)


examples/realm-java-compatibility/app/src/main/java/io/realm/kotlin/demo/javacompatibility/data/java/JavaRepository.kt
```diff
@@ -45,12 +45,12 @@ class JavaRepository(appContext: Context) : Repository {
...
-            RealmConfiguration.Builder()
-                .name("java.realm")
-                .allowWritesOnUiThread(true)
-                .build()
-        )
+        val config = RealmConfiguration.Builder()
+            .name("java.realm")
+            .allowWritesOnUiThread(true)
+            .build()
+        Realm.deleteRealm(config)
```
rorbech, line 52: Nit pick. Maybe a bit aggressive, maybe we could just `deleteIfMigrationRequired` ... or what it is called.

examples/realm-java-compatibility/app/src/main/java/io/realm/kotlin/demo/javacompatibility/data/kotlin/KotlinRepository.kt
```diff
@@ -33,9 +33,14 @@ class KotlinEntity : RealmObject {
...
 class KotlinRepository: Repository {
 
-    val realm = Realm.open(RealmConfiguration.Builder(setOf(KotlinEntity::class)).name("kotlin.realm").build())
+    val realm: Realm
 
     init {
+        val config = RealmConfiguration.Builder(setOf(KotlinEntity::class))
+            .name("kotlin.realm")
+            .build()
+        Realm.deleteRealm(config)
```
rorbech, line 42: Same

rorbech (approved):

examples/realm-java-compatibility/app/src/main/java/io/realm/kotlin/demo/javacompatibility/data/java/JavaRepository.kt
cmelchior: I added this because we are using `copyToRealm` in the Repository...something was causing the Realm to not be cleaned up which caused every subsequent test to fail with "Primary key already exists".

So while it is a work-around, it is similar to our tests using a random test directory. Using `deleteIfMigrationRequired` wouldn't help in this case.

rorbech: Ok, let's just wipe it 🧹

PR #1308, opened 2023-03-08
Remove all uses of freeze

### LINKED ISSUE #1294: Remove old memory model-related `freeze` calls
There are still quite a lot of `freeze` calls from pre-K/N new memory model times that should be removed. It is not strictly needed because the function is a no-op now though, but nice to have.

cmelchior:
Closes #1294 

This PR removes all uses of `freeze`. It was already a no-op when moving to the new memory model and we just kept it to be able to shift back easily. But there has been no pushback against moving to the new memory model, so might as well cleanup the code.


rorbech (approved):

Skaldebane:
Hi! This is all great, but the documentation on mongodb.com has not been updated to reflect this fact, which was confusing since I knew that "freezing" was no longer necessary in Kotlin/Native.

PR #1312, opened 2023-03-09
Fix leaking internal classes

clementetb:
Removes the leakage of internal classes by symlinking the sources


cmelchior:
Bit rotted. We should revisit this after having done unified networking.

PR #1315, opened 2023-03-10
[POC] Dataframe support

cmelchior:
This PR demonstrates how Realm Kotlin could interop with [Kotlin DataFrames](https://github.com/Kotlin/dataframe). This is particularly valuable for data science tasks. This also demonstrates that it is possible to pull in synced Realms into Jupyter Notebook, which would be a convenient way to share data between larger teams.

It consists of:

-  a `ext-dataframe` module that publishes a normal Kotlin Multiplatform library that can be consumed on Android and JVM (the only platforms supported by dataframes). The extension library exposes a number of extension methods on both `DataFrame` and `Realm` classes that make it possible to move data between the two worlds.

- a Jupyter Notebook example file in `packages/ext-dataframe/jupyter/example-realm.ipynb`. There is also a demonstration of pulling in a synced Realm. It uses a public Atlas App Services application that is set to ready-only.

- A json file making it easier to add Realm support to Jupyter. It will need to be pushed to https://github.com/Kotlin/kotlin-jupyter-libraries before people can reference it the normal way. The example notebook just references the local file.

- Github does a decent job at rendering the notebook (without the output) here: https://github.com/realm/realm-kotlin/blob/cm/dataframe-support/packages/ext-dataframe/jupyter/example-realm.ipynb

**TODO**
The extension is not ready as is and will in particular need work in the following areas:

- [ ] Figure out exactly what namespace this should be distributed under. I would prefer `io.realm.kotlinx.dataframe` but the way our publishing works right now doesn't allow this. So this PR uses `io.realm.kotlin:ext-dataframe`.
- [ ] We need a public `DynamicRealm` API. Right now I hacked our implementation in order to expose a `DynamicRealm`.
- [ ] Exposing a dynamic schema, requires as a minimum https://github.com/realm/realm-core/issues/6370.
- [ ] Currently, we require a schema for `RealmConfiguration`. Figure out how that should work for dynamic realms.
- [ ] There is currently no way to create a `MutableDynamicRealm`, so it isn't possible to write dataframes back into the Realm file.
- [ ] I haven't looked much into typed Realms since Jupyter is generally used for exploration, so providing the schema upfront might be impossible. It is particularly unclear how to apply our compiler plugin, and it might be the case it isn't even possible.
- [ ] Testing.

**Screenshots**
image
image
image


cmelchior:
This PR is waiting for a DynamicRealm API before we can do much with it. Closing for now, we can always pull it out from Git when needed.

PR #1328, opened 2023-03-29
Add GitHub issue templates

rorbech:
This adds a menu and forms for creating new issues. Just grabbed the Java templates and updated the link to the forum to show the kotlin issues instead.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -12,9 +12,10 @@ body:
     attributes:
       label: How frequently does the bug occur?
       options:
-        - Seen once
+        - -- select --
+        - Once
         - Sometimes
-        - All the time
+        - Always
     validations:
       required: true
 # Description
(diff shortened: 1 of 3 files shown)


bmunkholm (changes requested):
This doesn't look like the latest one from Java? Did you use an older version?
All dropdown options has to be exactly the same - character-by-charater.

rorbech:
> This doesn't look like the latest one from Java? Did you use an older version? All dropdown options has to be exactly the same - character-by-charater.

Argh, yes. I copied it from the `transformer-api` branch but that is spun from the `releases` branch and not `main` as I recalled. I have updated it from the latest `main` version. Thanks.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,106 @@
...
+# Environment
+  - type: input
+    id: platform
+    attributes:
+      label: Platform OS and version(s)
+      description: OS and version(s) are you seeing the issue on?
+    validations:
+      required: true
+  - type: textarea
+    id: java-build-environment
```
bmunkholm, line 97:
```suggestion
    id: kotlin-build-environment
```

bmunkholm (approved):

cmelchior (approved):

PR #1341, opened 2023-04-07
Fix issues related to using a Windows filesystem in tests

cmelchior:
In a number of places, we made assumptions about how the filesystem worked, which broke our tests on Windows. This PR fixes those tests.

### DIFF
--- a/buildSrc/src/main/kotlin/io/realm/RealmPublishPlugin.kt
+++ b/buildSrc/src/main/kotlin/io/realm/RealmPublishPlugin.kt
@@ -32,6 +32,7 @@ import org.gradle.kotlin.dsl.getByType
 import org.gradle.kotlin.dsl.withType
 import org.gradle.plugins.signing.SigningExtension
 import org.gradle.plugins.signing.SigningPlugin
+import java.io.File
 import java.net.URI
 import java.time.Duration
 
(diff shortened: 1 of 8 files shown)


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmConfigurationTests.kt
```diff
@@ -156,7 +156,7 @@ class RealmConfigurationTests {
 
     @Test
     fun directory_createIntermediateDirs() {
-        val realmDir = tmpDir + "/my/intermediate/dir"
+        val realmDir = tmpDir + "${PATH_SEPARATOR}my${PATH_SEPARATOR}intermediate${PATH_SEPARATOR}dir"
```
rorbech, line 159: Could also just use `setOf(tmpDir, ....).joinToString() { }` another time. Or maybe even an utility `pathOf(...)`. Would be a little easier to read.

rorbech (approved):
Good catches 🙈 Only some minor comments.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmConfigurationTests.kt
cmelchior: Ah yeah, good idea 👍

PR #1351, opened 2023-04-17
Return correct provider for JWT-authenticated users

### LINKED ISSUE #1350: JWT AuthenticationProvider not fully supported

Always

### Description

When checking for the provider of a user signed in with a JWT token, the sdk crashes.

The reason for the crash is that there is a kotlin [TODO() function in getProviderFromCore](https://github.com/realm/realm-kotlin/blob/ad695196c9f37dd0feecc76b9ac49f94fdf91477/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/UserImpl.kt#L195)
(Other mappings also contain TODO() functions).

Changing the line linked above to
`AuthProvider.RLM_AUTH_PROVIDER_CUSTOM -> AuthenticationProvider.JWT`
should fix the issue.
In [another location](https://github.com/realm/realm-kotlin/blob/ad695196c9f37dd0feecc76b9ac49f94fdf91477/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AuthenticationProviderImpl.kt#L41), this has already been done.

### Stacktrace & log output

```shell
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: <removed>, PID: 22790
    kotlin.NotImplementedError: An operation is not implemented.
        at io.realm.kotlin.mongodb.internal.UserImpl.getProviderFromCore(UserImpl.kt:196)
        at io.realm.kotlin.mongodb.internal.UserImpl.getProvider(UserImpl.kt:56)
...
 I/Process: Sending signal. PID: 22790 SIG: 9
```

Always

### Reproduction Steps

Access UserImpl::provider for a user logged in with a JWT token.

### Version

1.7.0

Both Atlas Device Sync and Atlas App Services

### Are you using encryption?

No

Android sdk 31

### Build environment

Android Studio version: 2021.2.1 Patch 1 (Chipmunk)
Android Build Tools version: 33.0.1
Gradle version: 7.3.3

rorbech:
Closes #1350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
  
 ### Fixed
 * [Sync] Client Reset on JVM on Linux would crash with `No built-in scheduler implementation for this platform. Register your own with Scheduler::set_default_factory()`
+* [Sync] Return correct provider for JWT-authenticated users. (Issue [#1350](https://github.com/realm/realm-kotlin/issues/1350))
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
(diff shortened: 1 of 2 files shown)


cmelchior (approved):
If CI is happy 👍 

It is a bit unfortunate that there is no great way to test this, but it is probably not worth it spending a lot of time trying to mock a login

PR #1354, opened 2023-04-19
Add support for authentication flows

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/749

Opening this PR with a few design decisions we need to agree on:

- [ ] In Realm Java, we only report loggedIn/loggedOut, but there also exists a REMOVED state (which is loggedOut with extra semantics). Should we also expose removed in this flow?
- [ ] Currently, I expose an `AuthenticationChange` with a `AuthenticationChange.Type` change. Is that a good approach? There exist two alternatives: 1) Instead reuse the `User.State` which has the same semantics 2) Our object notifications use sealed classes, should we do this instead? Which would also allow us to remove the `Type` enum? I'm probably slightly leaning toward this alternative.
- [ ] #namingIsHard, `authenticationChangeAsFlow` mirrors, kinda, the `SyncSession.connectionStateAsFlow()` as API, not sure if a better name exists?

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
@@ -23,6 +23,8 @@ import io.realm.kotlin.mongodb.exceptions.AuthException
 import io.realm.kotlin.mongodb.exceptions.InvalidCredentialsException
 import io.realm.kotlin.mongodb.internal.AppConfigurationImpl
 import io.realm.kotlin.mongodb.internal.AppImpl
+import io.realm.kotlin.mongodb.sync.ConnectionStateChange
+import kotlinx.coroutines.flow.Flow
 
 /**
  * An **App** is the main client-side entry point for interacting with an **Atlas App Services
@@ -100,6 +102,14 @@ public interface App {
      * communicating with App Services. See [AppException] for details.
      */
     public suspend fun login(credentials: Credentials): User
+
+    /**
+     * Create a [Flow] of [AuthenticationChange]-events to receive notifications of updates to all
+     * app user auth states like login and logout.
+     *
+     * @return a [Flow] of authentication events for users associated with this app.
+     */
+    public fun authenticationChangeAsFlow(): Flow<AuthenticationChange>;
 
     /**
      * Close the app instance and release all underlying resources.
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationChange.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationChange.kt
@@ -1,0 +1,40 @@
+package io.realm.kotlin.mongodb
+
+/**
+ * TODO
+ */
+public data class AuthenticationChange(
+    /**
+     * TODO
+     */
+    public val type: Type,
+    /**
+     * TODO
+     */
+    public val user: User
+) {
+
+    /**
+     * TODO
+     */
+    public enum class Type {
+        /**
+         * TODO
+         */
+        LOGGED_IN,
+        /**
+         * TODO
+         */
+        LOGGED_OUT
+    }
+
+    /**
+     * TODO
+     */
+    public fun didLogIn(): Boolean = (type == Type.LOGGED_IN)
+
+    /**
+     * TODO
+     */
+    public fun didLogOut(): Boolean = (type == Type.LOGGED_OUT)
+}
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppImpl.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppImpl.kt
@@ -25,10 +25,14 @@ import io.realm.kotlin.internal.interop.sync.NetworkTransport
 import io.realm.kotlin.internal.util.Validation
 import io.realm.kotlin.internal.util.use
 import io.realm.kotlin.mongodb.App
+import io.realm.kotlin.mongodb.AuthenticationChange
 import io.realm.kotlin.mongodb.Credentials
 import io.realm.kotlin.mongodb.User
 import io.realm.kotlin.mongodb.auth.EmailPasswordAuth
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.cancel
 
 // TODO Public due to being a transitive dependency to UserImpl
 public class AppImpl(
@@ -37,6 +41,7 @@ public class AppImpl(
 
     internal val nativePointer: RealmAppPointer
     private val networkTransport: NetworkTransport
+    private val authenticationChangeFlow = MutableSharedFlow<AuthenticationChange>()
 
     init {
         val appResources: Pair<NetworkTransport, NativePointer<RealmAppT>> = configuration.createNativeApp()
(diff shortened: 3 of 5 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationChange.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+public sealed interface AuthenticationChange {
+    /**
+     * TODO
+     */
+    public val user: User
+    /**
+     * TODO
+     */
+    public fun didLogIn(): Boolean
```
clementetb, line 14: How come that we like to expose the state as a boolean when we already have it as an specialized interface in: `LoggedIn` or `LoggedOut`?

clementetb:
I think that I would use a sealed interface, as we do with the other change notification interfaces. I would also remove the boolean as the status is already defined by the type of instance received. Naming is hard, I cannot think on something better, maybe we could shorten it to `AuthChange`?

```kotlin
public sealed interface AuthChange {
    public val user: User
}

public interface LoggedIn: AuthChange
public interface LoggedOut: AuthChange
public interface Removed: AuthChange

app.authChangeAsFlow().collect { change: AuthChange ->
    when(change) {
        is LoggedIn -> handleLogin(change.user)
        is LoggedOut -> handleLogout(change.user)
        is Removed -> handleRemove(change.user)
    }
}
```

I would leave it as simple as possible, we could add the booleans later on if required.

cmelchior:
PR should now reflect our discussion and is ready for review.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/UserImpl.kt
```diff
@@ -94,6 +94,7 @@ public class UserImpl(
         }
 
     override suspend fun logOut() {
+        val reportLoggedOut = loggedIn
```
rorbech, line 97: This is not really safe. Two threads calling this simultaneous would issue two events but should only emit one. I guess we would either need a return value from core or synchronize the operations for a use. Don't know if it is a critical issue though. Same for the other events.

cmelchior: Good catch. I already created https://github.com/realm/realm-core/issues/6516 which will also fix this. I'll think and see if there is something we can do to prevent this, but if it is super tricky I would probably just ignore this case. I doubt this is code that will see race conditions anyway.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppImpl.kt
```diff
@@ -73,10 +77,28 @@ public class AppImpl(
...
             )
             return channel.receive()
-                .getOrThrow()
+                .getOrThrow().also { user: User ->
+                    reportUserLoggedIn(user)
+                }
         }
     }
 
+    private suspend fun reportUserLoggedIn(user: User) {
```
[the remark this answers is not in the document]
cmelchior: I have merged the different methods.

We are emitting events inside a suspend function and the flow is suspending, so I don't think we need to use `checkForBufferOverFlow`?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/UserImpl.kt
cmelchior: I reworked the logic a little bit. The chance for a race condition should be less now, although still present. My stance would probably be that using locks for this would be overkill and we should wait for https://github.com/realm/realm-core/issues/6516 for a proper fix, but not sure what you think?

rorbech: Sounds ok to leave it as is for now. Maybe include a link to the core issue in a comment  above `reportLoggedOut`.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppImpl.kt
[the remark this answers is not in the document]
rorbech: Hmm. I guess you are right that we are just suspending if the recipient is not ready ... But is what we want? Would be quite tricky to see that the login/logout is blocked due to the observer of the `authencationChangeAsFlow()`-flow being locked. Don't know if there is an easy way to _channel_ the flow to follow the approach from our other flows. This is also way less active, so maybe just fine by adding some buffer space ... or is it then even less likely to spot if it doesn't happen on every `emit` and we don't throw if the buffer is full 🤔

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationChange.kt
```diff
@@ -0,0 +1,49 @@
...
+ *          is Removed -> handleRemove(change.user)
+ *       }
+ *   }
+ * ```
+ */
+public sealed interface AuthenticationChange {
+    /**
+     * A reference to the [User] this event happened to.
+     *
+     * *Warning:* This is the live user object, so the [User.state]] might have diverged from the
```
clementetb, line 28:
```suggestion
     * *Warning:* This is the live user object, so the [User.state] might have diverged from the
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationChange.kt
```diff
@@ -0,0 +1,49 @@
+package io.realm.kotlin.mongodb
```
clementetb, line 1: Missing ©

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AuthenticationChange.kt
```diff
@@ -0,0 +1,49 @@
...
+ * This sealed class describe the possible events that can be observed on the [Flow] created by
+ * calling [App.authenticationChangeAsFlow].
+ *
+ * The specific states are represented by these subclasses [LoggedIn], [LoggedOut] and
+ * [Removed].
+ *
+ * Changes can thus be consumed this way:
+ *
+ * ```
+ * app.authenticationChangeAsFlow().asFlow().collect { change: AuthenticationChange ->
```
clementetb, line 15:
```suggestion
 * app.authenticationChangeAsFlow().collect { change: AuthenticationChange ->
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AuthenticationChangeImpl.kt
```diff
@@ -0,0 +1,10 @@
+package io.realm.kotlin.mongodb.internal
```
clementetb, line 1: Missing ©

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
```diff
@@ -101,6 +102,14 @@ public interface App {
      */
     public suspend fun login(credentials: Credentials): User
 
+    /**
+     * Create a [Flow] of [AuthenticationChange]-events to receive notifications of updates to all
+     * app user authentication states, like login and logout.
```
clementetb, line 107:
```suggestion
     * app user authentication states: login, logout and removal.
```

Currently there are only three states, we could mention them all.

clementetb (approved):
Looks great, added some minor comments on documentation. Happy on green on CI.

rorbech (approved):
Maybe a bit pedantic, but I guess all APIs hitting the new  _authentication change updating_-operations (`login`, `logout`, `delete` and `remove`) that in rare circumstances could throw due to buffer overrun needs a `@throws IllegalStateException`-doc entry. 

Also some unaddressed leftover comments from Clemente 

Otherwise LGTM

PR #1356, opened 2023-04-19
Improved subscribe API

cmelchior:
This PR adds a new `subscribe` method making it easier to create subscriptions from queries and query results. 

Since the `subscribe` methods have a high chance of being called from a UI context, they are switching to the dispatcher provided by the `AppConfiguration`. This required reworking the logic around the dispatcher, as previously it was only used by the KtorNetworkTransport. 

Now that it can be used by all Sync APIs that want to delegate things to the background. I didn't want to use the notifier dispatcher since we risked it getting blocked by people trying to update subscriptions.

Due to 2 bugs in Coroutines, I had to bump the minimum supported version to 1.7.0: https://github.com/Kotlin/kotlinx.coroutines/issues/3768


packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmResultsTests.kt
```diff
@@ -68,6 +69,21 @@ class RealmResultsTests {
         assertEquals(1, realm.query<Parent>("name CONTAINS '1'").find().query("name CONTAINS '2'").count().find())
     }
 
+    @Test
+    fun query_returnBackingQuery() {
+        val query = realm.query<Parent>("name CONTAINS '1'")
+        val backingQuery = query.find().query("")
```
rorbech, line 75: Should this really be allowed in the public API?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmResultsExt.kt
```diff
@@ -0,0 +1,104 @@
...
+ * to the data being available on the device due to either latency or because a large dataset needs
+ * be downloaded.
+ *
+ * The default behaviour is that the first time `subscribe` is called, the query result will not
+ * be returned until data has been downloaded from the server. On subsequent calls to `subscribe`
+ * for the same query, the query will run immediately on the local database while any updates
+ * are downloaded in the background.
+ *
+ * @param name name of the subscription. This can be used to identify it later in the [SubscriptionSet].
+ * @param mode type of mode used to resolve the subscription. See [WaitForSync] for more details.
```
rorbech, line 86:
```suggestion
 * @param mode mode used to resolve the subscription. See [WaitForSync] for more details.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmResultsExt.kt
```diff
@@ -0,0 +1,104 @@
...
```
rorbech, line 41:
```suggestion
 * @param mode mode used to resolve the subscription. See [WaitForSync] for more details.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmResultsExt.kt
```diff
@@ -0,0 +1,104 @@
...
+ * This is only relevant for [WaitForSync.ALWAYS] and [WaitForSync.FIRST_TIME].
+ * @return The result of running the query against the local Realm file. The results returned will
+ * depend on which [mode] was used.
+ * @throws kotlinx.coroutines.TimeoutCancellationException if the specified timeout was hit before
+ * a query result could be returned.
+ * @Throws IllegalStateException if this method is called on a Realm that isn't using Flexible Sync.
+ */
+@ExperimentalFlexibleSyncApi
+public suspend fun <T : RealmObject> RealmResults<T>.subscribe(
+    name: String,
```
rorbech, line 97: Would it make sense to collapse the anonymous and named variant by allowing `name: String? = null` to reduce the API surface 🤔

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
```diff
@@ -0,0 +1,343 @@
...
+        realm.query<FlexParentObject>().find().subscribe()
+        assertEquals(SubscriptionSetState.COMPLETE, subs.state)
+        assertEquals(1, subs.size)
+        val sub: Subscription = subs.first()
+        assertNull(sub.name)
+        assertEquals("TRUEPREDICATE ", sub.queryDescription)
+        assertEquals(FlexParentObject::class.simpleName, sub.objectType)
+    }
+
+    // Verify that the realm query doesn't run against a frozen version previous to the Realm
```
rorbech, line 111:
```suggestion
    // Verify that the realm query doesn't run against a frozen version prior to the Realm
```

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
```diff
@@ -0,0 +1,343 @@
...
+        assertEquals(0, realm.query<FlexParentObject>().count().find())
+        // Check that subscribing to a query, will run the query on the data downloaded from
+        // the server and not just local data, due to WaitForSync.FIRST_TIME being the default.
+        val result = realm.query<FlexParentObject>("section = $0", section).subscribe()
+        assertEquals(1, result.size)
+        assertEquals(section, result.first().section)
+    }
+
+    @Test
+    fun realmQuery_subscribe_waitFirstTime() = runBlocking<Unit> {
```
rorbech, line 143: There aren't really any logic testing that it is actually `waitFirstTime`'ing here are there 🤔 It looks like that it is the above test that actually tests that, at least data wise. Are we testing the effect on the subscription set? ... and shouldn't it then show that we are not `COMPLETE` after updating an existing subscription.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
```diff
@@ -0,0 +1,343 @@
...
+        realm.query<FlexParentObject>().subscribe(name = "my-name", mode = WaitForSync.NEVER)
+        updatedSubs = realm.subscriptions
+        assertEquals(2, updatedSubs.size)
+        // Updating the subscription will happen in the background, but
+        // hopefully hasn't reached COMPLETE yet.
+        assertNotEquals(SubscriptionSetState.COMPLETE, updatedSubs.state)
+    }
+
+    @Test
+    fun realmQuery_subscribe_waitAlways() = runBlocking {
```
rorbech, line 185: Wouldn't it be safer to use the pattern from `realmQuery_subscribe_queryResultIsLatestVersion` to verify that you always get a result reflecting writes from another "device"?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
```diff
@@ -0,0 +1,343 @@
...
+        }
+        realm.syncSession.resume()
+        val results2 = realm.query<FlexParentObject>("section = $0", sectionId).subscribe(mode = WaitForSync.ALWAYS)
+        assertEquals(5, results2.size)
+    }
+
+    @Test
+    fun realmQuery_subscribe_timeOut_fails() = runBlocking<Unit> {
+        assertFailsWith<TimeoutCancellationException> {
+            realm.query<FlexParentObject>().subscribe(timeout = 1.nanoseconds)
```
rorbech, line 206: Should we also test the named-case?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
```diff
@@ -0,0 +1,343 @@
...
+        assertEquals(2, subs.size)
+        val lastSub = subs.last()
+        assertEquals("my-name", lastSub.name)
+        assertEquals("section == 42 and name == \"Jane\" ", lastSub.queryDescription)
+    }
+
+    @Test
+    fun realmResults_subscribe_timeOut_fails() = runBlocking<Unit> {
+        assertFailsWith<TimeoutCancellationException> {
+            realm.query<FlexParentObject>().find().subscribe(timeout = 1.nanoseconds)
```
rorbech, line 305: Again, any reason for not also testing the named variant?

rorbech (commented):
Nice. Only essential thing is whether to collapse the anonymous and named variant like we have for `MutableSubscriptionSet.add`? Otherwise just shattered some comments here and there.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmTests.kt
```diff
@@ -320,15 +320,14 @@ class RealmTests {
 
     @Test
     @Suppress("invisible_member")
-    fun writeAfterCloseThrows() = runBlocking {
+    fun writeAfterCloseThrows() = runBlocking<Unit> {
```
clementetb, line 323: We might consider start using runTest, from `kotlinx-coroutines-test`.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/Subscription.kt
```diff
@@ -30,6 +31,11 @@ import kotlin.reflect.KClass
  */
 public interface Subscription {
 
+    /**
+     * A unique id, generated by Realm, that identifies this subscription.
```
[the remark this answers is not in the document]
clementetb: https://www.scribbr.com/frequently-asked-questions/an-or-a-unique/#:~:text=A%20and%20an%20are%20different,instead%20of%20%E2%80%9Can%20unique.%E2%80%9D

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmQueryExt.kt
```diff
@@ -0,0 +1,150 @@
...
+ */
+@ExperimentalFlexibleSyncApi
+public suspend fun <T : RealmObject> RealmQuery<T>.subscribe(
+    mode: WaitForSync = WaitForSync.FIRST_TIME,
+    timeout: Duration = Duration.INFINITE
+): RealmResults<T> {
+    return createSubscriptionFromQuery(this, null, false, mode, timeout)
+}
+
+/**
```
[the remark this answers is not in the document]
cmelchior: Yes, good idea. The problem is that Dokka doesn't allow you to target different overloads of the same method. But it seems to work if I reorganize the file so the named variant is at the top, then I can reference it from other methods further down the file.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmQueryExt.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    if (query !is io.realm.kotlin.internal.query.ObjectQuery<T>) {
+        throw IllegalStateException("Only queries on objects are supported. This was: ${query::class}")
+    }
+    if (query.realmReference.owner !is RealmImpl) {
+        throw IllegalStateException("Calling `subscribe()` inside a write transaction is not allowed.")
+    }
+    val realm: Realm = query.getRealm()
+    val subscriptions = realm.subscriptions
+    val appDispatcher: CoroutineDispatcher = ((realm.configuration as SyncConfiguration).user.app as AppImpl).appNetworkDispatcher.dispatcher
```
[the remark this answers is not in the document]
cmelchior: `realm.subscriptions` will trigger an `IllegalStateException` if you are not using a Flexible Sync Realm, so this cast should be safe.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/MutableSubscriptionSetImpl.kt
```diff
@@ -102,7 +102,8 @@ internal class MutableSubscriptionSetImpl<T : BaseRealm>(
         return result
     }
 
-    override fun removeAll(): Boolean {
+    override fun removeAll(anonymousOnly: Boolean): Boolean {
+        // FIXME removeAll() need to check for anonymousOnly
```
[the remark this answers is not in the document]
cmelchior: I missed a test 🙈

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmResultsTests.kt
cmelchior: I was thinking about that, but ended up on it probably being okay, this is basically similar to query.find().query("TRUEPREDICATE"), except you don't end up with `<originalQuery> AND TRUEPREDICATE`

We could also expose a new method `RealmResults.query` that would return the backing query, but it didn't seem worth it to expand the API surface? But not sure what you think?

I did miss updating the `RealmResults.query` though, this has been fixed.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmResultsExt.kt
cmelchior: I remember why I split it, the reason was two-fold:

1) `name` and `updateExisting` kinda goes hand-in-hand. With all named parameters, you could set `name = null` and `updateExisting = true`, which wouldn't make sense.

2) We assume the primary use case is going to be the anonymous one and setting the mode, with all optional parameters is going to be slightly awkward. I guess we could move `mode/timeout` to the front of the parameters. But that felt a bit "off" to me. 

Not sure if that is argument enough to not collapse them though 🤔 The methods are marked experimental so we can change them later.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
cmelchior: This test calls `subscribe()` with has an implicit value of `waitFirstTime`, so I think the test is correct?

I.e. the subscription set should be COMPLETE with this setting, which is what the test asserts?

I guess it could be more visible if I made the mode explicit, but I wanted to test the default value on purpose.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
cmelchior: Similar to above, we have other tests showing that once a subscription set is COMPLETE, data is there.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
```diff
@@ -0,0 +1,343 @@
...
+                }
+                assertFailsWith<IllegalStateException> {
+                    query<FlexParentObject>().subscribe(name = "my-name")
+                }
+            }
+        }
+    }
+
+    @Test
+    fun realmResults_subscribe_waitFirstTime() = runBlocking {
```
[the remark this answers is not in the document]
cmelchior: See https://github.com/realm/realm-kotlin/pull/1356#discussion_r1229167588

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/WaitForSync.kt
```diff
@@ -0,0 +1,40 @@
+package io.realm.kotlin.mongodb.sync
+
+import io.realm.kotlin.mongodb.ext.subscribe
+import io.realm.kotlin.query.RealmQuery
+import io.realm.kotlin.query.RealmResults
+
+/**
+ * Enum defining the behaviour of when [RealmQuery.subscribe] and [RealmResults.subscribe] will
```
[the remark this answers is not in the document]
cmelchior: They do work in IntelliJ 🙈 , but yes, the Dokka created doesn't. I think it is important to mention that you can call this method on both RealmQuery and RealmSubscribe, and the dokka linking doesn't allow this. 

So instead I added a @see property at the bottom. Hope this is okay?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
rorbech: I see that we verify that it is `COMPLETE` due to `waitFirstTime`, but my point was that we should then also test that if we update the subscription then it is not `COMPLETE` immediately after the update, since it is only waiting the first time. Does that make sense?

cmelchior: Ah, yeah, good point 👍

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
rorbech: I don't get your comment. You are in fact asserting on data here 🤔 

My point was that it seemed safer to rely on writing data to another realm and asserting things after that was downloaded than relying on pause/resume, but 🤷

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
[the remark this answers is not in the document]
rorbech: See https://github.com/realm/realm-kotlin/pull/1356/files#r1229233588 🤪

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
cmelchior: Ah sorry, I was a bit fast there 🙈 ... Hmm, I'm not sure that uploading from another device is safer since we cannot control the timing of it, i.e. the data might already be downloaded as part of subscribing the first time, so we would be checking the wrong things:

```
      val results1 = realm.query<FlexParentObject>("section = $0", sectionId).subscribe(mode = WaitForSync.FIRST_TIME)
       uploadServerData(sectionId, 5)
      val results2 = realm.query<FlexParentObject>("section = $0", sectionId).subscribe(mode =  WaitForSync.ALWAYS)
```

The data uploaded would already be part of the subscription created first, so regardless of what the 2nd call does, the data might already be there. I agree the test is a bit awkward, but I don't see another way of doing it?

rorbech: If you do something like:
```
1. Subscribe to realm2 with some predicate A and check initial data
2. Write data that doesn't match predicate A but match predicate B to realm1
3. Update subscription with `ALWAYS` on realm2 to match predicate B
```
Wouldn't that ensure that data is always available when the subscription i 3. returns?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmResultsTests.kt
rorbech: Hmm. Not feeling strongly against it and I see your point about API surface, but ... Would it be a compromise to do a default `query: String = TRUE_PREDICATE` argument so it matched the `TypedRealm.query` method and then match that internally in the implementation instead of the empty string.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
cmelchior: The problem with 3) is that you cannot update the subscription to a new query, it would be considered "first time" that way.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmResultsTests.kt
cmelchior: Yes, that sounds fair 👍

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/shared/SubscriptionExtensionsTests.kt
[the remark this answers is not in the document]
cmelchior: Fixed

cmelchior:
@clementetb I should have addressed all concerns now, making this ready for round 2 🙏

clementetb (approved):
Looks great, some minor comments.

PR #1359, opened 2023-04-21
Support for bundled realm files with initialRealmFile

### LINKED ISSUE #577: Add support for `RealmConfiguration.assetFile()`
Similar to Realm Java we should support shipping bundled Realm files.

Proposed API:

```kotlin
// Path to file...semantics might differ based on target
RealmConfiguration.Builder.assetFile(path: String)
```

Since bundling files are different across platforms, this might be slightly tricky to do across all platforms, e.g. JAR's don't come with a build-in mechanism.

rorbech:
This PR adds support for initializing a realm file with a prebundled realm. 

```
val config = RealmConfiguration.Builder(....)
    .initialRealmFile("asset.realm", "d4cdef5d8da2b2e2ae692f2e2fead4ef8deeaccb8e4cf0790c2a2ae0dbaabbc4")
    .build()
```

There is currently no common convention for distributing asset and resource files across Kotlin Multiplatform project, so this PR requires the developer to place the `assetFile` according to the target platform conventions. The asset files are located through:
- Android: Located through android.content.res.AssetManager.open(assetFilename)
  - `<sourceset>/assets` 
- JVM: Class<T>.javaClass.classLoader.getResource(assetFilename)
  - `<sourceset>/resources` 
- Darwin: NSBundle.mainBundle.pathForResource(assetFilenameBase, assetFilenameExtension)
  - Add framework files to the main bundle through Xcode
  
Closes #577


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -170,6 +170,9 @@ public interface Configuration {
      */
     public val inMemory: Boolean
 
+    public val assetFile: String?
+    public val assetFileChecksum: String?
```
cmelchior, line 174: We need documentation for these properties

cmelchior: Should we perhaps move them into a dataclass? like `AssetFile(path, sha256checksum)` ? It would keep the number of properties in the configuration down and these are connected anyway? Similar to what we did for the `LogConfiguration`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -112,6 +113,13 @@ public class RealmImpl private constructor(
         var realmFileCreated = false
         try {
             runBlocking {
+                // FIXME We cannot ensure exclusive access to the realm file, so for now multiple
+                //  processes/threads could potentially race each other here
+                //  https://github.com/realm/realm-core/issues/6492
+                if (configuration.assetFile != null && !fileExists(configuration.path)) {
```
cmelchior, line 119: I think this is also racy in a single process. Would it make sense to put this inside a static mutex?

packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -46,6 +51,62 @@ public actual fun prepareRealmFilePath(directoryPath: String, filename: String):
...
+ * - Android: Through android.content.res.AssetManager.open(assetFilename)
+ * - JVM: Class<T>.javaClass.classLoader.getResource(assetFilename)
+ *
+ * @throws Exceptions.assetFileNotFound if the file is not found.
+ */
+public expect fun assetFileAsStream(assetFilename: String): InputStream
+
+@Suppress("NestedBlockDepth")
+public actual fun copyAssetFile(
+    realmFilename: String,
```
cmelchior, line 74: Maybe rename this to `realmFilePath` ?

packages/library-base/src/jvmMain/kotlin/io/realm/kotlin/internal/platform/SystemUtilsJvm.kt
```diff
@@ -15,5 +18,12 @@ public actual val DEVICE_MODEL: String = ""
 @Suppress("FunctionOnlyReturningConstant")
 public actual fun appFilesDirectory(): String = System.getProperty("user.dir") ?: "."
 
+public actual fun assetFileAsStream(assetFilename: String): InputStream {
+    // FIXME How to get hold of the right class loader
```
cmelchior, line 22: I assume we want the classloader associated with the Realm library? So probably `Realm.javaClass.classLoader`?

But I have never had to work that close with class loaders, so not 100% sure. I was going from https://www.baeldung.com/java-classloaders

packages/test-base/build.gradle.kts
```diff
@@ -215,3 +215,22 @@ kotlin {
         val iosTest by getting { dependsOn(nativeDarwinTest) }
     }
 }
+
+// FIXME Add appropriate pattern for getting test files into place for loading through the assetFile
```
cmelchior, line 219: Maybe I'm missing something, but isn't this good enough? JVM/Android have fixed locations in our build system, but ios/macos move around depending on where the apps are placed?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -635,6 +637,118 @@ class RealmTests {
         }
     }
 
+    @Test
+    fun assetFile() {
+        val config = RealmConfiguration.Builder(setOf(Parent::class, Child::class))
+            .directory(tmpDir)
+            .log(LogLevel.DEBUG)
+            // Need a separate name to avoid clashes with this.realm
```
cmelchior, line 645: You mean the Realm created in `setUp()`?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -635,6 +637,118 @@ class RealmTests {
...
+            .name("prefilled.realm")
+            .schemaVersion(1)
```
cmelchior, line 647: Is this needed?

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -635,6 +637,118 @@ class RealmTests {
         }
     }
 
+    @Test
+    fun assetFile() {
+        val config = RealmConfiguration.Builder(setOf(Parent::class, Child::class))
+            .directory(tmpDir)
+            .log(LogLevel.DEBUG)
```
cmelchior, line 644: Remove?

cmelchior (commented):
Preliminary review... but most things look good 👍

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
rorbech: It could make sense to bundle them up to reduce the number of properties, but I assume we should keep the 
```
public fun assetFile(assetFile: String, sha256checkSum: String? = null)
```
as is and not use the `AssetFileConfiguration`-data class as argument!? Then it becomes a bit tricky where to keep the main documentation. Should we use the `assetFile`-method as main entry for explaining the concepts, as this is where users would look in the first place and then just keep the other docs brief and reference this?

cmelchior: Yeah, I would keep this method 👍 And yes, I agree that since this is the main entry point, I would keep the bulk of the docs here and just keep it light on the final Configuration class

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
rorbech: Hmm. I guess we could try something like that and narrow the window for race conditions. But unless it is build into core or we lock the whole `init`-block then there is no way to guard against deleting it between it is copied from the asset and `configuration.openRealm`, in which case core will create a new file from scratch. Since people haven't seen this as an issue in realm-java, I suspect it is fine to just ensure that only one thread will try to copy it into place within a single process as suggested!?

cmelchior: Yes, that was also what I ment. Sorry if that wasn't clear. But right now it looks like there is no locking in place here at all?

cmelchior: Yes, that was also what I meant. Sorry if that wasn't clear. But right now it looks like there is no locking in place here at all?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
rorbech: Bundled the properties up in the `AssetFileConfiguration`, but kept `assetFile` and docs here.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
rorbech: I added a global mutex around copying the asset files in place. We could try to reduce that to only block others from writing the same `configuration.path` but I would assume that it is ok to prevent multiple asset files to be copied at the same time for now, as this only happens on first initialization. My immediate thought is that any improvements should be done in core instead.

packages/test-base/build.gradle.kts
rorbech: Yes, it works. But when ramping on this issue I thought that I could do a nice convention that work across all targets. Providing a better scheme for this should probably just be targeted in a separate issue.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
rorbech: Apparently not. It was somehow related to the way that I constructed the `asset.realm`. I have recreated it and now the schema version is in place without the need for this.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -374,6 +400,48 @@ public interface Configuration {
...
+         * [Realm.open] will fail with an [IllegalArgumentException].
+         * @param sha256checkSum a SHA256-checksum used to verify the integrity of the asset file.
+         * If this is specified and the checksum does not match the computed checksum of the
+         * [assetFile] when the realm is opened the first time [Realm.open] will fail with a
+         * [IllegalArgumentException].
+         *
+         * @throws IllegalArgumentException if called with an empty [assetFile].
+         */
+        public fun assetFile(assetFile: String, sha256checkSum: String? = null): S {
+            require(assetFile.isNotEmpty()) {
```
cmelchior, line 438: Should we also check sha256checksum in a similar way?

cmelchior (commented):
Looking good. I only have a few minor comments about naming and then we need to add a few tests for Sync as well.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -658,6 +659,123 @@ class RealmTests {
         }
     }
 
+    @Test
+    fun assetFile() {
+        val config = RealmConfiguration.Builder(setOf(Parent::class, Child::class))
+            .directory(tmpDir)
+            // Need a separate name to avoid clashes with this.realm initialized in setup()
+            .name("prefilled.realm")
+            .assetFile("asset.realm")
```
clementetb, line 668: I think that it might be worth to have some tests with assets in subdirectories.

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -137,4 +148,50 @@ private fun preparePath(directoryPath: String, dir: NSURL) {
     }
 }
 
+public fun assetFilePath(assetFileName: String): String? {
+    val pathParts: List<String> = assetFileName.split("[.|/]".toRegex())
+    val path: String? = NSBundle.mainBundle.pathForResource(pathParts[0], pathParts[1])
```
clementetb, line 153: Wouldn't it fail with "realm/db.realm"? The split would generate more than 2 parts, and the resource loading would fail to fetch the file 🤔

packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
```diff
@@ -46,6 +51,62 @@ public actual fun prepareRealmFilePath(directoryPath: String, filename: String):
     return File(directoryPath, filename).absolutePath
 }
 
+private fun copyStream(inputStream: InputStream, outputStream: OutputStream) {
+    @Suppress("MagicNumber")
+    val buf = ByteArray(4096)
```
clementetb, line 56: We could make the 4096 a constant FOUR_KB or something similar.

clementetb (approved):
Looks great! I think I spotted an issue when loading assets in native.

packages/library-base/src/nativeDarwin/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
rorbech: Yep. Good catch. I have added some tests that uses subdirs.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
rorbech: Yep. Good catch. I have added some tests that uses subdirs

packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
rorbech: It is only used for this copy and doesn't have connections to other parts, and does not even appear anywhere else in our codebase, so I would deem it ok to just suppress it.

rorbech:
> Looking good. I only have a few minor comments about naming and then we need to add a few tests for Sync as well.

I have added a complete round trip test with PBS to ensure that we are consistent on the order of:
- `initialRealmFile`
- `initialData`, and
- initial download of remote data

It is quite tricky to get right for flexible sync in a repeatable manner, so left that out as the internals for realm initialization is actually not different. Thoughts?

packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/SystemUtils.kt
cmelchior: Maybe rename to `FILE_COPY_BUFFER_SIZE` or something like that, which indicates the behaviour rather than the value

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -89,6 +90,21 @@ public data class LogConfiguration(
     public val loggers: List<RealmLogger>
 )
 
+/**
+ * Configuration for pre-bundled asset files used as initial state of the realm file.
+ */
+public data class InitialRealmFileConfiguration(
+    /**
+     * Path to the assetFile. This will be interpreted differently depending on the platform. See [SharedBuilder.initialRealmFile] for details.
```
cmelchior, line 98:
```suggestion
     * Path to the realm file. This will be interpreted differently depending on the platform. See [SharedBuilder.initialRealmFile] for details.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -172,6 +188,12 @@ public interface Configuration {
      */
     public val inMemory: Boolean
 
+    /**
+     * Configuration that holds details of a bundled asset file used as initial state of the realm
+     * file. See [SharedBuilder.initialRealmFile] for details.
```
cmelchior, line 193:
```suggestion
     * Configuration that holds details of a bundled asset file used as initial state of the realm
     * file. See [SharedBuilder.initialRealmFile] for details. `null` is returned if no initial realm 
     * file has been configured.
```

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmConfigurationTests.kt
```diff
@@ -430,6 +430,53 @@ class RealmConfigurationTests {
         assertFailsWithEncryptionKey(builder, 256)
     }
 
+    @Test
+    fun assetFile_defaultIsNull() {
```
cmelchior, line 434: all of these should probably be renamed to `initialRealmFile`

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/RealmTests.kt
```diff
@@ -658,6 +659,151 @@ class RealmTests {
         }
     }
 
+    @Ignore // Test to generate a realm file to use in initialRealmFile. Copy the generated file to
+    // - test-base/src/androidMain/assets/asset-fs.realm
```
cmelchior, line 663: Shouldn't it be /subdir/asset-fs.realm?

cmelchior (approved):
Nice :shipit: ...only a few minor comments

Issue #1138, opened 2022-11-18
Extend `@PersistedName` annotation to apply to classes


elle-j:
Similar to `@PersistedName` for fields (see #590), we should provide the option of mapping Kotlin class names to their underlying (persisted) name.

```kt
@PersistedName(name = "myPersistedName")
class Example : RealmObject {
    // ...
}
```

Other API's will also be affected by this:

* `RealmClass` should return the persisted name, not the public (Kotlin) name.
* RQL queries should support both the persisted and public names.

TODO:

// Step 1

[ ] Add a `CLASS` annotation target to `PersistedName.kt`.
[ ] Modify `AccessorModifierIrGeneration.modifyPropertiesAndCollectSchema` to account for persisted values (perhaps no changes needed here).
[ ] Modify `RealmModelSyntheticPropertiesGeneration` (see `addSchemaMethodBody()` where `irClass.name.identifier` is used, maybe other places as well).
[ ] Validate that persisted and public names are a unique in the Realm. They are not allowed to overlap. This check should happen at compile time.
[ ] Verify in Realm Studio that schema is modified correctly.

// Step 2

[ ] Make sure that `RealmClass` returns the persisted name, not the public (Kotlin) name in the model class.

// Step 3

[ ] Verify that queries can use both persisted and public names in queries.

lukaspieper:
Hello,

may I ask if you have a schedule for this issue? For properties this has been possible for some time now, but without support for classes the feature doesn't really feel complete yet. This is the last feature I'm missing to complete the migration from realm-java.

Would be very cool if this could complete the existing part in near future.

Thanks a lot.

rorbech:
Hi @lukaspieper. Thanks for the feedback, we are aware that this is a key point for some to migrate. Though it is indeed in our near future roadmap, the priorities and timeline are not fully laid out.

ravitejaavv:
Hello,
Is there any update on this issue?

ravitejaavv:
> Hi @lukaspieper. Thanks for the feedback, we are aware that this is a key point for some to migrate. Though it is indeed in our near future roadmap, the priorities and timeline are not fully laid out.

Its been more than 4 months. When do we expect this feature to be available ?

rorbech:
It is a top priority item, but as always it is hard to predict an exact ETA. Would hopefully arrive within the next weeks

PR #1360, opened 2023-04-21
Add support for @PersistedName on classes

cmelchior:
Closes https://github.com/realm/realm-kotlin/issues/1138

This PR introduces a new level of complexity to opening Realms as we can only detect schema mismatch as part of creating or opening the Realm.

Our current code set up the native RealmSchema when creating the RealmConfiguration, but the schema isn't validated until the Realm is opened. I opted for placing the checking code for this in the same location, but it also means that we can now throw IllegalArgumentException when creating the RealmConfiguration. 

Not 100% sure what the best approach is, so any feedback here would be nice.

### DIFF
--- a/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
+++ b/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
@@ -175,13 +175,13 @@ class PersistedNameTests {
             copyToRealm(PersistedNameSample())
         }
 
-        val dynamicSample = realm.asDynamicRealm().query(PersistedNameSample::class.simpleName!!)
+        val dynamicSample = realm.asDynamicRealm().query("AlternativePersistedNameSample")
             .find()
             .single()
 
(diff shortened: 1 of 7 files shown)


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/PersistedNameTests.kt
```diff
@@ -409,3 +449,20 @@ class PersistedNameChildSample : RealmObject {
...
 }
+
+@PersistedName("PersistedParent")
+class RealmParent(var id: Int) : RealmObject {
+    constructor() : this(0)
+    var child: RealmChild? = null
+}
+
+// Should conflict with RealmParent if included in the same Realm.
+class PersistedParent(var id: Int) : RealmObject {
```
rorbech, line 460: Nitpick ... but I guess we should also have a test of clashes between two classes with `@PersistedName` with the same name.

rorbech (approved):
Awesome. I think it makes good sense to error-out as fast as possible, even though it introduces exception at configuration construction time. Duplicate class names would be a developer error, so no reason for postponing it.

PR #1366, opened 2023-04-28
Reflect background updates in realm without listeners

rorbech:
This PR ensures that we will register a changelistener in the notifier, so that it is up-to-date when the user facing `Realm` is considering whether to advance its own version on an as-needed-basis.


cmelchior (approved):
If CI is happy :shipit:

PR #1373, opened 2023-05-02
Add support for compensating writes

### LINKED ISSUE #1372: Compensating writes
### Problem

Currently, there is no way for the client to know when the server undoes his writes due to any sync issue.

### Solution

Raise an exception when this case occurs so it can be captured.

### Alternatives

_No response_

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

Atlas Device Sync

clementetb:
Adds a new sync exception that allows surfacing any compensating write from the server.

Closes #1372

### DIFF
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/sync/SyncError.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/sync/SyncError.kt
@@ -28,7 +28,7 @@ data class SyncError constructor(
     val isFatal: Boolean,
     val isUnrecognizedByClient: Boolean,
     val isClientResetRequested: Boolean,
-    val compensatingWrites: List<CoreCompensatingWriteInfo>
+    val compensatingWrites: Array<CoreCompensatingWriteInfo>
 ) {
     // Constructor used by JNI so we avoid creating too many objects on the JNI side.
     constructor(
@@ -41,7 +41,7 @@ data class SyncError constructor(
         isFatal: Boolean,
         isUnrecognizedByClient: Boolean,
         isClientResetRequested: Boolean,
-        compensatingWrites: List<CoreCompensatingWriteInfo>
+        compensatingWrites: Array<CoreCompensatingWriteInfo>
     ) : this(
         SyncErrorCode.newInstance(category, value, message),
         detailedMessage,
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/sync/SyncError.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/sync/SyncError.kt
@@ -27,7 +27,8 @@ data class SyncError constructor(
     val recoveryFilePath: String?,
     val isFatal: Boolean,
     val isUnrecognizedByClient: Boolean,
-    val isClientResetRequested: Boolean
+    val isClientResetRequested: Boolean,
+    val compensatingWrites: List<CoreCompensatingWriteInfo>
 ) {
     // Constructor used by JNI so we avoid creating too many objects on the JNI side.
     constructor(
@@ -39,7 +40,8 @@ data class SyncError constructor(
         recoveryFilePath: String?,
         isFatal: Boolean,
         isUnrecognizedByClient: Boolean,
-        isClientResetRequested: Boolean
+        isClientResetRequested: Boolean,
+        compensatingWrites: List<CoreCompensatingWriteInfo>
     ) : this(
         SyncErrorCode.newInstance(category, value, message),
         detailedMessage,
@@ -47,6 +49,7 @@ data class SyncError constructor(
         recoveryFilePath,
         isFatal,
         isUnrecognizedByClient,
-        isClientResetRequested
+        isClientResetRequested,
+        compensatingWrites
     )
 }
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
@@ -68,19 +68,19 @@ public class BadFlexibleSyncQueryException internal constructor(message: String)
     SyncException(message)
 
 /**
- * TODO
+ * TODO documentation
  */
 public class CompensatingWriteException internal constructor(
     message: String,
-    compensatingWrites: List<CoreCompensatingWriteInfo>?
+    compensatingWrites: Array<CoreCompensatingWriteInfo>
 ) : SyncException(message) {
-    public val compensatingWrites: List<CompensatingWriteInfo> = compensatingWrites?.map {
+    public val compensatingWrites: List<CompensatingWriteInfo> = compensatingWrites.map {
         CompensatingWriteInfo(
             reason = it.reason,
             objectName = it.objectName,
             primaryKey = it.primaryKey.asRealmAny(),
         )
-    } ?: emptyList()
+    }
 
     private fun RealmValue.asRealmAny(): RealmAny? = when (val type = getType()) {
         ValueType.RLM_TYPE_NULL -> null
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
@@ -68,7 +68,8 @@ public class BadFlexibleSyncQueryException internal constructor(message: String)
     SyncException(message)
 
 /**
- * TODO documentation
+ * Thrown when the server undoes one or more client writes. Details on undone writes can be found in
+ * [compensatingWrites].
  */
 public class CompensatingWriteException internal constructor(
     message: String,
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
@@ -16,6 +16,15 @@
 
 package io.realm.kotlin.mongodb.exceptions
 
+import io.realm.kotlin.internal.RealmInstantImpl
+import io.realm.kotlin.internal.RealmUUIDImpl
+import io.realm.kotlin.internal.interop.RealmValue
+import io.realm.kotlin.internal.interop.ValueType
+import io.realm.kotlin.internal.interop.sync.CoreCompensatingWriteInfo
+import io.realm.kotlin.internal.realmValueToDecimal128
+import io.realm.kotlin.types.RealmAny
+import org.mongodb.kbson.BsonObjectId
+
 /**
  * This exception is considered the top-level exception or general "catch-all" for problems related
  * to using Device Sync.
@@ -26,9 +35,7 @@ package io.realm.kotlin.mongodb.exceptions
  *
  * @see io.realm.kotlin.mongodb.sync.SyncConfiguration.Builder.errorHandler
  */
-public open class SyncException : AppException {
-    internal constructor(message: String) : super(message)
-}
+public open class SyncException internal constructor(message: String) : AppException(message)
 
 /**
  * Thrown when something has gone wrong with Device Sync in a way that is not recoverable.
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 ### Enhancements
 * Support for initialization of a realm file with a bundled realm through `RealmConfiguration.Builder(...).initialRealmFile(...)` and `SyncConfiguration.Builder(...).initialRealmFile(...)`. (Issue [#577](https://github.com/realm/realm-kotlin/issues/577))
+* [Sync] The new sync exception `CompensatingWriteException` will be thrown when the server undoes one or more client writes. (Issue [#1372](https://github.com/realm/realm-kotlin/issues/1372))
```
```suggestion
* [Sync] The new sync exception `CompensatingWriteException` will be thrown in the `SyncSession.ErrorHandler` when the server undoes one or more client writes. (Issue [#1372](https://github.com/realm/realm-kotlin/issues/1372))
```

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/sync/SyncError.kt
```diff
@@ -27,7 +27,8 @@ data class SyncError constructor(
     val recoveryFilePath: String?,
     val isFatal: Boolean,
     val isUnrecognizedByClient: Boolean,
-    val isClientResetRequested: Boolean
+    val isClientResetRequested: Boolean,
+    val compensatingWrites: Array<CoreCompensatingWriteInfo>
```
cmelchior, line 31: Are we guaranteed that this is only contain information when we should throw a `CompensatingWriteException` ?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,56 @@ public open class SyncException : AppException {
...
  * [io.realm.kotlin.mongodb.sync.SubscriptionSet].
  */
-public class BadFlexibleSyncQueryException : SyncException {
-    internal constructor(message: String) : super(message)
+public class BadFlexibleSyncQueryException internal constructor(message: String) :
+    SyncException(message)
+
+/**
+ * Thrown when the server undoes one or more client writes. Details on undone writes can be found in
+ * [compensatingWrites].
```
cmelchior, line 72: What does this link to?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,56 @@ public open class SyncException : AppException {
...
+        ValueType.RLM_TYPE_BOOL -> RealmAny.create(getBoolean())
+        ValueType.RLM_TYPE_STRING -> RealmAny.create(getString())
+        ValueType.RLM_TYPE_BINARY -> RealmAny.create(getByteArray())
+        ValueType.RLM_TYPE_TIMESTAMP -> RealmAny.create(RealmInstantImpl(getTimestamp()))
+        ValueType.RLM_TYPE_FLOAT -> RealmAny.create(getFloat())
+        ValueType.RLM_TYPE_DOUBLE -> RealmAny.create(getDouble())
+        ValueType.RLM_TYPE_DECIMAL128 -> RealmAny.create(realmValueToDecimal128(this))
+        ValueType.RLM_TYPE_OBJECT_ID -> RealmAny.create(BsonObjectId(getObjectIdBytes()))
+        ValueType.RLM_TYPE_UUID -> RealmAny.create(RealmUUIDImpl(getUUIDBytes()))
+        else -> throw IllegalArgumentException("Unsupported primary key type: ${type.name}")
```
cmelchior, line 98: I would be slightly hesitant about throwing here. We don't have exhaustive tests for this (for obvious reasons), so if we at any point make a mistake here, we will accidentally crash apps. I would probably return something like `RealmAny.create("Unknown")` instead similar to what we do with errors we don't know exactly what is.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,56 @@ public open class SyncException : AppException {
...
+
+/**
+ * Thrown when the server undoes one or more client writes. Details on undone writes can be found in
+ * [compensatingWrites].
+ */
+public class CompensatingWriteException internal constructor(
+    message: String,
+    compensatingWrites: Array<CoreCompensatingWriteInfo>
+) : SyncException(message) {
+    public val compensatingWrites: List<CompensatingWriteInfo> = compensatingWrites.map {
```
cmelchior, line 78: Missing documentation for this property.

cmelchior: Since this property is inside a class already called `CompensatingWrite*`, it seems a bit double to have the property named `compensatingWrites` as well. Not sure if just called it `writes` or `reversedWrites` are better? 🤔

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,56 @@ public open class SyncException : AppException {
...
+        ValueType.RLM_TYPE_TIMESTAMP -> RealmAny.create(RealmInstantImpl(getTimestamp()))
+        ValueType.RLM_TYPE_FLOAT -> RealmAny.create(getFloat())
+        ValueType.RLM_TYPE_DOUBLE -> RealmAny.create(getDouble())
+        ValueType.RLM_TYPE_DECIMAL128 -> RealmAny.create(realmValueToDecimal128(this))
+        ValueType.RLM_TYPE_OBJECT_ID -> RealmAny.create(BsonObjectId(getObjectIdBytes()))
+        ValueType.RLM_TYPE_UUID -> RealmAny.create(RealmUUIDImpl(getUUIDBytes()))
+        else -> throw IllegalArgumentException("Unsupported primary key type: ${type.name}")
+    }
+
+    public inner class CompensatingWriteInfo(
```
cmelchior, line 101: This class and properties are missing documentation

cmelchior (commented):
Nice. I have a few doc comments. Also, it looks like we have no scope for this defined. Do we know how other SDKs are surfacing this. The error handler makes sense, but I think I heard someone talking about maybe making it a specific handler similar to ClientReset. I'm not 100% convinced that is worth it though, but we should probably reach out just to make sure.

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/sync/SyncError.kt
clementetb: I think so, we are only surfacing compensating writes when we have a `CompensatingWriteException`.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
clementetb: It links to the property that contains all compensating write info. It has now been renamed to `writes`

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,73 @@ public open class SyncException : AppException {
...
+     */
+    public val writes: List<CompensatingWriteInfo> = compensatingWrites.map {
+        CompensatingWriteInfo(
+            reason = it.reason,
+            objectName = it.objectName,
+            primaryKey = it.primaryKey.asRealmAny(),
+        )
+    }
+
+    private fun RealmValue.asRealmAny(): RealmAny? = when (getType()) {
```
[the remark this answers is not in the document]
clementetb: It is a slight different piece of code, it is supposed to be only valid for PK and there is no way to actually convert to RealmObjects, we lack of access to the mediator and so.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,73 @@ public open class SyncException : AppException {
...
+/**
+ * Thrown when the server undoes one or more client writes. Details on undone writes can be found in
+ * [writes].
+ */
+public class CompensatingWriteException internal constructor(
+    message: String,
+    compensatingWrites: Array<CoreCompensatingWriteInfo>
+) : SyncException(message) {
+    /**
+     * List of all the writes that has been reversed as part of triggering this exception.
```
cmelchior, line 79:
```suggestion
     * List of all the objects created that has been reversed as part of triggering this exception.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,73 @@ public open class SyncException : AppException {
...
+     * Class that describes the details for a reversed write.
+     */
+    public inner class CompensatingWriteInfo(
+        /**
+         * Reason for the compensating write.
+         */
+        public val reason: String,
+
+        /**
+         * Name of the object for which a write has been reversed.
```
cmelchior, line 114:
```suggestion
         * Name of the object class for which a write has been reversed.
```

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,73 @@ public open class SyncException : AppException {
...
+    public inner class CompensatingWriteInfo(
+        /**
+         * Reason for the compensating write.
+         */
+        public val reason: String,
+
+        /**
+         * Name of the object which write was reversed.
+         */
+        public val objectName: String,
```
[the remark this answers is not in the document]
cmelchior: Agree with Claus. `className` seems more appropriate, or `objectType` which is being used in `Subscription`.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
```diff
@@ -44,22 +51,73 @@ public open class SyncException : AppException {
...
+         */
+        public val reason: String,
+
+        /**
+         * Name of the object for which a write has been reversed.
+         */
+        public val objectName: String,
+
+        /**
+         * Primary key for the object for which a write has been reversed.
```
cmelchior, line 119:
```suggestion
         * Primary key value for the object for which a write has been reversed.
```

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FlexibleSyncIntegrationTests.kt
```diff
@@ -289,4 +296,45 @@ class FlexibleSyncIntegrationTests {
...
+                }
+            }
+            .build()
+
+        runBlocking {
+            val expectedPrimaryKey = BsonObjectId()
+
+            Realm.open(config1).use { realm ->
+                realm.subscriptions.update {
+                    add(realm.query<FlexParentObject>("_id = $0", BsonObjectId()))
```
cmelchior, line 319: I would move this below `expectedPrimaryKey` and assert that the two IDs are different

cmelchior (commented):
Some review comments doesn't seem to be adressed yet. Otherwise looking good.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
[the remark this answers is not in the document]
clementetb: Will use `objectType` as it is used already in subscription.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -69,10 +70,10 @@ internal fun <T, R> channelResultCallback(
 internal fun convertSyncError(error: SyncError): SyncException {
     // TODO In a normal environment we only expose the information in the SyncErrorCode.
     //  In debug mode we could consider using `detailedMessage` instead.
-    return convertSyncErrorCode(error.errorCode)
+    return convertSyncErrorCode(error.errorCode, error)
 }
 
-internal fun convertSyncErrorCode(syncError: SyncErrorCode): SyncException {
+internal fun convertSyncErrorCode(syncError: SyncErrorCode, error: SyncError? = null): SyncException {
```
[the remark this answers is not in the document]
clementetb: I think that I don't follow you, this weirdness in `convertSyncError` is because errors in `SyncSessionTransferCompletionCallback` are `SyncErrorCode` rather than `SyncError`. If we received an actual `SyncError` we would be able to have a common path.

As a solution I have created a wrapper based in `SyncError` to be used in `SyncSessionTransferCompletionCallback`. That way we can create a `SyncError` for a `SyncErrorCode` and have a common error handling path.

🤯

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FlexibleSyncIntegrationTests.kt
clementetb: We cannot move it there because you need at least a subscription to be able to write:

`[RLM_ERR_NO_SUBSCRIPTION_FOR_WRITE]: Cannot write to class FlexParentObject when no flexible sync subscription has been created.`

cmelchior: Sorry, I meant just the `BsonObjectID()` you create here. The test assumes it is different than the first one. It would be nice if that assumption was explicit.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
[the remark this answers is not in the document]
rorbech: But you could drop all this into a `primitiveRealmValueToAny` (or something like that) and rework the `fromRealmValue` of `realmAnyConverter` to call this new method in all other cases that for `RLM_TYPE_LINK` then we wouldn't have duplicate implementations and wouldn't miss to keep `RealmValue`->`RealmAny` conversion consistent if we update it. If that is to cumbersome for this PR (seems like there is also diverging behavior when you cannot convert it), then at least put this conversion method in `Converters.kt` with a descriptive name, so that we wouldn't redo it if we need something similar later on.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -122,14 +118,16 @@ internal fun convertSyncErrorCode(syncError: SyncErrorCode): SyncException {
...
+            when (errorCode.code) {
                 ProtocolSessionErrorCode.RLM_SYNC_ERR_SESSION_BAD_QUERY -> { // Flexible Sync Query was rejected by the server
                     BadFlexibleSyncQueryException(message)
                 }
                 ProtocolSessionErrorCode.RLM_SYNC_ERR_SESSION_PERMISSION_DENIED ->
                     // Permission denied errors should be unrecoverable according to Core, i.e. the
                     // client will disconnect sync and transition to the "inactive" state
                     UnrecoverableSyncException(message)
+                ProtocolSessionErrorCode.RLM_SYNC_ERR_SESSION_COMPENSATING_WRITE ->
+                    CompensatingWriteException(message, syncError!!.compensatingWrites)
```
rorbech, line 130: I guess `syncError` will not always be non-null!?
```suggestion
                    CompensatingWriteException(message, syncError.compensatingWrites)
```

packages/test-sync/src/androidAndroidTest/kotlin/io/realm/kotlin/test/mongodb/shared/FlexibleSyncIntegrationTests.kt
clementetb: I think it is not required. Random `ObjectId` creation has a counter (3 bytes) parameter that is incremented on each new `ObjectId`, so there should not be clashes between two consecutive random `ObjectId`.

cmelchior: Ah yeah, good point

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/exceptions/SyncExceptions.kt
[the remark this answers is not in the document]
rorbech: Looks way better now 🏆

cmelchior (approved):

PR #1376, opened 2023-05-04
Propagate callback user exceptions

### LINKED ISSUE #1228: Propagate exceptions at SDK level through core
Core has introduced `realm_register_user_code_callback_error` that should allow us to propagate exceptions at SDK level through core keeping its stacktrace.

clementetb:
This PR adds support for propagating user exceptions on the following callbacks:

- Migration
- Should compact on launch
- Realm data initialization

Client reset handlers do not support user exception propagation as it is not implemented in `SyncError` yet. https://github.com/realm/realm-core/issues/6569

Closes #1228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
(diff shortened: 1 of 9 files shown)


cmelchior (approved):
Awesome 🏆

cmelchior (approved):
Awesome 🏆

PR #1382, opened 2023-05-08
Compact database when less than 50 pct is used

### LINKED ISSUE #1380: compactOnLaunch does not compact when it should

Sometimes

### Description

I have a database with a very limited amount of used bytes, although the overall file size is well over 50 MB. I expected this database to compacted on launch by using `RealmConfiguration.Builder(....).compactOnLaunch()`, but that didn't happen.

Checking the documentation of `DEFAULT_COMPACT_ON_LAUNCH_CALLBACK`, I see this comment:

> The default implementation for determining if a file should be compacted or not. This implementation will only trigger if the file is above 50 MB and 50% or more of the space can be reclaimed.

But it seems that the current implementation is checking if less than 50% can be reclaimed, so the other way around than described in the comment.

Adding the last assertion to the unit test confirms this:
```
@Test
fun `doCompact test`() = runTest {
    assertEquals(false, LocalStorageInitializer.doCompact(50 * 1024 * 1024, 40 * 1024 *1024)) // not > 50MB and < 50% reclaimable -> should be false
    assertEquals(false, LocalStorageInitializer.doCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024)) // > 50MB and > 50% reclaimable -> should be true
    assertEquals(false, LocalStorageInitializer.doCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 3)) // > 50MB and > 50% reclaimable -> should be true
    assertEquals(true, LocalStorageInitializer.doCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 4)) // > 50MB and >= 50% reclaimable -> should be true
    assertEquals(true, LocalStorageInitializer.doCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 5)) // > 50MB and < 50% reclaimable -> should be false

    // adding the following test, more than 50MB filesize and 25 bytes used, should be compacted
    assertEquals(true, LocalStorageInitializer.doCompact(50 * 1024 * 1024 + 8, 25)) // > 50MB and > 50% reclaimable -> should be true
}
```

### Stacktrace & log output

```shell
N/A
```

Always

### Reproduction Steps

Use the unit test `doCompact test`, that currently exists in the codebase.

### Version

all versions since https://github.com/realm/realm-kotlin/commit/10acd38d4c3a8dae6c3ed9ba27aeb709df4965bd

Local Database only

### Are you using encryption?

No

Android 10+

### Build environment

Android Studio version: 2022.2.1
Android Build Tools version: 33.0.2
Gradle version: 7.6.1

Mardaneus86:
This PR fixes the logic behind the `DEFAULT_COMPACT_ON_LAUNCH_CALLBACK` so it adheres to the existing doc comments.

The database is now compacted when the default `compactOnLaunch()` implementation is used, when it is larger than 50MB and more than 50% can be reclaimed.

Resolves #1380

### DIFF
--- a/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
+++ b/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
@@ -69,10 +69,11 @@ class CompactOnLaunchTests {
     fun defaultCallback_boundaries() {
         val callback = Realm.DEFAULT_COMPACT_ON_LAUNCH_CALLBACK
         assertFalse(callback.shouldCompact(50 * 1024 * 1024, 40 * 1024 * 1024))
-        assertFalse(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024))
-        assertFalse(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 3))
+        assertTrue(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024))
+        assertTrue(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 3))
         assertTrue(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 4))
-        assertTrue(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 5))
+        assertFalse(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 5))
+        assertTrue(callback.shouldCompact(50 * 1024 * 1024 + 8, 25))
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
```diff
@@ -69,10 +69,11 @@ class CompactOnLaunchTests {
     fun defaultCallback_boundaries() {
         val callback = Realm.DEFAULT_COMPACT_ON_LAUNCH_CALLBACK
         assertFalse(callback.shouldCompact(50 * 1024 * 1024, 40 * 1024 * 1024))
-        assertFalse(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024))
-        assertFalse(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024 + 3))
+        assertTrue(callback.shouldCompact(50 * 1024 * 1024 + 8, 25 * 1024 * 1024))
```
clementetb, line 72: I think that this test readability could be improved, and would have help catching the issue earlier:
Add parameters names
Use some kind of constants/functions, like the duration framework

```suggestion
assertTrue(callback.shouldCompact(totalBytes = 50.MB + 1, usedBytes = 25.MB))
```

clementetb (commented):
Great find.

I have created the following PR as a replacement to get CI validation:

https://github.com/realm/realm-kotlin/pull/1384

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
Mardaneus86: Totally agree on the readability of the test. Didn't want to change the test more than strictly necessary. Thanks!

PR #1383, opened 2023-05-09
Update project to use Kotlin 1.8.21

cmelchior:
We decided to postpone moving from `ComponentRegistrar` to `CompilerPluginRegistrar`, since CompilerPluginRegistrar still doesn't support defining the order of plugins. This seems to be planned, however.

This PR also bumps the version of R8 internally. This is required for our minified tests as long as we are not using AGP 7.4.


buildSrc/build.gradle.kts
```diff
@@ -18,8 +18,6 @@
 plugins {
     `kotlin-dsl`
     `kotlin-dsl-precompiled-script-plugins`
-    // Workaround for https://youtrack.jetbrains.com/issue/KT-54634
```
rorbech, line 21: 🎉

CHANGELOG.md
```diff
@@ -14,18 +16,19 @@
 * File format: Generates Realms with file format v23.
 * Realm Studio 13.0.0 or above is required to open Realms created by this version.
 * This release is compatible with the following Kotlin releases:
-  * Kotlin 1.7.20 and above.
+  * Kotlin 1.8.0 and above.
```
rorbech, line 19: Should we spell out that our plugin is not FIR compatible yet ... 🤔

rorbech (approved):
LGTM

CHANGELOG.md
cmelchior: Yes, probably a good idea

PR #1384, opened 2023-05-09
Fix compact on launch to run when less than 50% is used

clementetb:
Replacement for https://github.com/realm/realm-kotlin/pull/1382 to get CI validation.

Closes https://github.com/realm/realm-kotlin/issues/1380

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,34 @@
+## 1.8.1-SNAPSHOT (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
+### Enhancements
+* None.
+
+### Fixed
+* The default compact-on-launch callback trigger 50% or more of the space could be reclaimed was reversed. (Issue [#1380](https://github.com/realm/realm-kotlin/issues/1380))
+
+### Compatibility
+* File format: Generates Realms with file format v23.
+* Realm Studio 13.0.0 or above is required to open Realms created by this version.
+* This release is compatible with the following Kotlin releases:
+  * Kotlin 1.7.20 and above.
+  * Ktor 2.1.2 and above.
+  * Coroutines 1.6.4 and above.
+  * AtomicFu 0.18.3 and above.
+  * The new memory model only. See https://github.com/realm/realm-kotlin#kotlin-memory-model-and-coroutine-compatibility
+* Minimum Gradle version: 6.7.1.
+* Minimum Android Gradle Plugin version: 4.0.0.
+* Minimum Android SDK: 16.
+
+### Internal
+* None.
+
+### Contributors
+*  [Tim Klingeleers](https://github.com/Mardaneus86)
+
+
 ## 1.8.0 (2023-05-01)
 
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 * None.
 
 ### Contributors
-*  [Tim Klingeleers](https://github.com/Mardaneus86)
+*  [Tim Klingeleers](https://github.com/Mardaneus86) for fixing the default `compactOnLaunch` logic. 
 
 
 ## 1.8.0 (2023-05-01)
--- a/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
+++ b/packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
@@ -69,10 +69,11 @@ class CompactOnLaunchTests {
     fun defaultCallback_boundaries() {
         val callback = Realm.DEFAULT_COMPACT_ON_LAUNCH_CALLBACK
         assertFalse(callback.shouldCompact(50 * 1024 * 1024, 40 * 1024 * 1024))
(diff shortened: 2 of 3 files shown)


packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
```diff
@@ -65,14 +64,59 @@ class CompactOnLaunchTests {
...
+                actual = callback.shouldCompact(
+                    totalBytes = fileSize,
+                    usedBytes = (fileSize * usage).toLong()
+                ),
+                message = "fileSize: $fileSize usage: $usage"
+            )
+        }
+
+        // File size <= 50MB, it will never reclaim
+        assert(
```
rorbech, line 88: I see the motivation for doing wrappers like this, but I tend to find it less clear to follow the assertion. Don't know the right balance though 😞

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
```diff
@@ -65,14 +64,59 @@ class CompactOnLaunchTests {
...
+        )
+        assert(
+            shouldCompact = false,
+            fileSize = 50.MB,
+            usage = 0.51
+        )
+
+        // File size > 50MB, only reclaims if usage >= 0.5
+        assert(
+            shouldCompact = true,
```
rorbech, line 106: Is this right? It sounds like it should be false here and true in the last test 🤔 Or am I just also flipping some operators in my head.

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
cmelchior: I kinda this way of writing the assertions, it makes it more easy to see what is being tested. (if you trust that the assertion code works :D)

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
```diff
@@ -65,14 +64,59 @@ class CompactOnLaunchTests {
...
+            fileSize = 50.MB,
+            usage = 0.5
+        )
+        assert(
+            shouldCompact = false,
+            fileSize = 50.MB,
+            usage = 0.51
+        )
+
+        // File size > 50MB, only reclaims if usage >= 0.5
```
cmelchior, line 104:
```suggestion
        // File size > 50MB, only reclaims if usage < 0.5
```

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
cmelchior: I think the comment just above this is wrong. We want to compact if you are using less than half of the space in the file.

CHANGELOG.md
```diff
@@ -1,3 +1,34 @@
...
+  * The new memory model only. See https://github.com/realm/realm-kotlin#kotlin-memory-model-and-coroutine-compatibility
+* Minimum Gradle version: 6.7.1.
+* Minimum Android Gradle Plugin version: 4.0.0.
+* Minimum Android SDK: 16.
+
+### Internal
+* None.
+
+### Contributors
+*  [Tim Klingeleers](https://github.com/Mardaneus86)
```
cmelchior, line 29:
```suggestion
*  [Tim Klingeleers](https://github.com/Mardaneus86) for fixing the default `compactOnLaunch` logic. 
```

cmelchior (approved):

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/CompactOnLaunchTests.kt
clementetb: I think that is more expressive as it uses the concepts of what we are testing, like file usage as a percentage. But it adds another layer of complexity to the test...

In my opinion, it should be ok writing our own asserters if we keep them simple and they facilitate writing and understanding tests. For example, there are asserters for comparing collection contents instead of comparing them manually.

But yes, we should also add some tests for any new asserter 😅

rorbech: It is a great improvement to the old tests. But the improvement lies more or less in the helper `MB` and `B` converters whereas hiding the assertions makes it more tricky to reason about what is actually asserted. I guess it also is just down to the naming. `assert` sounds like it is comparing two values, but it is actually doing all the invocations under the hood. `testShouldCompact` would be more appropriate.

But you could also just do
```
DEFAULT_COMPATCH_ON_LAUNCH_CALLBACK.let {
   assertFalse(shouldCompact(50.MB, 50.MB * 0.5))
   ...
}
```
If the `fileSize` and `usage` named parameters really brings so much clarity, then we should consider adding it to the callback methods in a future API.

rorbech (approved):

Issue #929, opened 2022-07-14
Query with multiple values: `in` operator


destman:
I am trying to build a query that will return objects with given set of values of _id property.
In Java i use  `in` opertaion and it works good.
In Kotlin i try to use this query: 

`realm.query<ModelBookmark>("_id in $0", ids)`

But it crashes with message that type not supported.
Right now i generate query that looks like "_id == "1" or _id == "2" ..." and it works...
Version of realm-kotlin is 1.1.0-SNAPSHOT (in 1.0.1 ByteArray still not works).

rorbech:
Hi @destman, could you supply the details of the crash and the type information of `_id` and `ids`

destman:
_id is a String (@PrimaryKey  var _id: String?)
ids is array of strings (Array<String>)

Crashlog:
`
kotlin.NotImplementedError: An operation is not implemented: Unsupported type for capiRealmValue `Array`
        at io.realm.kotlin.internal.interop.RealmInteropKt.capiRealmValue-28b4FhY(RealmInterop.kt:1518)
        at io.realm.kotlin.internal.interop.RealmInteropKt.access$capiRealmValue-28b4FhY(RealmInterop.kt:1)
        at io.realm.kotlin.internal.interop.MemScope.managedRealmValue-28b4FhY(RealmInterop.kt:1434)
        at io.realm.kotlin.internal.interop.RealmInterop$realm_query_parse$1.invoke(RealmInterop.kt:1051)
        at io.realm.kotlin.internal.interop.RealmInterop$realm_query_parse$1.invoke(RealmInterop.kt:1049)
        at io.realm.kotlin.internal.interop.RealmInteropKt.memScope(RealmInterop.kt:1451)
        at io.realm.kotlin.internal.interop.RealmInteropKt.access$memScope(RealmInterop.kt:1)
        at io.realm.kotlin.internal.interop.RealmInterop.realm_query_parse-6CC_B8E(RealmInterop.kt:1049)
        at io.realm.kotlin.internal.query.ObjectQuery$parseQuery$1.invoke(ObjectQuery.kt:175)
        at io.realm.kotlin.internal.query.ObjectQuery$parseQuery$1.invoke(ObjectQuery.kt:174)
        at io.realm.kotlin.internal.query.ObjectQuery.tryCatchCoreException(ObjectQuery.kt:184)
        at io.realm.kotlin.internal.query.ObjectQuery.parseQuery(ObjectQuery.kt:174)
[...]

PR #1389, opened 2023-05-10
Add support for list query arguments

rorbech:
This adds support for passing any `Iterable` as list arguments to the RQL `IN` operator. 

```
realm.query<QuerySample>("intField IN $0", listOf(1, 3))
realm.query<QuerySample>("intField IN $0", settOf(1, 3))
realm.query<QuerySample>("intField IN $0", 1 .. 3 step 2)
```

**Input for review**
I took some shortcuts for handling allocation and assignments of lists for query arguments. It is not as efficient as it could be, but given that we might want to tweak our scheme for transferring values to the C-API if we can generate the code, I chose to fix support for list arguments and will leave the optimizations for later. The current C-API wrappers does not offer identical was to work with lists of contiguous value types. It didn't find a uniform way to allocate a list of element and then update the elements in place with the current wrapper frameworks. Swig's `array_functions` copies values when accessed by index whereas Kotlin Native's cinterop only allows you to work with the element by reference. As our overall converter infrastructure is currently creating instances and passing it around, it was easier to just `memcpy` the elements into place in Kotlin Native than reworking the whole infrastructure to assign native references on JVM (and that would also not allow us to use the generated Swig array-functions for `valueArray`). 

Closes: #929


packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmValue.kt
```diff
@@ -69,7 +74,4 @@ actual value class RealmValue actual constructor(
     }
 }
 
-actual typealias RealmQueryArgT = realm_query_arg_t
-
-@JvmInline
-actual value class RealmQueryArgsTransport(val value: RealmQueryArgT)
+actual class RealmQueryArgumentList(val size: Long, val value: realm_query_arg_t)
```
cmelchior, line 77: Isn't `value` the wrong word here.  Isn't this more just like `ptr`?

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmValue.kt
```diff
@@ -90,6 +100,4 @@ actual value class RealmValue actual constructor(
     }
 }
 
-actual typealias RealmQueryArgT = realm_query_arg
-
-actual value class RealmQueryArgsTransport(val value: RealmQueryArgT)
+actual class RealmQueryArgumentList(val size: ULong, val value: realm_query_arg)
```
cmelchior, line 103: Same

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/QueryTests.kt
```diff
@@ -2213,6 +2213,34 @@ class QueryTests {
             }
     }
 
+    @Test
+    fun query_inListArgument() {
```
cmelchior, line 2217: I don't think we need to be exhaustive here, but it would be nice with a test for this that takes objects as arguments, just as a smoke test since it involves are few more moving parts.

cmelchior (approved):
Awesome... I can see why this is a bit suboptimal, but given that this code doesn't run in any hot path, it is probably okay 👍

packages/test-base/src/androidAndroidTest/kotlin/io/realm/kotlin/test/shared/QueryTests.kt
cmelchior: Actually we also need to test strings with `'` and `"`

clementetb (approved):

rorbech:
I have added and extended some test cases (list with single element, strings and objects) in https://github.com/realm/realm-kotlin/pull/1389/commits/f82058ed350d7917da4c80f4ab287fed8bff6abd#diff-db6b549cfdb4b7e0512e5d470bdd52902d32f0647ff5d9b8e1a9f72b7b952546 but some fail, so will have to await resolution of core issues.

rorbech:
Still fails for object links. Created https://github.com/realm/realm-core/issues/6688 to track that.

cmelchior:
This is now waiting for https://github.com/realm/realm-core/pull/6705 to be merged.

rorbech:
https://github.com/realm/realm-core/pull/6705 has been merged, so should try to update core. There seem to be a number of API changes, so not just a submodule update though.

Issue #1374, opened 2023-05-02
Full-text Search


cmelchior:
Add support for Full-text search as it is currently implemented in Core.

PR #1394, opened 2023-05-15
Update to Core 13.11.0

cmelchior:
This unblocks #1374 as well as fixes a bug in the PBS -> FLX migration.

This PR currently depends on the Kotlin 1.8 upgrade as there were some issues applying it directly to `main`. The root cause for this was not found, but making this PR depends on Kotlin 1.8 also seemed fine.


cmelchior:
This PR is now ready for review.

clementetb (approved):
🚀

PR #1395, opened 2023-05-17
Add support for building the project locally on Windows.

cmelchior:
This PR adds support for building the project locally on Windows.

I guess it will be a bit hard to verify though if you don't have a Windows machine 😄.

I have tested locally, where building and running tests work fine (on JVM and Android).


clementetb (approved):
LGTM

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -15,6 +15,51 @@
...
  */
 
+
+/**
+ * Enum describing operating systems we can build on.
+ *
+ * We need to track this in order to control which Kotlin Multiplatform tasks we can safely
+ * create on the given Host OS.
+ */
+enum class OperatingSystem {
```
rorbech, line 25: Nice infrastructure. Would be great if we could find a cleaner way (than the current if-else-statements) to add all the targets/tasks based on this infrastructure. But at least building blocks for future clean ups 😅

rorbech (approved):
Nice 🎉  Have tried to trace that all things are properly included for normal (non-Windows 🤪) builds, and seems so. But it is a bit hard to follow, so we should probably test a published snapshot from a Windows machine as soon as we can to catch any missing pieces in good time before next release.

PR #1396, opened 2023-05-22
Disable multiplexing

cmelchior:
See https://github.com/realm/realm-core/issues/6656. Hopefully this should be enough to prevent crashes until a fix can be found.

No changelog as no release has been made with this yet.


rorbech (approved):

clementetb (commented):
LGTM

PR #1399, opened 2023-05-26
Migrate to new Android source sets for tests introduced in Kotlin 1.8

cmelchior:
This PR adds support for the new `androidUnitTest` and `androidInstrumentedTest` source sets.

It also provides a path towards how end users can us JVM code for their own `androidUnitTests` (by using dependency substitution), but finalizing that part is left out of this PR. This PR is only for our own internal benefit.

Some hacks are still required to make this work. I tried to describe them in this code comment:

```
            // Instrumentation tests do not depend on common by default:
            // https://kotlinlang.org/docs/whatsnew18.html#the-relation-between-android-and-common-tests
            // But adding support for this only using `dependsOn(commonTest)` will prevent us
            // from selectively running unit tests on device from the IDE.
            //
            // So in order to work around these limitations the following strategy is used.
            //
            // 1. A symlink between all commonTest files and androidInstrumentedTest is created.
            //    This symlink is called `common` as we need to include both shared tests as well
            //    as some helper classes.
            // 2. We need to duplicate all test dependencies in `androidInstrumentedTest` that is
            //    used in `commonTest`.
            //
            // This approach satisfies both our IDE and CI requirements with minimal changes, with
            // two downsides:
            //
            // 1. We need to duplicate test dependencies (which alsmost never changes)
            // 2. We get an IDE warning about tests in `androidInstrumentedTests` having the wrong
            //    package since the symlink put them in a different location than the package name
            //    would imply (This is acceptable).
            //
            // Improvements to this situation is tracked here:
            // https://youtrack.jetbrains.com/issue/KT-46452/Allow-to-run-common-tests-as-Android-Instrumentation-tests

```

But basically this PR:

- Move all shared tests to `commonTest`
- Removed all current `shared` symlinks from jvm and native targets
- Reintroduced a new `common` symlink in `androidInstrumentedTest`
- Added dependency substitution when selecting "android" in commonTest. This means that while you run the _android_ variant, Realm will be running the JVM library and Cinterop.


packages/test-sync/build.gradle.kts
```diff
@@ -154,19 +167,59 @@ kotlin {
     sourceSets {
         val androidMain by getting {
             dependencies {
-                implementation(kotlin("stdlib"))
+                // implementation(kotlin("stdlib"))
```
clementetb, line 170: Could we remove this?

clementetb (approved):
Looks great! 💎

clementetb (approved):
Looks great! 💎

packages/test-sync/src/androidInstrumentedTest/kotlin/io/realm/kotlin/test/common
```diff
@@ -0,0 +1 @@
+./../../../../../../commonTest/kotlin/io/realm/kotlin/test
```
rorbech, line 1: Why do we both have a `common` link and a `shared` package 🤔

cmelchior: Fixed

rorbech (approved):
Looks way better with the updated directory structure ... so when CI is green 🥳

PR #1400, opened 2023-05-27
Add support for running Android Unit Tests on the JVM

cmelchior:
Automatically apply the dependency substitution used for our own tests in https://github.com/realm/realm-kotlin/pull/1399 to user projects.

We need to do the dependency substitution manually in our tests as we are not applying the plugin there but just calling the compiler plugin.

With the plugin doing this, running Android tests on the JVM should "just work" for users.

I added a step to our CI pipeline that verifies that this behavior works on Android projects. Kotlin Multiplatform just delegate to Android for this, so I don't think we need to also test Multiplatform projects.

### DIFF
--- a/packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmPlugin.kt
+++ b/packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmPlugin.kt
@@ -52,6 +52,8 @@ open class RealmPlugin : Plugin<Project> {
 
         project.configurations.all { conf: Configuration ->
             // Ensure that android unit tests uses the Realm JVM variant rather than Android.
+            // This is a bit britle. See https://github.com/realm/realm-kotlin/issues/1404 for
+            // a potential improvement.
             if (conf.name.endsWith("UnitTestRuntimeClasspath")) {
                 conf.resolutionStrategy.dependencySubstitution { ds: DependencySubstitutions ->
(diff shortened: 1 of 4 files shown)


clementetb (approved):

packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmPlugin.kt
```diff
@@ -47,6 +50,22 @@ open class RealmPlugin : Plugin<Project> {
         ) { /* Do nothing */ }
         getBuildEventsRegistry().onTaskCompletion(serviceProvider)
 
+        project.configurations.all { conf: Configuration ->
+            // Ensure that android unit tests uses the Realm JVM variant rather than Android.
+            if (conf.name.endsWith("UnitTestRuntimeClasspath")) {
```
rorbech, line 55: It would be nice if we could somehow use dependency attributes to only direct the 
```attribute(Attribute.of("org.jetbrains.kotlin.platform.type", String::class.java), "jvm")``` 
dependencies to the `-jvm` modules ... but I haven't been able to find a way to enforce attributes on the target dependencies.

rorbech (approved):
Awesome. 

I guess I share the concern from https://kotlinlang.slack.com/archives/C19FD9681/p1685089661499199 about hardcoding the name of the unit test configuration, but haven't been able to come up with a better solution.

PR #1406, opened 2023-06-02
Move sync tests to new sourceset layout

cmelchior:
In #1399 I only moved the base tests, but forgot the sync tests 🙈 . This PR moves them as well


packages/test-sync/src/jvmTest/kotlin/io/realm/kotlin/test/mongodb/jvm/RealmTests.kt
```diff
@@ -43,7 +43,7 @@ class RealmTests {
         val configuration = SyncConfiguration.create(user, TestHelper.randomPartitionValue(), setOf(ParentPk::class, ChildPk::class))
         Realm.open(configuration).close()
         app.close()
-        var counter = 5 // Wait 5 seconds for threads to settle
+        var counter = 10 // Wait 10 seconds for threads to settle
```
clementetb, line 46: We could use `repeat(10) {delay(1.seconds)}`

clementetb (approved):
Looks great

PR #1407, opened 2023-06-02
Explore API possible for asymmetric sync

cmelchior:
We might want to add Asymmetric Sync sooner, rather than later. This PR explores what it would mean for our type system: https://www.mongodb.com/docs/realm/sdk/dotnet/sync/asymmetric-sync/

Initially, I thought it would be simple, but there are some edge cases around the Deletable interface that makes it a bit tricky. This will need to be discussed.


cmelchior:
Deprecated in favor of #1420

Issue #1401, opened 2023-05-30
Assertion failed: m_method_id != nullptr


DamianJaeger:
#### SDK and version
SDK : Kotlin
Version: 1.8.0

#### Observations
* How frequent do the crash occur? On every app start
* Does it happen in production or during dev/test? dev
* Can the crash be reproduced by you? Yes
* Can you provide instructions for how we can reproduce it? No

#### Crash log / stacktrace
```
11:37:49.472 27037-27179 REALM                          <app-package-name>       E  /Users/realm/workspace-realm-kotlin/releases/packages/cinterop/src/jvm/jni/java_method.cpp:33: [realm-core-13.10.0] Assertion failed: m_method_id != nullptr with (method_name, signature) =  ["asThrowable", "(Lio/realm/kotlin/internal/interop/RealmCoreException;)Ljava/lang/Throwable;"]
                                                                                 <backtrace not supported on this platform>
                                                                                 !!! IMPORTANT: Please report this at https://github.com/realm/realm-core/issues/new/choose
11:37:49.473 27037-27179 libc                           <app-package-name>       A  Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 27179 (SyncThread-<app-name>), pid 27037 (<app-package-name>)
11:37:49.651 27188-27188 DEBUG                          pid-27188             A  Cmdline: <app-package-name>
11:37:49.651 27188-27188 DEBUG                          pid-27188             A  pid: 27037, tid: 27179, name: SyncThread-<app-name>  >>> <app-package-name> <<<
11:37:49.654 27188-27188 DEBUG                          pid-27188             A        realm/realm-core#1 pc 0000000000728305  /data/app/~~1bsexmCci6vx6HbC46nkhA==/<app-package-name>-QuDZ7iD8i2uRoGbwSLuzeQ==/lib/x86_64/librealmc.so (BuildId: d04f2a18ff690e7bec1a86a262aa1081a82ff7c2)
11:37:49.654 27188-27188 DEBUG                          pid-27188             A        realm/realm-core#2 pc 0000000000728660  /data/app/~~1bsexmCci6vx6HbC46nkhA==/<app-package-name>-QuDZ7iD8i2uRoGbwSLuzeQ==/lib/x86_64/librealmc.so (BuildId: d04f2a18ff690e7bec1a86a262aa1081a82ff7c2)
[...]

PR #1410, opened 2023-06-07
Fix async server errors resulting in a native crash.

cmelchior:
Fixes part of #1401 (still investigating the persisted name issue)

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
@@ -224,7 +224,10 @@ internal fun convertAppError(appError: AppError): Throwable {
                         ServiceException(msg)
                     }
                 }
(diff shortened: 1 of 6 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -224,7 +224,13 @@ internal fun convertAppError(appError: AppError): Throwable {
                         ServiceException(msg)
                     }
                 }
-                ErrorCode.RLM_ERR_INVALID_SESSION -> CredentialsCannotBeLinkedException(msg)
+                ErrorCode.RLM_ERR_INVALID_SESSION -> {
+                    if (msg.contains("a user already exists with the specified provider")) {
```
rorbech, line 228: Hmm. This looks a bit shady, but I imagine that you also felt it when writing it, so I assume that there is no way around it 🤔

rorbech (approved):
LGTM. Only thing would be a better differentiation of the exception thrown on `RLM_ERR_INVALID_SESSION`, but if there is no way around it 🤷

clementetb (approved):

PR #1415, opened 2023-06-07
Fix incremental compilation not working across object references.

cmelchior:
This PR fixes the 2nd bug hit in #1401.

It does it by reworking how the `RealmObjectCompanion` work. Instead of putting the name of a Child relationship into the Parent, the Parent will now instead reference the Child KClass and get the name at runtime.

This approach will work even if @PersistedName on Child is changed, which would not otherwise trigger a recompilation of Parent.

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/schema/CompilerPluginBridgeUtils.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/schema/CompilerPluginBridgeUtils.kt
@@ -1,0 +1,55 @@
+package io.realm.kotlin.internal.schema
+
+import io.realm.kotlin.internal.interop.CollectionType
+import io.realm.kotlin.internal.interop.PropertyInfo
+import io.realm.kotlin.internal.interop.PropertyType
+import io.realm.kotlin.internal.realmObjectCompanionOrNull
+import io.realm.kotlin.types.TypedRealmObject
+import kotlin.reflect.KClass
+
+/**
+ * Helper method the compiler plugin can use to create PropertyInfo instances based on KClass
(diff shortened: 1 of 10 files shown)


packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/RealmModelSyntheticPropertiesGeneration.kt
```diff
@@ -133,7 +137,7 @@ class RealmModelSyntheticPropertiesGeneration(private val pluginContext: IrPlugi
     val classInfoCreateMethod = classInfoClass.lookupCompanionDeclaration<IrSimpleFunction>(CLASS_INFO_CREATE)
 
     private val propertyClass = pluginContext.lookupClassOrThrow(PROPERTY_INFO)
-    val propertyCreateMethod = propertyClass.lookupCompanionDeclaration<IrSimpleFunction>(PROPERTY_INFO_CREATE)
+    val propertyCreateMethod = pluginContext.referenceFunctions(PROPERTY_INFO_CREATE).first()
```
rorbech, line 140: Private?

rorbech (approved):
Nice fix 🏅

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/schema/CompilerPluginBridgeUtils.kt
```diff
@@ -0,0 +1,55 @@
+package io.realm.kotlin.internal.schema
```
clementetb, line 1: missing ©

clementetb (approved):
Nice catch 🎣

Issue #1202, opened 2023-01-06
UnsatisfiedLinkError: couldn't find "librealmc.so"


KennethRichards:
**SDK**: 
realm-kotlin

**Version**: 
'io.realm.kotlin:library-sync:1.5.0', 'io.realm.kotlin:gradle-plugin:1.5.0'

**Observations**
- How frequent do the crash occur?
Subset of users

- Does it happen in production or during dev/test?
Production

- Can the crash be reproduced by you?
No. I was unable to reproduce in the Android Emulator on Nexus 7 - Android 8.1
        
- Can you provide instructions for how we can reproduce it?
Seems to be occuring on app launch when calling `App.create()`

### From Firebase Crashlytics:

**Nexus 5X - Android 8.1.0**
```
Fatal Exception: java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.crystalyze.crystalyze-mp0dbfOJe8A8n38Sjr4jkw==/base.apk"],nativeLibraryDirectories=[/data/app/com.crystalyze.crystalyze-mp0dbfOJe8A8n38Sjr4jkw==/lib/x86, /system/lib, /vendor/lib]]] couldn't find "librealmc.so"
       at java.lang.Runtime.loadLibrary0(Runtime.java:1011)
       at java.lang.System.loadLibrary(System.java:1657)
       at io.realm.kotlin.internal.interop.realmc.<clinit>(realmc.java:16)
       at io.realm.kotlin.internal.interop.RealmInterop.realm_network_transport_new(RealmInterop.java:49)
       at io.realm.kotlin.mongodb.internal.AppConfigurationImpl.initializeRealmAppConfig(AppConfigurationImpl.java:49)
       at io.realm.kotlin.mongodb.internal.AppConfigurationImpl.<init>(AppConfigurationImpl.java:49)
       at io.realm.kotlin.mongodb.AppConfiguration$Builder.build(AppConfiguration.java:68)
       at io.realm.kotlin.mongodb.App$Companion.create(App.java:68)
```

**Galaxy Tab A - Android 7.1.2**
```
[...]

PR #1433, opened 2023-06-26
Disable Relinker on API 23 and above

cmelchior:
An attempt at fixing #https://github.com/realm/realm-kotlin/issues/1202,  but since we have never been able to reproduce it it's just a guess. But regardless, Relinker should not be needed at those API levels anyway.

Also bumps to the latest core version and removes the version checks which are just getting in our way for every core release, without providing any real value.


rorbech (approved):
LGTM

PR #1439, opened 2023-06-29
Use async open for FLX Realms

### LINKED ISSUE #1438: FLX enabled Realms should also use async open when waitForInitialRemoteData is used
The current behavior is that only partition-based realms are using async open because we assumed it wasn't helpful for flx-enabled Realms. However, that turned out to be false, because when a connection is established we are uploading the schema to the server instead of waiting for it to be sent. This caused contention on the server as all devices had to read from a single collection at the same time.

The server has implemented mitigation for this, but we should also enable async open to prevent the problem from happening in the first place.

cmelchior:
Closes #1438 

I also discovered that we didn't check whether a realm file existed or not before triggering `waitForInitialRemoteData`, this meant that if `waitForInitialRemoteData` was enabled, the device would always require a network connection before opening the realm file, even after being opened for the first time.

It isn't really possible to write a test for this since it would require us to be able to cut the internet connection between opening the Realm for the first and second time. But I verified the behavior manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,10 @@
 * None.
 
 ### Enhancements
-* [Sync] Optimized the opening of Flexible Sync Realms when  `waitForInitialRemoteData` is used. (Issue [#1438](https://github.com/realm/realm-kotlin/issues/1438))
+* [Sync] Optimized the opening of Flexible Sync Realms when `waitForInitialRemoteData` is used. (Issue [#1438](https://github.com/realm/realm-kotlin/issues/1438))
 
 ### Fixed
-* None
+* [Sync] Using `SyncConfiguration.waitForInitialRemoteData()` would require a network connection, even after opening the realm file for the first time. (Issue [#1439](https://github.com/realm/realm-kotlin/pull/1439)) 
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,32 @@
+## 1.10.1-SNAPSHOT (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
+### Enhancements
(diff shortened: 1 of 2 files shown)


blagoev:
Perhaps 
```
adb shell svc wifi enable/disable
adb shell svc data enable/disable 
```
can be used, but even then a test like this should be run in isolation from the rest.

clementetb (approved):
LGTM.

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
+## 1.10.1-SNAPSHOT (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
+### Enhancements
+* [Sync] Optimized the opening of Flexible Sync Realms when `waitForInitialRemoteData` is used. (Issue [#1438](https://github.com/realm/realm-kotlin/issues/1438))
```
rorbech, line 7: Do we need to put in a warning about that it is now required to be online when opening a FLX Realm the first time. Or maybe there would be some other condition/side effect of subscriptions, etc. that already essentially caused same constraint?

rorbech (approved):
Seems good to me, but can't really grasp if we changed the behavior if offline when opening the realm for the first time. If so we should maybe mention that.

CHANGELOG.md
cmelchior: The new behavior still depends on `waitForInitialRemoteData`, since we are also writing the subscriptions and making sure they were valid before continuing, the end user will still see the same timeout if disconnected, it will just happen earlier in our internal code now.

rorbech: Ahh, yes. Seems to be in `initializeRealmData`.

cmelchior: Yes, I agree the flow is a bit confusing. It is an unfortunate side-effect of combining Sync And Local Realms in the same bootstrap flow + taking into account the various bootstrap options for both.

PR #1452, opened 2023-07-17
Fix RealmInstant accepting invalid constructor parameters

clementetb:
RealmInstant requires that both constructor's parameters be positive or negative, but never a mix of them.


packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmInstantTests.kt
```diff
@@ -207,6 +200,17 @@ class RealmInstantTests {
         }
     }
 
+    @Test
+    fun mismatchingSign_throws() {
```
rorbech, line 204: @clementetb Did you actually try to see if enabling [`REALM_ENABLE_ASSERTIONS` ](https://github.com/realm/realm-core/issues/6762#issuecomment-1624347161) would catch the same error?

clementetb: No, I didn't. Enabling core assertions won't catch the issue, because Core is not involved in `RealmInstant` object creation. Assertions would only fail when writing values in the DB.

PR #1453, opened 2023-07-17
Add Proxy support with custom request headers and custom authorization name

clementetb:
Adds support for custom request headers and custom authorization name, required for proxy setups.

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
@@ -61,15 +61,15 @@ public interface AppConfiguration {
     public val syncRootDirectory: String
 
     /**
+     * Authorization header name used for Atlas App services requests.
+     */
+    public val authorizationHeaderName: String
+
+    /**
      * Custom configured headers that will be sent alongside other headers when
      * making network requests towards Atlas App services.
      */
     public val customRequestHeaders: Map<String, String>
-
-    /**
-     * Authorization header name used for Atlas App services requests.
-     */
-    public val authorizationHeaderName: String
 
     /**
      * The name of app. This is only used for debugging.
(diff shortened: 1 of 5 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
```diff
@@ -119,7 +132,7 @@ public interface AppConfiguration {
      * @param appId the application id of the App Services Application.
      */
     public class Builder(
-        private val appId: String
+        private val appId: String,
```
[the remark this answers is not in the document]
clementetb: I have enabled trailing commas in the IDE so it will automatically include them 👯

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
```diff
@@ -268,6 +287,37 @@ public interface AppConfiguration {
...
+        public fun authorizationHeaderName(name: String): Builder = apply {
+            authorizationHeaderName = name
+        }
+        /**
+         * Adds an extra HTTP header to append to every request to an Atlas App Services Application.
+         *
+         * @param name the name of the header.
+         * @param value the value of header.
+         */
+        public fun addCustomRequestHeader(name: String, value: String): Builder = apply {
```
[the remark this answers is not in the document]
clementetb: I am merging both custom request headers methods into a single one that would expose the map via a lambda. That way all map methods will be available.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
```diff
@@ -268,6 +286,32 @@ public interface AppConfiguration {
...
+            authorizationHeaderName = name
+        }
+
+        /**
+         * Update the custom headers that would be appended to every request to an Atlas App Services Application.
+         *
+         * @param block lambda with the the custom header map update instructions.
+         */
+        public fun customRequestHeaders(
+            block: MutableMap<String, String>.() -> Unit,
```
cmelchior, line 310: This is actually a pretty neat way of doing this 👍

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
```diff
@@ -268,6 +286,32 @@ public interface AppConfiguration {
...
+        ): Builder = apply {
+            customRequestHeaders.block()
```
cmelchior, line 312: Should we also check if the header name is valid or non-empty? At least we threw IllegalArgumentException in the Java implementation for that input.

cmelchior (approved):
Looking good. Only have a minor comment about checking empty header fields

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/AppConfiguration.kt
clementetb: yes, we should check for empty contents.

cmelchior (approved):

PR #1454, opened 2023-07-17
Add implementation for equals, hascode, and toString to unmanaged RealmList and RealmSet

clementetb:
Adds equals, hashcode, and toString to unmanaged lists and sets. It would default to the MutableList and MutableSet behaviors.

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
@@ -46,13 +46,21 @@ import kotlin.reflect.KClass
 /**
  * Implementation for unmanaged lists, backed by a [MutableList].
  */
-internal class UnmanagedRealmList<E> : RealmList<E>, InternalDeleteable, MutableList<E> by mutableListOf() {
+internal class UnmanagedRealmList<E>(
+    private val backingList: MutableList<E> = mutableListOf()
+) : RealmList<E>, InternalDeleteable, MutableList<E> by backingList {
     override fun asFlow(): Flow<ListChange<E>> =
         throw UnsupportedOperationException("Unmanaged lists cannot be observed.")
 
     override fun delete() {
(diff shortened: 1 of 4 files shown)


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
```diff
@@ -46,13 +46,21 @@ import kotlin.reflect.KClass
 /**
  * Implementation for unmanaged lists, backed by a [MutableList].
  */
-internal class UnmanagedRealmList<E> : RealmList<E>, InternalDeleteable, MutableList<E> by mutableListOf() {
+internal class UnmanagedRealmList<E>(
+    private val backingList: MutableList<E> = mutableListOf()
```
rorbech, line 50: Any specific reason for keeping a separate reference to this? You could also just implement `equals` and `hashCode` by delegating to `MutableList` with 

```
override fun equals(other: Any?): Boolean = super<MutableList>.equals(other)
override fun hashCode(): Int = super<MutableList>.hashCode()
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmSetInternal.kt
```diff
@@ -46,14 +46,22 @@ import kotlin.reflect.KClass
 /**
  * Implementation for unmanaged sets, backed by a [MutableSet].
  */
-internal class UnmanagedRealmSet<E> : RealmSet<E>, InternalDeleteable, MutableSet<E> by mutableSetOf() {
+internal class UnmanagedRealmSet<E>(
+    private val backingSet: MutableSet<E> = mutableSetOf()
```
rorbech, line 50: Same.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
clementetb: Because I didn't know that that expression was possible.

PR #1459, opened 2023-07-20
Fix `RealmLog` does not log sync events

clementetb:
Sync events are not logged when `RealmLog.level` is set. Any level previously defined is ignored, except if it is defined via the deprecated `AppConfiguration.Builder.log`.

The reason the level is ignored is because whenever an App or Realm configuration is set the log level is set to a default level of `WARN`.

This PR fixes the issue by setting the default level to the `RealmLog.level`, this way, if the user does not define any level via the deprecated method, we will set the level to whatever was already defined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 ### Fixed
 * `RealmInstant` could be instantiated with invalid arguments. (Issue [#1443](https://github.com/realm/realm-kotlin/issues/1443))
 * `equals` and `hashCode` on unmanaged `RealmList` and `RealmSet` resulted in incorrect values. (Issue [#1454](https://github.com/realm/realm-kotlin/pull/1454))
-* [Sync] HTTP requests were not logged when the log level was set in `RealmLog.level`. (Issue [#1456](https://github.com/realm/realm-kotlin/pull/1456)) 
+* [Sync] HTTP requests were not logged when the log level was set in `RealmLog.level`. (Issue [#1456](https://github.com/realm/realm-kotlin/pull/1456))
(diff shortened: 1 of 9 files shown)


nirinchev:
I don't have any idea how the Kotlin SDK opens apps or defines the logger, but if I understand the issue description correctly, then it must be doing it the wrong way. In the implementation of [`SyncManager::do_get_logger`](https://github.com/realm/realm-core/blob/37478156fa61e60f93e29309ad706a010de71906/src/realm/object-store/sync/sync_manager.cpp#L307-L312) we can see that the app-supplied `log_level` is only used if the the log factory on the app configuration is set. If the user didn't specify custom logger/loglevel through the deprecated methods, then the default logger should be used which should ignore any `log_level` set on the app config.

If the Kotlin SDK does set a logger factory on the app config, even if the user doesn't use the deprecated methods, this is a mistake as it means the app logger will ignore changes to the `RealmLog.level` (e.g. you create the app with `RealmLog.level` set to `info`, but then change it to `detail`).

clementetb:
As discussed, this behavior should be possible (but it is not):
```
RealmLog.level = info;
app = App.create(AppConfig(...));
user = app.login(...);
realm = Realm(FlexibleSyncConfig(user));

// see only info messages

RealmLog.level = debug;

// expect to see debug messages
```

nirinchev:
I wrote a .NET test to verify the Core behavior is correct and it seems to be. This is the PR: https://github.com/realm/realm-dotnet/pull/3389 and I can see that once I change the log level, I see messages at the new level.

clementetb:
I confirm that it has the right behavior in Kotlin too, I was testing with an app without the latest local snapshot 🤦

PR #1462, opened 2023-07-25
Add tests for persisted name links in lists

rorbech:


clementetb:
> It works well for changing the name of an entity but only when this entity is not part of a RealmList.

I interpreted this like:

```kotlin
class Parent {
    var children: RealmList<Child> = realmListOf()
}

@PersistedName("RenamedChild")
class Child {
    var name: String = ""
}
```

rorbech:
But that is also part of the test. In my test I am just also renaming the `children` property, but should still show that the parent can have a list of `Child`ren even though `Child` is renamed.

clementetb:
Arg, I only read the incoming changes 🤦🏻‍♂️

clementetb (approved):

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/PersistedNameTests.kt
```diff
@@ -386,15 +386,17 @@ class PersistedNameTests {
...
+                        this.children.add(Child().apply { children.add(Child()) })
                     }
                 )
             }
 
             assertEquals(1, realm.query<Parent>().count().find())
-            assertEquals(1, realm.query<Child>().count().find())
+            assertEquals(3, realm.query<Child>().count().find())
+
+            assertEquals("child1", realm.query<Parent>().first().find()!!.child!!.name)
```
cmelchior, line 398: We should probably also check that you can access the list of children

cmelchior (approved):
Only a minor comment about the test and CI needs to be green 🚀

PR #1463, opened 2023-07-25
Fix realm scheduler leak

### LINKED ISSUE #1445: JNI ERROR (app bug): global reference table overflow (max=51200)

Always

### Description

Hi there!

We have an app that is supposed to run continously on a wall-mounted terminal as a way for employees to check in and out at the start and end of their work day. It also does some periodic work in the background to refresh data that might have been updated.
We recently switched from realm-java to realm-kotlin and started experiencing random crashes, once every couple of days. According to the crash report, the global reference table overflowed.
It seems that the global reference table is filled with thousands of unique instances of io.realm.kotlin.internal.interop.JVMScheduler.

### Stacktrace & log output

```shell
F libc    : Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 19354 (...), pid 929 (com.company.terminal.app)
F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
F DEBUG   : Build fingerprint: 'Zebra/CC610/CC610LC:11/11-20-18.00-RG-U00-STD-HEL-04/82:user/release-keys'
F DEBUG   : Revision: '0'
F DEBUG   : ABI: 'arm64'
F DEBUG   : Timestamp: 2023-06-15 10:59:48+0200
F DEBUG   : pid: 929, tid: 19354, name: ...  >>> com.company.terminal.app <<<
F DEBUG   : uid: 10278
F DEBUG   : signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
F DEBUG   : Abort message: 'JNI ERROR (app bug): global reference table overflow (max=51200)global reference table dump:
F DEBUG   :   Last 10 entries (of 51200):
F DEBUG   :     51199: 0x19e39e00 io.realm.kotlin.internal.interop.RealmInterop$$ExternalSyntheticLambda0
F DEBUG   :     51198: 0x150b64e8 io.realm.kotlin.internal.ConfigurationImpl$$ExternalSyntheticLambda0
F DEBUG   :     51197: 0x19f32350 io.realm.kotlin.internal.interop.JVMScheduler
F DEBUG   :     51196: 0x19f32328 io.realm.kotlin.internal.interop.RealmInterop$$ExternalSyntheticLambda0
F DEBUG   :     51195: 0x150b64e8 io.realm.kotlin.internal.ConfigurationImpl$$ExternalSyntheticLambda0
F DEBUG   :     51194: 0x19e30940 io.realm.kotlin.internal.interop.RealmInterop$$ExternalSyntheticLambda0
F DEBUG   :     51193: 0x150b64e8 io.realm.kotlin.internal.ConfigurationImpl$$ExternalSyntheticLambda0
F DEBUG   :     51192: 0x19f2b4d8 io.realm.kotlin.internal.interop.JVMScheduler
F DEBUG   :     51191: 0x19f2b4b0 io.realm.kotlin.internal.interop.RealmInterop$$ExternalSyntheticLambda0
F DEBUG   :     51190: 0x150b64e8 io.realm.kotlin.internal.ConfigurationImpl$$ExternalSyntheticLambda0
F DEBUG   :   Summary:
F DEBUG   :     49810 of io.realm.kotlin.internal.interop.JVMScheduler (49810 unique instances)
F DEBUG   :       357 of java.lang.Class (272 unique instances)
F DEBUG   :       323 of io.realm.kotlin.internal.ConfigurationImpl$$ExternalSyntheticLambda0 (12 unique instances)
F DEBUG   :       323 of io.realm.kotlin.internal.interop.RealmInterop$$ExternalSyntheticLambda0 (323 unique instances)
F DEBUG   :       301 of java.nio.DirectByteBuffer (301 unique instances)
F DEBUG   :        21 of android.os.Binder (21 unique instances)
F DEBUG   :        20 of android.os.Handler$MessengerImpl (20 unique instances)
F DEBUG   :         5 of android.view.ViewRootImpl$W (5 unique instances)
F DEBUG   :         4 of java.lang.ref.WeakReference (4 unique instances)
F DEBUG   :         4 of android.content.ContentProvider$Transport (4 unique instances)
F DEBUG   :         3 of android.opengl.EGLContext (2 unique instances)
F DEBUG   :         3 of android.opengl.EGLDisplay (2 unique instances)
F DEBUG   :         3 of android.opengl.EGLSurface (2 unique instances)
F DEBUG   :         2 of dalvik.system.PathClassLoader (1 unique instances)
F DEBUG   :         2 of java.lang.String (2 unique instances)
F DEBUG   :         2 of java.lang.ThreadGroup (2 unique instances)
F DEBUG   :         1 of com.qualcomm.qti.Performance$PerfServiceDeathRecipient
F DEBUG   :         1 of java.nio.ByteOrder
F DEBUG   :         1 of dalvik.system.VMRuntime
F DEBUG   :         1 of android.app.ActivityThread$ApplicationThread
F DEBUG   :         1 of android.app.LoadedApk$ReceiverDispatcher$InnerReceiver
F DEBUG   :         1 of android.app.job.JobServiceEngine$JobInterface
F DEBUG   :         1 of android.database.ContentObserver$Transport
F DEBUG   :         1 of android.graphics.HardwareRenderer$ProcessInitializer$1
F DEBUG   :         1 of android.hardware.display.DisplayManagerGlobal$DisplayManagerCallback
F DEBUG   :         1 of android.hardware.input.InputManager$InputDevicesChangedListener
F DEBUG   :         1 of android.os.PersistableBundle$1
F DEBUG   :         1 of android.telephony.TelephonyManager$DeathRecipient
F DEBUG   :         1 of android.view.WindowManagerGlobal$1
F DEBUG   :         1 of android.view.accessibility.AccessibilityManager$1
F DEBUG   :         1 of android.view.inputmethod.InputMethodManager$1
F DEBUG   :         1 of android.view.inputmethod.InputMethodManager$ControlledInputConnectionWrapper
F DEBUG   :         1 of android.net.TetheringManager$TetheringCallbackInternal
```

Always

### Reproduction Steps

After some testing we found that opening Realm to read an entry creates those JVMScheduler instances, but closing the realm doesn't clean them up.

The problem can be reproduced using this sample code:

```kotlin
class MyRealmClass : RealmObject {
    @PrimaryKey
    var id = ObjectId()
    var field: String? = null
}

@Test
fun test() = runBlocking {
    val config = RealmConfiguration.create(schema = setOf(MyRealmClass::class))

    with (Realm.open(config)) {
        write { MyRealmClass().also { it.field = "Hello World" } }
        close()
    }
    CoroutineScope(EmptyCoroutineContext).launch {
        while (true) {
            delay(60000)
            dumpReferenceTables()
        }
    }
    while (true) {
        delay(10)
        with(Realm.open(config)) {
            query<MyRealmClass>().first().find()?.field
            close()
        }
    }
}

private fun dumpReferenceTables() =
        android.os.Debug::class.functions
            .firstOrNull { it.name == "dumpReferenceTables" }
            ?.apply { isAccessible = true }
            ?.call()
```

The table dump should show that every minute several 100 new instances of the JVMScheduler are added.

### Version

1.10.1

Local Database only

### Are you using encryption?

No

Android 8-13. Other versions not tested.

### Build environment

Android Build Tools version: 7.4.2
Gradle version: 7.5

clementetb:
Closes #1445

It has been reported that we are leaking Realm scheduler instances. It can be reproduced by opening and closing a Realm in a loop:

```kotlin
    @Test
    fun test() = runBlocking {
        val config = RealmConfiguration.create(schema = emptySet())

        while (true) {
            Realm.open(config).close()
        }
    }
```

JVM and Android would fail with `JNI ERROR (app bug): global reference table overflow`


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/util/CoroutineDispatcherFactory.kt
```diff
@@ -81,24 +83,31 @@ public sealed interface DispatcherHolder {
      */
     public val dispatcher: CoroutineDispatcher
 
+    /**
+     * Realm scheduler, we need to hold a reference to prevent being GC'ed
+     */
+    public val realmScheduler: RealmSchedulerPointer
```
rorbech, line 89: I think that we should rename the `DispatchHolder` to something that also indicates that it holds on to the scheduler. `RealmExecutionContext` or something in that direction 🤪  #naming

rorbech (commented):
The infrastructural changes wrapping the scheduler to a Kotlin object looks great. But I don't feel like the actual scheduler should be hooked up on the `DispatchHolder`, because that is actually just that - a dispatch holder - and is also used for the network transport that doesn't require a scheduler. We should probably have a SDK-layer scheduler object that holds a reference to the dispatch holder and the native pointer of the associated native scheduler.

I also think that we should just also explicitly release the native pointer to the scheduler when we close the realm. There should already be implementation in place to guard for releasing it twice.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -516,21 +515,75 @@ actual object RealmInterop {
...
 
+    actual fun realm_create_scheduler(): RealmSchedulerPointer {
+        // If there is no notification dispatcher use the default scheduler.
+        // Re-verify if this is actually needed when notification scheduler is fully in place.
+        val scheduler = checkedPointerResult(realm_wrapper.realm_scheduler_make_default())
+        return CPointerWrapper<RealmSchedulerT>(scheduler)
+    }
+
+    actual fun realm_create_scheduler(dispatcher: CoroutineDispatcher): RealmSchedulerPointer {
+        printlntid("createSingleThreadDispatcherScheduler")
```
cmelchior, line 534: Debug output that should be removed? Same for a lot of other places

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -180,22 +180,26 @@ actual object RealmInterop {
...
 
-        // create a custom Scheduler for JVM if a Coroutine Dispatcher is provided other wise
-        // pass null to use the generic one
-        val realmPtr = LongPointerWrapper<LiveRealmT>(
-            realmc.open_realm_with_scheduler(
-                (config as LongPointerWrapper).ptr,
-                if (dispatcher != null) JVMScheduler(dispatcher) else null
-            )
-        )
+        realmc.realm_config_set_scheduler(config.cptr(), scheduler.cptr())
```
cmelchior, line 200: Does Core automatically the scheduler when passed in this way?

cmelchior (commented):
Minor comments. Otherwise agree with Claus, using the name "DispatcherHolder" doesn't seem right given the expanded responsibility, and we should probably manually close scheduler. We have seen "leaks" in other places because the GC might be slow to run. Especially on the JVM.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
rorbech: This is it a no-op lambda unless redefined. It was left this way as an easy way to reenable debugging, as it is quite tricky to otherwise get an overview of all involved scheduling-related operations.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -180,22 +180,26 @@ actual object RealmInterop {
         realmc.realm_config_set_in_memory(config.cptr(), inMemory)
     }
 
-    actual fun realm_open(config: RealmConfigurationPointer, dispatcher: CoroutineDispatcher?): Pair<LiveRealmPointer, Boolean> {
+    actual fun realm_create_scheduler(): RealmSchedulerPointer =
+        LongPointerWrapper(realmc.realm_scheduler_make_default(), false)
```
clementetb, line 184: Making it managed, we will release it manually when closing realms.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Realm.kt
```diff
@@ -118,9 +118,11 @@ public interface Realm : TypedRealm {
             }
             if (!fileExists(configuration.path)) return false
             val config = (configuration as InternalConfiguration)
-            val (dbPointer, _) = RealmInterop.realm_open(config.createNativeConfiguration())
+            val scheduler = RealmInterop.realm_create_scheduler()
+            val (dbPointer, _) = RealmInterop.realm_open(config.createNativeConfiguration(), scheduler)
             return RealmInterop.realm_compact(dbPointer).also {
                 RealmInterop.realm_close(dbPointer)
+                scheduler.release()
```
cmelchior, line 125: We will not correctly release the scheduler if `realm_open()` throws an exception (which there is a high chance it will do). There is also a risk if `realm_close()` fails, but I'm not sure that can happen without something really weird being wrong.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/ConfigurationImpl.kt
```diff
@@ -107,10 +107,12 @@ public open class ConfigurationImpl constructor(
 
     override suspend fun openRealm(realm: RealmImpl): Pair<FrozenRealmReference, Boolean> {
         val configPtr = realm.configuration.createNativeConfiguration()
-        val (dbPointer, fileCreated) = RealmInterop.realm_open(configPtr)
+        val scheduler = RealmInterop.realm_create_scheduler()
+        val (dbPointer, fileCreated) = RealmInterop.realm_open(configPtr, scheduler)
         val liveRealmReference = LiveRealmReference(realm, dbPointer)
         val frozenReference = liveRealmReference.snapshot(realm)
         liveRealmReference.close()
+        scheduler.release()
```
cmelchior, line 115: We will leak this if `realm_open` throws. Also, how do we feel about having to close this manually and not as part of `Realm.close()`, but unfortunately when setting up the LiveRealmReference this way, it doesn't seem like that is possible.

cmelchior (commented):
Nice refactor, untangling the scheduler from the dispatcher holder makes sense.  However now it looks like we might leak it if there are crashes.

It would be nice if we could somehow tie the `LiveRealmPointer` and the `RealmSchedulerPointer` lifecycle together, but I don't have a great idea on how to do that will not make all usages of the LiveRealmPointer rather annoying to use.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
rorbech: But why is it manually managed? Seems way better to make it managed by the GC as we can then either `close`/`release` it explicitly or the GC will do it if we forget/can't. Currently this will not be release in the `MemoryTests.garbageCollectorShouldFreeNativeResources` where we don't do an explicit `Realm.close`. (Or maybe it will now with this PR? ... This was where the whole PR started 🙈)

rorbech (commented):
Think we finally got the dispatcher/scheduler connection right. Sorry for the confusion. 

But I don't get why the scheduler cannot be GC-managed ? 🤔

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -2093,13 +2097,11 @@ private class JVMScheduler(dispatcher: CoroutineDispatcher) {
     val scope: CoroutineScope = CoroutineScope(dispatcher)
 
     fun notifyCore(schedulerPointer: Long) {
-        val function: suspend CoroutineScope.() -> Unit = {
+        scope.launch(
+            context = scope.coroutineContext, // FIXME REQUIRED? this is something additional
```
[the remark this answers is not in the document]
clementetb: To test the behavior without these parameters. The original code had them and it might be redundant.

packages/cinterop/src/nativeDarwinTest/kotlin/io/realm/kotlin/test/CinteropTest.kt
```diff
@@ -178,9 +179,12 @@ class CinteropTest {
                 SchemaMode.RLM_SCHEMA_MODE_AUTOMATIC
             )
             RealmInterop.realm_config_set_schema_version(nativeConfig, 1)
-
-            val (realm, fileCreated) = RealmInterop.realm_open(nativeConfig)
-            assertEquals(1L, RealmInterop.realm_get_num_classes(realm))
+            RealmInterop.realm_create_scheduler()
+                .releaseAfter { scheduler ->
+                    val (realm, fileCreated) = RealmInterop.realm_open(nativeConfig, scheduler)
```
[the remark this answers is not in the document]
clementetb: I don't know if it has some use. If it is not, it might be required in the future.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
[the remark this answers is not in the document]
rorbech: They are just not in the original part of the diff when I am reviewing 👀

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Realm.kt
```diff
@@ -118,10 +119,17 @@ public interface Realm : TypedRealm {
...
-                RealmInterop.realm_close(dbPointer)
-            }
+
+            return RealmInterop.realm_create_scheduler()
+                .releaseAfter { scheduler ->
+                    val (dbPointer, _) = RealmInterop.realm_open(
+                        config = config.createNativeConfiguration(),
+                        scheduler = scheduler
+                    )
+                    RealmInterop.realm_compact(dbPointer).also {
```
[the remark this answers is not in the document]
clementetb: I am leaving as it it. It would be great to have a scoped-like `releaseAfter`, but the issue as you point out, is that the Realm pointer needs to be closed and then released.

It is easier to add a try-finally for this specific instance. We could consider it later if we have more occurrences for this pattern.

packages/test-base/src/commonMain/kotlin/io/realm/kotlin/test/StandaloneDynamicMutableRealm.kt
```diff
@@ -29,13 +30,17 @@ import io.realm.kotlin.internal.interop.RealmInterop
...
  */
-internal class StandaloneDynamicMutableRealm(configuration: InternalConfiguration) :
+internal class StandaloneDynamicMutableRealm(
+    configuration: InternalConfiguration,
+    private val scheduler: RealmSchedulerPointer = RealmInterop.realm_create_scheduler(),
+) :
     DynamicMutableRealmImpl(
         configuration,
-        RealmInterop.realm_open(configuration.createNativeConfiguration(), null)
+        RealmInterop.realm_open(configuration.createNativeConfiguration(), scheduler)
```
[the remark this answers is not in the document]
clementetb: Yeah, I thought the same. I have added a guard just in case.

rorbech (approved):
LGTM - We should probably have renamed the `scheduler` properties to `executionContext` or something like that to avoid the `scheduler.scheduler`, but will leave it for now.

PR #1471, opened 2023-08-01
Fix corner cases when creating subscriptions

### LINKED ISSUE #1466: Fix SubscriptionSet behaviour when updating subscriptions
From Slack:

> I have two problems with this code:
I'm not convinced query error messages are reported by SyncSession.downloadAllServerChanges so if I used an invalid query with WaitForSync.ALWAYS, I won't get an error. This is arguably a bug in Core, but it is the current behavior.
If I use a named subscription and change the query and use WaitForSync.FIRST_TIME, I'd expect that I'd wait for sync, since even though the name is the same, the query is different. Right now, Kotlin sees that there's a sub with that name and just skips the wait.

We should fix the behavior in both of those cases.

cmelchior:
Closes #1466 

This PR is also fixing a discrepancy between `RealmQuery.description()` and `Subscription.queryDescription()`, so you can now compare the two strings and have it work as expected.


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/ext/RealmQueryExt.kt
```diff
@@ -60,7 +60,9 @@ import kotlin.time.Duration
  * depend on which [mode] was used.
  * @throws kotlinx.coroutines.TimeoutCancellationException if the specified timeout was hit before
  * a query result could be returned.
- * @Throws IllegalStateException if this method is called on a Realm that isn't using Flexible Sync.
+ * @throws IllegalStateException if this method is called on a Realm that isn't using Flexible Sync.
+ * @throws io.realm.kotlin.mongodb.exceptions.BadFlexibleSyncQueryException if the server did not
```
clementetb, line 64: Shall we test that we throw `BadFlexibleSyncQueryException` correctly?

clementetb (approved):
LGTM 🚀

PR #1473, opened 2023-08-01
Add automatic resolution of embedded object constraints during migration

### LINKED ISSUE #1464: Orphaned embedded objects are not deleted when migrating from Object to Embedded Object

-- select --

### Description

According this ticket migration from object to embedded should now be automatic in core and orphaned children should automatically be deleted: https://github.com/realm/realm-core/pull/5737

This doesn't seem to work in the Kotlin SDK. I have a unit test where I have an orphaned object and do an upgrade. Instead of deleting the orphaned object, the migration fails and the app crashes. I have no such issues on iOS

### Stacktrace & log output

```shell
java.lang.IllegalStateException: [RLM_ERR_ILLEGAL_OPERATION]: Cannot convert 'MyEmbeddedObject:' to embedded: at least one object has no incoming links and would be deleted.
at io.realm.kotlin.internal.interop.CoreErrorConverter.asThrowable(CoreErrorConverter.kt:44)
at io.realm.kotlin.internal.interop.realmcJNI.open_realm_with_scheduler(Native Method)
at io.realm.kotlin.internal.interop.realmc.open_realm_with_scheduler(realmc.java:1757)
at io.realm.kotlin.internal.interop.RealmInterop.realm_open(RealmInterop.kt:194)
at io.realm.kotlin.internal.interop.RealmInterop.realm_open$default(RealmInterop.kt:183)
at io.realm.kotlin.internal.ConfigurationImpl.openRealm$suspendImpl(ConfigurationImpl.kt:110)
at io.realm.kotlin.internal.ConfigurationImpl.openRealm(Unknown Source:0)
at io.realm.kotlin.internal.RealmImpl$1.invokeSuspend(RealmImpl.kt:125)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:284)
at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:85)
at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source:1)
at io.realm.kotlin.internal.platform.CoroutineUtilsSharedJvmKt.runBlocking(CoroutineUtilsSharedJvm.kt:22)
at io.realm.kotlin.internal.platform.CoroutineUtilsSharedJvmKt.runBlocking$default(CoroutineUtilsSharedJvm.kt:21)
at io.realm.kotlin.internal.RealmImpl.<init>(RealmImpl.kt:107)
at io.realm.kotlin.internal.RealmImpl.<init>(Unknown Source:0)
at io.realm.kotlin.internal.RealmImpl$Companion.create$io_realm_kotlin_library(RealmImpl.kt:285)
at io.realm.kotlin.Realm$Companion.open(Realm.kt:82)
```

Always

### Reproduction Steps

_No response_

### Version

1.10.1

-- select --

### Are you using encryption?

-- select --

Android

### Build environment

Android Studio version: ...
Android Build Tools version: ...
Gradle version: ...

rorbech:
This adds configuration option to enable automatic handling of embedded object constraints during automatic migration through:

```
class RealmConfiguration {

    class Builder {
        /**
         * Sets the migration to handle schema updates with automatic migration of data.
         *
         * @param migration the [AutomaticSchemaMigration] instance to handle schema and data
         * migration in the event of a schema update.
         * @param resolveEmbeddedObjectConstraints a flag to indicate whether realm should resolve
         * embedded object constraints after migration. If this is `true` then all embedded objects
         * without a parent will be deleted and every embedded object with multiple references to it
         * will be duplicated so that every referencing object will hold its own copy of the
         * embedded object.
         *
         * @see RealmMigration
         * @see AutomaticSchemaMigration
         */
        public fun migration(
            migration: AutomaticSchemaMigration,
            resolveEmbeddedObjectConstraints: Boolean = false
        ): Builder =
            apply {
                this.migration = migration
                this.automaticEmbeddedObjectConstraintsResolution = resolveEmbeddedObjectConstraints
            }
    }
}
```

The above API was chosen in favor of adding additional properties/methods to the `AutomaticSchemaMigration` interface, as this would require users to instantiate it as anynomous instances and that seems very tedious compared to the standalone functional interface. Alternative it could also have been wrapped in a completely new type, but that just bloats the name space.  

Closes #1464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Enhancements
 * Support for performing geospatial queries using the new classes: `GeoPoint`, `GeoCircle`, `GeoBox`, and `GeoPolygon`. See `GeoPoint` documentation on how to persist locations. (Issue [#1403](https://github.com/realm/realm-kotlin/pull/1403))
+* Support for automatic resolution of embedded object constraints during migration through `RealmConfiguration.Builder.migration(migration: AutomaticSchemaMigration, resolveEmbeddedObjectConstraints: Boolean)`. (Issue [#1464](https://github.com/realm/realm-kotlin/issues/1464)
 
 ### Fixed
 * None.
--- a/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -186,6 +186,7 @@ expect object RealmInterop {
     fun realm_config_get_encryption_key(config: RealmConfigurationPointer): ByteArray?
     fun realm_config_set_should_compact_on_launch_function(config: RealmConfigurationPointer, callback: CompactOnLaunchCallback)
     fun realm_config_set_migration_function(config: RealmConfigurationPointer, callback: MigrationCallback)
+    fun realm_config_set_automatic_backlink_handling(config: RealmConfigurationPointer, enabled: Boolean)
     fun realm_config_set_data_initialization_function(config: RealmConfigurationPointer, callback: DataInitializationCallback)
     fun realm_config_set_in_memory(config: RealmConfigurationPointer, inMemory: Boolean)
     fun realm_schema_validate(schema: RealmSchemaPointer, mode: SchemaValidationMode): Boolean
--- a/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -172,6 +172,10 @@ actual object RealmInterop {
         realmc.realm_config_set_migration_function(config.cptr(), callback)
     }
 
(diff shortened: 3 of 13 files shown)


sipersso:
Ok... well as long as I can enable it I guess it isn't an issue. But people migrating from Java to Kotlin might get a nasty surprise if they are not aware of this difference in implementations.

cmelchior (approved):
Nice 🚀 .. I was a bit surprised to see the duplicated method, but I assume it is because we need to constrain it to the `AutomaticRealmMigration` interface rather than the `RealmMigration` interface?

rorbech:
> Nice 🚀 .. I was a bit surprised to see the duplicated method, but I assume it is because we need to constrain it to the `AutomaticRealmMigration` interface rather than the `RealmMigration` interface?

Yes. The option will probably only make sense for an automatic migration, so didn't want to add it to the base method.

PR #1476, opened 2023-08-07
Remove all deprecated APIs from the compiler plugin.

cmelchior:
This PR reworks the compiler plugin, so it no longer uses deprecated APIs. This should move us closer to K2 support, but I'm keeping support out of this for now.

- All APIs using descriptors have been changed to IR-based APIs.
- All APIs marked with `@FirIncompatiblePluginAPI` has been reworked.
- Both of the two above required changing from using `FqNames` to `ClassId` and `CallableId`.

Regarding testing. Our library is built and tested with Kotlin 1.8.21. We do have a minimal sample project being built with 1.8.20, but I'll see if all tests run on 1.8.20 as well, just to be sure we are not breaking backwards compatibility.

### DIFF
--- a/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/Identifiers.kt
+++ b/packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/Identifiers.kt
@@ -16,9 +16,9 @@
 
 package io.realm.kotlin.compiler
 
-import io.realm.kotlin.compiler.FqNames.CLASS_APP_CONFIGURATION
 import io.realm.kotlin.compiler.FqNames.PACKAGE_MONGODB
 import io.realm.kotlin.compiler.FqNames.PACKAGE_MONGODB_INTERNAL
+import org.jetbrains.kotlin.name.CallableId
 import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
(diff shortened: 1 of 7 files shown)


cmelchior:
Using Kotlin 1.8.0, Gradle 7.3.3, and AGP 7.2.0, I hit a few cases not covered by our minimal sample project. I had to backport a single helper method and work around another case where a helper method was no longer available. But with these changes all tests run locally.

packages/plugin-compiler/src/main/kotlin/io/realm/kotlin/compiler/Identifiers.kt
```diff
@@ -86,69 +86,78 @@ internal object Names {
...
+    val TYPED_REALM_OBJECT_INTERFACE = ClassId(FqNames.PACKAGE_TYPES, Name.identifier("TypedRealmObject"))
+    val EMBEDDED_OBJECT_INTERFACE = ClassId(FqNames.PACKAGE_TYPES, Name.identifier("EmbeddedRealmObject"))
+    val ASYMMETRIC_OBJECT_INTERFACE = ClassId(FqNames.PACKAGE_TYPES, Name.identifier("AsymmetricRealmObject"))
 
-    val BASE_REALM_OBJECT_INTERFACE = FqName("io.realm.kotlin.types.BaseRealmObject")
-    val REALM_OBJECT_INTERFACE = FqName("io.realm.kotlin.types.RealmObject")
-    val TYPED_REALM_OBJECT_INTERFACE = FqName("io.realm.kotlin.types.TypedRealmObject")
-    val EMBEDDED_OBJECT_INTERFACE = FqName("io.realm.kotlin.types.EmbeddedRealmObject")
-    val ASYMMETRIC_OBJECT_INTERFACE = FqName("io.realm.kotlin.types.AsymmetricRealmObject")
+    val CLASS_APP_CONFIGURATION = ClassId(FqName("io.realm.kotlin.mongodb"), Name.identifier("AppConfiguration"))
```
rorbech, line 116:
```suggestion
    val CLASS_APP_CONFIGURATION = ClassId(PACKAGE_MONGODB, Name.identifier("AppConfiguration"))
```

rorbech (approved):
LGTM - Only some minor suggestions, but you may skip them. Seems like we still have identifiers here and there anyway, so wouldn't bother if we don't normalize all of them in this go.

PR #1489, opened 2023-08-23
Rework sync exceptions based on new error codes.

cmelchior:
All the old SyncClient, SyncProtocol, and SyncSession errors are gone and have been rolled up into the general RealmError struct. This PR reworks our error handling to accommodate this.

Since we are hiding the real errors behind our exception hierarchy, there should be no visible change to end users. The only exception is the exception message which might now contain different categories and codes, but the behavior should be the same. 

Some notable changes:

- A `CoreError` class has been added that mirrors the C struct `realm_error_t`.
- `CategoryFlags` have been expanded and are now also used for Sync Errors where it previously was only used for local exceptions. 
- Websocket errors have been exposed as an internal enum. They are not currently used, but will be needed for the Platform Networking project.


cmelchior:
Moving to another branch

PR #1490, opened 2023-08-25
Rework internal sync error handling

cmelchior:
All the old SyncClient, SyncProtocol, and SyncSession errors are gone and have been rolled up into the general RealmError struct. This PR reworks our error handling to accommodate this.

Since we are hiding the real errors behind our exception hierarchy, there should be no visible change to end users. The only exception is the exception message which might now contain different categories and codes, but the behavior should be the same.

Some notable changes:

- A `CoreError` class has been added that mirrors the C struct realm_error_t.
- `CategoryFlags` have been expanded and are now also used for Sync Errors where it previously was only used for local exceptions.
- Websocket errors have been exposed as an internal enum. They are not currently used, but will be needed for the Platform Networking project.


packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/CoreError.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Wrapper for C-API `realm_error_t`.
+ * See https://github.com/realm/realm-core/blob/master/src/realm.h#L231
+ */
+class CoreError(
+    categoriesNativeValue: Int,
+    val errorCodeNativeValue: Int,
+    messageNativeValue: String?,
+    // These are represent in the C-API, but not populated by Core.
```
rorbech, line 11: Should we get rid of them just to avoid confusion later?

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/CoreError.kt
```diff
@@ -0,0 +1,69 @@
...
+        )
+    }
+
+    /**
+     * Returns a description of the most important category defined in [categoryFlags].
+     * If no known categories are found, the integer values for all the categories is returned
+     * as debugging information.
+     */
+    val description: String = CATEGORY_ORDER.firstOrNull { category ->
+        category in this
```
rorbech, line 62: I would probably also have gone for this if it was me writing it, but if feels a bit magic and would maybe be easier to review with an explicit `contains` call.

PR #1491, opened 2023-08-25
Add support for automatic resetting incremental backoff when network status changes

### LINKED ISSUE #1479: Add SyncSession.reconnect
It appears we missed `reconnect` when closing https://github.com/realm/realm-kotlin/issues/427... it still has use cases. 

We ended up not exposing a `Sync` class as we didn't really have any other functionality, but maybe adding it to `SyncSession` is fine? 🤔 

We should probably hook this up into the Android connection listener to automatically trigger it when enabling/disabling flight mode.

cmelchior:
Closes #1479 

This PR exposes a number of methods relevant to controlling the network connection and session lifecycle. It does this through a new `Sync` class that is exposed through `App.Sync`, similar to Realm Java.

Before this PR, reconnecting was handled purely by Core, which used incremental backoff. In the worst case, this meant that the Sync connection was not established until 5 minutes after the device was online again (the limit in Core) 

With this PR a new `Sync.reconnect()` method is added that makes it possible to trigger a manual reconnect if users have better insight into when the network condition changed, e.g. through the `ConnectivityManager` on Android. 

It also adds a best-effort implementation for doing this automatically: https://github.com/realm/realm-kotlin/pull/1491/files#diff-835b655f827ffc568b4fd6c04722b17faf9ec56fe73289379bfdad6723d5a00bR79. 

Note, that automatic reconnect detection is only available on Android, iOS, and macOS. JVM has no APIs we can hook into.

TODO
- [x] Add tests
- [x] Figure out how this works on Darwin. Answer: Core is responsible for setting this up on Darwin platforms, thus no changes are needed here and automatic reconnect worked, also before this PR.


cmelchior:
Finally ready for review. For some reason, this PR exposed some race conditions in our test setup caused by us not shutting down all test apps correctly. So a long list of changes here are just associated with fixing this (Basically all apps now have a unique ID that will make it easier to find and fix threading test errors in the future)

packages/library-sync/src/androidMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncInitializer.kt
```diff
@@ -0,0 +1,118 @@
...
+ * in cinterop.o allow Realm to access context properties.
+ */
+class RealmSyncInitializer : Initializer<Context> {
+
+    companion object {
+        @Suppress("DEPRECATION") // Should only be called below API 21
+        fun isConnected(cm: ConnectivityManager?): Boolean {
+            return cm?.let {
+                val networkInfo: NetworkInfo? = cm.activeNetworkInfo
+                networkInfo != null && networkInfo.isConnectedOrConnecting || isEmulator()
```
rorbech, line 44: Why are emulator always connected?

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/TestApp.kt
```diff
@@ -109,13 +123,23 @@ open class TestApp private constructor(
     }
 
     override fun close() {
+        if (isClosed) {
```
rorbech, line 126: Should we also `realm_app_sync_client_wait_for_sessions_to_terminate`?

rorbech, line 126: And synchronize (aka `SynchronizableObject.withLock { }`) the variable?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/ApiKeyAuthTests.kt
```diff
@@ -49,7 +49,7 @@ class ApiKeyAuthTests {
 
     @BeforeTest
     fun setup() {
-        app = TestApp(appName = TEST_APP_PARTITION)
+        app = TestApp(this::class.simpleName, appName = TEST_APP_PARTITION)
```
rorbech, line 52: I guess not for this PR, but would be awfully nice if we could just inject the test name from a rule or something like that. This seems like a really annoying thing to have to always write ... but I guess less annoying than trying to debug without it 😅

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/TestApp.kt
```diff
@@ -50,6 +51,15 @@ val TEST_APP_CLUSTER_NAME = SyncServerConfig.clusterName
 val TEST_SERVER_BASE_URL = SyncServerConfig.url
 const val DEFAULT_PASSWORD = "password1234"
 
+// Expose a try-with-resource pattern for Test Apps
+inline fun App.use(action: (App) -> Unit) {
```
rorbech, line 55: Given that we also forget to wrap apps in this ourselves, should we then consider to rework the `TestApp`-constructor to a factory method with a block that always executed in a try-finally block 🤔

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/CredentialsTests.kt
```diff
@@ -60,7 +60,7 @@ class CredentialsTests {
 
     @BeforeTest
     fun setup() {
-        app = TestApp()
+        app = TestApp(this::class.simpleName)
```
rorbech, line 63: Isn't there a way to get the actual name of the test instead of the suite name?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncClientTests.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    fun sync() {
+        assertNotNull(app.sync)
+    }
+
+    // There is no way to test reconnect automatically, so just verify that code path does not crash.
+    @Test
+    fun reconnect_noRealms() {
+        app.sync.reconnect()
```
rorbech, line 53: Works for me, but there could maybe be some specific log statement?

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/TestApp.kt
rorbech: I guess we can't do this because the app is mostly setup in `@BeforeTest` and `use`d in the actual `@Test`

packages/library-sync/src/androidMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncInitializer.kt
cmelchior: No idea 😅 ... I pulled this code from somewhere else and has never really questioned it 🙈

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/TestApp.kt
cmelchior: I created https://github.com/realm/realm-kotlin/issues/1496 to track this

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/CredentialsTests.kt
cmelchior: I created https://github.com/realm/realm-kotlin/issues/1496 to track this

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/sync/Sync.kt
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Calling this method will block until all sync sessions for a given [App] has terminated.
+     *
+     * Closing a Realm will terminate the sync session, but it is not synchronous as Realms
+     * communicate with their sync session using an asynchronous communication channel. This
+     * has the effect that trying to delete a Realm right after closing it will sometimes throw
+     * an [IllegalStateException]. Using this method can be a way to ensure it is safe to delete
+     * the file.
+     */
+    public fun waitForSessionsToTerminate()
```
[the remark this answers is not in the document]
cmelchior: When Realms are opened. It will wait until all Realms are closed... I tried to think of a way to test it, but I cannot come up with any test that doesn't suffer from race conditions.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncClientTests.kt
cmelchior: Unfortunately not. We don't have anything in the logs, and I would be hesitant to add it just for tests.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmInstantImpl.kt
```diff
@@ -32,7 +32,7 @@ public data class RealmInstantImpl(override val seconds: Long, override val nano
     }
 }
 
-internal fun RealmInstant.toDuration(): Duration {
+public fun RealmInstant.toDuration(): Duration {
```
rorbech, line 35: Not for this PR, but we could consider to make this a public-public extension.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/AppImpl.kt
```diff
@@ -65,18 +68,16 @@ public class AppImpl(
         // "isOnline" messages in short order. So in order to prevent resetting the network
         // too often we throttle messages, so a reconnect can only happen ever 5 seconds.
         RealmLog.debug("Network state change detected. ConnectionAvailable = $connectionAvailable")
-        val nowMs: Long = RealmInstant.now().let { timestamp ->
-            timestamp.epochSeconds * 1000L + timestamp.nanosecondsOfSecond / 1_000_000L
-        }
-        if (connectionAvailable && (lastOnlineStateReportedMs == null || nowMs - lastOnlineStateReportedMs!! > reconnectThresholdMs)
+        val now: Duration = RealmInstant.now().toDuration()
+        if (connectionAvailable && (lastOnlineStateReported == null || now.minus(lastOnlineStateReported!!) > reconnectThreshold)
```
rorbech, line 72: 🎉  ... I guess you could make it more Kotlin idiomatic with something like
```
if (connectionAvailable && (lastOnlineStateReported?.let { now - it > reconnectThreshold } != false))
```
but don't mind keeping it as is.

rorbech (approved):
LGTM and great clean up. Would be interesting to see if this clears up the GHA test clean up issues on Windows 🧐

PR #1494, opened 2023-08-30
Serialization of RealmAny with collections

rorbech:
This adds a suggested workaround for support of serialization of `RealmAny`s with nested collections. 

It seems that statically annotated serializers cannot be mutually recursive, so `RealmListKSerializer`, etc. cannot be used to serialize `RealmList<RealmAny?>` from inside the `RealmAny` serializer. This PR is a suggested workaround for this until we get to the root of the issue, but requires users to manual add the contextual serializers as shown in the tests in https://github.com/realm/realm-kotlin/compare/cr/collections-in-mixed...cr/collection-in-mixed-serialization?expand=1#diff-64803515612f60c2bc368eac00a24dab29eaee3b05952461ffb092790cc55c76R92.


rorbech:
Seems to be related to this issue https://github.com/Kotlin/kotlinx.serialization/issues/1815

PR #1514, opened 2023-09-12
Remove Realm Plugin setting the bytecode target to 1.8

cmelchior:
From Kotlin 1.8.0 the minimum supported Java target is 1.8. This means there is no longer any reason we are setting this in our Realm Plugin. In fact, it might silently modify the behavior of people's compiled code if they are not setting the jvmTarget themselves: https://kotlinlang.org/docs/whatsnew18.html#updated-jvm-compilation-target

I tested this with various combinations of  Java (11 and 17)  and `build.gradle` settings and it doesn't seem to cause build failures.

At the same time, I also unified how we are setting the Kotlin and Java bytecode versions, so they can now all be controlled through our buildSrc Config file.

Regarding the target version, I kept it at 1.8 since we don't know what version of the JVM people are running on the desktop.  On Android the minimum the 11, but it didn't seem worth it to try to differentiate the two.

Since we are running Java 11 as a minimum on our developer machines and CI, we could probably set the `sourceCompatibility` to 11 as well, but the confusion didn't seem worth it. Also, we don't really have any Java sources to speak of anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
-* Realm will now longer set the JVM bytecode to 1.8 when applying the Realm plugin. (Issue [#XXXX]())
+* Realm will no longer set the JVM bytecode to 1.8 when applying the Realm plugin. ([#1513](https://github.com/realm/realm-kotlin/issues/1513))
 
 ### Fixed
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
-* None.
+* Realm will now longer set the JVM bytecode to 1.8 when applying the Realm plugin. (Issue [#XXXX]())
 
 ### Fixed
 * None.
--- a/buildSrc/src/main/kotlin/realm-lint.gradle.kts
+++ b/buildSrc/src/main/kotlin/realm-lint.gradle.kts
@@ -57,6 +57,7 @@ allprojects {
 
         description = "Check Kotlin code style."
         classpath = ktlint
+        jvmArgs = listOf("--add-opens=java.base/java.lang=ALL-UNNAMED")
         mainClass.set("com.pinterest.ktlint.Main")
         args = listOf(
             "src/**/*.kt",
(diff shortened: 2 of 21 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None.
 
 ### Enhancements
-* None.
+* Realm will now longer set the JVM bytecode to 1.8 when applying the Realm plugin. (Issue [#XXXX]())
```
rorbech, line 7:
```suggestion
* Realm will now longer set the JVM bytecode to 1.8 when applying the Realm plugin. (Issue [#1513](https://github.com/realm/realm-kotlin/issues/1513))
```

rorbech (approved):
LGTM ... if CI is green.

CHANGELOG.md
nhachicha, line 7:
```suggestion
* Realm will no longer set the JVM bytecode to 1.8 when applying the Realm plugin. (Issue [#XXXX]())
```

packages/cinterop/build.gradle.kts
```diff
@@ -113,11 +113,7 @@ val nativeLibraryIncludesIosSimulatorArm64Release =
     includeBinaries(releaseLibs.map { "$absoluteCorePath/build-simulator-arm64/lib/$it" })
 
 kotlin {
-    jvm {
-        compilations.all {
-            kotlinOptions.jvmTarget = Versions.jvmTarget
-        }
-    }
+    jvm()
```
nhachicha, line 116: Is it getting the correct target default?

buildSrc/src/main/kotlin/realm-lint.gradle.kts
```diff
@@ -57,6 +57,7 @@ allprojects {
 
         description = "Check Kotlin code style."
         classpath = ktlint
+        jvmArgs = listOf("--add-opens=java.base/java.lang=ALL-UNNAMED")
```
nhachicha, line 60: why do we need this? for refelction?

cmelchior: Some APIs changed or modularized after Java 11, so if you are using Java 17, ktlint will crash. This will fix it.

packages/cinterop/build.gradle.kts
cmelchior: Yes, it is being set by https://github.com/realm/realm-kotlin/pull/1514/files#diff-5ea5ab9e6b701c3d35d3fd2fad81ff600102db520494994cc7448e7d3e032d9dR32 ... which now controls everything.

PR #1515, opened 2023-09-13
Cleanup resources better during Client Reset and closing Realms.

cmelchior:
This PR fixes some problems I discovered on Windows.

- During Client Reset we were not closing a realm instance correctly, which will prevent the Realm file from being deleted later. This fixed most of our Client Reset Tests on Windows, which all timed out during tearDown.
- Manually closing the SyncSession Pointer when closing the Realm make shutdown more predictable. Before this, it was GC dependant.
- I switched to a more accurate pattern for testing client resets rather than us "faking" it.
- I haven't added a changelog entry since we are still leaking something during a manual client reset. I'm still trying to hunt that one down.

### DIFF
--- a/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
+++ b/packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
@@ -906,7 +906,7 @@ after_client_reset(void* userdata, realm_t* before_realm,
     realm_t* after_realm_ptr = realm_from_thread_safe_reference(after_realm, &scheduler);
     auto after_pointer = wrap_pointer(env, reinterpret_cast<jlong>(after_realm_ptr), false);
     env->CallVoidMethod(static_cast<jobject>(userdata), java_after_callback_function, before_pointer, after_pointer, did_recover);
-
+    realm_close(after_realm_ptr);
     if (env->ExceptionCheck()) {
         std::string exception_message = get_exception_message(env);
         std::string message_template = "An error has occurred in the 'onAfter' callback: ";
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
@@ -46,7 +46,6 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlin.reflect.KClass
 
@@ -55,8 +54,6 @@ import kotlin.reflect.KClass
 public class RealmImpl private constructor(
     configuration: InternalConfiguration,
 ) : BaseRealmImpl(configuration), Realm, InternalTypedRealm, Flowable<RealmChange<Realm>> {
-
-    private val realmPointerMutex = Mutex()
 
     public val notificationScheduler: LiveRealmContext =
         configuration.notificationDispatcherFactory.createLiveRealmContext()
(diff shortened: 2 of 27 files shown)


packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncClientResetIntegrationTests.kt
```diff
@@ -1180,23 +1182,30 @@ class SyncClientResetIntegrationTests {
...
+        app: TestApp,
+        user: User,
+        builder: SyncConfiguration.Builder
+    ) {
+        var testRealm: Realm? = null
         val channel = Channel<ClientResetEvents>(2)
         val config = builder.syncClientResetStrategy(object : RecoverUnsyncedChangesStrategy {
             override fun onBeforeReset(realm: TypedRealm) {
-                fail("Should not call onBeforeReset")
+                @Suppress("TooGenericExceptionThrown")
```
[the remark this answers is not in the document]
cmelchior: No, this error is translated on the C++ side as well. This is now being triggered because we are using an actual client reset rather than the "fake" one.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -906,7 +906,7 @@ after_client_reset(void* userdata, realm_t* before_realm,
     realm_t* after_realm_ptr = realm_from_thread_safe_reference(after_realm, &scheduler);
     auto after_pointer = wrap_pointer(env, reinterpret_cast<jlong>(after_realm_ptr), false);
     env->CallVoidMethod(static_cast<jobject>(userdata), java_after_callback_function, before_pointer, after_pointer, did_recover);
-
+    realm_close(after_realm_ptr);
```
nhachicha, line 909: Shouldn't we also close the pointers for Darwin? https://github.com/realm/realm-kotlin/blob/main/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt#L2534-L2538

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -261,27 +260,39 @@ public class RealmImpl private constructor(
...
+    }
+
     override fun close() {
         // TODO Reconsider this constraint. We have the primitives to check is we are on the
         //  writer thread and just close the realm in writer.close()
         writer.checkInTransaction("Cannot close the Realm while inside a transaction block")
-        runBlocking {
-            realmPointerMutex.withLock {
+        realmReferenceLock.withLock {
+            if (isClosed()) {
```
nhachicha, line 274:
```suggestion
            if (isClosed.value) {
```

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/FlexibleSyncIntegrationTests.kt
```diff
@@ -64,7 +63,7 @@ class FlexibleSyncIntegrationTests {
 
     @BeforeTest
     fun setup() {
-        app = TestApp(this::class.simpleName, appName = TEST_APP_FLEX, logLevel = LogLevel.ALL)
```
nhachicha, line 67: It could be helpful to keep it all for CI failures?

nhachicha (approved):
Nice fix 🎉 minor comments

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/FlexibleSyncIntegrationTests.kt
cmelchior: I don't remember having all the logs available was ever useful on CI. Just having the final exception seems to do the trick?

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
cmelchior: Hmm, we probably should. I'm only closing them here because it prevents Windows from removing the file, but you are correct. We should also behave nice on Darwin.

PR #1518, opened 2023-09-13
Fix SyncConfig tests that would break on Windows

cmelchior:
Some tests in `SyncConfigTests` assumed some things about path separators that it shouldn't

### DIFF
--- a/packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncConfigTests.kt
+++ b/packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncConfigTests.kt
@@ -29,8 +29,8 @@ import io.realm.kotlin.entities.sync.flx.FlexChildObject
 import io.realm.kotlin.entities.sync.flx.FlexEmbeddedObject
 import io.realm.kotlin.entities.sync.flx.FlexParentObject
 import io.realm.kotlin.ext.query
-import io.realm.kotlin.internal.platform.PATH_SEPARATOR
 import io.realm.kotlin.internal.platform.createDefaultSystemLogger
+import io.realm.kotlin.internal.platform.pathOf
 import io.realm.kotlin.internal.platform.runBlocking
 import io.realm.kotlin.internal.platform.singleThreadDispatcher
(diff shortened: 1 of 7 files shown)


packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncConfigTests.kt
```diff
@@ -29,6 +29,7 @@ import io.realm.kotlin.entities.sync.flx.FlexChildObject
 import io.realm.kotlin.entities.sync.flx.FlexEmbeddedObject
 import io.realm.kotlin.entities.sync.flx.FlexParentObject
 import io.realm.kotlin.ext.query
+import io.realm.kotlin.internal.platform.PATH_SEPARATOR
```
rorbech, line 32: Hmm. Not for this PR, but unfortunate #naming ... `File.pathSeparator` is usually for separating multiple path entries and not the directory separator 🙃

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncConfigTests.kt
```diff
@@ -406,7 +407,7 @@ class SyncConfigTests {
 
     private fun verifyName(builder: SyncConfiguration.Builder, expectedFileName: String) {
         val config = builder.build()
-        val suffix = "/mongodb-realm/${config.user.app.configuration.appId}/${config.user.identity}/$expectedFileName"
+        val suffix = "${PATH_SEPARATOR}mongodb-realm${PATH_SEPARATOR}${config.user.app.configuration.appId}${PATH_SEPARATOR}${config.user.id}${PATH_SEPARATOR}$expectedFileName"
```
rorbech, line 410: Nitpick and not necessarily for this PR, but I tend to prefer `listOf("...", "...").joinToString(PATH_SEPARATOR)`. Maybe we could even make a `pathOf(...)`-utility method.

rorbech (approved):
LGTM. Just some minor nit picks. There is also an ignored `deleteRealm()` test in `SyncRealmTests.kt` that we might want to fix in the same go. Just for consistency.

PR #1519, opened 2023-09-13
Avoid initial version pinning

clementetb:
During the Realm initialization, we pinned the initial version then it was never released because the `VersionTracker` was not cleaned up.

In this PR we only track in the user-facing Realm the initial snapshot, and we do so until the notifier or writer has a snapshot available. We keep the version tracker, but now is triggered on each notifier/writer update, to allow flushing of the initial realm snapshot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ### Fixed
 * `Realm.getNumberOfActiveVersions` now returns the actual number of active versions. (Core issue [#6960](https://github.com/realm/realm-core/pull/6960))
+* Fixed memory leak on Darwin caused by a reference cycle between resources and the GC cleaner. (Issue [#1530](https://github.com/realm/realm-kotlin/pull/1530))
+* Fixed memory leaks on the JVM platform, see PR for more information. (Issue [#1526](https://github.com/realm/realm-kotlin/pull/1526))
 * Removed pin on the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
 
 ### Compatibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Fixed
 * `Realm.getNumberOfActiveVersions` now returns the actual number of active versions. (Core issue [#6960](https://github.com/realm/realm-core/pull/6960))
-* Removed pinning the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
+* Removed pin on the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,30 @@
+## 1.11.2 (YYYY-MM-DD)
+
+### Enhancements
+* None.
+
+### Fixed
+* `Realm.getNumberOfActiveVersions` now returns the actual number of active versions. (Core issue [#6960](https://github.com/realm/realm-core/pull/6960))
+* Removed pin on the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
+
+### Compatibility
+* File format: Generates Realms with file format v23.
+* Realm Studio 13.0.0 or above is required to open Realms created by this version.
+* This release is compatible with the following Kotlin releases:
+  * Kotlin 1.8.0 and above. The K2 compiler is not supported yet.
+  * Ktor 2.1.2 and above.
+  * Coroutines 1.7.0 and above.
+  * AtomicFu 0.18.3 and above.
+  * The new memory model only. See https://github.com/realm/realm-kotlin#kotlin-memory-model-and-coroutine-compatibility
+* Minimum Kbson 0.3.0.
+* Minimum Gradle version: 6.8.3.
+* Minimum Android Gradle Plugin version: 4.1.3.
+* Minimum Android SDK: 16.
+
+### Internal
+* Updated to Realm Core 13.20.0, commit a3717e492ec606e42995169cf706a82fc1cd0698.
+
+
 ## 1.11.1 (2023-09-07)
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Fixed
 * `Realm.getNumberOfActiveVersions` now returns the actual number of active versions. (Core issue [#6960](https://github.com/realm/realm-core/pull/6960))
-* Removed pin on the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
+* Removed pinning the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealmHolder.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealmHolder.kt
@@ -17,6 +17,8 @@
 package io.realm.kotlin.internal
 
 import io.realm.kotlin.VersionId
+import kotlinx.atomicfu.AtomicBoolean
+import kotlinx.atomicfu.atomic
 
 /**
  * A **live realm holder** encapsulated common properties of [SuspendableWriter] and
@@ -24,9 +26,9 @@ import io.realm.kotlin.VersionId
  * references when advancing the version of [RealmImpl].
  */
 internal abstract class LiveRealmHolder<out LiveRealm> {
-
     abstract val realmInitializer: Lazy<LiveRealm>
     abstract val realm: io.realm.kotlin.internal.LiveRealm
+    val isInitialized: AtomicBoolean = atomic(false)
 
     /**
      * Current version of the frozen snapshot reference of the live realm. This is not guaranteed
@@ -34,21 +36,21 @@ internal abstract class LiveRealmHolder<out LiveRealm> {
      * request a more recent GC-tracked snapshot from the [LiveRealmHolder] through [snapshot].
      */
     val version: VersionId?
-        get() = if (realmInitializer.isInitialized()) { realm.snapshotVersion } else null
+        get() = if (isInitialized.value) { realm.snapshotVersion } else null
 
     /**
      * Returns a GC-tracked snapshot from the underlying [realm]. See [LiveRealm.gcTrackedSnapshot]
      * for details of the tracking.
      */
     val snapshot: FrozenRealmReference?
-        get() = if (realmInitializer.isInitialized()) {
+        get() = if (isInitialized.value) {
             realm.gcTrackedSnapshot()
         } else null
 
     /**
      * Dump the current snapshot and tracked versions of the LiveRealm used for debugging purpose.
      */
-    fun versions(): VersionData? = if (realmInitializer.isInitialized()) {
+    fun versions(): VersionData? = if (isInitialized.value) {
         realm.versions()
     } else {
         null
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealmHolder.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealmHolder.kt
@@ -27,7 +27,6 @@ internal abstract class LiveRealmHolder<out LiveRealm> {
 
     abstract val realmInitializer: Lazy<LiveRealm>
     abstract val realm: io.realm.kotlin.internal.LiveRealm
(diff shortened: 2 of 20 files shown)


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -226,38 +231,46 @@ public class RealmImpl private constructor(
...
-
-            // Consider versions of current realm, notifier and writer to identify if we should
-            // advance the user facing realms version to a newer frozen snapshot.
-            val version = value1?.version()
-            val notifierSnapshot = notifier.version
-            val writerSnapshot = writer.version
+    /**
+     * Removes the local reference to start relying on the notifier - writer for snapshots.
+     */
+    private fun removeInitialRealmReference() {
```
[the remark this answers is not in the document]
clementetb: `_skipClosingReferences` acts as a mutex mechanism. Only one thread would be able to get that it requires to clear the version tracker.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -226,38 +231,48 @@ public class RealmImpl private constructor(
...
```
rorbech, line 237: Why not just `realmReferenceLock.withLock { versionTracker.closeExpiredReferences() }` here? Or even from the notifier flow if you really need to trigger it on signals from the notifier?

rorbech (commented):
Hmm. I feel that this is a bit convoluted. There are some pretty nice clean ups splitting the version tracker's _track_ and _clean_ and how to select the most recent of `notifier` and `writer` versions. But couldn't we just rely on existing signalling and these new addition to solve the use case? Happy to jump on a call and discuss.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
```diff
@@ -17,12 +17,14 @@
...
 import io.realm.kotlin.VersionId
+import io.realm.kotlin.internal.interop.NativePointer
 import io.realm.kotlin.internal.interop.RealmInterop
-import io.realm.kotlin.internal.interop.RealmPointer
+import io.realm.kotlin.internal.interop.RealmT
 import io.realm.kotlin.internal.platform.WeakReference
 import kotlinx.atomicfu.AtomicRef
 import kotlinx.atomicfu.atomic
 
+internal typealias IntermediateReference = Pair<NativePointer<out RealmT>, WeakReference<RealmReference>>
```
rorbech, line 27: Couldn't this be?
```suggestion
internal typealias IntermediateReference = Pair<RealmPointer, WeakReference<RealmReference>>
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
```diff
@@ -32,14 +34,33 @@ import kotlinx.atomicfu.atomic
 internal class VersionTracker(private val owner: BaseRealmImpl, private val log: ContextLogger) {
     // Set of currently open realms. Storing the native pointer explicitly to enable us to close
     // the realm when the RealmReference is no longer referenced anymore.
-    private val intermediateReferences: AtomicRef<Set<Pair<RealmPointer, WeakReference<RealmReference>>>> = atomic(mutableSetOf())
+    private val intermediateReferences: AtomicRef<MutableSet<IntermediateReference>> =
```
rorbech, line 37: I actually think that this is from the old memory model, so don't think we need to "atominice" this anymore. And as long as you ensure synchronization elsewhere you could just operate on the mutable set. At least I think, so probably worth an experiment. If so, then `trackReference` could just add the single reference and  close expired could be trimmed to something like
```
intermediateReferences.removeAll { (pointer, ref) ->
    val reference = ref.get()
    return if (reference == null) {
        log.trace("$owner CLOSE-FREED ${RealmInterop.realm_get_version_id(pointer)}")
        RealmInterop.realm_close(pointer)
        true
    } else false
}
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -131,13 +130,14 @@ public class RealmImpl private constructor(
...
-                versionTracker.trackAndCloseExpiredReferences(frozenReference)
-                _realmReference.value = frozenReference
+                versionTracker.trackReference(frozenReference)
+                initialRealmReference.value = frozenReference
                 configuration.initializeRealmData(this@RealmImpl, realmFileCreated)
             }
 
             realmScope.launch {
                 notifier.realmChanged().collect {
+                    removeInitialRealmReference()
```
rorbech, line 140: It is not enough to closeExpiredReferences() in `removeInitialRealmReference` only the first time that you advance/clear the initial reference, as this might still be referenced. In that case it would not be closed and we will never revisit this and never close it.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/LiveRealmHolder.kt
```diff
@@ -17,38 +17,43 @@
...
  */
 internal abstract class LiveRealmHolder<out LiveRealm> {
-
-    abstract val realmInitializer: Lazy<LiveRealm>
     abstract val realm: io.realm.kotlin.internal.LiveRealm
 
+    /**
+     * Indicates whether the live realm has been initialized or not.
+     */
+    val isInitialized: AtomicBoolean = atomic(false)
```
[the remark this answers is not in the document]
clementetb: I was runnning into some race conditions so I made it atomic, seems it is no longer an issue.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
rorbech, line 140: This is executing on the `notifier` thread so you need to synchronize it to be sure that it is up to date.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -72,19 +74,22 @@ public class RealmImpl private constructor(
     private val notifier = SuspendableNotifier(
         owner = this,
         scheduler = notificationScheduler,
+        onSnapshotAvailable = ::removeInitialRealmReference
```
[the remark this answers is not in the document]
clementetb: Yes, I liked to have instant notifications from both realms but it is convoluted. Now, I am leveraging the realm-changed notifications from the notifier realm to trigger the initial realm reference cleanup.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
rorbech: Would it somehow be possible to pass the initial frozen version into the notifier's version tracker instead? Then we could completely eliminate the version tracker here 🤔

rorbech (commented):
There are some issues with the implementation. 

Also noted that the `VersionTracker.versions` are actually not printing the garbage collecting versions even though they have not been closed 🙈 We should probably change it so that it was distinct if the version was no longer reference and/or closed.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
clementetb: I experimented removing the atomicity, but then it ran into race conditions. But that might be because I was initiating a new set. I will give it another try.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
clementetb: My bad, yes it is supposed to be triggered on each notification. I am starting to think of adding the GC auto-close.

clementetb: That was my initial design, but then it kind of convolute testing on the notifier version tracker.

Because the notifier loads later, you might find one or two active versions, user-facing and notifier pending to be released.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
clementetb: Yes, it can!

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
clementetb: Make it a regular set.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -226,41 +229,38 @@ public class RealmImpl private constructor(
...
+        return initialRealmReference.value.let { localReference ->
+            // Find whether the user-facing, notifier or writer has the latest snapshot.
+            // Sort is stable, it will try to preserve the following order.
+            listOf(
+                { localReference } to localReference?.uncheckedVersion(),
+                { notifier.snapshot } to notifier.version,
+                { writer.snapshot } to writer.version
+            ).sortedByDescending {
+                it.second
+            }.first().first()
```
rorbech, line 255: Nit-pick but `.first().first.invoke()` would make it a bit clearer that you are actually invoking the lambda.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionInfo.kt
```diff
@@ -22,7 +22,7 @@ import io.realm.kotlin.VersionId
  * Version meta data for an overall [Realm]-instance with [VersionData] for the user-facing [Realm]
  * and the underlying [SuspendableNotifier]'s and [SuspendableWriter]'s live realms.
  */
-public data class VersionInfo(val main: VersionData?, val notifier: VersionData?, val writer: VersionData?) {
-    val all: Set<VersionId> = setOf(main, notifier, writer).mapNotNull { it?.versions }.flatten().toSet()
-    val allTracked: Set<VersionId> = setOf(main, notifier, writer).mapNotNull { it?.active }.flatten().toSet()
+public data class VersionInfo(val notifier: VersionData?, val writer: VersionData?) {
```
rorbech, line 25: Why don't we track the _initial version_ anymore? Isn't it as important to know if the initial version is released as it is with the versions tracked by the notifier and writer 🤔

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
rorbech: ??

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/VersionTrackingTests.kt
```diff
@@ -211,6 +208,41 @@ class VersionTrackingTests {
             samples.map { it.list.version() }.joinToString { it.toString() }
         )
     }
+
+    @Test
+    @Suppress("invisible_member", "invisible_reference")
+    fun initialVersionDereferencedAfterFirstWrite() {
+        (realm as RealmImpl).let { realm ->
+            realm.initialRealmReference.value
+
```
rorbech, line 217:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
+## 1.11.2 (YYYY-MM-DD)
+
+### Enhancements
+* None.
+
+### Fixed
+* `Realm.getNumberOfActiveVersions` now returns the actual number of active versions. (Core issue [#6960](https://github.com/realm/realm-core/pull/6960))
+* Removed pin on the initial realm version after opening a Realm. (Issue [#1519](https://github.com/realm/realm-kotlin/pull/1519))
```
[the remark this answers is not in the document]
rorbech: Resolved without any acknowledgement or comments 🤔

rorbech (approved):
Looks good, but had some minor comments.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionTracker.kt
clementetb: It is a note to remember to use a read-only set.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/VersionInfo.kt
clementetb: I have moved it into a separate test case: `initialVersionDereferencedAfterFirstWrite`

clementetb: I removed because I was testing it now in a different test case: initialVersionDereferencedAfterFirstWrite. But it is true that it is better to keep it.

PR #1523, opened 2023-09-18
Fix error in RealmAny.equals when comparing similarly typed object

rorbech:


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 ### Fixed
 * `Realm.getNumberOfActiveVersions` now returns the actual number of active versions. (Core issue [#6960](https://github.com/realm/realm-core/pull/6960))
+* Fix error in `RealmAny.equals` that would sometime return true when comparing RealmAnys wrapping same type but different values.
```
```suggestion
* Fix error in `RealmAny.equals` that would sometimes return `true` when comparing RealmAnys wrapping same type but different values. (Issue [#1523](https://github.com/realm/realm-kotlin/pull/1523))
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmAnyImpl.kt
```diff
@@ -127,27 +127,27 @@ internal class RealmAnyImpl<T : Any> constructor(
         if (other.type != this.type) return false
         if (clazz == ByteArray::class) {
             if (other.internalValue !is ByteArray) return false
-            if (!other.internalValue.contentEquals(this.internalValue as ByteArray)) return false
+            return other.internalValue.contentEquals(this.internalValue as ByteArray)
         } else if (internalValue is BsonObjectId) {
```
cmelchior, line 131: We only support the KBSON variant for RealmAny, right? And not also our deprecated variant?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmAnyTests.kt
```diff
@@ -421,6 +422,66 @@ class RealmAnyTests {
...
+                    assertNotEquals(
+                        RealmAny.create(byteArrayOf(1, 2)), RealmAny.create(byteArrayOf(2, 1))
+                    )
+                }
+                RealmAny.Type.TIMESTAMP -> {
+                    val now = RealmInstant.now()
+                    assertEquals(RealmAny.create(now), RealmAny.create(now))
+                    assertNotEquals(RealmAny.create(RealmInstant.from(1, 1)), RealmAny.create(now))
+                }
+                RealmAny.Type.FLOAT -> {
```
cmelchior, line 454: The values here should be something that cannot be implicitly cast to a Long

cmelchior (changes requested):
Ups 🙈 ...but this PR doesn't 100% fix the issue ;)

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmAnyTests.kt
```diff
@@ -421,6 +422,70 @@ class RealmAnyTests {
...
+                }
+                RealmAny.Type.UUID -> {
+                    val value = RealmUUID.random()
+                    assertEquals(RealmAny.create(value), RealmAny.create(value))
+                    assertNotEquals(RealmAny.create(RealmUUID.random()), RealmAny.create(value))
+                }
+                RealmAny.Type.OBJECT -> {
+                    val realmObject = Sample()
+                    assertEquals(RealmAny.create(realmObject), RealmAny.create(realmObject))
+                    assertNotEquals(RealmAny.create(Sample()), RealmAny.create(realmObject))
```
cmelchior, line 483: I would also add a test that exercises `if (other.clazz != this.clazz) return false`

cmelchior (commented):
Looking fine, but there seems to be a CI error that looks slightly concerning. Not 100% why this PR would cause that though.

rorbech:
> Looking fine, but there seems to be a CI error that looks slightly concerning. Not 100% why this PR would cause that though.

This is failing because the serialization round trip will loose the nano-second precision as it is not representable in EJson and the RealmAny comparison does not catch the inequality prior to the update.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmAnyImpl.kt
rorbech: Right. There is only an entry point for `org.mongodb.kbson.BsonObjectId`. I have removed any logic guarding this and relied on type checking and internal value's `equals` method for this.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmAnyImpl.kt
```diff
@@ -127,27 +127,27 @@ internal class RealmAnyImpl<T : Any> constructor(
...
+                is Number -> other.internalValue.toLong() == this.internalValue.toLong()
                 else -> return false
             }
         } else if (internalValue is Char) { // We are comparing chars
-            when (other.internalValue) {
-                is Char -> if (other.internalValue.code.toLong() != internalValue.toLong()) return false
-                is Number -> if (other.internalValue.toLong() != this.internalValue.toLong()) return false
+            return when (other.internalValue) {
+                is Char -> other.internalValue.code.toLong() == internalValue.toLong()
+                is Number -> other.internalValue.toLong() == this.internalValue.toLong()
```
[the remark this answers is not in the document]
rorbech: No, indeed not. Good catch. I have removed the specializations around `Char` and `Number` and are now just relying on type checks and inner value's `equals`.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmAnyTests.kt
rorbech: The `toLong()` actually truncates it, so can't really choose other good numbers. I removed the `toLong()` implementation in `RealmAny.equals` and are now relying on inner value's `equals` so should behave according to Kotlins semantics now.

cmelchior (approved):

PR #1526, opened 2023-09-25
Fix JVM memory leaks

clementetb:
This PR fixes some native memory leaks caused when handling core:

- Schema creation leaked strings buffers.
- Querying the schema leaked property pointer buffers.
- Converting sync errors leaked a map.
- Retrieving all user identities leaked an array.
- Collection changesets leaked indexes and change arrays.
- Query parsing leaked arguments buffers.


rorbech (commented):
Nice catches. Wouldn't it make sense to encapsulate all these allocations (`PropertyInfo`, `PropertyInfo[]`, `ClassInfo`, `ClassInfo[]`, etc.) into the `MemTrackingAllocator` so that we could just use our normal `inputScope { }` for these kind of blocks?

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -117,7 +118,35 @@ actual object RealmInterop {
...
+                realmc.delete_propertyArray(propertyArray)
+
+                // end clean properties
+
+                realmc.realm_class_info_t_cleanup(classInfo)
+                classInfo.delete()
+            }
+
+            realmc.delete_propertyArrayArray(cproperties)
+            realmc.delete_classArray(cclasses)
```
cmelchior, line 148: Not sure I get why we need to do cleanup this way? Also, this is quite a lot of JNI traversals to do cleanup. Couldn't we move this entire block into a helper function in `realm_api_helpers`?

cmelchior: Also, doesn't `realmc.classArray_getitem` cause a new allocation rather than returning the old one?

cmelchior (commented):
Great find. I do have some questions about some of the implementations though.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
clementetb: But now that you mention, in `realm_get_class` we could free the allocated struct afterwards.

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -1609,62 +1652,78 @@ actual object RealmInterop {
...
+                    realm.cptr(),
+                    classKey.key,
+                    query,
+                    args.size,
+                    args.head
+                )
             )
-        )
+        } finally {
+            realmc.delete_queryArgArray(args.head)
```
rorbech, line 1668: I don't think we release everything here. The query argument array holds a pointer to a value array allocated in `JvmMemAllocator.allocRealmValueList`. Maybe not showing up in the benchmarks if you are not testing with list query-arguments (`query("x in $0", [1, 2, 3]`)

rorbech (commented):
As per discussion I am OK with just fixing this now, and making a nicer scoped solution when optimizing further ... but I think we are missing a leak of list query-arguments 🧐

packages/cinterop/src/jvm/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
cmelchior: This is still doing a ton of JNI traversals. My comment was about us moving the entire for-loop into C++. I assume that is possible?

clementetb: Yes, it is possible. I wanted to use, if possible, swig APIs, and then in a separate PR move this logic to c++ to measure the actual performance gain from doing it.

cmelchior: Ah, fair enough 👍

nhachicha (approved):
LGTM 👍

PR #1530, opened 2023-09-26
Fix darwin native leak

clementetb:
A reference cycle between `CPointerWrapper` and its cleaner was causing a leak were the GC was not reclaiming any native memory. 

This PR breaks the cycle by moving the `CPointerWrapper` and cleaner shared data into a new data class.

See [LeakingCleaner](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native.ref/create-cleaner.html).

### DIFF
--- a/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -111,7 +111,6 @@ import realm_wrapper.realm_property_info_t
 import realm_wrapper.realm_query_arg_t
 import realm_wrapper.realm_release
 import realm_wrapper.realm_results_t
-import realm_wrapper.realm_scheduler_notify_func_t
 import realm_wrapper.realm_scheduler_t
 import realm_wrapper.realm_set_t
 import realm_wrapper.realm_string_t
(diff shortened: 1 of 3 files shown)


packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -160,11 +159,22 @@ private fun <T : CPointed> checkedPointerResult(pointer: CPointer<T>?): CPointer
...
 }
 
-// FIXME API-INTERNAL Consider making NativePointer/CPointerWrapper generic to enforce typing
+/**
+ * Class with a pointer reference and its status. It breaks the reference cycle between CPointerWrapper
+ * and its GC cleaner, otherwise the cleaner would never be invoked.
+ *
+ * See leaking cleaner: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native.ref/create-cleaner.html
+ */
+data class ReleasablePointer(
```
cmelchior, line 168: After reading those docs, I'm still not 100% sure why we need this class 🤔 

Couldn't we just move all the logic to the `release` method:

```
    val cleaner = if (managed) {
        createCleaner(this) { it.release() }
    } else null

    override fun release() {
        if (released.compareAndSet(expect = false, update = true)) {
            realm_release(_ptr)
        }
    }
```

I think it was me that added the old logic, and I don't remember if there was a problem not doing that in the first place 🙈

cmelchior (commented):
Great find 🥇 I do have a question about the implementation though.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
clementetb: The original issue is that the cleaner lambda referenced a property of the object to clean, causing the object to never be garbage collected.

This class breaks the cycle, the cleaner lambda now points to it, allowing the GC to collect the original reference.

cmelchior (approved):
Great find. Just needs a changelog update, otherwise 👍

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
nhachicha: I think in the original implementation we were referencing `released: AtomicBoolean = atomic(false)` inside the cleaner lambda (`released.compareAndSet(expect = false, update = true)`) which is equivalent to capturing `this` which will leak as per the doc example (`createCleaner(this)`).

nhachicha (approved):
Good catch 👍

PR #1533, opened 2023-10-03
Add geo types sync integration tests.

clementetb:
Add tests to validate we can define sync subscriptions using geo queries.


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/UserImpl.kt
```diff
@@ -49,7 +49,7 @@ public class UserImpl(
     override val loggedIn: Boolean
         get() = RealmInterop.realm_user_is_logged_in(nativePointer)
     override val provider: AuthenticationProvider
-        get() = AuthenticationProviderImpl.fromId(RealmInterop.realm_user_get_auth_provider(nativePointer))
+        get() = identities.first().provider
```
cmelchior, line 52: Talking to JS, we should deprecate this method as it doesn't really make sense anymore and instead point people towards `user.identities`.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+import kotlinx.coroutines.delay
+import org.mongodb.kbson.ObjectId
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFails
+import kotlin.time.Duration.Companion.seconds
+
+private val GEO_SCHEMA = setOf(SyncRestaurant::class, Location::class)
```
cmelchior, line 33: We cannot use this anymore. All schemas need to be the same after the server introduced the new developer mode. I have a branch you can work from.

cmelchior: See https://github.com/realm/realm-kotlin/pull/1535

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+
+@OptIn(ExperimentalGeoSpatialApi::class)
+class GeoSpatialTests {
+    private lateinit var app: TestApp
+
+    @BeforeTest
+    fun setup() {
+        app = TestApp(this::class.simpleName, appName = TEST_APP_FLEX, logLevel = LogLevel.ALL)
```
cmelchior, line 41: LogLevel.ALL seems a bit much?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+        val config =
+            SyncConfiguration.Builder(
+                user = user,
+                schema = GEO_SCHEMA
+            ).initialSubscriptions(rerunOnOpen = true) {
+            }.build()
+
+        Realm.open(config).use {
+            it.writeBlocking {
+                assertFails {
```
cmelchior, line 91: I don't think the write will fail as long as the class is in the schema, instead you will probably see a compensating write?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+    ) {
+        val section = ObjectId()
+
+        // User #1 will try to write some data and assert some failure conditions.
+        val user1 = createRandomUser()
+        val config =
+            SyncConfiguration.Builder(
+                user = user1,
+                schema = GEO_SCHEMA
+            ).initialSubscriptions(rerunOnOpen = true) {
```
cmelchior, line 160: Why rerun on open?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+                        "section = $0 AND location GEOWITHIN $1",
+                        section,
+                        bounds
+                    )
+                )
+            }.waitForInitialRemoteData().build()
+
+        Realm.open(config).use { realm ->
+            runBlocking {
+                realm.subscriptions.waitForSynchronization(timeout = 30.seconds)
```
cmelchior, line 172: This shouldn't be needed if you have `waitForInitialRemoteData`. It should take into account initial subscriptions.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+                    SyncRestaurant().apply {
+                        this.section = section
+                        this.location = validLocation
+                    }
+                )
+            }
+
+            runBlocking {
+                realm.syncSession.uploadAllLocalChanges(timeout = 30.seconds)
+                delay(30000)
```
cmelchior, line 194: Why this?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
```diff
@@ -0,0 +1,229 @@
...
+                    it.query<SyncRestaurant>(
+                        "section = $0 AND location GEOWITHIN $1",
+                        section,
+                        bounds
+                    )
+                )
+            }.waitForInitialRemoteData().build()
+
+        Realm.open(config).use { realm ->
+            runBlocking {
```
cmelchior, line 171: There are a lot of runBlocking calls in here. I would probably just move it outside the `Realm.open()` and then cleanup the code here.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
clementetb: It fails with `RLM_ERR_NO_SUBSCRIPTION_FOR_WRITE`. It is core preemptively catching the error.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
clementetb: I forgot to remove it. I wanted to check why the data is deleted after setting an invalid location on a polygon shape, while Circle and Box undo the changes.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/GeoSpatialTests.kt
clementetb: Initially, I reused the Realms and then I did not remove this line.

cmelchior (approved):
:shipit:

PR #1534, opened 2023-10-03
Remove dependency on KAPT

### LINKED ISSUE #1513: Cannot compile with Kotlin 1.9.10

Always

### Description

Used Android project with Kotlin v`1.8.22` and Realm-Kotlin v`1.10.0` with no issues.

When migrated to use Kotlin `1.9.x`, specifically `1.9.10`, cannot compile due to this error:
```
'compileDebugJavaWithJavac' task (current target is 17) and 'kaptGenerateStubsDebugKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

Tbe `build.gradle` in every Android Module contains:
```groovy
android {
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_17
        targetCompatibility JavaVersion.VERSION_17
    }
    kotlinOptions {
        jvmTarget = '17'
    }
}
```

without reference to Module containing dependency: `io.realm.kotlin:library-base`, the project compiles.

### Stacktrace & log output

```shell
Execution failed for task ':database:kaptGenerateStubsDebugKotlin'.
> 'compileDebugJavaWithJavac' task (current target is 17) and 'kaptGenerateStubsDebugKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

Always

### Reproduction Steps

_No response_

### Version

1.11.1, 1.10.0

Local Database only

### Are you using encryption?

No

Android SDK 33

### Build environment

Android Studio version: Android Studio Giraffe | 2022.3.1 Patch 1
Android Build Tools version: 33
Gradle version: 8.3

cmelchior:
Closes #1513 

Fixes the error in #1513, but it will just move the error to another. Most likely 

```
> 'compileDebugJavaWithJavac' task (current target is 1.8) and 'compileDebugKotlinAndroid' task (current target is 17) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 * None.
 
 ### Enhancements
-* Realm will no longer set the JVM bytecode to 1.8 when applying the Realm plugin. ([#1513](https://github.com/realm/realm-kotlin/issues/1513))
+* Realm will no longer set the JVM bytecode to 1.8 when applying the Realm plugin. (Issue [#1513](https://github.com/realm/realm-kotlin/issues/1513))
+* The Realm Gradle Plugin no longer has a dependency on KAPT. (Issue [#1513](https://github.com/realm/realm-kotlin/issues/1513))
 
 ### Fixed
 * `Realm.close()` is now idempotent.
--- a/packages/gradle-plugin/build.gradle.kts
+++ b/packages/gradle-plugin/build.gradle.kts
@@ -17,7 +17,6 @@ import kotlin.text.toBoolean
 
 plugins {
     kotlin("jvm")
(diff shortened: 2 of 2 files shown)


cmelchior:
Yes, setting the toolchain in the Kotlin block fixes the issue, but it is not something we can do for people. They have to decide which java version to support. We should probably have it added to our documentation though.

PR #1542, opened 2023-10-11
Tests GC realm pointer release

clementetb:
Adds a test case to catch whether we are releasing realm pointers.

### DIFF
--- a/packages/cinterop/src/nativeDarwinTest/kotlin/io/realm/kotlin/test/CinteropTest.kt
+++ b/packages/cinterop/src/nativeDarwinTest/kotlin/io/realm/kotlin/test/CinteropTest.kt
@@ -106,11 +106,16 @@ class CinteropTest {
         assertFalse(releasablePointer.released.value)
 
         // Trigger GC and wait for some time to allow it to collect the object
-        GC.collect()
-        platform.posix.sleep(5u)
+        for (i in 0..5) {
+            GC.collect()
+            platform.posix.sleep(5u)
 
-        // The pointer has not been reclaimed
-        assertTrue(releasablePointer.released.value)
+            // if reclaimed stop looping
+            if (releasablePointer.released.value) break
+        }
+
+        // The pointer has been reclaimed
+        assertTrue(releasablePointer.released.value, "Pointer was not reclaimed")
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


packages/cinterop/src/nativeDarwinTest/kotlin/io/realm/kotlin/test/CinteropTest.kt
```diff
@@ -79,6 +82,36 @@ import kotlin.test.assertTrue
...
+        }()
+
+        // The pointer has not been reclaimed
+        assertFalse(releasablePointer.released.value)
+
+        // Trigger GC and wait for some time to allow it to collect the object
+        GC.collect()
+        platform.posix.sleep(5u)
+
+        // The pointer has not been reclaimed
```
cmelchior, line 112:
```suggestion
        // The pointer has been reclaimed
```

packages/cinterop/src/nativeDarwinTest/kotlin/io/realm/kotlin/test/CinteropTest.kt
```diff
@@ -79,6 +82,36 @@ import kotlin.test.assertTrue
...
+                CPointerWrapper<RealmSchemaT>(realmSchemaNew)._ptr
+            }
+        }()
+
+        // The pointer has not been reclaimed
+        assertFalse(releasablePointer.released.value)
+
+        // Trigger GC and wait for some time to allow it to collect the object
+        GC.collect()
+        platform.posix.sleep(5u)
```
cmelchior, line 110: How stable is this? In other places, we have put in loops that test this a few times just to avoid flaky tests. I would probably recommend doing the same here.

cmelchior (approved):
A small recommendation to make the test more stable, but otherwise this looks good 👍

packages/test-base/src/nativeDarwinTest/kotlin/io/realm/kotlin/test/darwin/MemoryTests.kt
```diff
@@ -57,63 +55,6 @@ class MemoryTests {
         deleteTempDir(tmpDir)
     }
 
-    // TODO Only run on macOS, filter using https://developer.apple.com/documentation/foundation/nsprocessinfo/3608556-iosapponmac when upgrading to XCode 12
```
rorbech, line 60: Why are we removing this test? The other test is verifying that garbage collecting is actually cause a `realm_release` but that is just a subset of ensure that we actually release all resource when not referencing a realm anymore.

PR #1545, opened 2023-10-12
Bump core to version 13.23.1

clementetb:
Bumps core to version 13.23.0.

Core has removed `realm_user_get_auth_provider`, in reaction to this we have annotated `User.provider` as deprecated and now returns the first element auth provider at `User.identities`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ GET requests. (Issue [#1517](https://github.com/realm/realm-kotlin/pull/1517))
 * Minimum Android SDK: 16.
 
 ### Internal
-* Updated to Realm Core 13.23.0, commit f8604b9fd3d2982008a1d3f5ff35e52ee9098d5b.
+* Updated to Realm Core 13.23.1, commit c569bec4d04da84030d94f376437bc4efda3686b.
 
 
 ## 1.11.1 (2023-09-07)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.12.0-SNAPSHOT (YYYY-MM-DD)
 
-To downgrade a Sync app from this version, you'll need to manually delete the metadata folder located at `$[SYNC-ROOT-DIRECTORY]/mongodb-realm/[APP-ID]/server-utility/metadata/`.
+This release upgrades the Sync metadata in a way that is not compatible with older versions. To downgrade a Sync app from this version, you'll need to manually delete the metadata folder located at `$[SYNC-ROOT-DIRECTORY]/mongodb-realm/[APP-ID]/server-utility/metadata/`. This will log out all users.
 
 ### Breaking Changes
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ GET requests. (Issue [#1517](https://github.com/realm/realm-kotlin/pull/1517))
 * Minimum Android SDK: 16.
 
 ### Internal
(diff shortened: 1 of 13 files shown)


cmelchior (approved):
We should probably target `main` with this. But otherwise 👍  if CI is green.

nhachicha (approved):

cmelchior:
We should also fix https://github.com/realm/realm-kotlin/issues/1546 as part of this

clementetb:
closes https://github.com/realm/realm-kotlin/issues/1546

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/User.kt
```diff
@@ -83,6 +83,7 @@ public interface User {
      * Returns the provider type used to log the user in.
      * If a user logs out, the authentication provider last used to log the user in will still be returned.
      */
+    @Deprecated("Property not stable, users might have multiple providers.", ReplaceWith("identities"))
```
rorbech, line 86:
```suggestion
    @Deprecated("Users might have multiple providers. This will return the provider of the first identity of the user", ReplaceWith("identities"))
```

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/AppTests.kt
```diff
@@ -391,7 +391,7 @@ class AppTests {
             Realm.open(syncConfig).close()
 
             // Create a configuration pointing to the metadata Realm for that app
-            val lastSetSchemaVersion = 6L
+            val lastSetSchemaVersion = 7L
```
rorbech, line 394: Hmm. I don't really know the full consequence of this, but from this comment it seems that we should at least warn about not being able to downgrade 🤔

clementetb: We are opening the metadata directly instead of opening a higher schema version with an app to trigger a schema downgrade.

I am going to create a ticket to validate what is the expected behavior when downgrading an App metadata schema.

clementetb: https://github.com/realm/realm-kotlin/issues/1549

rorbech: Ahh sorry, forgot the link to the comment I was referring to: https://github.com/realm/realm-kotlin/pull/1528#discussion_r1355148801

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 1.12.0-SNAPSHOT (YYYY-MM-DD)
 
+To downgrade a Sync app from this version, you'll need to manually delete the metadata folder located at `$[SYNC-ROOT-DIRECTORY]/mongodb-realm/[APP-ID]/server-utility/metadata/`.
```
cmelchior, line 3:
```suggestion
This release upgrades the Sync metadata in a way that is not compatible with older versions. To downgrade a Sync app from this version, you'll need to manually delete the metadata folder located at `$[SYNC-ROOT-DIRECTORY]/mongodb-realm/[APP-ID]/server-utility/metadata/`. This will log out all users.
```

cmelchior (approved):
Small suggestion for the changelog, otherwise 👍

Issue #713, opened 2022-03-03
Internal refactor wishlist


cmelchior:
This issue is a meta issue for tracking and discussing the most important internal refactors we would like to do in the short-team:

So far I have compiled the following list:

* (Christian) We need to figure out how to split `cinterop` so it supports both `base` and `sync` builds. Instead of right now where sync native code is getting shipped with the base library. The networking refactor in core, should in theory remove the need to depend on us shipping SSL, but it is unclear what their timeline is.

* (Christian) `RealmInterop` is too huge an annoying to reason about. I would like to split it into more manageable chunks that mirror the ObjectStore classes: `Realm`, `Results, `List`, `Object` or something along those lines.

* (Christian) Add Benchmarking setup. The first step should be to set up benchmarks for Android (since that is our slowest platform), then later for iOS.

* (Claus) Bundle all managed properties into one composite object. It will significantly reduce the amount of code in our compiler plugin, help us enforce invariants and improve type-safety in our internal code

Feel free to edit or clarify if I'm missing something

rorbech:
> (Christian - 1-3?) We need to figure out how to split `cinterop` so it supports both `base` and `sync` builds. Instead of right now where sync native code is getting shipped with the base library. The networking refactor in core, should in theory remove the need to depend on us shipping SSL, but it is unclear what their timeline is.
> 
> * Claus has some ideas on this.

According to my initial investigations (more that a year ago 🙈) it should be possible to achieve a concept like build variants through custom compilations: https://kotlinlang.org/docs/multiplatform-configure-compilations.html

With the current setup (different modules for `library-base` and `library-sync`) it should also be possible to configure that `library-base`-dependency of `library-sync` to override the `cinterop`-dependency to such a new `cinterop-sync` variant.

rorbech:
> * (Claus - 1) Bundle all managed properties into one composite object. Initial thought is to substitute `RealmObject`'s inheritance from `RealmObjectInteral` to a composition of an `managed` internal property. It will significantly reduce the amount of code in our compiler plugin, help us enforce invariants and improve type-safety in our internal code. Could probably also do something similar with `RealmObjectCompanion` to keep public visible things at a minimum.

[...]

PR #1547, opened 2023-10-13
Add support for Keypath filtering in notifications

### LINKED ISSUE #661: Add support for keypath filtering
From Realm Java we had this issue: https://github.com/realm/realm-java/issues/4786

Which is about object level notifications only triggering if top-level properties are modified, but not if nested objects are modified. For RealmResults this also works for nested objects.

This is a highly surprising behavior and something we should fix: https://github.com/realm/realm-core/issues/5212

One work-around is adding support for keypath filtering, which also has other use cases where you want to filter notifications.

We should add support for this to our query API.

A suggested API could be something like:

```
RealmQuery.asFlow(keyPaths: Array<String>)
RealmSingleQuery.asFlow(keyPaths: Array<String>)
```

We need Core support before this can happen: https://github.com/realm/realm-core/issues/5216

I don't believe Core currently has support for wildcards, e.g `*` or `myList.*`, which would probably also be helpful, but we should push for that in Core instead.

cmelchior:
Closes #661 

This PR adds support for keypath filtering which makes it possible to select which fields should trigger the notifications, and enable notifications more than 4 levels deep. 

TODO
- [x] Add tests for query flows
- [x] Add tests for single object query flows
- [x] Figure out how to do our internal interface, I.e. some do not support keypaths like Realm and aggregate flows. Either add two different interfaces or have 1 but just ignore the keypath for some notifications (internally)
- [x] Add support for passing the keypath to the C-API.
- [x] Should the public API use `List<String>()`, `List<String>?` or `vararg String`. Right now I'm leaning towards `List<String>?` but should be discussed.
- [x] Merge https://github.com/realm/realm-core/pull/7087 in Core


cmelchior:
This PR depends on https://github.com/realm/realm-core/pull/7087 being merged first.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -219,40 +221,45 @@ realm_on_dictionary_change_func_t get_on_dictionary_change() {
...
 
     switch (collection_type) {
-        case RLM_COLLECTION_TYPE_NONE: return realm_object_add_notification_callback(
+        case RLM_COLLECTION_TYPE_NONE:
+            return realm_object_add_notification_callback(
                     reinterpret_cast<realm_object_t*>(collection_ptr),
                     user_data, // Use the callback as user data
                     user_data_free,
-                    NULL, // See https://github.com/realm/realm-kotlin/issues/661
+                    (key_path_array_ptr == 0) ? NULL : reinterpret_cast<realm_key_path_array_t*>(key_path_array_ptr),
```
clementetb, line 241: Is this nullability check required?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/ObjectBoundRealmResults.kt
```diff
@@ -78,5 +78,8 @@ internal class ObjectBoundRealmResults<E : BaseRealmObject>(
  * Binds a flow to an object lifecycle. It allows flows on queries to complete once the object gets
  * deleted. It is used on sub-queries and backlinks.
  */
-internal fun <T> Flow<T>.bind(reference: RealmObjectReference<out BaseRealmObject>): Flow<T> =
-    this.terminateWhen(reference.asFlow()) { it is DeletedObject }
+internal fun <T> Flow<T>.bind(
+    reference: RealmObjectReference<out BaseRealmObject>,
+    keyPaths: List<String>?
```
clementetb, line 83: We could use a null default value on keypaths.

rorbech, line 83: Do we actually need the `keyPaths` here to register if the object is deleted? It sounds a bit strange if the deletion event is not triggered with out a specific path.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
```diff
@@ -113,9 +114,17 @@ internal class ManagedRealmList<E>(
         return operator.set(index, element)
     }
 
-    override fun asFlow(): Flow<ListChange<E>> {
+    override fun asFlow(keyPaths: List<String>?): Flow<ListChange<E>> {
         operator.realmReference.checkClosed()
-        return operator.realmReference.owner.registerObserver(this)
+        val keyPathInfo = keyPaths?.let {
```
rorbech, line 119: Hmm. Maybe worth examining the `contract` approach if we have a lot of these. Or maybe just use the existing 
```
Validation.checkType(... ).let { Pair(it.classkey, keyPaths) }
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/KeyPathFlowable.kt
```diff
@@ -21,7 +21,17 @@ import kotlinx.coroutines.flow.Flow
...
+ * allow you to pass in keypaths to restrict notifications to certain properties and one variant
+ * for flows on entities that doesn't support this, e.g. flows on query aggregates or the Realm as
+ * whole.
  */
+
+internal interface KeyPathFlowable<T> {
+    fun asFlow(keyPaths: List<String>? = null): Flow<T>
+}
+
 internal interface Flowable<T> {
```
rorbech, line 35: If we don't really have a connection of 1-1 connection between `Flowable` to `FlowableTests` then I guess `Flowable` could just go since only implemented by `RealmImpl` now, right?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/QueryTests.kt
```diff
@@ -2946,6 +2946,86 @@ class QueryTests {
...
+            channel.receiveOrFail().let { resultsChange ->
+                assertIs<UpdatedResults<*>>(resultsChange)
+                assertEquals(1, resultsChange.list.size)
+            }
+            realm.writeBlocking {
+                // Should not trigger notification
+                findLatest(obj)!!.intField = 42
+            }
+            realm.writeBlocking {
+                // Should not trigger notification
```
rorbech, line 2977:
```suggestion
                // Should trigger notification
```

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/BacklinksNotificationsTests.kt
```diff
@@ -408,4 +409,176 @@ class BacklinksNotificationsTests : RealmEntityNotificationTests {
...
+        val observer = async {
+            result.asFlow(listOf("objectBacklinks.intField")).collect {
+                c.trySend(it)
+            }
+        }
+        assertIs<InitialResults<Sample>>(c.receiveOrFail())
+        realm.write {
+            // Update field that should not trigger a notification
+            findLatest(result.first())!!.booleanField = false
+        }
```
rorbech, line 484: We should somehow assert that nothing happens. The channel result will be dropped on new events since the channel only have size 1, so theoretically this could emit a notification but the test could still succeed.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/BacklinksNotificationsTests.kt
```diff
@@ -408,4 +409,176 @@ class BacklinksNotificationsTests : RealmEntityNotificationTests {
...
+                }
+                else -> fail("Unexpected change: $resultsChange")
+            }
+        }
+        observer.cancel()
+        c.close()
+    }
+
+    @Test
+    @Ignore // Already covered by RealmResultsNotificationTests
```
rorbech, line 568: Shouldn't we test that we actually resolve the keypath the right place, and since that place is different how can we have coverage of the in `RealmResultsNotificationTests`? Similarly for the methods below.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmDictionaryNotificationsTests.kt
```diff
@@ -428,4 +431,192 @@ class RealmDictionaryNotificationsTests : RealmEntityNotificationTests {
...
+        val observer = async {
+            dict.asFlow(listOf("stringField")).collect {
+                c.trySend(it)
+            }
+        }
+        assertIs<InitialMap<String, RealmDictionaryContainer>>(c.receiveOrFail())
+        realm.write {
+            // Update field that should not trigger a notification
+            findLatest(dict.values.first()!!)!!.id = 42
+        }
```
rorbech, line 456: Again, we need some validation that we actually don't emit this and discard it.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmListNotificationsTests.kt
```diff
@@ -469,6 +472,190 @@ class RealmListNotificationsTests : RealmEntityNotificationTests {
...
+        val observer = async {
+            list.asFlow(listOf("stringField")).collect {
+                c.trySend(it)
+            }
+        }
+        assertIs<InitialList<RealmListContainer>>(c.receiveOrFail())
+        realm.write {
+            // Update field that should not trigger a notification
+            findLatest(list.first())!!.id = 42
+        }
```
rorbech, line 498: Event could just be overwritten and discarded.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/utils/FlowableTests.kt
```diff
@@ -49,6 +49,24 @@ interface FlowableTests {
...
+    fun keyPath_topLevelProperty()
+
+    @Test
+    fun keyPath_nestedProperty()
+
+    @Test
+    fun keyPath_propertyBelowDefaultLimit()
+
+    @Test
+    fun keyPath_unknownTopLevelProperty()
```
rorbech, line 62: What is the difference between these three `unknownTopLevelProperty`, `unknownNestedProperty` and `invalidNestedProperty`? Aren't they conceptually the same?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/utils/FlowableTests.kt
```diff
@@ -49,6 +49,24 @@ interface FlowableTests {
     @Test
     fun closingRealmDoesNotCancelFlows()
 
+    @Test
+    fun keyPath_topLevelProperty()
```
rorbech, line 53: What is the difference between `topLevelProperty` and `nestedProperty`? Aren't they conceptually the same?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/utils/FlowableTests.kt
```diff
@@ -49,6 +49,24 @@ interface FlowableTests {
...
+
+    @Test
+    fun keyPath_nestedProperty()
+
+    @Test
+    fun keyPath_propertyBelowDefaultLimit()
```
rorbech, line 59: I guess we should/could also have a test that asserts that we don't receive events below the default level if we don't register a key path. Just to ensure that we track changed behavior if it ever were to change.

rorbech (commented):
Really nice. I have a number of comments around internals but most notably around tests.

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -56,6 +56,7 @@ const val UUID_BYTES_SIZE = 16
 interface CapiT
 interface RealmConfigT : CapiT
 interface RealmSchemaT : CapiT
+interface RealmObjectSchemaT : CapiT
```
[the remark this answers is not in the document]
cmelchior: Ups, leftover. Will remove

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmNotificationsTests.kt
```diff
@@ -189,13 +189,49 @@ class RealmNotificationsTests : FlowableTests {
...
     }
 
     @Test
     @Ignore
     override fun closingRealmDoesNotCancelFlows() {
-        TODO("Wait for a Global change listener to become available")
+        // Keypaths not supported by Realm notifications
+    }
+
+    @Test
```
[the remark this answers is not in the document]
cmelchior: Hmm, good point. I will refactor that.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/utils/FlowableTests.kt
rorbech: Maybe related to properties that actually exists but does not support it? Ex. decending into a `RealmAny` fields, etc.

cmelchior: Only on the surface, but there are subtle differences. It came from when I was working on the implementation in Core.

1. At the topLevel `invalidProperty` doesn't really exist as we can listen to changes on all properties
2. At the nested level `unknownProperty` means that only the leaf node is not found, e.g. with the path `parent.child.prop` `parent.child` is a valid and `child` is a an Object/Collection with properties, only `prop` is not found on the class defined by `child`.
3. For invalidProperties, and the same path `parent.child.prop`. Here it is `child` that is invalid if it is e.g. a primitive type that doesn't have properties by itself. That also makes `prop` invalid but for a different reason than the 2nd case.

Maybe the name of the tests can be improved 🤔 

You can see the Core implementation here with the two different cases resulting in different errors:

https://github.com/realm/realm-core/pull/7087/files#diff-5e9c3010cf9e06dc7e87de9ac08da8359d523716d926dd8b44bc041e6bb49921R91

https://github.com/realm/realm-core/pull/7087/files#diff-5e9c3010cf9e06dc7e87de9ac08da8359d523716d926dd8b44bc041e6bb49921R114

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/utils/FlowableTests.kt
cmelchior: See https://github.com/realm/realm-kotlin/pull/1547#discussion_r1400692045

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmElementQuery.kt
```diff
@@ -60,8 +60,12 @@ public interface RealmElementQuery<T : BaseRealmObject> : Deleteable {
      *
      * **It is not allowed to call [asFlow] on queries generated from a [MutableRealm].**
      *
-     * @param keyPath TODO
+     * @param keyPaths An optional list of properties that defines when a change to the object will
+     * result in a change being emitted. Nested properties can be defined using a dotted
+     * syntax, e.g. `parent.child.name`. If no keypaths are provided, changes to all top-level
+     * properties and nested properties 4 levels down will trigger a change.
```
clementetb, line 66:
```suggestion
     * properties and nested properties up to 4 levels down will trigger a change.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmSingleQuery.kt
```diff
@@ -72,9 +72,13 @@ public interface RealmSingleQuery<T : BaseRealmObject> : Deleteable {
      * the elements in a timely manner the coroutine scope will be cancelled with a
      * [CancellationException].
      *
-     * @param keyPaths TODO
+     * @param keyPaths An optional list of properties that defines when a change to the object will
+     * result in a change being emitted. Nested properties can be defined using a dotted
+     * syntex, e.g. `parent.child.name`. If no keypaths are provided, changes to all top-level
+     * properties and nested properties 4 levels down will trigger a change.
```
clementetb, line 78:
```suggestion
     * properties and nested properties up to 4 levels down will trigger a change.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmList.kt
```diff
@@ -72,9 +72,14 @@ public interface RealmList<E> : MutableList<E>, Deleteable {
      * the elements in a timely manner the coroutine scope will be cancelled with a
      * [CancellationException].
      *
+     * @param keyPaths An optional list of properties that defines when a change to the object will
+     * result in a change being emitted. Nested properties can be defined using a dotted
+     * syntax, e.g. `parent.child.name`. If no keypaths are provided, changes to all top-level
+     * properties and nested properties 4 levels down will trigger a change.
```
clementetb, line 78:
```suggestion
     * properties and nested properties up to 4 levels down will trigger a change.
```

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/KeyPathFlowable.kt
```diff
@@ -21,7 +21,17 @@ import kotlinx.coroutines.flow.Flow
...
  * [Flow].
+ *
+ * This comes in two variants: One variant for flows on things like Objects, Results and Lists that
+ * allow you to pass in keypaths to restrict notifications to certain properties and one variant
+ * for flows on entities that doesn't support this, e.g. flows on query aggregates or the Realm as
+ * whole.
  */
+
+internal interface KeyPathFlowable<T> {
+    fun asFlow(keyPaths: List<String>? = null): Flow<T>
```
clementetb, line 32: Is there any reason for using `List` and not `vararg`?

packages/jni-swig-stub/src/main/jni/realm_api_helpers.h
```diff
@@ -137,6 +143,12 @@ realm_sync_thread_error(realm_userdata_t userdata, const char* error);
 realm_scheduler_t*
 realm_create_generic_scheduler();
 
+realm_key_path_array_t*
```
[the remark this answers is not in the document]
cmelchior: Yes, this was needed before the C-API was modified, but now we can just use it directly. I will remove it.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -1060,6 +1067,22 @@ realm_create_generic_scheduler() {
     return new realm_scheduler_t { realm::util::Scheduler::make_dummy() };
 }
 
+realm_key_path_array_t*
```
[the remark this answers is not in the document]
cmelchior: See https://github.com/realm/realm-kotlin/pull/1547#discussion_r1401659516

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/utils/FlowableTests.kt
```diff
@@ -49,6 +49,24 @@ interface FlowableTests {
     @Test
     fun closingRealmDoesNotCancelFlows()
 
+    @Test
```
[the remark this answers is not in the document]
cmelchior: Yes, these was moved into a `KeyPathFlowable` interface. QueryTests are somewhat of a special case that I don't see this interface covering easily. As the current test class covers both Results and ObjectBound tests. That is maybe an indication that we should split `QueryTests` into multiple classes, but we should leave that to #713 IMO.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/query/ScalarQuery.kt
```diff
@@ -96,7 +96,8 @@ internal class CountQuery<E : BaseRealmObject> constructor(
     override fun asFlow(): Flow<Long> {
         realmReference.checkClosed()
         return realmReference.owner
-            .registerObserver(this)
+            // TODO
```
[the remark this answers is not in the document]
cmelchior: Leftover. It has been removed

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/KeyPathFlowable.kt
cmelchior: Yes, good point.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/KeyPathFlowable.kt
cmelchior: This is the internal interface that just mirrors the public API which uses `List`. I was considering whether to use `vararg` or `List` in the public API but ended up with List as it makes it easier to extend the API later. I.e. there is the potential that projections will end up adding arguments to the `asFlow` method as well.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmList.kt
```diff
@@ -72,9 +72,14 @@ public interface RealmList<E> : MutableList<E>, Deleteable {
      * the elements in a timely manner the coroutine scope will be cancelled with a
      * [CancellationException].
      *
+     * @param keyPaths An optional list of properties that defines when a change to the object will
```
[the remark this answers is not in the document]
cmelchior: I expanded the documentation. Hopefully it is more clear now.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/query/RealmResults.kt
```diff
@@ -69,7 +69,8 @@ public interface RealmResults<T : BaseRealmObject> : List<T>, Deleteable, Versio
      * the elements in a timely manner the coroutine scope will be cancelled with a
      * [CancellationException].
      *
+     * @param keyPath TODO
```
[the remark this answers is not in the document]
cmelchior: Added the same docs as for lists, slightly reworded

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmSet.kt
```diff
@@ -52,9 +52,14 @@ public interface RealmSet<E> : MutableSet<E>, Deleteable {
      * the elements in a timely manner the coroutine scope will be cancelled with a
      * [CancellationException].
      *
+     * @param keyPaths An optional list of properties that defines when a change to the object will
```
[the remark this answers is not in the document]
cmelchior: I expanded the documentation. Hopefully it is more clear now.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmSetInternal.kt
```diff
@@ -159,8 +160,16 @@ internal class ManagedRealmSet<E> constructor(
         }
     }
 
-    override fun asFlow(): Flow<SetChange<E>> {
-        return operator.realmReference.owner.registerObserver(this)
+    override fun asFlow(keyPaths: List<String>?): Flow<SetChange<E>> {
+        val keyPathInfo = keyPaths?.let {
+            val classKey = if (operator is RealmObjectSetOperator) {
```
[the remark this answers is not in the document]
cmelchior: Forgot about contracts. Pretty neat way to make this work. I added a helper method for this and refactored the relevant places.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmListInternal.kt
cmelchior: I used the approach outlined here: https://github.com/realm/realm-kotlin/pull/1547#discussion_r1401890531

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/ObjectBoundRealmResults.kt
cmelchior: Good point. Removing it.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/BacklinksNotificationsTests.kt
cmelchior: Good point, I added an assert catching this case.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -143,7 +143,9 @@ bool migration_callback(void *userdata, realm_t *old_realm, realm_t *new_realm,
 // TODO OPTIMIZE Abstract pattern for all notification registrations for collections that receives
 //  changes as realm_collection_changes_t.
 realm_notification_token_t *
-register_results_notification_cb(realm_results_t *results, jobject callback) {
+register_results_notification_cb(realm_results_t *results,
+                                 int64_t key_path_array_ptr,
```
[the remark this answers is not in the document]
cmelchior: We cannot in this case, as this is methods being called from the JVM which doesn't have that concept. We are using `0` in that case

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
cmelchior: Strictly no, but it makes it easier to understand what is going on IMO

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/ObjectBoundRealmResults.kt
cmelchior: After reviewing Clauses comments, this was removed altogether.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmResultsNotificationsTests.kt
```diff
@@ -387,4 +390,223 @@ class RealmResultsNotificationsTests : FlowableTests {
...
+                    this.intField = 1
+                    this.stringField = "parent"
+                    this.nullableObject = Sample().apply {
+                        this.stringField = "child"
+                        this.nullableObject = Sample().apply {
+                            this.stringField = "child-child"
+                            this.nullableObject = Sample().apply {
+                                this.stringField = "child-child-child"
+                                this.nullableObject = Sample().apply {
+                                    this.stringField = "child-child-child"
```
nhachicha, line 482:
```suggestion
                                    this.stringField = "child-child-child-child"
```

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmResultsNotificationsTests.kt
```diff
@@ -387,4 +390,223 @@ class RealmResultsNotificationsTests : FlowableTests {
...
```
nhachicha, line 542:
```suggestion
                                    this.stringField = "child-child-child-child"
```

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmSetNotificationsTests.kt
```diff
@@ -342,4 +345,267 @@ class RealmSetNotificationsTests : RealmEntityNotificationTests {
...
+            // Update field that should not trigger a notification
+            findLatest(set.first())!!.id = 42
+        }
+        realm.write {
+            // Update field that should trigger a notification
+            findLatest(set.first())!!.stringField = "Foo"
+        }
+        c.receiveOrFail().let { listChange ->
+            assertIs<UpdatedSet<RealmSetContainer>>(listChange)
+            when (listChange) {
```
nhachicha, line 379: Where's the update to `stringField`?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmSetNotificationsTests.kt
```diff
@@ -342,4 +345,267 @@ class RealmSetNotificationsTests : RealmEntityNotificationTests {
...
+            // Update field that should not trigger a notification
+            findLatest(set.first())!!.id = 1
+        }
+        realm.write {
+            // Update field that should trigger a notification
+            findLatest(set.first())!!.objectSetField.first().stringField = "Bar"
+        }
+        c.receiveOrFail().let { setChange ->
+            assertIs<UpdatedSet<RealmSetContainer>>(setChange)
+            when (setChange) {
```
nhachicha, line 426: where's the update to `objectSetField` ?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmSetNotificationsTests.kt
```diff
@@ -342,4 +345,267 @@ class RealmSetNotificationsTests : RealmEntityNotificationTests {
...
+            findLatest(list.first())!!.stringField = "Parent change"
+        }
+        realm.write {
+            // Update field that should trigger a notification
+            val obj = findLatest(list.first())!!.objectSetField.first().objectSetField.first().objectSetField.first().objectSetField.first().objectSetField.first()
+            obj.stringField = "Bar"
+        }
+        c.receiveOrFail().let { listChange ->
+            assertIs<UpdatedSet<RealmSetContainer>>(listChange)
+            when (listChange) {
```
nhachicha, line 572: where's the update to `stringField`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/types/RealmSet.kt
```diff
@@ -52,9 +52,16 @@ public interface RealmSet<E> : MutableSet<E>, Deleteable {
      * the elements in a timely manner the coroutine scope will be cancelled with a
      * [CancellationException].
      *
+     * @param keyPaths An optional list of model class properties that defines when a change to
+     * objects inside the set will result in a change being emitted. Nested properties can be
+     * defined using a dotted syntax, e.g. `parent.child.name`. If no keypaths are provided, changes
+     * to all top-level properties and nested properties up to 4 levels down will trigger a change.
+     * Keypaths are only supported for sets of objects.
```
nhachicha, line 59: why only supported for sets of objects? it's also supported for List, RealmResults etc...

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/query/ScalarQuery.kt
```diff
@@ -216,7 +216,8 @@ internal class SumQuery<E : BaseRealmObject, T : Any> constructor(
     override fun asFlow(): Flow<T> {
         realmReference.checkClosed()
         return realmReference.owner
-            .registerObserver(this)
+            // TODO
```
nhachicha, line 219: left over?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmMapInternal.kt
```diff
@@ -61,7 +63,7 @@ internal abstract class ManagedRealmMap<K, V> constructor(
     internal val parent: RealmObjectReference<*>,
     internal val nativePointer: RealmMapPointer,
     val operator: MapOperator<K, V>
-) : AbstractMutableMap<K, V>(), RealmMap<K, V>, CoreNotifiable<ManagedRealmMap<K, V>, MapChange<K, V>>, Flowable<MapChange<K, V>> {
```
nhachicha, line 64: why removing `Flowable` interface?

packages/jni-swig-stub/realm.i
```diff
@@ -397,6 +398,66 @@ import static io.realm.kotlin.internal.interop.realm_errno_e.*;
...
+
+/* This allows a C function to return a char ** as a Java String array */
+%typemap(out) char ** {
+    int i;
+    int len=0;
+    jstring temp_string;
+    const jclass clazz = jenv->FindClass("java/lang/String");
+
+    while ($1[len]) len++;
+    jresult = jenv->NewObjectArray(len, clazz, NULL);
```
nhachicha, line 440: check for OOM

nhachicha (approved):
Nice 🎉 minor comments 👍

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmMapInternal.kt
cmelchior: It was an internal interface that matched our public API. But with the KeyPath changes that is no longer the case and is actually harmful instead now as Notifier cannot use it to handle all flows the same.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/BacklinksNotificationsTests.kt
```diff
@@ -408,4 +409,242 @@ class BacklinksNotificationsTests : RealmEntityNotificationTests {
...
+                c.trySend(it)
+            }
+        }
+        assertIs<InitialResults<Sample>>(c.receiveOrFail())
+        realm.write {
+            // Update field that should not trigger a notification
+            findLatest(results.first())!!.booleanField = true
+        }
+        realm.write {
+            // Update field that should trigger a notification
```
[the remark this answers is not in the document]
cmelchior: Yes, because I used the keypath `listOf("objectBacklinks.objectBacklinks.objectBacklinks.objectBacklinks.objectBacklinks.stringField")` which only consider changes through the backlinks as things that should trigger a notification

rorbech (approved):
Some minor comments about nested `runBlocking`, etc.. Just wanted to highlight it, but don't think we need to bother fixing it and rerunning CI ... unless there is other reasons for doing so. 

:shipit:

PR #1550, opened 2023-10-17
Fix new temp dirs not different on native

clementetb:
Creating two temporary directories in native with the same prefix value via `PlatformUtils.createTempDir` would result in the same directory, JVM would result in different directories.

This incongruence might cause issues while writing test cases.

### DIFF
--- a/packages/test-base/src/nativeDarwin/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
+++ b/packages/test-base/src/nativeDarwin/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
@@ -16,6 +16,7 @@
 
 package io.realm.kotlin.test.platform
 
+import io.realm.kotlin.test.util.Utils
 import kotlinx.cinterop.ULongVar
 import kotlinx.cinterop.alloc
 import kotlinx.cinterop.cValue
@@ -35,7 +36,8 @@ import kotlin.time.Duration
 actual object PlatformUtils {
     actual fun createTempDir(prefix: String, readOnly: Boolean): String {
         // X is a special char which will be replace by mkdtemp template
-        val mask = prefix.replace('X', 'Z', ignoreCase = true)
+        val suffix = "-${Utils.createRandomString(4)}"
+        val mask = prefix.plus(suffix).replace('X', 'Z', ignoreCase = true)
         val path = "${platform.Foundation.NSTemporaryDirectory()}$mask"
         platform.posix.mkdtemp(path.cstr)
         if (readOnly) {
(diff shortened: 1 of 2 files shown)


packages/test-base/src/nativeDarwin/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
```diff
@@ -35,7 +36,8 @@ import kotlin.time.Duration
 actual object PlatformUtils {
     actual fun createTempDir(prefix: String, readOnly: Boolean): String {
         // X is a special char which will be replace by mkdtemp template
-        val mask = prefix.replace('X', 'Z', ignoreCase = true)
+        val suffix = "-${Utils.createRandomString(4)}"
```
rorbech, line 39: Any reason not to just add an `xxxxxx`-suffix (after replacing X->Z) and rely on `mkdtemp`?

packages/test-base/src/nativeDarwin/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
```diff
@@ -34,10 +34,12 @@ import kotlin.time.Duration
...
     actual fun createTempDir(prefix: String, readOnly: Boolean): String {
-        // X is a special char which will be replace by mkdtemp template
-        val mask = prefix.replace('X', 'Z', ignoreCase = true)
-        val path = "${platform.Foundation.NSTemporaryDirectory()}$mask"
-        platform.posix.mkdtemp(path.cstr)
+        // Currently we cannot template using platform.posix.mkdtemp
+        // the return value is not of use.
+        val suffix = "-${Utils.createRandomString(6)}"
+        val path = "${platform.Foundation.NSTemporaryDirectory()}$prefix$suffix"
+        platform.posix.mkdir(path, 448.toUShort())
```
nhachicha, line 41: maybe using the const like in https://github.com/realm/realm-kotlin/pull/1550/files#diff-0e15d894eafc8be3bb7b226a9802e7a2ec938f153158929d9377c6b5635316e1R44 is more readable than `448`

nhachicha (approved):
👍

packages/test-base/src/nativeDarwin/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
clementetb: I wasn't able to make `mkdtemp` work. It returns a pointer to the new path, but I cannot retrieve the actual data.

I have moved the implementation to use `mkdir` with our template.

cmelchior (approved):

PR #1552, opened 2023-10-18
Workaround for client reset flaky tests

clementetb:
Workaround to prevent crashes while testing sync client reset integration tests.

There is a race condition where the sync client might try to use a scheduler that has been released causing a null pointer exception in Core.


rorbech (approved):

rorbech:
This PR is superseded by the actual fix in https://github.com/realm/realm-kotlin/pull/1559

Issue #665, opened 2022-02-04
Streamline exceptions from C++


cmelchior:
I was testing exceptions in some callbacks in JNI.

We have this code in `throw_as_java_exception`:

```
jclass error_type_class = (jenv)->FindClass("io/realm/internal/interop/CoreErrorUtils");
```

Which will crash in `FindClass` if you have a pending JavaException already, overriding the initial exception (bad!).

I also noticed we have a lot of `throw std::runtime_error("An unexpected Error was thrown from Java. See LogCat");` in our code.

This means you have to dig through the logs in order to find the correct message.

The most problematic cases are the ones where C++ calls back through JNI, possible multiple times. In those cases we want C++ to abort as quickly as possible while still maintaining the original error context.

We should look to see if it is possible to somehow get the best of both worlds, which I think would be:

1) Detect that a Java exception happened
2) Collect stacktrace.
3) Wrap it in a C++ exception we can throw which will cancel any C++ code.
4) Before returning to Java, we unpack the exception and its metadata and convert it back to a Java exception.

clementetb:
We should also rethink how we handle JVM exceptions from native. The current behaviour is to run `jni_check_exception` right after executing the method. This function will silently clear any pending exceptions, preventing them from reaching the user or the logs. It will prevent the app from hard crashing.

Because of the separation between the user and SDK code, a crash during the execution of such methods means an issue at the SDK/OS level, so once we receive such exceptions we should stop operating with that Realm instance as it might be in an illegal state. 

On the other hand, Realm-kotlin on Darwin platform will hard-crash if it encounters an exception when invoking SDK code from core.

We should align the behaviour with Darwin, and even consider some way to capture and propagate these exceptions to user code to allow users to gracefully exit these situations.

PR #1560, opened 2023-11-02
Delete local references

clementetb:
There were some instances where a sync app would crash after enabling log-level `Trace` or `All`.

The cause is that objects instantiated in JNI were not reclaimed and could cause an Out Of Memory exception or a local reference table overflow.

This PR attempts to release any local instances by using the local frames.

Testing was done locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 * Fix craches caused by posting to a released scheduler. (Issue [#1543](https://github.com/realm/realm-kotlin/issues/1543))
-* [Sync] Fix crash while syncing data when LogLevel was set to `Trace` or `All`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
+* [Sync] Fix crash when syncing data if the log level was set to `LogLevel.TRACE` or `LogLevel.ALL`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Fixed
 * Fix craches caused by posting to a released scheduler. (Issue [#1543](https://github.com/realm/realm-kotlin/issues/1543))
 * Fix NPE when applying query aggregators on classes annotated with `@PersistedName`. (Issue [1569](https://github.com/realm/realm-kotlin/pull/1569))
-* [Sync] Fix crash when syncing data if the log level was set to `LogLevel.TRACE` or `LogLevel.ALL`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
+* [Sync] Fix crash when syncing data if the log level was set to `LogLevel.TRACE` or `LogLevel.ALL`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560)) 
 
 ### Compatibility
 * File format: Generates Realms with file format v23.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,29 @@
+## 1.12.1 (YYYY-MM-DD)
+
+### Breaking Changes
+* None.
+
+### Enhancements
+* None
+
+### Fixed
+* [Sync] Fix crash while syncing data when LogLevel was set to `Trace` or `All`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
+
+### Compatibility
+* File format: Generates Realms with file format v23.
+* Realm Studio 13.0.0 or above is required to open Realms created by this version.
+* This release is compatible with the following Kotlin releases:
+  * Kotlin 1.8.0 and above. The K2 compiler is not supported yet.
+  * Ktor 2.1.2 and above.
+  * Coroutines 1.7.0 and above.
+  * AtomicFu 0.18.3 and above.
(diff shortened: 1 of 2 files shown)


packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -183,6 +196,7 @@ realm_on_object_change_func_t get_on_object_change() {
         jenv->CallVoidMethod(static_cast<jobject>(userdata),
                              on_change_method,
                              reinterpret_cast<jlong>(changes));
+        jni_check_exception(jenv);
```
cmelchior, line 199: I don't think we have any tests for these, but then again, I suspect it will also be very tricky do to since it is happening inside the Notifier?

cmelchior (approved):

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
clementetb: Yes, it would be tricky to test that we catch this, to do so, there should be an issue at [object change callback](https://github.com/realm/realm-kotlin/blob/be75e66560737e6b4b13be4941f9ac3f09e02747/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/SuspendableNotifier.kt#L121) with the realm/changeflow.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -42,6 +54,8 @@ inline jobject create_java_exception(JNIEnv *jenv, realm_error_t error) {
                                               error_type_class,
                                               "asThrowable",
                                               "(IILjava/lang/String;Ljava/lang/String;Ljava/lang/Throwable;)Ljava/lang/Throwable;", true);
+
+    jenv->PushLocalFrame(3);
```
nhachicha, line 58: How's this approach different from calling `DeleteLocalRef` on each local reference? 
also we need to check for OOM for each `PushLocalFrame` call

clementetb: Added OOM checks.

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -34,6 +34,26 @@ jobject wrap_pointer(JNIEnv* jenv, jlong pointer, jboolean managed = false) {
...
+            realm_register_user_code_callback_error(jenv->NewGlobalRef(exception));
+        return false;
+    }
+    return true;
+}
+
+inline void push_local_frame(JNIEnv *jenv, jint frame_size) {
+    if (jenv->PushLocalFrame(frame_size) != 0) {
+        jni_check_exception(jenv);
+        jenv->ExceptionDescribe();
```
nhachicha, line 52: The exception is already cleared in https://github.com/realm/realm-kotlin/pull/1560/files#diff-c71ff0791782da90014db6f7000bdeea6c883ca9729f17bd1b11041d1e7711faR40 so this won't print anything to stderr? 
I think instead of calling `jenv->ExceptionClear();` inside `jni_check_exception` we could call `jenv->ExceptionDescribe();` which has the side effect of also clearing the exception

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
```diff
@@ -34,6 +34,26 @@ jobject wrap_pointer(JNIEnv* jenv, jlong pointer, jboolean managed = false) {
                            managed);
 }
 
+inline jboolean jni_check_exception(JNIEnv *jenv = get_env(), bool registrable_callback_error = false) {
+    if (jenv->ExceptionCheck()) {
+        jthrowable exception = jenv->ExceptionOccurred();
+        jenv->ExceptionClear();
+        // setting the user code error is only propagated on certain callbacks.
+        if (registrable_callback_error)
```
nhachicha, line 42: I think we should link to https://github.com/realm/realm-kotlin/issues/665 to throw the exception if we don't want to rely on the `registrable_callback_error` mechanism, otherwise this method calls acts as catch and swallow ...

packages/jni-swig-stub/src/main/jni/realm_api_helpers.cpp
clementetb: Yes, it is my bad. It should be inside.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 ### Fixed
 * Fix craches caused by posting to a released scheduler. (Issue [#1543](https://github.com/realm/realm-kotlin/issues/1543))
 * Fix NPE when applying query aggregators on classes annotated with `@PersistedName`. (Issue [1569](https://github.com/realm/realm-kotlin/pull/1569))
-* [Sync] Fix crash when syncing data if the log level was set to `LogLevel.TRACE` or `LogLevel.ALL`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
+* [Sync] Fix crash when syncing data if the log level was set to `LogLevel.TRACE` or `LogLevel.ALL`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
```
nhachicha, line 12:
```suggestion
* [Sync] Fix crash when syncing data if the log level was set to `LogLevel.TRACE` or `LogLevel.ALL`. (Issue [#1560](https://github.com/realm/realm-kotlin/pull/1560))
```

nhachicha (approved):
LGTM 👍

Issue #1555, opened 2023-10-23
Querying min/max of entity with class name not matching @PersistedName value causes NPE


DamianJaeger:
### How frequently does the bug occur?

Always

### Description

Entity:
```
@PersistedName("DBFeedPostEntity")
class FeedPostEntity constructor(
    @PrimaryKey @PersistedName("_id") var id: ObjectId = ObjectId(),
    ...
    @PersistedName("feed") var feed: FeedEntity? = null,
    @PersistedName("orderIndex") var orderIndex: Long = -1,
) : RealmObject {
    constructor() : this(id = ObjectId())
}
```

Query:
```
query(FeedPostEntity::class,
    query = "feed._id == $0",
    ObjectId(feedId)
).min<Long>("orderIndex").find()
```

This causes a NPE. Note that the class name and the value of @PersistedName do not match. By renaming the class to `DBFeedPostEntity`, the NPE can be avoided and everything works as expected. 

The behavior is the same for `max` queries.

### Stacktrace & log output

```shell
java.lang.NullPointerException
	at io.realm.kotlin.internal.query.ObjectQuery.min(ObjectQuery.kt:139)
        [...]
        at io.realm.kotlin.internal.SuspendableWriter$write$2.invokeSuspend(SuspendableWriter.kt:110)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
	at java.lang.Thread.run(Thread.java:1012)
```

### Can you reproduce the bug?

Always

### Version

1.10.2

### What Atlas App Services are you using?

Both Atlas Device Sync and Atlas App Services

### Are you using encryption?

No

### Platform OS and version(s)

Android 13

### Build environment

Android Studio version: 2022.3.1
Android Build Tools version: 34
Gradle version: 8.1.2

PR #1569, opened 2023-11-13
Allow aggregators on properties annotated with @PersistedName

clementetb:
Closes https://github.com/realm/realm-kotlin/issues/1555

Aggregators on properties annotated with @PersistedName triggered an NPE. The cause was that we used the public names to retrieve the schema metadata when they were only accessible via the persisted one.

This PR fixes the issue by retrieving the object schema via the classkey, and then allowing retrieving properties via both persisted and public names.


cmelchior (approved):
Nice 👍

PR #1571, opened 2023-11-14
DynamicObjects can now access fields by public name.

clementetb:
Fixes CI

### DIFF
--- a/packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/PersistedNameTests.kt
+++ b/packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/PersistedNameTests.kt
@@ -232,6 +232,8 @@ class PersistedNameTests {
 
         assertNotNull(dynamicSample)
         assertEquals("Realm", dynamicSample.getValue("persistedNameStringField"))
+        // We can access property via the public name because the dynamic Realm is build upon a typed
+        // Realm via the extension function `asDynamicRealm`.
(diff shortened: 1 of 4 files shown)


rorbech (commented):
This doesn't look right. How can we know the `@PersistedName` mapping when there is not necessarily a schema definition?

clementetb:
Then it won't provide access to a public name because there won't be any. 

Right now this behavior is a by-product of building the dynamic realm out of a typed realm config. It has a schema definition, and thus has public names available.

I don't think this should be an issue as it only affects an internal function, and even if it was public we would allow users to access data using a public name if required.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/PersistedNameTests.kt
```diff
@@ -232,9 +232,7 @@ class PersistedNameTests {
 
         assertNotNull(dynamicSample)
         assertEquals("Realm", dynamicSample.getValue("persistedNameStringField"))
-        assertFailsWithMessage<IllegalArgumentException>("Schema for type 'AlternativePersistedNameSample' doesn't contain a property named 'publicNameStringField'") {
-            dynamicSample.getValue("publicNameStringField")
-        }
+        assertEquals("Realm", dynamicSample.getValue("publicNameStringField"))
```
rorbech, line 235: It would probably be appropriate with a comment here that this is only due to the native of `asDynamicRealm` that the public name is available here.

rorbech (approved):
Great. One minor comment.

Issue #1564, opened 2023-11-08
RealmInstant.now() requires Android SDK 26 minSDK is 16


nhachicha:
### How frequently does the bug occur?

Always

### Description

Reported here https://kotlinlang.slack.com/archives/C6J79CSDP/p1699445259140619
We need to be able to catch this kind of API usage via the `min-android-sample` 

### Can you reproduce the bug?

Always

### Version

v1.12

### What Atlas App Services are you using?

Local Database only

### Are you using encryption?

No

### Platform OS and version(s)

all

PR #1572, opened 2023-11-14
RealmInstant.now implementation on Android uses API 26 > minSDK 16

nhachicha:
Fixes https://github.com/realm/realm-kotlin/issues/1564


packages/library-base/src/androidMain/kotlin/io/realm/kotlin/internal/platform/SystemUtilsAndroid.kt
```diff
@@ -36,3 +39,12 @@ public actual fun assetFileAsStream(assetFilename: String): InputStream = try {
 // Returns the default logger for the platform
 public actual fun createDefaultSystemLogger(tag: String, logLevel: LogLevel): RealmLogger =
     LogCatLogger(tag, logLevel)
+
+public actual fun currentTime(): RealmInstant {
+    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
clementetb, line 44: We should document that there is some precision loss depending on the SDK version used.

Issue #1563, opened 2023-11-03
[macos] Crash when trying to clean up scheduler used by multiple realms


rorbech:
Trying to release dispatcher/scheduler used across multiple realms in https://github.com/realm/realm-kotlin/pull/1559/commits/f3e4ce362d56231c77bbf6dffcf8240bf5756fd2#diff-291d3886fa0a21b98dc0ebd5ff0ce3846d78dab19a8d3ccffe6386899f0879a4R79 causes 

```
io.realm.kotlin.test.common.utils.PlatformUtilsTests.createTempDir_createDifferentDirs STANDARD_OUT
    Uncaught Kotlin exception: kotlin.IllegalStateException: Dispatcher multipleSchedulersOnSameThread was closed, attempted to schedule: DispatchedContinuation[MultiWorkerDispatcher@584bb7d0, Continuation @ 0]

io.realm.kotlin.test.darwin.CoroutineTests.dispatchQueueScheduler SKIPPED

io.realm.kotlin.test.darwin.MemoryTests.closeShouldFreeMemory STANDARD_OUT
    NEW_MEMORY_MODEL: true
        at 0   test.kexe                           0x10950b47b        kfun:kotlin.Throwable#<init>(kotlin.String?){} + 107 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:25:37)
        at 1   test.kexe                           0x109504ca7        kfun:kotlin.Exception#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:23:44)
        at 2   test.kexe                           0x109504ee7        kfun:kotlin.RuntimeException#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:34:44)
        at 3   test.kexe                           0x1095055a7        kfun:kotlin.IllegalStateException#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/acafc8c59a79cc1/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:70:44)
[...]

PR #1579, opened 2023-11-21
Fix corner cases in tests discovered after running on Github Actions

cmelchior:
This PR is an attempt to cleanup https://github.com/realm/realm-kotlin/pull/881

- It pulls out a number of smaller fixes to out test suit which resulted in errors when running tests on Windows.
- Improves the CI check to check for both Jenkins and GHA
- Increases the wait period for some tests that timed out when running on GHA

### DIFF
--- a/packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/SystemNotificationTests.kt
+++ b/packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/SystemNotificationTests.kt
@@ -76,5 +76,9 @@ class SystemNotificationTests {
             writer1.write { copyToRealm(Sample()) }
             writer2.write { copyToRealm(Sample()) }
         }
+        writer1.close()
+        writer2.close()
+        liveRealmContext.close()
+        baseRealm.close()
     }
 }
--- a/packages/test-base/src/jvmMain/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
+++ b/packages/test-base/src/jvmMain/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
@@ -16,7 +16,6 @@
 
 package io.realm.kotlin.test.platform
 
-import java.io.File
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
--- a/packages/test-base/src/jvmMain/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
+++ b/packages/test-base/src/jvmMain/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
@@ -19,7 +19,9 @@ package io.realm.kotlin.test.platform
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Path
+import java.nio.file.Paths
 import java.nio.file.attribute.PosixFilePermission
+import java.util.stream.Collectors
 import kotlin.io.path.absolutePathString
 import kotlin.time.Duration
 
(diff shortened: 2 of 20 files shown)


cmelchior:
This PR is now ready for review

packages/test-base/src/jvmMain/kotlin/io/realm/kotlin/test/platform/PlatformUtils.kt
```diff
@@ -40,7 +41,22 @@ actual object PlatformUtils {
     }
 
     actual fun deleteTempDir(path: String) {
-        File(path).deleteRecursively()
+        val rootPath: Path = Paths.get(path)
+        val pathsToDelete: List<Path> =
+            Files.walk(rootPath).sorted(Comparator.reverseOrder()).collect(Collectors.toList())
```
nhachicha, line 46: Is the sort needed to delete files first?

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/jvm/RealmTests.kt
```diff
@@ -81,12 +81,13 @@ class RealmTests {
         // Closing a Realm should also cleanup our default (internal) dispatchers.
         // The core notifier and the finalizer thread will never be closed.
         val expectedThreadCount = initialThreads.size + 1 /* core-notifier */ + if (finalizerRunning) 0 else 1
-        var counter = 5 // Wait 5 seconds for threads to settle
-        while (totalThreadCount() != expectedThreadCount && counter > 0) {
+        var counter = 10 // Wait 5 seconds for threads to settle
```
nhachicha, line 84:
```suggestion
        var counter = 10 // Wait 10 seconds for threads to settle
```

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/jvm/RealmTests.kt
```diff
@@ -81,12 +81,13 @@ class RealmTests {
...
+        while (newThreads().any { !it.isDaemon } && counter > 0) {
             delay(1000)
             counter--
         }
-        assertEquals(expectedThreadCount, totalThreadCount(), "Unexpected thread count after closing realm: ${newThreads()}")
+        val totalThreadCount = totalThreadCount()
```
nhachicha, line 89: For the sake of debugging, we should print the stack trace of all non-daemon thread to find where they're are blocking

nhachicha (approved):
Nice catches 🎣

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/SystemNotificationTests.kt
```diff
@@ -76,5 +76,9 @@ class SystemNotificationTests {
             writer1.write { copyToRealm(Sample()) }
             writer2.write { copyToRealm(Sample()) }
         }
+        writer1.close()
+        writer2.close()
+        liveRealmContext.close()
```
rorbech, line 81: Is this actually green on CI? I tried to clean this up as part of another PR but ended up splitting into this separate issue #1563

rorbech: Hmm. Maybe closing the writers individually first make it work.

cmelchior: Yes, this is working on GHA.

packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/jvm/RealmTests.kt
cmelchior: I thought about this, but it would massively expand the output and so far all the cases we had for this, it was enough to get the name of the thread to figure out what was going on. So for now I think I would prefer to keep as as is.

Issue #1580, opened 2023-11-22
Add support for user error callback propagation during client reset.


clementetb:
A user-level exception might arise during client reset callbacks. Currently, these exceptions are swallowed and logged by the SDK but ideally, they should be propagated to the user with the full stacktrace.

See https://github.com/realm/realm-core/issues/7098 and https://github.com/realm/realm-core/pull/7148.

PR #1581, opened 2023-11-22
Propagate user-level exceptions during client reset

clementetb:
User-level exceptions were not propagated to the users via the exception channels. The exception message was printed on the stderr channel and swallowed.

This PR fixes this issue allowing the user-level exceptions to propagate all the way through the proper channels. Now it would be available as the `ClientResetFailed` throwable cause.

Depends on https://github.com/realm/realm-core/pull/7148
Closes https://github.com/realm/realm-kotlin/issues/1580


cmelchior (approved):
Looks like we still miss to update the changelog with the new Core version being used and CI needs to be green, but otherwise :shipit:

rorbech (commented):
I am not follow the updates to how the global refs on the `error_user_code` are managed. Might be that I have missed something, but will have to check it out to get the insights. Or do you have any magic one liner that can give the overview?

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -158,9 +158,9 @@ private fun throwOnError() {
                 errorCodeNativeValue = error.error.value.toInt(),
                 messageNativeValue = error.message?.toKString(),
                 path = error.path?.toKString(),
-                userError = error.usercode_error?.asStableRef<Throwable>()?.get()
+                userError = error.user_code_error?.asStableRef<Throwable>()?.get()
             ).also {
-                error.usercode_error?.let { disposeUserData<Throwable>(it) }
+                error.user_code_error?.let { disposeUserData<Throwable>(it) }
```
rorbech, line 163: Why are we releasing this just after creating it?

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -638,9 +638,9 @@ actual object RealmInterop {
                             errorCodeNativeValue = err.error.value.toInt(),
                             messageNativeValue = err.message?.toKString(),
                             path = err.path?.toKString(),
-                            userError = err.usercode_error?.asStableRef<Throwable>()?.get()
+                            userError = err.user_code_error?.asStableRef<Throwable>()?.get()
```
rorbech, line 641: Seems weird to create them to just dispose them right after 🤔 But aren't these pointers already guaranteed to be stable refs generated in `stableUserDataWithErrorPropagation`?

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -2534,8 +2534,11 @@ actual object RealmInterop {
                         isFatal = is_fatal,
                         isUnrecognizedByClient = is_unrecognized_by_client,
                         isClientResetRequested = is_client_reset_requested,
-                        compensatingWrites = compensatingWrites
-                    )
+                        compensatingWrites = compensatingWrites,
+                        userError = user_code_error?.asStableRef<Throwable>()?.get()
+                    ).also {
+                        user_code_error?.let { disposeUserData<Throwable>(it) }
```
rorbech, line 2540: Same

rorbech (commented):
> I am not follow the updates to how the global refs on the `error_user_code` are managed. Might be that I have missed something, but will have to check it out to get the insights. Or do you have any magic one liner that can give the overview?

I see the connection now. I was just confused as I though it had something to do with the removed `PopLocalFrame`s. But after settling that these were removed by mistake I get it. 

But still think there is an erroneous `asStableRef` on Darwin.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
clementetb: `asStableRef` does not create the pointer, only wraps a `CPointer` as a stable reference. 

The `StableRef` is created [here](https://github.com/realm/realm-kotlin/blob/5f8e6c7e74c01841caf387b9812280efef7ecd7e/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt#L284), `StableRef.create` actually creates the underlying pointer that is released [here](https://github.com/realm/realm-kotlin/blob/5f8e6c7e74c01841caf387b9812280efef7ecd7e/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt#L161).

PR #1585, opened 2023-11-29
Improve tests using Channels

cmelchior:
This PR streamlines how we are using Channels in tests by introducing a new TestChannel factory method which will ensure we construct the same channel across all tests.

This channel has a few changes compared to the previous one:
- It has an unlimited buffer, which means that `send` should never suspend and we should never accidentally overwrite an event.
- When the channel is canceled, it will check if some elements have not been delivered and throw an assertion if that is true (as it indicates the writer of the test doesn't fully understand the event flow)

I also refactored `Channel.receiveOrFail()`, `onReceive` in channels does not consume elements and we suspect it might be the cause of some of our flaky tests (but hasn't been 100% confirmed).

All uses of `trySend()` in tests has also been replaced by `send()` since `trySend()` might fail silently causing us to miss an error.

### DIFF
--- a/packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmNotificationsTests.kt
+++ b/packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmNotificationsTests.kt
@@ -27,11 +27,11 @@ import io.realm.kotlin.notifications.RealmChange
 import io.realm.kotlin.notifications.UpdatedRealm
 import io.realm.kotlin.test.common.utils.FlowableTests
 import io.realm.kotlin.test.platform.PlatformUtils
+import io.realm.kotlin.test.util.TestChannel
 import io.realm.kotlin.test.util.receiveOrFail
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.async
 import kotlinx.coroutines.channels.BufferOverflow
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.buffer
 import kotlinx.coroutines.sync.Mutex
@@ -73,7 +73,7 @@ class RealmNotificationsTests : FlowableTests {
     @Test
     override fun initialElement() {
         runBlocking {
-            val c = Channel<RealmChange<Realm>>(1)
+            val c = TestChannel<RealmChange<Realm>>()
             val startingVersion = realm.version()
             val observer = async {
                 realm.asFlow().collect {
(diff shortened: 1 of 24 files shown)


packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmNotificationsTests.kt
```diff
@@ -179,7 +179,6 @@ class RealmNotificationsTests : FlowableTests {
                 assertEquals(VersionId(startingVersion.version + 2), realmChange.realm.version())
             }
 
-            realm.write { /* Do nothing */ }
```
nhachicha, line 182: unecessary?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncedRealmTests.kt
```diff
@@ -220,7 +221,10 @@ class SyncedRealmTests {
             partitionValue = partitionValue
         )
         val realm1 = Realm.open(config1)
-        val c = Channel<RealmChange<Realm>>(1)
+        // We don't fully control all events from the server, so do not check the
```
nhachicha, line 224:
```suggestion
        // We don't fully control all events from the server, so do not check 
```

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/nonlatin/NonLatinTests.kt
```diff
@@ -69,12 +69,12 @@ class NonLatinTests {
...
             assertNotNull(oid)
 
-            val channel = Channel<ObjectIdPk>(1)
+            val channel = TestChannel<ObjectIdPk>()
             val job = async {
                 realm.query<ObjectIdPk>("_id = $0", BsonObjectId(oid)).first()
                     .asFlow().collect {
                         if (it.obj != null) {
-                            channel.trySend(it.obj!!)
+                            channel.send(it.obj!!)
```
nhachicha, line 77: Why not use `trySendOrFail`?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmNotificationsTests.kt
cmelchior: No, it will trigger a notification on observer1, so we can verify that observer2 was correctly canceled.

PR #1605, opened 2023-12-13
[CI] Sync tests with Device farm

nhachicha:
This uses BAAS CLI to run sync tests for Android (via Device Farm) JVM, macOS and iOS

- [ ] TODO: Replaying device farm sync test needs to start a new BAAS instance inorder to use the new IP adress
image


.github/actions/run-android-device-farm-sync-test/action.yml
```diff
@@ -0,0 +1,72 @@
+name: 'Run Android tests on Device Farm'
```
cmelchior, line 1: Unless I am missing it, the only difference between this action and `run-android-device-farm-test` seems to be the package name: `io.realm.sync.testapp.test` vs. `io.realm.testapp.test`. Isn't there a way we can combine them?

I guess `app_auxiliary_files` is also a difference, but couldn't we just make that optional?

.github/workflows/pr.yml
```diff
@@ -552,6 +553,28 @@ jobs:
...
+      - name: Install JSON parser
+        run: brew install jq
+
+      # checkout BAAS CLI repo
+      - name: Checkout BAAS repo
+        run: |
+          echo ${{ secrets.BAAS_CLI }} | gh auth login --with-token
+          gh repo clone 10gen/baasaas
+          
+
```
cmelchior, line 565:
```suggestion
```

cmelchior (approved):
Nice... Just restarted the build to make sure it still passes. 

The only comment I have is that this seems to duplicate a lot of code that would be nice to have centralized. But if it is too difficult to combine them, we can also leave that as a future task.

.github/actions/run-android-device-farm-sync-test/action.yml
nhachicha: Updated with common action used for sync and non-sync device farm tests

cmelchior (approved):
Looking good. We can fix the other tests on the main sync branch

cmelchior:
Merging this to unblock merges coming from the other direction

PR #1607, opened 2023-12-15
Add new publishCIPackages target in Gradle to make it easier to control which targets to build on a given platform

cmelchior:
With this change, it is now possible to define which targets to build using. It does this by introducing a new target `publishCIPackages`, which can be configured using a number of new Gradle properties:

```
realm.kotlin.mainHost=true
realm.kotlin.targets=<CommaSeperatedListOfPublications>
realm.kotlin.buildRealmCore=true
realm.kotlin.copyNativeJvmLibs=false
```

With this it is possible to get fine-grained control over exactly what gets built on a CI runner. 

This PR is mostly moving things out of #881, so we can keep the review more focused.

### DIFF
--- a/packages/build.gradle.kts
+++ b/packages/build.gradle.kts
@@ -64,6 +64,7 @@ tasks.register("publishCIPackages") {
         "compilerPlugin",
         "gradlePlugin"
     )
+
     val mainHostTarget: Set<String> = setOf("metadata") // "kotlinMultiplatform"
 
     val isMainHost: Boolean? = if (project.properties.containsKey("realm.kotlin.mainHost"))  {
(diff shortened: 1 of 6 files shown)


packages/build.gradle.kts
```diff
@@ -33,6 +33,136 @@ allprojects {
...
+        ?.filter { it.isNotEmpty() }
+        ?.toSet()
+    userTargets?.forEach {
+        if (!availableTargets.contains(it)) {
+            project.logger.error("Unknown publication: $it")
+            throw IllegalArgumentException("Unknown publication: $it")
+        }
+    }
+    // Configure which platforms publications we do want to publish
+    val wantedTargets: Collection<String> = when (isMainHost) {
```
rorbech, line 88: This logic is not that clear. Wouldn't you also have to filter out `mainHostTarget` from `userTargets` if `isMainHost==false` to be safe.  

Maybe just define a default for `isMainHost` and then have
```
val targets = (userTargets?: availableTargets).let {
   when (isMainHost) {
     true -> it + mainHostTarget
     false -> it - mainHostTarget
  }
}  
``` 
Otherwise maybe add some comments, so that it is clear what we expect.

Issue #1608, opened 2023-12-15
Flaky test: io.realm.kotlin.test.common.notifications.BacklinksNotificationsTests.keyPath_invalidNestedProperty


cmelchior:
This this on MacOs: https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-kotlin/detail/PR-1607/11/pipeline/200

```
io.realm.kotlin.test.common.notifications.BacklinksNotificationsTests.keyPath_invalidNestedProperty STANDARD_OUT

    Uncaught Kotlin exception: kotlin.IllegalStateException: Dispatcher notifier-default.realm was closed, attempted to schedule: DispatchedContinuation[MultiWorkerDispatcher@2e133260, Continuation @ null]

        at 0   test.kexe                           0x104c8a73b        kfun:kotlin.Throwable#<init>(kotlin.String?){} + 107 (/opt/buildAgent/work/f43969c6214a19e7/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:28:37)

        at 1   test.kexe                           0x104c843a7        kfun:kotlin.Exception#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/f43969c6214a19e7/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:23:44)

        at 2   test.kexe                           0x104c84577        kfun:kotlin.RuntimeException#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/f43969c6214a19e7/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:34:44)

        at 3   test.kexe                           0x104c84ab7        kfun:kotlin.IllegalStateException#<init>(kotlin.String?){} + 103 (/opt/buildAgent/work/f43969c6214a19e7/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:70:44)

        at 4   test.kexe                           0x105076e6b        kfun:kotlinx.coroutines.MultiWorkerDispatcher.dispatch#internal + 1131 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt:123:23)

[...]

PR #1559, opened 2023-11-02
Guard notifying outdated scheduler

rorbech:
TODOs:
- [x] Await that core fix is merged https://github.com/realm/realm-core/pull/7102

Closes #1543
Closes #1558
Closes #1561

PR #1611, opened 2023-12-20
Fix various dispatcher issues

cmelchior:
Close https://github.com/realm/realm-kotlin/issues/1608

- This PR works around a race condition with Core notifications and dispatchers which I do not think we can solve otherwise. 
- Fixes a race condition with events in a test

See also https://github.com/Kotlin/kotlinx.coroutines/issues/3993

### DIFF
--- a/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
+++ b/packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
@@ -3603,18 +3603,41 @@ actual object RealmInterop {
         }
 
         override fun notify(work_queue: CPointer<realm_work_queue_t>?) {
-            scope.launch {
-                try {
-                    printlntid("on dispatcher")
-                    lock.withLock {
-                        if (!cancelled) {
-                            realm_wrapper.realm_scheduler_perform_work(work_queue)
+            try {
+                scope.launch {
+                    try {
+                        printlntid("on dispatcher")
+                        lock.withLock {
+                            if (!cancelled) {
+                                realm_wrapper.realm_scheduler_perform_work(work_queue)
+                            }
                         }
+                    } catch (e: Exception) {
+                        // Should never happen, but is included for development to get some indicators
+                        // on errors instead of silent crashes.
+                        e.printStackTrace()
                     }
-                } catch (e: Exception) {
-                    // Should never happen, but is included for development to get some indicators
-                    // on errors instead of silent crashes.
-                    e.printStackTrace()
+                }
+            } catch (ex: IllegalStateException) {
+                // A work-around for https://github.com/realm/realm-kotlin/issues/1608
+                // As the Core listeners are separated from Coroutines, there is a chance
+                // that we have closed the Kotlin dispatcher while Core is in the process
+                // of sending notifications. If this happens we might end up in this
+                // `notify` method with the dispatcher being closed.
+                //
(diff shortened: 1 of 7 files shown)


packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -3603,18 +3603,41 @@ actual object RealmInterop {
...
+                        // on errors instead of silent crashes.
+                        e.printStackTrace()
                     }
-                } catch (e: Exception) {
-                    // Should never happen, but is included for development to get some indicators
-                    // on errors instead of silent crashes.
-                    e.printStackTrace()
+                }
+            } catch (ex: IllegalStateException) {
+                // A work-around for https://github.com/realm/realm-kotlin/issues/1608
```
rorbech, line 3622: I guess we have had some debate on which way to organize the dispatcher/scheduler relationship (https://github.com/realm/realm-kotlin/pull/1559 and https://github.com/realm/realm-kotlin/pull/1463#pullrequestreview-1551509397), but this seems like a very unsafe way of guarding this. We already have infrastructure in place for being able to cancel the `SingleThreadDispatcherScheduler`, so seems way safer to expose the `cancel` to the upper layers, so that we follow the exact logic on when we expect this to happen (when we close the realm?). I guess that wouldn't work if people inject their own dispatcher and just close it, but we could document that they shouldn't close the dispatcher before closing the realm.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/notifications/RealmNotificationsTests.kt
```diff
@@ -108,16 +108,16 @@ class RealmNotificationsTests : FlowableTests {
         realm.write { /* Do nothing */ }
         val nextVersion = realm.version()
 
+        c1.receiveOrFail(message = "Failed to receive update event on Channel 1").let { realmChange ->
```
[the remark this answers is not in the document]
cmelchior: There was a race condition where the write on line 108 resulted in an update also being sent to the second observer, which would crash the test with a non-empty queue at the end. By consuming the update before listening for the 2nd initial event, this cannot happen.

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
cmelchior: I modified the implementation, so we now have two locks instead, one to control the dispatcher and one for the scheduler.

cmelchior:
Ready for a second round @rorbech

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/util/CoroutineDispatcherFactory.kt
```diff
@@ -120,6 +120,10 @@ public class LiveRealmContext(
     }
 
     override fun close() {
+        // Warning: It is important to release the scheduler before closing the
```
rorbech, line 123: I guess this will work if we are guaranteed the `user_data_free` callback from core synchronously from `realm_release(scheduler)`. It would probably be safer for future changes with a separate abstraction of the scheduler that would set `dispatcherClosing` before `realm_release(scheduler)` ... but this is way better that the previous implementation that was relying on the message from the coroutine-exception 🙈

rorbech (approved):

PR #1636, opened 2024-01-18
Passing encryption key via a callback

nhachicha:
This PR adds the ability to use pass in the AES encryption key from a native memory region. After opening the Realm the caller can reset this native memory region to reduce the window where the key is available clear in memory.

- [ ] Changelog 
- [ ] Test will not pass on Android (need to add a helper method in JNI to allocate native memory for the `openEncryptedRealmWithEncryptionKeyCallback` test)

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
@@ -106,6 +106,18 @@ public data class InitialRealmFileConfiguration(
     val checksum: String?
 )
 
+public interface EncryptionKeyCallback {
+    /**
+     * Provides the native memory address of the 64 byte array containing the key used to encrypt and decrypt the Realm file.
+     */
+    public fun keyPointer(): Long
+
+    /**
+     * This callback will be invoked by Realm after it's open. This hint to the user that the key provided in [keyPointer] can now be released.
+     */
+    public fun releaseKey()
+}
+
 /**
  * Base configuration options shared between all realm configuration types.
  */
@@ -152,6 +164,13 @@ public interface Configuration {
      * @return null on unencrypted Realms.
      */
     public val encryptionKey: ByteArray?
+
+    /**
+     * Native memory address of the 64 byte array containing the key used to encrypt and decrypt the Realm file.
+     *
+     * @return null on unencrypted Realms.
+     */
+    public val encryptionKeyAsCallback: EncryptionKeyCallback?
 
     /**
      * Callback that determines if the realm file should be compacted as part of opening it.
@@ -234,6 +253,7 @@ public interface Configuration {
         protected var writeDispatcher: CoroutineDispatcher? = null
         protected var schemaVersion: Long = 0
         protected var encryptionKey: ByteArray? = null
+        protected var encryptionKeyAsCallback: EncryptionKeyCallback? = null
         protected var compactOnLaunchCallback: CompactOnLaunchCallback? = null
         protected var initialDataCallback: InitialDataCallback? = null
         protected var inMemory: Boolean = false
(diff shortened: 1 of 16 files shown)


packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -138,6 +137,20 @@ public class RealmImpl private constructor(
...
+                    // if we're using an encryption key as a callback, we preemptively open the notifier and writer Realm
+                    // with the given configuration because the key might be deleted from memory after the Realm is open.
+
+                    // These touches the notifier and writer lazy initialised Realms to open them with the provided configuration.
+                    launch(notificationScheduler.dispatcher) {
+                        notifier.realm.version().version
+                    }
+                    launch(writeScheduler.dispatcher) {
+                        writer.realm.version().version
+                        it.releaseKey()
```
nhachicha, line 150: maybe wait for modifier job and the writer job to complete before calling `it.releaseKey()`

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/EncryptionTests.kt
```diff
@@ -122,4 +131,59 @@ class EncryptionTests {
                 }
             }
     }
+
+    @OptIn(ExperimentalEncryptionCallbackApi::class)
+    @Test
+    fun openEncryptedRealmWithEncryptionKeyCallback() = runBlocking {
```
cmelchior, line 137: It would also be nice with tests for invalid pointers. Do we throw a correct exception in that case or ends up with a native crash?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -106,6 +107,19 @@ public data class InitialRealmFileConfiguration(
     val checksum: String?
 )
 
+@ExperimentalEncryptionCallbackApi
+public interface EncryptionKeyCallback {
+    /**
+     * Provides the native memory address of the 64 byte array containing the key used to encrypt and decrypt the Realm file.
```
cmelchior, line 113: I would also document that this can be called multiple times and that release is only called once, so they do not accidentially create a new Pointer for every call to this, but only release it once.

This was also why I thought it might make a better API if this was an symmetric API, i.e. called 3 times, and released 3 times, but I do agree that the current behavior is easier to implement for the sake of an POC.

cmelchior (commented):
Nice...great description in the docs. I do have a few comments about the API docs + some missing tests, but it is minor

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/RealmImpl.kt
```diff
@@ -138,6 +140,24 @@ public class RealmImpl private constructor(
...
             realmScope.launch {
+                @OptIn(ExperimentalEncryptionCallbackApi::class)
+                configuration.encryptionKeyAsCallback?.let {
+                    // if we're using an encryption key as a callback, we preemptively open the notifier and writer Realm
+                    // with the given configuration because the key might be deleted from memory after the Realm is open.
+
+                    // These touches the notifier and writer lazy initialised Realms to open them with the provided configuration.
+                    awaitAll(
+                        async(notificationScheduler.dispatcher) {
+                            notifier.realm.version().version
```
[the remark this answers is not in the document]
nhachicha: I'll add a TODO. There might be other feedbacks we can incorporate after the POC

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/EncryptionTests.kt
nhachicha: Invalid pointers can cause a segmentation fault that will crash the entire test suite. I'll add a comment on the class doc about key validity.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/Configuration.kt
```diff
@@ -106,6 +107,24 @@ public data class InitialRealmFileConfiguration(
...
+     * This can be called multiple times internally, so the key needs to be the same for between calls.
```
cmelchior, line 114:
```suggestion
     * This can be called multiple times internally, so the key needs to be the same between calls.
```

cmelchior (approved):
Looking good. CI will be fixed by https://github.com/realm/realm-kotlin/pull/1633

Issue #1617, opened 2023-12-27
Investigate and support new behavior by Conveyor for packing desktop apps.


cmelchior:
See https://hydraulic.dev/blog/11-in-jar-signing.html

While we added support for this in 1.5.0 (see https://github.com/realm/realm-kotlin/issues/1105), it appears that the defaults might have changed and we should investigate if there is anything we can do on our side to make it easier to work with.

Internal HELP ticket about the same is here: https://jira.mongodb.org/browse/HELP-53529

PR #1637, opened 2024-01-19
Using library version to unpack JVM libs

nhachicha:
Using library version to build a path where to unpack native libraries for JVM targets

Fixes https://github.com/realm/realm-kotlin/issues/1617


packages/cinterop/build.gradle.kts
```diff
@@ -792,3 +749,12 @@ tasks.named("clean") {
...
 }
+
+// Generate an object holding the current release version, to be used by the JVM SoLoader path construction
+buildkonfig {
+    packageName = "io.realm.kotlin.jvm"
+    objectName = "LibraryConfig"
+    defaultConfigs {
+        buildConfigField(Type.STRING, "version", Realm.version)
+    }
+}
```
cmelchior, line 760:
```suggestion
}

```

packages/cinterop/build.gradle.kts
```diff
@@ -792,3 +749,12 @@ tasks.named("clean") {
         delete(project.file(".cxx"))
     }
 }
+
+// Generate an object holding the current release version, to be used by the JVM SoLoader path construction
+buildkonfig {
```
cmelchior, line 754: This is already being generated by `cinterop` here: https://github.com/realm/realm-kotlin/blob/main/packages/cinterop/build.gradle.kts#L796, so I do not think you need to add this plugin for it?

packages/cinterop/src/jvmMain/kotlin/io/realm/kotlin/jvm/SoLoader.kt
```diff
@@ -53,97 +44,51 @@ class SoLoader {
...
-    }
-
-    private fun defaultAbsolutePath(libraryName: String, libraryHash: String): File {
+    private fun defaultAbsolutePath(libraryName: String): File {
         return File(
-            platform.defaultSystemLocation + File.separator +
-                libraryHash + File.separator +
+            platform.defaultSystemLocation +
+                File.separator +
+                LibraryConfig.version +
```
cmelchior, line 75: See comment about this already being created by our cinterop library

cmelchior (changes requested):
Nice. I do have some reservations about the CHANGELOG though and I think you are duplicating creating a version string.

cmelchior (approved):
Nice 💯

PR #1643, opened 2024-01-29
Add specific exception for auth errors being sent to to the Sync Session error handler

### LINKED ISSUE #1640: When to do re-login after token expiration
### Problem

I've understanding that the app should implement an error handler and react on token expiration to keep the device sync going on. The piece of information the error handler gets is an exception that happed. 

Which exception is it to which the app should react? 

### Solution

This could be a documentation issue.

Not sure if there's a missing exception derived from SyncException, which could indicate the need of re-login.

### Alternatives

_No response_

### How important is this improvement for you?

I would like to have it but have a workaround

### Feature would mainly be used with

Atlas Device Sync

cmelchior:
Closes #1640 

This adds a specific exception class for auth errors being sent to the SyncSession error handler. This is relevant as this indicates the user's credentials are no longer valid and they need to log in again.

Some open questions:

- [ ] Should it be possible to update the refresh token on an already open Realm? (this will require changes in Core)
- [ ] It is a bit annoying to create an "auth" type exception under `SyncException` when we already have `AuthException` under `AppException`. But relaxing the interface would be a breaking change and besides, this is the only "App" type exception that would surface here anyway.
- [ ] Naming... alternatives considered:  `SyncSessionExpiredException`, `SyncAuthException`, `SyncSessionAuthExpired`. Choosing `AuthExpiredException` was somewhat arbitrary, except it felt short and to the point. Not sure if `Sync` or `Session` should somehow be part of the name to separate it more from the generic `AuthException` that is an `AppException`. All our other specific exception types end with `Exception`

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
@@ -17,6 +17,7 @@ import io.realm.kotlin.mongodb.exceptions.FunctionExecutionException
 import io.realm.kotlin.mongodb.exceptions.InvalidCredentialsException
 import io.realm.kotlin.mongodb.exceptions.ServiceException
 import io.realm.kotlin.mongodb.exceptions.SyncException
+import io.realm.kotlin.mongodb.exceptions.SyncSessionAuthExpired
 import io.realm.kotlin.mongodb.exceptions.UnrecoverableSyncException
 import io.realm.kotlin.mongodb.exceptions.UserAlreadyConfirmedException
 import io.realm.kotlin.mongodb.exceptions.UserAlreadyExistsException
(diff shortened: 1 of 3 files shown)


cmelchior:
@nirinchev believes that Core would automatically trigger a reconnect if you log the same user back in again. If this happens, users do not need to close the Realm. We need to test the behaviour.

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -79,6 +80,7 @@ internal fun convertSyncError(syncError: SyncError): SyncException {
     val errorCode = syncError.errorCode
     val message = createMessageFromSyncError(errorCode)
     return when (errorCode.errorCode) {
+        ErrorCode.RLM_ERR_AUTH_ERROR -> AuthExpiredException(message)
```
cmelchior, line 83: We should also send this to the AuthenticationListener

rorbech: We should probably update the `AuthenticationListener` to take advantage of https://github.com/realm/realm-core/pull/7302 instead of issuing these events ourselves.

rorbech: > We should also send this to the AuthenticationListener

We should probably update the `AuthenticationListener` to take advantage of https://github.com/realm/realm-core/pull/7302 instead of issuing these events ourselves.

PR #1651, opened 2024-02-05
Additional tests of collections in mixed behavior

rorbech:


rorbech:
Incorporated in #1666 instead.

PR #1663, opened 2024-02-17
Automate release process to Maven Central and Gradle Plugin Portal

nhachicha:
Fixes https://jira.mongodb.org/browse/RKOTLIN-709


rorbech (approved):
LGTM

.github/workflows/pr.yml
```diff
@@ -1642,4 +1642,70 @@ jobs:
...
+    needs: [check-cache]
+    outputs:  
+      is_publish_build: ${{ steps.check_release.outputs.is_publish_build }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get Git tag
+        run: |
+          gitTag=$(git describe --tags --exact-match HEAD) || echo "NONE"
```
clementetb, line 1659: If the tag does not exist the tag would be "NONE", shouldn't we stop the release if not gittag is defined?

clementetb (commented):
Looks good except for that gittag comment.

clementetb (commented):
Looks good except for that gittag comment.

.github/workflows/pr.yml
nhachicha: If the gittag is `NONE` `is_publish_build` will be `false` in the next step which will skip the release 
image

PR #1664, opened 2024-02-20
Guard analytics errors from breaking build

rorbech:


.github/workflows/pr.yml
```diff
@@ -227,11 +228,11 @@ jobs:
...
     if: |
-      always() && 
-      !cancelled() && 
-      !contains(needs.*.result, 'failure') && 
+      always() &&
+      !cancelled() &&
+      !contains(needs.*.result, 'failure') &&
       !contains(needs.*.result, 'cancelled') &&
-      needs.check-cache.outputs.jni-windows-lib-cache-hit != 'true'
+      needs.check-cache.outputs.jni-macos-lib-cache-hit != 'true'
```
nhachicha, line 235: nice catch 🎉

packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmCompilerSubplugin.kt
```diff
@@ -237,6 +242,20 @@ class RealmCompilerSubplugin : KotlinCompilerPluginSupportPlugin, AnalyticsError
             options
         }
     }
+
+    /**
+     * Wrapper that ignores error if `failOnAnalyticsError=true`.
+     */
+    private fun <R> muteErrors(block: () -> R): R? {
```
nhachicha, line 249: Why this needs to return?

nhachicha (approved):

packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmCompilerSubplugin.kt
rorbech: It is used guard collecting of target information and make them `null` if an error happens [here](https://github.com/realm/realm-kotlin/pull/1664/files#diff-aa43ef78792f7359466e6dafb449349572d491ed2f6f630cdd02e41880286075R220)

PR #1669, opened 2024-02-29
Consistent equals method for collection types

rorbech:


rorbech:
Abandoning this and keeping the non-performant but consistant structural equality. Cannot keep symmetry for `Realm[Collection].equals` if we don't honor the structural equality check from the standard Kotlin Collection types (Set, List, Map) unless we remove these from our `Realm[Collection]`-types.

PR #1674, opened 2024-03-05
RPM-568: Add a workflow that verifies PR title

nirinchev:


.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,22 @@
+# NOTE: This is a common file that is overwritten by realm/ci-actions sync service
```
rorbech, line 1: Is this added the right way ... Shouldn't it just flow from the sync service?

.github/workflows/check-pr-title.yml
```diff
@@ -0,0 +1,22 @@
...
+    name: Check PR Title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: false
+      - name: Enforce PR title
+        uses: realm/ci-actions/title-checker@main
+        with:
+          regex: R[A-Z]{2,5}-[0-9]{1,6}
```
rorbech, line 20: This does not seem to match `RKOTLIN` 🤔

.github/workflows/check-pr-title.yml
nirinchev: The sync service is broken right now due to SSO, so adding it manually

.github/workflows/check-pr-title.yml
nirinchev: Ugh, I thought RSWIFT was the longest one 🤦‍♂️

rorbech:
Not for this PR, but maybe we should think of grouping all the ci-action scripts so it was obvious where there are coming from and what role they play. Ideally with a prefix that gathered them last in the list of workflows 😬

rorbech:
> Not for this PR, but maybe we should think of grouping all the ci-action scripts so it was obvious where there are coming from and what role they play. Ideally with a prefix that gathered them last in the list of workflows 😬

If put in a separate folder it would also be easier not to trigger all the other workflows when they change.

PR #1692, opened 2024-03-15
[RKOTLIN-1038] Implement new log category for RealmLogger

clementetb:
closes https://github.com/realm/realm-kotlin/issues/1691

RealmLog would see the current `level` property deprecated in favor of a setter and getter functions.

```kotlin
RealmLog {
    @Deprecated
    public var level: LogLevel

    public fun setLevel(level: LogLevel, category: LogCategory)
    public fun getLevel(category: LogCategory): LogLevel
}

```

The custom logger interface also gets modified adding a new method that has the category in its signature. Backward compatibility would be ensured to the current deprecated methods.

```kotlin
RealmLogger {
    public fun log(
        category: LogCategory, 
        level: LogLevel, 
        throwable: Throwable?, 
        message: String?, 
        vararg args: Any?
    )
}

### CODE REVIEW

File: packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt [hunk 1]
```diff
@@ -38,12 +39,23 @@ public interface RealmLogger {
...
-    public fun log(level: LogLevel, throwable: Throwable?, message: String?, vararg args: Any?)
+    public fun log(category: LogCategory, level: LogLevel, throwable: Throwable?, message: String?, vararg args: Any?) {
+        log(level, throwable, message, *args)
+    }
+
+    /**
+     * Log an event.
+     */
+    @Deprecated("Use log(category: LogCategory, level: LogLevel, throwable: Throwable?, message: String?, vararg args: Any?)", ReplaceWith("log"))
+    public fun log(level: LogLevel, throwable: Throwable?, message: String?, vararg args: Any?) {
```
rorbech, line 50: I guess we cannot leave the category-variant without a default implementation, as that would be a breaking change if users need to implement that, but feels like there is something wrong here. This would just recurse between the two log methods if ever called. I think it would be safer to have a 
```
// Don't overwrite this, but overwrite the category-variant instead
TODO("log() shouldn't be called. Overwrite the category-variant instead")
```
or something like that, so people would be guided the right way faster.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt
```diff
@@ -38,12 +39,23 @@ public interface RealmLogger {
     /**
      * Log an event.
      */
-    public fun log(level: LogLevel, throwable: Throwable?, message: String?, vararg args: Any?)
+    public fun log(category: LogCategory, level: LogLevel, throwable: Throwable?, message: String?, vararg args: Any?) {
+        log(level, throwable, message, *args)
```
rorbech, line 43: Should we consider padding the category on the messages in the default implementation with: `message = "[$category]$message"`?

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmConfigurationTests.kt
```diff
@@ -246,7 +246,7 @@ class RealmConfigurationTests {
     fun defaultLogLevel() {
         val config: RealmConfiguration = RealmConfiguration.Builder(schema = setOf(Sample::class))
             .build()
-        assertEquals(LogLevel.WARN, config.log.level)
+        assertEquals(LogLevel.WARN, RealmLog.level)
```
rorbech, line 249: Don't we still need to assert that `config.log.level` is correct? Maybe assert both.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/LogCategory.kt
```diff
@@ -0,0 +1,314 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.kotlin.log
+
+/*
+ * Core does not expose the different categories in compile time.
```
rorbech, line 20: I think we should move all the internal stuff to an internal package/class/file. It confuses users if they end up here.

packages/cinterop/src/commonMain/kotlin/io/realm/kotlin/internal/interop/Callback.kt
```diff
@@ -43,7 +43,7 @@ interface SyncSessionTransferCompletionCallback {
 
 interface LogCallback {
     // Passes core log levels as shorts to avoid unnecessary jumping between the SDK and JNI
-    fun log(logLevel: Short, category: String?, message: String?)
+    fun log(logLevel: Short, categoryValue: String, message: String?)
```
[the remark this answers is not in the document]
clementetb: It is an internal api so it should be exposed to the users...

packages/cinterop/src/nativeDarwin/kotlin/io/realm/kotlin/internal/interop/RealmInterop.kt
```diff
@@ -2511,7 +2513,7 @@ actual object RealmInterop {
         realm_wrapper.realm_set_log_callback(
             staticCFunction { userData, category, logLevel, message ->
                 val userDataLogCallback = safeUserData<LogCallback>(userData)
-                userDataLogCallback.log(logLevel.toShort(), category?.toKString(), message?.toKString())
+                userDataLogCallback.log(logLevel.toShort(), category!!.toKString(), message?.toKString())
```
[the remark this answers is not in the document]
clementetb: I am sure it cannot be null, it is a closed set of loggers, all named unless core does something shady... 

I am gonna double check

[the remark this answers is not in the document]
clementetb: Yes, it is safe. We have tests that validate that we map all valid category values.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/LogLevel.kt
```diff
@@ -20,3 +21,31 @@ public enum class LogLevel(public val priority: Int) {
     WTF(6),
     NONE(7);
 }
+
+internal fun LogLevel.toCoreLogLevel(): CoreLogLevel {
```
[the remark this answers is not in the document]
clementetb: I disagree we could make functions internal to a public file, like we do with `RealmLog` is a public singleton and has internal methods.

We created the internal package to place all the internal classes that had to be made public because of the base/sync split.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt
clementetb: Great point, the categories could even be used as tags.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt
clementetb: I wanted to avoid a breaking change but I introduced a cycle. I have update the logic.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt
rorbech: It might be that there is some issues with mixing it up with the tags. On Android there are other constraints on the tags, etc., so seems safest to just pad it with `[$category] $message`

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt
rorbech: The new implementation looks even more confusing than the old. I don't think it guides users if they don't see any output from `fun log(level: LogLevel, message: String)`!?

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/LogLevel.kt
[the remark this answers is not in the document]
rorbech: The reason for the `internal` packages was to split it so that it was obvious what the public API was and that users wouldn't end up browsing internal code unless they intentionally navigated to it from the public API. We haven't been completely strict on this, but I don't see why we should not try to follow it 🤷

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLogger.kt
clementetb: Yes, there is a limit on tag length... :(

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/ContextLogger.kt
```diff
@@ -80,11 +81,29 @@ public class ContextLogger(public val context: String? = null) {
         doLog(LogLevel.WTF, null, { contextPrefix + message }, *args)
     }
 
+    private inline fun checkPriority(
+        level: LogLevel,
+        block: () -> Unit,
+    ) {
+        if (level.priority >= RealmLog.getLevel(SdkLogCategory).priority) {
```
rorbech, line 88: We should probably cache the SDK category level in the SDK, so that we didn't have to round trip core on each  check.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This release will bump the Realm file format from version 23 to 24. Opening a fi
 ### Enhancements
 * Support for RealmLists and RealmDictionaries in `RealmAny`. (Issue [#1434](https://github.com/realm/realm-kotlin/issues/1434))
 * Optimized `RealmList.indexOf()` and `RealmList.contains()` using Core implementation of operations instead of iterating elements and comparing them in Kotlin. (Issue [#1625](https://github.com/realm/realm-kotlin/pull/1666) [RKOTLIN-995](https://jira.mongodb.org/browse/RKOTLIN-995)).
+* Add support for filtering logs by category. (Issue [#1691](https://github.com/realm/realm-kotlin/issues/1691) [JIRA](https://jira.mongodb.org/browse/RKOTLIN-1038))
```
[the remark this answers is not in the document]
rorbech: We should also highlight that the `LogConfiguration` is gone.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncedRealmTests.kt
```diff
@@ -1238,44 +1238,6 @@ class SyncedRealmTests {
         }
     }
 
-    @Test
```
rorbech, line 1241: We have seen errors that we messed up the logging so that users only received log messages if they set the correct log levels at the right moment, but maybe that was an issue with the previous two competing concepts (RealmLog vs. LogConfiguration) 🤔 ... I guess it should be consistent and safe now

rorbech (commented):
Nice clean ups. I like the idea of always logging to a specific logger instead of `RealmLog`. 

There are some minors and then we need to bump the version and highlight the breaking changes.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/LogCategory.kt
```diff
@@ -0,0 +1,308 @@
...
+     * Category hierarchy:
+     * ```
+     * Storage
+     * ├─► Transaction
+     * ├─► Query
+     * ├─► Object
+     * └─► Notification
+     * ```
+     */
+    public val Storage: StorageLogCategory = StorageLogCategory
```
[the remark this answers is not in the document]
clementetb: It is a per name convention: https://kotlinlang.org/docs/coding-conventions.html#property-names

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/RealmLog.kt
```diff
@@ -44,19 +52,55 @@ public object RealmLog {
     // copy this reference before using it.
     private var loggers: MutableList<RealmLogger> = mutableListOf()
 
+    // Log level that would be set by default
+    private val defaultLogLevel = LogLevel.WARN
+
+    /**
+     * The current [LogLevel]. Changing this will affect all registered loggers.
```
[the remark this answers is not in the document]
clementetb: Yes, it should have been removed.

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/log/LogCategory.kt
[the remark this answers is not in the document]
rorbech: The [Kotlin conventions](https://kotlinlang.org/docs/coding-conventions.html#property-names) says that it is OK to use capital names for object references.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncedRealmTests.kt
clementetb: Yes, before there were some conditions were log levels were overwritten after setting a config due to this LogConfiguration vs RealmLog.

It is not an issue now, sync logs are received depending on the category level.

packages/test-base/src/commonTest/kotlin/io/realm/kotlin/test/common/RealmLogTests.kt
```diff
@@ -210,61 +214,101 @@ class RealmLogTests {
         }
     }
 
+    @Test
```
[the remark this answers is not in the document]
clementetb: Yes, I though I removed them.

Issue #1690, opened 2024-03-15
Missing Executable Files for Windows and Linux in cinterop-jvm 1.14.0


guiyanakuang:
### How frequently does the bug occur?

Always

### Description

Missing Executable Files for Windows and Linux in cinterop-jvm 1.14.0

https://repo1.maven.org/maven2/io/realm/kotlin/cinterop-jvm/1.14.0/cinterop-jvm-1.14.0.jar

<img width="170" alt="image" src="https://github.com/realm/realm-kotlin/assets/4069905/2c2b545f-aacc-4470-9a1e-02b484c4d301">

### Can you reproduce the bug?

Always

### Version

1.14.0

### What Atlas App Services are you using?

Local Database only

### Are you using encryption?

No

### Platform OS and version(s)

windows

rorbech:
Hi @guiyanakuang. Thanks for the report. Seems like the native libraries slipped from our new release script. We will look into it.

rorbech:
Fixed in `1.14.1`.

PR #1694, opened 2024-03-18
[RKOTLIN-1037] Copy missing JVM binaries (Win/Linux) when releasing

nhachicha:
- Fixes https://github.com/realm/realm-kotlin/issues/1690 
- Building cinterop-jvm Linux shared library in Release mode

### DIFF
--- a/.github/workflows/include-deploy-release.yml
+++ b/.github/workflows/include-deploy-release.yml
@@ -108,5 +108,5 @@ jobs:
          "${{ secrets.DOCS_S3_ACCESS_KEY }}" "${{ secrets.DOCS_S3_SECRET_KEY }}" \
          "${{ secrets.SLACK_URL_RELEASE }}" "${{ secrets.SLACK_URL_CI }}" \
          "${{ secrets.GRADLE_PORTAL_KEY }}" "${{ secrets.GRADLE_PORTAL_SECRET }}" \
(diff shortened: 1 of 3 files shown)


.github/workflows/include-deploy-release.yml
```diff
@@ -108,5 +108,5 @@ jobs:
          "${{ secrets.DOCS_S3_ACCESS_KEY }}" "${{ secrets.DOCS_S3_SECRET_KEY }}" \
          "${{ secrets.SLACK_URL_RELEASE }}" "${{ secrets.SLACK_URL_CI }}" \
          "${{ secrets.GRADLE_PORTAL_KEY }}" "${{ secrets.GRADLE_PORTAL_SECRET }}" \
-         '-PsignBuild=true -PsignSecretRingFileKotlin="${{ secrets.GPG_SIGNING_KEY_BASE_64_DBG }}" -PsignPasswordKotlin=${{ secrets.GPG_PASS_PHRASE_DBG }}'
+         '-PsignBuild=true -PsignSecretRingFileKotlin="${{ secrets.GPG_SIGNING_KEY_BASE_64_DBG }}" -PsignPasswordKotlin=${{ secrets.GPG_PASS_PHRASE_DBG }} -Prealm.kotlin.copyNativeJvmLibs=linux,windows'
```
rorbech, line 111: Shouldn't we also include `macos`? I see that the Macos `dylib`-file was already included in previous release, but I can't really grasp why. Was it unintentionally (or intentionally) rebuilt as part of the release 🤔

nhachicha: no, macOS file doesn't need to be copied because it is already built. I checked manually with a staged build on https://oss.sonatype.org/ and all of them are uploaded

rorbech (approved):

PR #1710, opened 2024-03-26
[RKOTLIN-1062] Remove CMake required version

clementetb:
Removes the fixed cmake version required by cinterop to rely on Cores cmake minimum version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 - Minimum R8: 8.0.34.
 
 ### Internal
-- None
+- Remove CMake required version.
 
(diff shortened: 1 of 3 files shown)


rorbech (approved):
There are more to it than this to control the cmake version. We also have a [hardcoded version for docker image to build the Linux so-files](https://github.com/realm/realm-kotlin/blob/main/packages/cinterop/src/jvmMain/generic.Dockerfile#L24) and a `VERSION_CMAKE` GHA-variable. We should somehow align all of them ... but I guess this unblocks us for now, so if CI is happy.

PR #1721, opened 2024-04-11
Support for compiling the SDK with Kotlin 2

rorbech:
Due to binary incompatibilities of the Kotlin compile API across 1.9 and 2.0 we cannot supply a cross version compatible distribution. This PR will however update the code base so that it is possible to compile the SDK itself with both 1.9 and 2.0 to make it easier to maintain and supply a 2.0-compatible branch. 

**Note** This PR does not enable support for K2 in consumer project nor does it change the project to use K2 itself.


packages/gradle-plugin/src/main/kotlin/io/realm/kotlin/gradle/RealmCompilerSubplugin.kt
```diff
@@ -318,9 +318,12 @@ private fun gatherTargetInfo(kotlinCompilation: KotlinCompilation<*>): TargetInf
...
                     Family.MINGW,
-                    Family.ANDROID,
-                    Family.WASM,
-                    Family.ZEPHYR -> null // Not supported yet
+                    Family.ANDROID -> null // Not supported yet
+                    // Remaining options are removed in Kotlin 2, so cannot reference them but need
+                    // an else clause to be exhaustive
+                    // Family.WASM,
+                    // Family.ZEPHYR,
+                    else -> null
```
kneth, line 326: How do we ensure we catch new target added in the future?

rorbech: We can't both ignore non-existing symbols and ensure that we are exhaustive. This anyway a bit shady. We don't fetch minimum SDK for non-Cocoapods Kotlin Native builds. I have added a TODO about revisiting it when we skip Kotllin 1.9 support.

PR #1727, opened 2024-04-23
Support compiling the SDK with Kotlin 2 (#1721)

rorbech:


rorbech:
Just cherry-picking from `main` so will merge without review.

PR #1733, opened 2024-04-30
[RKOTLIN-1023] Add support for changing App Services base URL

clementetb:
With this PR we add support for changing the App Services URL allowing roaming between Atlas and Edge server.

Closes https://github.com/realm/realm-kotlin/issues/1659

### DIFF
--- a/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
+++ b/packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
@@ -17,6 +17,7 @@
 package io.realm.kotlin.mongodb
 
 import io.realm.kotlin.internal.util.Validation
+import io.realm.kotlin.mongodb.annotations.ExperimentalEdgeServerApi
 import io.realm.kotlin.mongodb.auth.EmailPasswordAuth
 import io.realm.kotlin.mongodb.exceptions.AppException
 import io.realm.kotlin.mongodb.exceptions.AuthException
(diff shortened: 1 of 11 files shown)


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/App.kt
```diff
@@ -84,6 +84,19 @@ public interface App {
...
+     */
+    public val baseUrl: String
+
+    /**
+     * Sets the App Services base url. Changing the URL would trigger a client reset.
+     *
+     * @param baseUrl The new App Services base url. Use [AppConfiguration.DEFAULT_BASE_URL] to set it
+     * to the default value.
+     */
+    public suspend fun updateBaseUrl(baseUrl: String)
```
clementetb, line 98: Passing an empty string to `updateBaseUrl` would result in the base URL being set to a default value set in core. 

Because the default URL is already set in core, we could replace the actual URL value in the SDK with an empty string. This way we could avoid maintaining the URL on the SDK. 

```
AppConfiguration.DEFAULT_BASE_URL = ""
```

We would address this on a different PR.

nirinchev: We could also make `baseUrl` optional and allow users to supply `null` here to the same effect of supplying empty string. I feel this is a bit more obvious and the C API already works that way.

nirinchev (approved):

rorbech (changes requested):
According to the scope the `updateBaseUrl` API should be guarded by an ` @ExperimentalEdgeServerApi`

rorbech (approved):

PR #1748, opened 2024-05-20
[RKOTLIN-1082] Remove RealmObjectId

clementetb:
Removes deprecated Realm version of `ObjectId` 

closes https://github.com/realm/realm-kotlin/issues/1749


packages/test-base/src/jvmTest/kotlin/io/realm/kotlin/test/compiler/Utils.kt
```diff
@@ -47,7 +47,7 @@ internal const val OBJECT_CLASS = "SampleClass"
 
 private val COLLECTION_CODE = """
 import io.realm.kotlin.types.EmbeddedRealmObject
-import io.realm.kotlin.types.ObjectId
+
```
rorbech, line 50: Will this fly with `ktlint`?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/serializer/BsonEncoderTests.kt
```diff
@@ -143,13 +141,6 @@ class BsonEncoderTests {
                             bsonValue = BsonBinary(BsonBinarySubType.UUID_STANDARD, uuid.bytes),
                         )
                     }
-                ObjectId::class -> ObjectId.create().let { objectId ->
```
rorbech, line 146: Shouldn't we then have a test for `BsonObjectId`? I guess this is a test of our addition to the serializers, so maybe not required, but we are testing it for `BsonDecimal128`. https://github.com/realm/realm-kotlin/pull/1748/files#diff-57eed610740b3e30ea24c8fe99231780bb1e17c15dbcc0dd3adb9d5654b3a803R64 has a comment that `BsonObjectId` is tested elsewhere, but I can't really see how that should be the case from following the code. 

Since serialization is just reusing kbson's serializer I don't think it is a blocker, but more of a nice to have.

rorbech (approved):
Cool 🎉 Some minor comments.

PR #1768, opened 2024-05-29
[RKOTLIN-1096] Map `is_fatal` to `UnrecoverableSyncException`

clementetb:
closes https://github.com/realm/realm-kotlin/issues/1767


packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
```diff
@@ -78,26 +78,27 @@ internal fun <T, R> channelResultCallback(
...
-            // Flexible Sync Query was rejected by the server
-            BadFlexibleSyncQueryException(message)
-        }
-        ErrorCode.RLM_ERR_SYNC_COMPENSATING_WRITE -> CompensatingWriteException(message, syncError.compensatingWrites)
+            ErrorCode.RLM_ERR_INVALID_SUBSCRIPTION_QUERY -> {
+                // Flexible Sync Query was rejected by the server
+                BadFlexibleSyncQueryException(message)
+            }
 
-        ErrorCode.RLM_ERR_SYNC_PROTOCOL_INVARIANT_FAILED,
```
nhachicha, line 90: Are these 3 all now under Fatal `UnrecoverableSyncException`?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncedRealmTests.kt
```diff
@@ -336,13 +335,12 @@ class SyncedRealmTests {
...
                     // Make sure that the test eventually fail. Coroutines can cancel a delay
                     // so this doesn't always block the test for 10 seconds.
-                    delay(10 * 1000)
+                    delay(10_000)
                     channel.send(AssertionError("Realm was successfully opened"))
                 }
             }
 
             val error = channel.receiveOrFail()
-            assertTrue(error is UnrecoverableSyncException, "Was $error")
```
nhachicha, line 345: Isn't there a test that checks `UnrecoverableSyncException`?

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
clementetb: At least `RLM_ERR_SYNC_PERMISSION_DENIED` isn't, but shouldn't now any fatal (unrecoverable) errors have`is_fatal` set to true?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncedRealmTests.kt
clementetb: That is the only test that mentioned `UnrecoverableSyncException`. I am going to add some tests using `realm_sync_session_handle_error_for_testing`

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/RealmSyncUtils.kt
clementetb: Checked with core, `is_fatal` is a better property to identify unrecoverable errors.

rorbech: Seems safest to keep this, unless you are absolutely certain that core throws the three error codes with `is_fatal` set. If not we will suddenly throw them as `SyncExceptions` instead.

CHANGELOG.md
```diff
@@ -1,3 +1,33 @@
+## 2.0.1 (YYYY-MM-DD)
+
+> [!NOTE]
```
nhachicha, line 3: We already had this note for `2.0.0` should we keep it here as well?

nhachicha (approved):

CHANGELOG.md
rorbech: I included it in 2.0.0 even though it was also in 1.16.0. The reason was that I though people would maybe fast track from earlier releases to 2.0.0. But I assume that people would always read the changelog for the major bump, so don't think we need to included it any further.

PR #1770, opened 2024-05-30
[RKOTLIN-1100] Clean up build system

clementetb:
This PR removes all the known warnings by addressing or suppressing them, some other changes in this PR:

- Bumped minimum R8.
- Ignore temp test files in the project directory.
- Updated to Detekt.
- Refactor `CoreConnectionState` because native method `byValue`deprecation.

Closes  https://github.com/realm/realm-kotlin/issues/1773


packages/jni-swig-stub/build.gradle.kts
```diff
@@ -62,13 +72,6 @@ realmPublish {
     }
 }
 
-java {
-    withSourcesJar()
-    withJavadocJar()
```
rorbech, line 67: I am not sure that we can remove this. There are some subtleties about publishing to Maven central that requires some of these. But maybe it is only the sources - Let's see when we release 🙈

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/Converters.kt
```diff
@@ -269,76 +271,75 @@ internal interface CoreIntConverter : StorageTypeConverter<Long> {
 }
 
 internal object ByteConverter : CoreIntConverter, CompositeConverter<Byte, Long>() {
-    override inline fun fromPublic(value: Byte?): Long? = byteToLong(value)
```
rorbech, line 272: Are these critical to remove. While introducing the converter infrastructure I inspected the generated byte code and it seemed to have quite significant impact on the number of instructions if the calls were inlined. Don't know if that is measurable though, so maybe not a issue. But I would just be more comfortable to leave them until verified that it doesn't have an impact.

packages/settings.gradle.kts
```diff
@@ -24,6 +24,7 @@ dependencyResolutionManagement {
         getPropertyValue("testRepository")?.let {
             maven("file://${rootDir.absolutePath}/$it")
         }
+        maven("https://oss.sonatype.org/content/repositories/snapshots")
```
rorbech, line 27: I am a bit hesitant to leave this here, as you might end up compiling stuff against public snapshots instead of locally distributed once. Do we really need it, or could we just have it commented out for easy access when needed.

rorbech (commented):
Awesome cleanup - Had some minor questions that might need to be addressed, but not sure 😅

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/internal/Converters.kt
clementetb: The warning is that these inlines only work well if called directly and not via the interface they are implementing. 

Now that you point to it I found some places where they are invoked directly. Im restoring them and supressing the warning.

rorbech: They are also called directly from the generated IR code. Or at least was when I implemented it in the first place.

packages/jni-swig-stub/build.gradle.kts
clementetb: I have added them back.

rorbech (approved):

PR #1778, opened 2024-06-07
[RKOTLIN-1102] Remove nullability check in `SubscriptionSetImpl.waitForSynchronization`

clementetb:
closes https://github.com/realm/realm-kotlin/issues/1777

Removes a nullability check on the synchronization error message that is causing a null pointer exception crash

### DIFF
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/exceptions/RealmException.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/exceptions/RealmException.kt
@@ -9,7 +9,7 @@ package io.realm.kotlin.exceptions
  */
 public open class RealmException : RuntimeException {
     public constructor() : super()
-    public constructor(message: String) : super(message)
-    public constructor(message: String, cause: Throwable) : super(message, cause)
-    public constructor(cause: Throwable) : super(cause)
+    public constructor(message: String?) : super(message)
+    public constructor(message: String?, cause: Throwable) : super(message, cause)
+    public constructor(cause: Throwable?) : super(cause)
 }
--- a/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/exceptions/RealmException.kt
+++ b/packages/library-base/src/commonMain/kotlin/io/realm/kotlin/exceptions/RealmException.kt
@@ -10,6 +10,6 @@ package io.realm.kotlin.exceptions
 public open class RealmException : RuntimeException {
     public constructor() : super()
     public constructor(message: String?) : super(message)
-    public constructor(message: String?, cause: Throwable) : super(message, cause)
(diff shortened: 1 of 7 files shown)


packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/CredentialsTests.kt
```diff
@@ -369,7 +369,7 @@ class CredentialsTests {
             payload = mapOf("mail" to TestHelper.randomEmail(), "id" to 0)
         )
 
-        assertFailsWithMessage<AuthException>("unauthorized") {
+        assertFailsWithMessage<AuthException>("Authentication failed") {
```
rorbech, line 372: Hmm. Weird that this has to go the other way. Was updated on the server two weeks ago and seemed pretty consistent with `unauthorized` throughout the server 🤔 But I guess it is ok, since it is just a test assertion 🤷

rorbech (approved):
LGTM

packages/library-base/src/commonMain/kotlin/io/realm/kotlin/exceptions/RealmException.kt
```diff
@@ -9,7 +9,7 @@ package io.realm.kotlin.exceptions
  */
 public open class RealmException : RuntimeException {
     public constructor() : super()
-    public constructor(message: String) : super(message)
-    public constructor(message: String, cause: Throwable) : super(message, cause)
-    public constructor(cause: Throwable) : super(cause)
+    public constructor(message: String?) : super(message)
+    public constructor(message: String?, cause: Throwable) : super(message, cause)
```
nhachicha, line 13: `cause` is also nullable https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-runtime-exception/#constructors

packages/library-sync/src/commonMain/kotlin/io/realm/kotlin/mongodb/internal/SubscriptionSetImpl.kt
```diff
@@ -127,7 +127,7 @@ internal class SubscriptionSetImpl<T : BaseRealm>(
                     if (result) {
                         return true
                     } else {
-                        throw BadFlexibleSyncQueryException(errorMessage!!)
+                        throw BadFlexibleSyncQueryException(errorMessage)
```
nhachicha, line 130: is this nullable because `RealmInterop.realm_sync_subscriptionset_error_str` can return null/empty?

clementetb: Yes, it looks like it.

PR #1782, opened 2024-06-18
Disable workaround for core 7297

clementetb:


clementetb:
Fixed by https://github.com/realm/realm-kotlin/pull/1791

PR #1785, opened 2024-06-19
Test app initializer rework

rorbech:
This PR reworks our test app initializer, so that it is easier to do variations of the setup.


clementetb (approved):
Awesome it looks great. The only concern is that I cannot rerun `roundtripCollectionsInMixed` without resetting the server.

rorbech:
> Awesome it looks great. The only concern is that I cannot rerun `roundtripCollectionsInMixed` without resetting the server.

Should be fixed now.

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/SyncedRealmTests.kt
```diff
@@ -765,31 +767,37 @@ class SyncedRealmTests {
             }
         ).let { config ->
             Realm.open(config).use { realm ->
+                println("download changes")
```
clementetb, line 770: Do we want these `println`s?

nhachicha (approved):

Issue #1715, opened 2024-04-04
Improved mechanism for unpacking of JVM native libs suitable for local development


rorbech:
After #1637 we only copy native libs into a suitable location if the files does not exists. This prevents updates of the cache if unpacking new libraries with the same version which is what we do during development. We need an improved mechanism that takes this into account. 

I think it is safest to go back to the previous approach with a checksum and then find some way for users to programatically opt out from the verification if required.

PR #1786, opened 2024-06-19
Always unpacking JVM native lib when using SNAPSHOT version

nhachicha:
Fixes https://github.com/realm/realm-kotlin/issues/1715

This was tested locally by adding a JVM test that checks the schema version 
```Kotlin
assertEquals(realm.schemaVersion(), 42)
```

Then on AS edit the Core method `realm_get_schema_version` (inside https://github.com/realm/realm-core/blob/master/src/realm/object-store/c_api/schema.cpp#L52-L56) to return a different version which should compile a new native shared library to be extracted which has the new value. Observe that the test passes with the modified version .

Note: Obviously If you edit the `buildSrc/src/main/kotlin/Config.kt` to use a named version locally this work around won't work

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - None.
 
 ### Enhancements
-- None.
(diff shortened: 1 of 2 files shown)


rorbech (approved):
LGTM

PR #1791, opened 2024-07-02
Remove centos 7 docker image

clementetb:
Centos 7 has been deprecated as of 30th June 2024. Our CI job `build-jvm-linux-native-lib` used it in a docker image.

This PR removes the docker image and integrates this build job in GHA using a ubuntu-22.04 image.

This is the same image as used by core: https://github.com/realm/ci/blob/master/realm/docker/build-ubuntu/Dockerfile

This PR includes the changes from https://github.com/realm/realm-kotlin/pull/1782:

Some schema synchronization issues caused the app not being properly initialized. We have a workaround that addresses them, but unfortunately there are still some race conditions.

Tests now define the schema in the server via the admin API instead of relying on the development mode to do the synchronization. It allows to first set up the server, enable sync, and wait for sync to be enabled. Once enabled we wait to be ready to accept new clients.


.github/workflows/pr.yml
```diff
@@ -93,7 +92,7 @@ jobs:
           retention-days: 1
 
   build-jvm-linux-native-lib:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
nhachicha, line 95: What version of `libc` it has? the goal with old `centos:7` is to build it on the smallest libc Core supports. So maybe we can still use/install the old libc via ubuntu

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/SchemaProcessor.kt
```diff
@@ -0,0 +1,204 @@
...
+import io.realm.kotlin.internal.interop.CollectionType
+import io.realm.kotlin.internal.interop.PropertyInfo
+import io.realm.kotlin.internal.interop.PropertyType
+import io.realm.kotlin.internal.realmObjectCompanionOrNull
+import io.realm.kotlin.internal.schema.RealmClassImpl
+import io.realm.kotlin.schema.RealmClassKind
+import io.realm.kotlin.types.BaseRealmObject
+import kotlin.reflect.KClass
+
+// TODO REname methods and classes
```
rorbech, line 29: For this PR to rename?

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/SchemaProcessor.kt
```diff
@@ -0,0 +1,204 @@
...
+        val companion = clazz.realmObjectCompanionOrNull()!!
+        val realmSchema = companion.io_realm_kotlin_schema()
+        realmSchema.cinteropClass.name to realmSchema
+    }
+
+    val processedSchemas: MutableMap<String, SchemaData> = mutableMapOf()
+    val processedRelationships: MutableMap<String, Map<String, SchemaRelationship>> = mutableMapOf()
+
+    init {
+        // TODO CHECK embedded CYCLES
```
rorbech, line 68: What happens with cycles right now? If not supported could we at least easily break/throw, so that we don't have dig in if it happens.

rorbech (approved):
LGTM - Only some minor comments.

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/SchemaProcessor.kt
clementetb: I forgot to add it...

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/SchemaProcessor.kt
clementetb: It was a reminder for my future self to clean up names, it has been done but the comment was not removed.

packages/test-sync/src/commonMain/kotlin/io/realm/kotlin/test/mongodb/util/TestAppInitializer.kt
```diff
@@ -105,88 +128,9 @@ object TestAppInitializer {
...
-                        "firstName",
-                        "lastName",
-                        "age"
-                    ],
-                    "title": "SyncPerson"
-                }
-            }
-            """.trimIndent()
-        )
+        while (!app.initialSyncComplete()) {
```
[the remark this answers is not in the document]
clementetb: Doesn't the test have a general timeout?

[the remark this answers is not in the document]
rorbech: Don't think so

PR #1779, opened 2024-06-10
[RKOTLIN-1114] Fix sync client config not applied

clementetb:
This PR includes the changes from the previous PR https://github.com/realm/realm-kotlin/pull/1801 as it ended up addressing some of the flaky tests from that. 
 
TODOs:
- [x] ~~Depends on https://github.com/realm/realm-kotlin/pull/1801 for green client reset and flx progress listener tests~~ Now includes all changes from #1801 
- [x] Await that https://github.com/realm/realm-core/pull/7891 is merged and bump to an official Core release

Closes #1811

PR #1801, opened 2024-07-09
Bump core 14.10.4

clementetb:
- Bumps core to version 14.10.3
- Fixes GHA static analysis jobs not failing after an exception ocurred.
- Adds workaround for `encryptedMetadataRealm_openWithWrongKeyThrows`: copies the Realm file to validate encryption.
- Applies flip on `is_fatal` flag after being fliped in `realm_sync_session_handle_error_for_testing`.
- Fixes expected message assertions in client reset exceptions, as the now report a larger message.
- `uploadProgressListener_changesOnly` disabled until https://github.com/realm/realm-core/issues/7869 is fixed.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -605,7 +605,7 @@ jobs:
 
       - name: Setup Android SDK
         env: 
-          JAVA_HOME: ${{ env.JAVA_HOME_17_x64 }}
+          JAVA_HOME: ${{ env.JAVA_HOME_17_X64 }}
         uses: android-actions/setup-android@v3
     
       - name: Install NDK
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -547,7 +547,9 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: ${{ vars.VERSION_JAVA_DISTRIBUTION }}
-          java-version: ${{ vars.VERSION_JAVA }}
+          java-version: |
+            17
+            ${{ vars.VERSION_JAVA }}
 
       - name: Setup Gradle and task/dependency caching
         uses: gradle/actions/setup-gradle@v3
@@ -602,6 +604,8 @@ jobs:
           echo '#!/bin/bash\nccache clang++ "$@"%"' > /usr/local/bin/ccache-clang++          
 
       - name: Setup Android SDK
+        env: 
+          JAVA_HOME: ${{ env.JAVA_HOME_17_x64 }}
         uses: android-actions/setup-android@v3
     
       - name: Install NDK
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -122,7 +122,7 @@ jobs:
           key: jni-linux-lib-${{ needs.check-cache.outputs.packages-sha }}
 
       - name: Setup Java 11
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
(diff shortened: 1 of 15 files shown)


.github/workflows/pr.yml
```diff
@@ -547,7 +547,9 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: ${{ vars.VERSION_JAVA_DISTRIBUTION }}
-          java-version: ${{ vars.VERSION_JAVA }}
+          java-version: |
+            17
```
nhachicha, line 551: we should add this version into `Repository variables` 
image
This version is only needed for sdkmanager and we should not build with it the library

CHANGELOG.md
```diff
@@ -4,10 +4,28 @@
 * None.
 
 ### Enhancements
-* None.
+* Reduce the size of the local transaction log produced by creating objects, improving the performance of insertion-heavy transactions (Core issue [realm/realm-core#7734](https://github.com/realm/realm-core/pull/7734)).
+* Performance has been improved for range queries on integers and timestamps. Requires that you use the "BETWEEN" operation in MQL or the Query::between() method when you build the query. (Core issue [realm/realm-core#7785](https://github.com/realm/realm-core/pull/7785))
```
nhachicha, line 8:
```suggestion
* Performance has been improved for range queries on integers and timestamps. Requires that you use the "BETWEEN" operation in RQL or the Query::between() method when you build the query. (Core issue [realm/realm-core#7785](https://github.com/realm/realm-core/pull/7785))
```
?

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/AppTests.kt
```diff
@@ -438,12 +438,20 @@ class AppTests {
...
+            PlatformUtils.copyFile(metadataDir + "sync_metadata.realm", metadataDir + "copy_sync_metadata.realm")
+
             val wrongKey = TestHelper.getRandomKey()
             val config = RealmConfiguration
                 .Builder(setOf())
-                .name("sync_metadata.realm")
+                .name("copy_sync_metadata.realm")
                 .directory(metadataDir)
                 .encryptionKey(wrongKey)
+                .schemaVersion(7)
```
nhachicha, line 454: Why do we need to specify the schema version? the test is validating the encryption key...

nhachicha (approved):
Minor things 👍

packages/test-sync/src/commonTest/kotlin/io/realm/kotlin/test/mongodb/common/AppTests.kt
rorbech: Because the schema version that the client is using to write the metadata realm is greater than 0 which is used by default. If we don't set it we will get a migration error when trying to open the realm.

.github/workflows/pr.yml
rorbech: This PR is preceded by https://github.com/realm/realm-kotlin/pull/1779. But the explicit reference to Java 17 is maintained. This syntax will cause `actions/setup-java` to install both `17` and `vars.VERSION_JAVA` so this is exactly allowing us to run the sdkmanager with Java 17 while building the rest with `vars.VERSION_JAVA` pointed to by `JAVA_HOME`.

CHANGELOG.md
rorbech: Have addressed this in the preceding PR #1779

PR #1817, opened 2024-08-07
Adding insertToRealm: a fast and more memory efficient way to copy data into Realm

nhachicha:


caeduk:
@nhachicha did this make it into realm 2.3.0 ?

Issue #1825, opened 2024-08-23
Kotlin 2.0.20 support


AndreyFetisov:
### How frequently does the bug occur?

Always

### Description

'2.1.0' fails with the below:

### Stacktrace & log output

```shell
Task :app:compileDebugKotlin FAILED
e: java.lang.NoSuchMethodError: 'java.lang.RuntimeException org.jetbrains.kotlin.backend.common.CompilationExceptionKt.wrapWithCompilationException(java.lang.Throwable, java.lang.String, org.jetbrains.kotlin.ir.declarations.IrFile, org.jetbrains.kotlin.ir.IrElement)'
	at io.realm.kotlin.compiler.RealmModelLowering.lowerFromModuleFragment(RealmModelLoweringExtension.kt:83)
	at io.realm.kotlin.compiler.RealmModelLoweringExtension.generate(RealmModelLoweringExtension.kt:53)
	at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.applyIrGenerationExtensions(convertToIr.kt:437)
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.runActualizationPipeline(convertToIr.kt:246)
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.convertToIrAndActualize(convertToIr.kt:130)
	at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.convertToIrAndActualize(convertToIr.kt:99)
	at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.convertToIrAndActualize$default(convertToIr.kt:72)
	at org.jetbrains.kotlin.cli.jvm.compiler.pipeline.JvmCompilerPipelineKt.convertToIrAndActualizeForJvm(jvmCompilerPipeline.kt:196)
	at org.jetbrains.kotlin.cli.jvm.compiler.pipeline.JvmCompilerPipelineKt.convertAnalyzedFirToIr(jvmCompilerPipeline.kt:169)
	at org.jetbrains.kotlin.cli.jvm.compiler.pipeline.JvmCompilerPipelineKt.compileModulesUsingFrontendIrAndLightTree(jvmCompilerPipeline.kt:140)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:148)
[...]

PR #1826, opened 2024-08-27
Kotlin 2.0.20 support

KitsuneAlex:
Fixes #1825

PR #1830, opened 2024-09-10
Fix compiler crash when using Kotlin 2.0.20 

nhachicha:
Fixes https://github.com/realm/realm-kotlin/issues/1825 via https://github.com/realm/realm-kotlin/pull/1826 
Credit @KitsuneAlex


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## 2.2.0-SNAPSHOT (YYYY-MM-DD)
 
+> [!WARNING]
+> This is the last version compatible with Sync feature.
```
nirinchev, line 4: This is probably imprecise as I expect there may be more patch/minor versions we release in case we identify critical bugs that need to be addressed.

clementetb (approved):
🚀

KitsuneAlex:
Thank you all so much ^^

Issue #1544, opened 2023-10-12
App not Responding when Realm Configuration is built


rishabh876:
### How frequently does the bug occur?

Sometimes

### Description

Many of our users are facing a lot of ANRs on app launch. This happens when Koin dependies are injected on Application onCreate. From the stacktrace it looks like its always stuck in RealmConfiguration build Step.

I have not been able to reproduce this ANR so far.

### Stacktrace & log output

```shell
Thread 1 "main" tid=1 Runnable  


at kotlin.reflect.jvm.internal.impl.load.kotlin.header.ReadKotlinClassHeaderAnnotationVisitor$KotlinMetadataArgumentVisitor$2.<init> (ReadKotlinClassHeaderAnnotationVisitor.java:215)
  at kotlin.reflect.jvm.internal.impl.load.kotlin.header.ReadKotlinClassHeaderAnnotationVisitor$KotlinMetadataArgumentVisitor.stringsArrayVisitor (ReadKotlinClassHeaderAnnotationVisitor.java:215)
  at kotlin.reflect.jvm.internal.impl.load.kotlin.header.ReadKotlinClassHeaderAnnotationVisitor$KotlinMetadataArgumentVisitor.visitArray (ReadKotlinClassHeaderAnnotationVisitor.java:196)
  at kotlin.reflect.jvm.internal.impl.descriptors.runtime.components.ReflectClassStructure.processAnnotationArgumentValue (ReflectKotlinClass.kt:235)
  at kotlin.reflect.jvm.internal.impl.descriptors.runtime.components.ReflectClassStructure.processAnnotationArguments (ReflectKotlinClass.kt:184)
  at kotlin.reflect.jvm.internal.impl.descriptors.runtime.components.ReflectClassStructure.processAnnotation (ReflectKotlinClass.kt:166)
  at kotlin.reflect.jvm.internal.impl.descriptors.runtime.components.ReflectClassStructure.loadClassAnnotations (ReflectKotlinClass.kt:85)
  at kotlin.reflect.jvm.internal.impl.descriptors.runtime.components.ReflectKotlinClass$Factory.create (ReflectKotlinClass.kt:56)
  at kotlin.reflect.jvm.internal.impl.descriptors.runtime.components.ReflectKotlinClassFinder.findKotlinClass (ReflectKotlinClassFinder.kt:34)
[...]

PR #1851, opened 2024-10-31
Dont use kotlin reflection (#1544)

kzotin:
Kotlin reflection is super slow on some old devices (~20seconds on cold start, see #1544 ), which is a blocker for us after migrating from realm-java to kotlin.

We've got such device, and fix was verified there

### DIFF
--- a/packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/RealmObject.kt
+++ b/packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/RealmObject.kt
@@ -19,15 +19,36 @@ package io.realm.kotlin.internal.platform
 import io.realm.kotlin.internal.RealmObjectCompanion
 import io.realm.kotlin.types.BaseRealmObject
 import kotlin.reflect.KClass
-import kotlin.reflect.full.companionObjectInstance
+
 
 // TODO OPTIMIZE Can we eliminate the reflective approach? Maybe by embedding the information
 //  through the compiler plugin or something similar to the Native findAssociatedObject
 @PublishedApi
-internal actual fun <T : Any> realmObjectCompanionOrNull(clazz: KClass<T>): RealmObjectCompanion? =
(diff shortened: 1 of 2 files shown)


kzotin:
PS. I've submitted the CLA form yesterday

stavfx:
@cla-bot check

stavfx:
@kzotin I think you signed the form as "@kzotin" instead of "kzotin" and that's why the cla-bot is still failing

MrJohn10:
@cla-bot check

clementetb:
Would you mind adding a changelog entry to the PR?

kzotin:
Updated the changelog

packages/library-base/src/jvm/kotlin/io/realm/kotlin/internal/platform/RealmObject.kt
```diff
@@ -19,15 +19,36 @@ package io.realm.kotlin.internal.platform
...
+            null
+        }
+    }?.objectInstance as? RealmObjectCompanion
+
+    if (companion != null) {
+        reflectionCache[clazz] = companion
+    }
+
+    return companion
+}
```
santaevpavel, line 49: We tested this fix in our project. The fix makes realm configuration creating much faster, but it slows down some queries. If `realmObjectCompanionOrNull` called with `Double::class` then this function works much slower (more than on 50%). Most likely because of `Class.forName()` is called two times on every `realmObjectCompanionOrNull` function call.

tuba: Double::class implements Parcelable in your case?) Twice call can be only for classes that implements Parcelable and only once when warm up the cache

LouisCAD:
Hello, thanks to https://github.com/XilinJia/krdb, we published a fork on Maven Central yesterday, supporting all targets (iOS and macOS included), and we included a fix for this issue.

https://github.com/Infomaniak/realm-kotlin/releases/tag/3.2.8-2

